PR #15, opened 2018-07-26
Deprecate machine role and update vendor

Lion-Wei:


m1093782566:
Why do we need to update vendor?

dims:
@m1093782566 until we hit alpha in cluster-api it would be good to keep updating the vendored code. Actually someone is looking at a bot to propose PR(s) that update the vendored code too.

/lgtm

PR #35, opened 2018-10-08
Call Parse() to avoid noisy logs

chaosaffe:
Prevents glog from complaining about logging before parsing flags
See https://github.com/kubernetes/kubernetes/issues/17162 for more info

**Release note**:

```release-note
NONE
```


m1093782566:
/lgtm

/approve

THANKS for catching this.

PR #38, opened 2018-10-09
Run gofmt

flaper87:
This PR just fixes a fmt error


Lion-Wei:
@flaper87 Thanks, and we should open an issue to trace works about improve the ci. Which may include 
- gofmt check
- golint check
- add ut and ut test
- auto generate check(vendor and generated code)

Those works gonna take some time, I think.

PR #39, opened 2018-10-09
Use users $HOME/.ssh for openstack_tmp SSH key

chaosaffe:
Uses the users `$HOME` directory to store and reference the temporary SSH key generated to access the master nodes. This eliminates the need for MacOS users to create a fake `/root`, as well as enabling Linux users to run commands without root privileges.

N/A

N/A

**Release note**:

```release-note
NONE
```


cloud/openstack/deployer.go
```diff
@@ -56,8 +57,13 @@ func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clu
 		return "", err
 	}
 
+	homeDir, ok := os.LookupEnv("HOME")
```
flaper87: Should we try to read `MACHINE_CONTROLLER_SSH_HOME` before building the path ourselves?

chaosaffe: `MACHINE_CONTROLLER_SSH_HOME` is only defined in `generate-yaml.sh` and is not exported. This is just intermediary work to allow MacOS users to use the provider to bootstrap without a lot of overhead (and remove the dependence on running as `root` on Linux). Once #36 lands I intend to add a patch to enable the user to configure the directory to use via a flag. 
I know it is fairly common in the OpenStack world to use env vars, but I would really like to see us avoid using implicit inputs (i.e. env vars) and opt for explicit inputs (i.e. flags, config files)

flaper87: wfm! thanks

Issue #16, opened 2018-08-28
Use standard clouds.yaml file as config file


flaper87:
We're currently using a custom config file for `cluster-api-provider-openstack`, which is different from both, the `cloud-provider-openstack` config file and the [standard `clouds.yaml` file](https://docs.openstack.org/python-openstackclient/pike/configuration/index.html). The current config uses the following format:

```
user-name:
password: 
domain-name: 
tenant-id: 
region:
auth-url:
```
In order to provider a better user experience, it'd be better to either adopt the config file format used by the `cloud-provider-openstack` or, even better, the `clouds.yaml` format (gophercloud has support for the latter already). Here's an example of what the yaml would look like:

```
clouds:
  my_cloud:
    insecure: true
    verify: false
    identity_api_version: 3
    auth:
      auth_url: http://10.0.0.14:5000/v3
      project_name: admin
      username: admin
      password: $SUPER_PASSWORD
      project_domain_name: Default
      user_domain_name: Default
    region: RegionOne
```

Thoughts?

/cc @Lion-Wei

Lion-Wei:
Thanks, Actually I'm not very familiar with openstack cloud configs. But I feel use standard `clouds.yaml` file would be a better way.
So user offer a `clouds.yaml` configfile, and we generate it to a secret?

flaper87:
@Lion-Wei yes! We create a secret just like the `clouds.yaml` file and we change the `openstack-machine-controller` to use the same format as well. If you agree, I can start hacking on this

m1093782566:
@flaper87 

Contributions are always welcome!

flaper87:
@m1093782566 sure thing, thanks.

What I'd like to do, before coding anything, is evaluating whether this approach is something the rest of the team would be ok with. If so, I'm more than happy to work on it.

chaosaffe:
@flaper87 Have you made any commits on this? If not I am interested in picking this up.

flaper87:
@chaosaffe 

I started looking into this and then I thought it'd be nice to have this code shared across the cluster-api-provider and the cloud-provider OpenStack. I discussed the idea with some other folks from this SIG and the feedback seemed to be good. 

If you have time available to get this done in a short time, then feel free to take it as I'm not going to be able to do it myself until the end of next week. But I was definitely hoping to get it done before mid-Oct (just to give you an idea :D)

Thanks a lot for reaching out, let me know what you decide :smile: :cake:

[...]

PR #45, opened 2018-10-10
Use clouds.yaml for auth

iamemilio:
First Iteration Support for using Clouds.yaml to connect to openstack cluster:

In response to #16


cloud/openstack/clients/machineservice.go
```diff
@@ -50,6 +52,28 @@ type CloudConfig struct {
 	CAFile     string `json:"ca-file"`
 }
 
+// Converts clientconfig.Cloud to CloudConfig struct
+func ToClientConfig(cloud *clientconfig.Cloud) (*CloudConfig, error) {
+	conf := new(CloudConfig)
```
flaper87: Do we still need this struct?

Considering we're moving away from the custom config file to `clouds.yaml` it might be good to just use types provided by `gophercloud` rather than keeping this struct around. For example, `CloudConfig` doesn't define project specific options, keystone API versions, etc. This is part of what this issue is supposed to fix.

Here's an example of how this is managed in `terraform-provider-openstack`: https://github.com/flaper87/terraform-provider-openstack/blob/master/openstack/config.go#L73-L118

Lion-Wei:
Thanks, whenever ready for review, just remove the WIP, I'll give a fully review. : )
/assign

cloud/openstack/clients/machineservice.go
iamemilio: As an initial fix I went for more of a band-aid solution, but yeah, if functionality is the problem it makes sense to migrate to gophercloud completely.

flaper87: Functionality and format are the issue here :D Thanks a lot

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -81,14 +68,13 @@ type InstanceListOpts struct {
 	Name string `q:"name"`
 }
 
-func NewInstanceService(cfg *CloudConfig) (*InstanceService, error) {
-	authUrl := gophercloud.NormalizeURL(cfg.AuthURL)
+func NewInstanceService(cfg *clientconfig.Cloud) (*InstanceService, error) {
 	opts := &gophercloud.AuthOptions{
```
flaper87: Could you add all the struct fields? https://github.com/flaper87/terraform-provider-openstack/blob/master/openstack/config.go#L103-L115

That way we won't be leaving functionality out :smile:

iamemilio:
> Oh, btw! Do you think you could add some tests for this? Like parsing the `clouds.yaml` file. We could use something like that since we don't even have a proper e2e test right now.

I was going to ask about that actually haha

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -81,14 +68,13 @@ type InstanceListOpts struct {
...
-		IdentityEndpoint: authUrl,
-		Username:         cfg.Username,
-		Password:         cfg.Password,
-		DomainName:       cfg.DomainName,
-		TenantID:         cfg.TenantID,
+		IdentityEndpoint: cfg.AuthInfo.AuthURL,
+		Username:         cfg.AuthInfo.Username,
+		Password:         cfg.AuthInfo.Password,
+		DomainName:       cfg.AuthInfo.DomainName,
+		TenantID:         cfg.AuthInfo.ProjectDomainID,
```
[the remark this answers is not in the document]
iamemilio: It's also very confusing.

pkg/cloud/openstack/clients/machineservice.go
iamemilio: @flaper87 clientconfig.Cloud has all the fields of that struct and then some. https://github.com/gophercloud/utils/blob/259b9dbae8cbb5aa5b61d08555d98a7c522e736e/openstack/clientconfig/results.go#L18

is it worth duplicating?

iamemilio:
note: will squash on final commit

pkg/cloud/openstack/clients/machineservice.go
flaper87: We clarified this on IRC. Resolving this convo :+1:

pkg/cloud/openstack/clients/machineservice_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestMachineServiceInstance(t *testing.T) {
+	clientOpts := new(clientconfig.ClientOpts)
+	cloud, err := clientconfig.GetCloudFromYAML(clientOpts)
+	if err != nil {
+		t.Errorf("Couldn't create clientconfig")
+	}
+	_, err = NewInstanceService(cloud)
+	if !(strings.Contains(err.Error(), "i/o timeout")) {
```
flaper87: @Lion-Wei @chaosaffe I was helping @iamemilio with this test and this looks like a good start, although it's probably not the best way to do it. We should use `mock` instead, or just write the test differently.

flaper87:
@iamemilio I just remembered that we also have to update the `generate-yaml.sh` script: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/generate-yaml.sh#L79-L84

It's generating a config file based on the old format, we need to update it to use the new one.

pkg/cloud/openstack/clients/clouds.yaml
```diff
@@ -0,0 +1,13 @@
+clouds:
+  openstack:
+    auth:
+      auth_url: https://example.com:5000/v3
+      username: "test@test.com"
```
chaosaffe: nit: quotes are not required here

pkg/cloud/openstack/clients/clouds.yaml
```diff
@@ -0,0 +1,13 @@
...
+      password: "here'syourpassword"
```
chaosaffe: nit: quotes aren't required here either

iamemilio:
@flaper87 @chaosaffe  do we want to generate a clouds.yaml or just populate one that they already have?

chaosaffe:
Do we need to populate anything? I think they can just download a complete one from Horizon, etc... and use that?

flaper87:
> @flaper87 @chaosaffe do we want to generate a clouds.yaml or just populate one that they already have?

If we're not generating one, then the `generate-yaml.sh` file should be smart enough to go and pick it up itself. We can also just pass it through the CLI.

`generate-yaml.sh --clouds clouds.yaml` and we need to make sure it's [mounted in the right place](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c25411e8a4cc9e7166a2cb932d9204ff30350e02/config/manager/manager.yaml#L62) with [the right name](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/provider-components.yaml.template#L19)

pkg/cloud/openstack/clients/clouds.yaml
flaper87: but are also not prohibited :wink:

chaosaffe:
I am all for passing a path into `generate-yaml.sh`

iamemilio:
I was thinking the same thing. We can just pass an absolute path. Then it can just wrap the file in a secret like the rest of them.

flaper87:
This looks good! let's squash the commits, remove the WIP and merge it! Thanks a bunch @iamemilio

Gopkg.lock
```diff
@@ -2,112 +2,87 @@
 
 
 [[projects]]
-  digest = "1:6b21090f60571b20b3ddc2c8e48547dffcf409498ed6002c2cada023725ed377"
```
chaosaffe: Looks like these changes are due to an old version of `dep` 
@iamemilio can you please update your `dep` and run `dep ensure` again please?

PR #60, opened 2018-10-16
 Recommend using v1.12.1

flaper87:
Deployments using v1.10.0 are broken unless `--feature-gates=CustomResourceSubresources=true` is passed to the cluster creation and `.spec.subresources.status` is added to the CRD definition.

Rather than doing that and forcing the use of an older version of k8s, let's take this opportunity to update the minimum required version to the latest available.

More about this here: https://github.com/kubernetes-sigs/cluster-api/issues/533#issuecomment-430127733

/cc @Lion-Wei @chaosaffe

### DIFF
--- a/cmd/clusterctl/examples/openstack/machines.yaml.template
+++ b/cmd/clusterctl/examples/openstack/machines.yaml.template
@@ -19,8 +19,8 @@ items:
         securityGroups:
         - default
     versions:
-      kubelet: 1.9.4
-      controlPlane: 1.9.4
+      kubelet: 1.12.1
+      controlPlane: 1.12.1
 - apiVersion: "cluster.k8s.io/v1alpha1"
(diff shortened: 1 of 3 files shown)


cmd/clusterctl/examples/openstack/machines.yaml.template
```diff
@@ -19,8 +19,8 @@ items:
         securityGroups:
         - default
     versions:
-      kubelet: 1.9.4
-      controlPlane: 1.9.4
+      kubelet: 1.12.1
```
chaosaffe: Just wanted to note that 1.12 is only required in the bootstrap cluster, not in any of the child clusters ;)

chaosaffe:
/lgtm
Tested with minikube 0.30.0 without issue

PR #61, opened 2018-10-16
Add kuryr support

luis5tb:
This patch adds kuryr support by ensuring that VMs can be created
on top of trunk ports if kuryr is enabled.

TODO:
- testing


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -187,7 +223,45 @@ func (is *InstanceService) InstanceCreate(config *openstackconfigv1.OpenstackPro
...
+	// get instance port id
+	allInterfaces, err := attachinterfaces.List(computeClient, id).allPages()
+	if err != nil {
+		return err
+	}
+	instanceInterfaces, err := attachinterfaces.ExtractInterfaces(allInterfaces)
+	if err != nil {
+		return err
+	}
+	portID := InstanceService[0].PortID
```
celebdor: Shouldn't this be:
```
var portID string
if len(InstanceInterfaces) > 0 {
    portID = instanceInterfaces[0].PortID
}
```

luis5tb: yep!

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -187,7 +223,45 @@ func (is *InstanceService) InstanceCreate(config *openstackconfigv1.OpenstackPro
...
+	}
+	allTrunks, err := trunks.List(is.networkingClient, listOpts).AllPages()
+	if err != nil {
+		return err
+	}
+	trunkInfo, err := trunks.ExtractTrunks(allTrunks)
+	if err != nil {
+		return err
+	}
+	if len(trunkInfo) != 1 {
```
celebdor: By definition, there can't be more than one. So if len(trunkInfo) != 1, it is 0, so the error message should be fmt.Errorf("No trunk found for port ID: %s", portID)

luis5tb: true, and actually, if there is no trunk it is ok too

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -159,25 +171,48 @@ func (is *InstanceService) GetAcceptableFloatingIP() (string, error) {
...
-	var createOpts servers.CreateOpts
 	if config == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace.")
 	}
 	userData := base64.StdEncoding.EncodeToString([]byte(cmd))
-	createOpts = servers.CreateOpts{
+
+	// create server port
+	portCreateOpts := ports.CreateOpts{
+		Name:           name,
```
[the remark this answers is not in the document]
luis5tb: done

chaosaffe:
Sorry, we dragged our feet a little on this one, and there have been quite some changed to how we handle networks.
@luis5tb would you mind rebasing and we can take a look once the changes are done?

Thanks!

luis5tb:
> Sorry, we dragged our feet a little on this one, and there have been quite some changed to how we handle networks.
> @luis5tb would you mind rebasing and we can take a look once the changes are done?
> 
> Thanks!

Sure! I'll do asap!

chaosaffe:
@luis5tb the tests appear to be failing as your code isn't `gofmt`ed. Can you please gofmt and push?

celebdor:
@chaosaffe Now it is properly formatted, could you take a new look?

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -269,18 +273,64 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
 		}
 	}
 	userData := base64.StdEncoding.EncodeToString([]byte(cmd))
-	createOpts = servers.CreateOpts{
+	clusterTag := ""
+	for _, tag := range config.Tags {
+		if string.Contains(tag.Name, "ClusterID") {
```
chaosaffe: It appears you have the wrong package name here, which is causing the tests to fail as the code is unbuildable since there isn't a `string` package
```suggestion
		if strings.Contains(tag.Name, "ClusterID") {
```

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -19,17 +19,22 @@ package clients
 import (
 	"encoding/base64"
 	"fmt"
+	"string"
```
chaosaffe:
```suggestion
	"strings"
```

flaper87:
@luis5tb @celebdor needs rebase :(

flaper87:
Also, could you guys remove the WIP prefix if this is ready for review? Otherwise folks will skip this patch :D

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -269,18 +273,64 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
 		}
 	}
 	userData := base64.StdEncoding.EncodeToString([]byte(cmd))
-	createOpts = servers.CreateOpts{
+	clusterTag := ""
+	for _, tag := range config.Tags {
+		if strings.Contains(tag.Name, "ClusterID") {
```
flaper87: Can you elaborate on why you're adding the tag only for the `ClusterID`? Shouldn't we also support adding other tags?

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -75,6 +81,13 @@ type NetworkParam struct {
 	Filter Filter `json:"filter,omitempty"`
 }
 
+type TagParam struct {
```
flaper87: I like this `TagParam` here and it'd be great if we could use it for the cluster-level resources too. We've recently added tags to networks created by the cluster controller in #178 it'd be great to update that code to use this TagParam and allow for custom tags to be added.

This said, I think we should have a default, tho. A tag that can't be overwritten by the custom tags set in the Cluster/Machine spec. I'm happy to keep using `cluster-api-provider-openstack=<cluster-name>` for this. Thoughts?

/cc @chrigl

chrigl: I like the idea of having a way to add custom tags to network resources, and I am with @flaper87  that we might want to have defaults that are always applied. But from the users perspective, I think we should leave `cluster-api-provider-openstack` and the `<cluster-name>` be two separate tags. So it would be more convenient to filter resources. e.g. such things would be possible

```
# all resources owned by cluster-api-provider openstack
openstack network list --tags cluster-api-provider-openstack

# all owned resources and cluster-name = default/test1
openstack network list --tags cluster-api-provider --tags default/test1

# all owned resources but *not* cluster-name = default/test1
openstack network list --tags cluster-api-provider-openstack --not-tags=default/test1
```

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -290,7 +340,52 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
+	}
+
+	// get and delete trunks
+	for _, port := range instanceInterfaces {
+		listOpts := trunks.ListOpts{
+			PortID: port.PortID,
+		}
+		allTrunks, err := trunks.List(is.networkClient, listOpts).AllPages()
+		if err != nil {
+			return err
```
chrigl: This code will error out if the installation of OpenStack does not support trunking... with the error "Resource not found". Even if trunks are not used to create VMs. Further cleanups will not happen.

In my test, I don't have an OpenStack with trunking, but I created a node (via machine object) and manually attached another device of a different subnet. Deletion of this node caused a left over port.

The error results in another reconciliation. But in the next loop, the machine is already gone, so no interfaces can be listed. Maybe it would be good to remove the VM at the very last step in this function. At least using the openstack cli tools, it is possible to delete ports that are attached to a running VM. However, trunking not being supported is not an error here.

Looping happens until
`I0114 15:47:26.492098    2447 actuator.go:195] Skipped deleting a VM that is already deleted.`

pkg/apis/openstackproviderconfig/v1alpha1/types.go
luis5tb: Ok, so I'll keep those 2 and add the extra ones (if existing)

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -19,17 +19,22 @@ package clients
...
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
```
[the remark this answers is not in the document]
luis5tb: I'll do! It is preferred on a different commit, right? (same PR)

pkg/cloud/openstack/clients/machineservice.go
luis5tb: This was added so that resources created by CAPO can be deleted by the installer (following installer logic)

luis5tb: We can simply add all, and that should work too

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -269,18 +273,64 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
+	clusterTag := ""
+	for _, tag := range config.Tags {
+		if strings.Contains(tag.Name, "ClusterID") {
+			clusterTag = tag.Name + tag.Value
+			break
+		}
+	}
+
+	var ports_list []servers.Network
+	for _, net := range nets {
```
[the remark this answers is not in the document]
luis5tb: So, I should first check if it exists and then replace the tags?

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -269,18 +273,64 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
+		// create server port
+		portCreateOpts := ports.CreateOpts{
+			Name:           name,
+			NetworkID:      net.UUID,
+			SecurityGroups: &config.SecurityGroups,
```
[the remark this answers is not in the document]
luis5tb: Why is that? you can also pass here the SG name, right?

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
chrigl: Yes, so far we always did this in the same PR.

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
chrigl: It is not just about tags. It is about all resources. Before creating a new port, you should check for an existing one. I know… this looks pretty ugly... but the code can fail for whatever reason at any point in time. Even if only the node it runs on dies in the exact moment when a port has been created, but the machine not. The most un-nice thing IMHO is that we must use names to find potential existing resources.
Another example for failing is... parts of OpenStack are not available. Let's assume the Neutron Network API does work, but the Nova Compute API does not work at the time.
This will result in
```
Call to InstanceCreate:
Create port
Create VM => return err

## Reconciler kicks in.
Call to InstanceCreate:
Create port
Create VM => return err

## Reconciler kicks in.
...
```
So the code would create a port at each reconciliation loop. Workflow must be something like

````
Call to InstanceCreate:
Use existing port OR Create new port
Create VM => return err

Call to InstanceCreate
Use existing port
Create VM => return err
```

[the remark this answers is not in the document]
luis5tb: yes yes! I got this! preparing the new version checking for ports on the given net with the name

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
chrigl: This is how neutron works. https://developer.openstack.org/api-ref/network/v2/?expanded=create-port-detail

When passing SGs to servers.Create, OpenStack takes care of the name <=> uuid translation. When creating ports directly, only uuids are supported.

Btw. When using `openstack --debug port create --network net-test --security-group test-group blahfasel`, the OpenStack client fetches all SecurityGroups and does the translation, and then posts to the ports endpoint with only uuids included.

I would be totally fine to implement this the same way.

pkg/cloud/openstack/clients/machineservice.go
luis5tb: I thought you could not delete the port of a VM (at least not the last one). Another option is checking if the VM exists, and only delete it if it is there, otherwise continue to remove the ports/trunks

luis5tb: Umm, on a second thought... how will we not the associated ports... I'll check if all the ports can be deleted from a running VM

luis5tb: ok, it seems you can delete all the ports... so VM should be deleted last! thanks!

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
luis5tb: But that should be handled on the gopghercloud side, isn't it? And based on the testing examples it seems that it is already handled:
https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/ports/testing/requests_test.go#L185-L204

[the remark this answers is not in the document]
chrigl: Hmm. Seems this is not the case. I am currently only able to test against OpenStack Pike... 

I tested with the code below against an OpenStack Pike, and get `'kube-group' is not a valid UUID`.

I think, because the official openstack cli also converts names to UUID before creating the port, I would argue that this is not anything `gophercloud` should do, but the enduser should. It would be also fine if the enduser is the one who defines the yaml. But then it must be documented prominently.

For me the test mentioned test should be changed.

```
package main

import (
        "flag"
        "fmt"
        "log"
        "os"

        "github.com/gophercloud/gophercloud"
        "github.com/gophercloud/gophercloud/openstack"
        "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
        "github.com/gophercloud/utils/openstack/clientconfig"
)

func getNetworkClient() (*gophercloud.ServiceClient, error) {
        opts, err := clientconfig.AuthOptions(&clientconfig.ClientOpts{})
        if err != nil {
                return nil, err
        }

        provider, err := openstack.AuthenticatedClient(*opts)
        if err != nil {
                return nil, fmt.Errorf("create providerClient err: %v", err)
        }

        client, err := openstack.NewNetworkV2(provider, gophercloud.EndpointOpts{})
        if err != nil {
                return nil, err
        }

        return client, nil
}


func main() {
        networkID := flag.String("net-id", "", "a valid network id")
        groupName := flag.String("security-group-name", "", "a valid security group name")
        groupID := flag.String("security-group-id", "", "a valid security group id")

        flag.Parse()

        if *networkID == "" {
                log.Fatal("Please provide a net-id")
        }
        if *groupName == "" {
                log.Fatal("Please provide a security-group-name")
        }
        if *groupID == "" {
                log.Fatal("Please provide a security-group-id")
        }

        client, err := getNetworkClient()
        if err != nil {
                log.Fatalf("Unable to get network client: %v", err)
        }

        port, err := ports.Create(client, ports.CreateOpts{
                Name:      "test-port-with-uuid",
                NetworkID: *networkID,
                SecurityGroups: &[]string{
                        *groupID,
                },
        }).Extract()
        if err != nil {
                log.Fatalf("Unable to create port with UUID in SecurityGroups: %v", err)
        }
        fmt.Printf("Created port with UUID in SecurityGroups: %v\n", port)

        port, err = ports.Create(client, ports.CreateOpts{
                Name:      "test-port-with-name",
                NetworkID: *networkID,
                SecurityGroups: &[]string{
                        *groupName,
                },
        }).Extract()

        if err != nil {
                log.Fatalf("Unable to create port with Name in SecurityGroups: %v", err)
        }

        fmt.Printf("Created port with Name in SecurityGroups: %v\n", port)
}
```

```
$ dep status
PROJECT                             CONSTRAINT     VERSION        REVISION  LATEST   PKGS USED
github.com/gophercloud/gophercloud  branch master  branch master  a9f9006   a9f9006  8
github.com/gophercloud/utils        branch master  branch master  34f5991   34f5991  1
gopkg.in/yaml.v2                    v2.2.2         v2.2.2         51d6538   v2.2.2   1

$ go run main.go -net-id=92cbf083-e057-4974-b24d-dce55279f4f9 -security-group-name=kube-group -security-group-id=f2c2ff1a-0785-4c23-b316-7d8c1b823b20

Created port with UUID in SecurityGroups: &{51e35d87-802d-4818-8c6f-0b300d153cd5 92cbf083-e057-4974-b24d-dce55279f4f9 test-port-with-uuid  true DOWN fa:16:3e:64:a9:c1 [{f44ee19c-bbe5-414a-81f1-34e323c5e37d 192.168.0.12}] bf8e5f9815c54afea53962fd4a13c745 bf8e5f9815c54afea53962fd4a13c745  [f2c2ff1a-0785-4c23-b316-7d8c1b823b20]  [] []}
2019/01/15 21:42:17 Unable to create port with Name in SecurityGroups: Bad request with: [POST https://network.fra.optimist.innovo.cloud/v2.0/ports], error message: {"NeutronError": {"message": "Invalid input for security_groups. Reason: 'kube-group' is not a valid UUID.", "type": "HTTPBadRequest", "detail": ""}}
exit status 1

$ openstack security group show kube-group
+-----------------+--------------------------------------+
| Field           | Value 
+-----------------+--------------------------------------+
| created_at      | 2018-10-25T14:38:07Z
| id              | f2c2ff1a-0785-4c23-b316-7d8c1b823b20
| name            | kube-group
```

luis5tb:
@chrigl: strange! it didn't saw me any conflict, anyway, rebased!

flaper87:
@luis5tb needs rebase again. Let me know and I'll merge it right away

PR #67, opened 2018-10-22
Remove hardcoded machine username

### LINKED ISSUE #66: Hardcoded ubuntu username
Allow for setting the ssh username in the providerConfig, if needed.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/machineactuator.go#L50

flaper87:
Allow for passing the machine username in the providerConfig. This is
mostly needed by the `clusterctl` command and we should probably find a
better way to get the kubeconfig data. Until then, let's not hardcode
the username in the implementation.

Closes #66

### DIFF
--- a/pkg/apis/openstackproviderconfig/v1alpha1/types.go
+++ b/pkg/apis/openstackproviderconfig/v1alpha1/types.go
@@ -27,6 +27,10 @@ type OpenstackProviderConfig struct {
 	Flavor string `json:"flavor"`
 	// The name of the image to use for your server instance.
 	Image string `json:"image"`
+
+	// The machine ssh username
+	SshUserName string `json:"sshUserName,omitempty"`
+
 	// A networks object. Required parameter when there are multiple networks defined for the tenant.
 	// When you do not specify the networks parameter, the server attaches to the only network created for the current tenant.
 	Networks []NetworkParam `json:"networks,omitempty"`
--- a/pkg/cloud/openstack/deployer.go
+++ b/pkg/cloud/openstack/deployer.go
@@ -22,7 +22,10 @@ import (
 	"os"
 	"strings"
 
+	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
+
+	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
 	clustercommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	"sigs.k8s.io/cluster-api/pkg/util"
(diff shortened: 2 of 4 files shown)


pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -27,6 +27,10 @@ type OpenstackProviderConfig struct {
 	Flavor string `json:"flavor"`
 	// The name of the image to use for your server instance.
 	Image string `json:"image"`
+
+	// The machine ssh username
+	SshUserName string `json:"sshUserName",omitempty`
```
chaosaffe:
```suggestion
	SshUserName string `json:"sshUserName,omitempty"`
```

pkg/cloud/openstack/deployer.go
```diff
@@ -62,15 +65,28 @@ func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clu
...
+		fmt.Sprintf("%s@%s", machineConfig.SshUserName, ip),
 		"echo STARTFILE; sudo cat /etc/kubernetes/admin.conf"))
 	parts := strings.Split(result, "STARTFILE")
 	if len(parts) != 2 {
 		return "", nil
 	}
 	return strings.TrimSpace(parts[1]), nil
 }
+
+func (d *DeploymentClient) providerconfig(providerConfig clusterv1.ProviderConfig) (*openstackconfigv1.OpenstackProviderConfig, error) {
```
chaosaffe: What does this func have to do with the `*DeploymentClient`? Seems like this shouldn't belong on that struct

flaper87: kinda, sorta, maybe! It's a utility to unmarshal the providerConfig raw data into a providerConfig struct. I'd agree this is prob not the best place to put it but I thought we could keep it here while I work on cleaning up other parts of the code.

I copied this function from the OpenStackClient implementation, which makes it even worse as I basically duplicated code :P it's just temporary, tho.

pkg/apis/openstackproviderconfig/v1alpha1/types.go
flaper87: doh! lemme fix that

pkg/cloud/openstack/deployer.go
chaosaffe: Since it doesn't require `*DeploymentClient` for input, nor does it mutate `*DeploymentClient` in any way it is much better to leave this func unattached to a struct, or better to move into the `openstackproviderconfig` package to give a consistent func for unmarshalling

flaper87: Yes, as I said earlier. I don't this function belongs to this struct but, I'm trying to take baby steps and introduce small refactors as we don't have good test coverage yet.

Small refactor -> let ppl test things -> repeat :smile:

PR #75, opened 2018-10-25
Get only the last line of the kubeadm token cmd

### LINKED ISSUE #74: getKubeadmToken produces gibberish output when offline
This call produces gibberish when working offline and no kubernetes-version is specified: 

```
TOKEN=I1025 09:05:57.001785      15 version.go:89] could not fetch a Kubernetes version from the internet: unable to ge
t URL "https://dl.k8s.io/release/stable-1.txt": Get https://dl.k8s.io/release/stable-1.txt: net/http: request canceled
while waiting for connection (Client.Timeout exceeded while awaiting headers)
I1025 09:05:57.001886      15 version.go:94] falling back to the local client version: v1.12.1
pjpft6.5o7555anlw2y2b3a
```

FWIW, we should not be relying on kubeadm to generate tokens but, this is what we have now so we ought to fix it. We could pass `kubernetes-version` to the call if it was a valid option in the `TokenParams` struct. Sadly it isn't.

flaper87:
Depending on the arguments passed to kubeadm, it may or may not output
more data to stdout. Since we're shelling out and depending on the
kubeadm cli to generate the token, this commit fixes the token parsing
by splitting the output into lines and getting the very last one.

This is far from ideal, as in we should not be depending on kubeadm to
generate the token. Until that is fixed, this patch seems to be a good
compromise.

Closes #74

### DIFF
--- a/pkg/cloud/openstack/machineactuator.go
+++ b/pkg/cloud/openstack/machineactuator.go
@@ -445,7 +445,9 @@ func (oc *OpenstackClient) getKubeadmToken() (string, error) {
 		glog.Errorf("unable to create token: %v", err)
 		return "", err
 	}
-	return strings.TrimSpace(output), err
+
+	soutput := strings.Split(output, "\n")
+	return strings.TrimSpace(soutput[len(soutput)-1]), err
 }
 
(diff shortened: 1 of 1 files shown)


chaosaffe:
Future work: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/78

PR #84, opened 2018-10-30
Add templates for CentOS

flaper87:
Refactor generate-yaml.sh to allow for generating templates for
multiple OSs.

I consider generate-yaml.sh a temporary script that we're using to ease
development and not a production ready tool. Eventually it should go
away/moved to avoid sending the wrong message as far as these examples
go.


chaosaffe:
We certainly need some better mechanism for this, I am hoping that work around https://docs.google.com/document/d/1mMdKnNi4Yio1y4GAzU_6CMz21XbAkM__ZIIEZ_i8OHo/edit will come up soon.
/lgtm

PR #89, opened 2018-10-31
Generate bootstrap token

### LINKED ISSUE #78: Find a better way to generate tokens
We're currently depending on `kubeadm` to generate tokens. This is far from ideal as it requires to shell out and it doesn't work on environments where kubeadm is not present.

chrigl:

Therefore, I added new dependency to k8s.io/cluster-bootstrap.

Shelling out to kubeadm from token generation is now removed from
machineactuator.go. I used the tools of cluster-bootstrap to generate a
kubeadm compliant token name and create a secret out of it.

This implementation does not use the ways kubeadm currently uses,
because it would pull a hole bunch of other dependencies in, which are
IMHO not needed at that point.

Fixes #78

1. Pointers to kubeadm 
* https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/bootstraptoken/node/token.go
* https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/types.go#L201

**Release note**:

```release-note

```


pkg/cloud/openstack/machineactuator.go
```diff
@@ -436,18 +441,32 @@ func (oc *OpenstackClient) providerconfig(providerConfig clusterv1.ProviderConfi
...
+	if err != nil {
+		return "", err
 	}
-	output, err := oc.kubeadm.TokenCreate(tokenParams)
+
+	expiration := time.Now().UTC().Add(TokenTTL)
+	tokenSecret, err := bootstrap.GenerateTokenSecret(token, expiration)
 	if err != nil {
-		glog.Errorf("unable to create token: %v", err)
+		// If this happens, there is a bug somewhere
```
flaper87: Should this panic and exit then? What happens if we get here and the execution continues? Any horrible side-effect?

pkg/cloud/openstack/machineactuator.go
```diff
@@ -436,18 +441,32 @@ func (oc *OpenstackClient) providerconfig(providerConfig clusterv1.ProviderConfi
...
 
-	soutput := strings.Split(output, "\n")
-	return strings.TrimSpace(soutput[len(soutput)-1]), err
+	err = oc.client.Create(context.TODO(), tokenSecret)
+	if err != nil {
+		return "", err
+	}
+
+	if tokenSecret == nil {
+		return "", errors.New("unable to create token")
```
flaper87: When can this happen? It sounds like if there's a problem generating the token `oc.client.Create` would return an error, right?

pkg/cloud/openstack/machineactuator.go
```diff
@@ -52,6 +56,8 @@ const (
 
 	TimeoutInstanceCreate       = 5 * time.Minute
 	RetryIntervalInstanceStatus = 10 * time.Second
+
+	TokenTTL = 30 * time.Minute
```
flaper87: 30mins might not be enough in some scenarios where either network is not super fast or the nodes bootstrap process takes longer. Sounds horrible but it's possible :smile:

I'd go from 60 mins on.

pkg/bootstrap/token.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	tokenapi "k8s.io/cluster-bootstrap/token/api"
+	tokenutil "k8s.io/cluster-bootstrap/token/util"
+)
+
+// GeneratesTokenSecret returns a Secret conform to kubeadms bootstrap tokens
+func GenerateTokenSecret(token string, expiration time.Time) (*v1.Secret, error) {
```
flaper87: Sorry for asking but, was this function copied from somewhere? If so, I'd put a reference, mostly as a way to document the logic behind this function (although it's kinda explicit).

pkg/cloud/openstack/machineactuator.go
chrigl: I thought about this as well, and think to panic is better, because the admin sees the problem in kubernetes (crash loop), and does not need to dig into logs.

chrigl:
@flaper87 I think it would make sense to propose this to cluster-api. What do you think, should we prove the code here and then try to bring is to cluster-api?

chaosaffe:
Agreed, the common standard is to implement things in the providers and then lift them into cluster-api proper

flaper87:
@chrigl yup, absolutely! Can we create an issue for this? I'd like us to be able to track the "future migration" of this work to the `cluster-api` repo.

PR #91, opened 2018-11-01
OS_CLOUD templating re-work

iamemilio:
**What this PR does / why we need it**: Re-works the way that OS_CLOUD gets set in manager.yaml to avoid having to template it with a seperate file. Also fixes **critical** bug that reads the template into provider-components in place of the actual manager file.

Review Request: @chaosaffe


iamemilio:
@roberthbailey  Hi, you are listed as one of the reviewers. Could you please look this patch over.

Issue #106, opened 2018-11-19
Implement basic cluster actuator


chrigl:
So far, the cluster actutor is not yet implemented, except for basic types.

I would like to see a way to set up Networks, SecrityGroups and Loadbalancers for the cluster. In a first step, a single subnet for all nodes (master and worker) would IMHO be enough. Created Machines should use the created infrastructure.

Todo:
* Create Network with subnet with a configured CIDR
* Create a router connected to a configured external network, as well as the just created subnet
* Create SecurityGroups for master and nodes
* Create a loadbalancer with floating IP

~~LB Floating IP~~ and NetworkID must go to cluster status. SecurityGroups also somehow, but I am currently not sure yet how. Edit: Floating IP not needed because there is an `APIEndpoint` on `ClusterStatus`.

The loadbalancer should be used as the entry point for the apiserver, so we are able to create HA control planes.

If we are done with this, we can update the machine actuator to use the infrastructure.

See also: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/actuators/cluster/actuator.go

chrigl:
I will start working on it this week if there are no concerns.

gdoctor:
@chrigl I am currently working on the security group aspect, very similar to the aws implementation if you want to keep that our of your scope for now. You could just let me know what you had in mind for security groups within the cluster actuator, because right now most of that logic exists in the machine actuator. Up to you

chrigl:
@gdoctor I definitely appreciate help on this. I think, all building blocks necessary for a cluster should go into the cluster actuator, because they are not tight to a specific machine. I count SecurityGroups also as building block, because it is created once (at cluster level) and then used individually in the machines.

I am starting with getting the types in `apis/openstackproviderconfig/v1alpha1` to support a cluster config and status, and get this landed in master. I want to have as many small-ish PRs as possible. You could build on top of this, then... Does this sound legit?

gdoctor:
@chrigl sorry this got lost for me during the holidays here in the US. So is your idea that the cluster actuator actually creates the Security Groups within OpenStack? And then the machine actuator would  manage which security groups exist, are added, or are removed on a per machine/machineset basis? I see this working well

chrigl:
@gdoctor No problem. Hoping you had a great vacation :)

[...]

PR #107, opened 2018-11-20
Added OpenstackClusterProviderConfig

chrigl:

Added OpenstackClusterProviderConfig. Plus some minor refactorings to make OpenstackClusterProviderConfig work.

Building block for further development of the cluster actuator.

Fixes #

Nothing yet, but relates to #106 

**Release note**:

```release-note

```


chaosaffe:
@chrigl SGTM
Can you please update the examples and we can get this merged?

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -62,8 +67,24 @@ type RootVolume struct {
...
+// +k8s:openapi-gen=true
+type OpenstackClusterProviderConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// NodeCIDR is the OpenStack Subnet to be created
+	NodeCIDR string `json:"nodeCidr,omitempty"`
+	// Floating IP which will be assigned to a created loadbalancer. If not provided,
+	// no loadbalancer will be created.
+	FloatingIP string `json:"floatingIP,omitempty"`
```
flaper87: I wonder if we should pick a better name for this. I hate to nitpick on names but as a user, it might not be entirely obvious to me that setting a floatingIP ends up creating a load balancer. Thoughts? Maybe call it LoadBlanacerIP?

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -62,8 +67,24 @@ type RootVolume struct {
...
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// OpenstackClusterProviderConfig is the providerConfig for OpenStack in the cluster object
+// +k8s:openapi-gen=true
+type OpenstackClusterProviderConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// NodeCIDR is the OpenStack Subnet to be created
+	NodeCIDR string `json:"nodeCidr,omitempty"`
```
flaper87: Can you expand the docs here? It's not clear to me why we need the NodeCIDR here.

chrigl:
/hold

First getting yaml (#110) and rename to Spec (#109) in and then continuing on this one.

chrigl:
@flaper87 Just a more general question that came up when working with this for the network reconciler...

Currently we get this out of it:
```
    providerSpec:
      value:
        apiVersion: openstackproviderconfig/v1alpha1
        kind: OpenstackClusterProviderConfig
        externalNetworkId: 54258498-a513-47da-9369-1a644e4be692
        floatingIP: 10.0.0.1
        nodeCidr: 10.1.0.0/24
```

Which feels quite a bit odd in comparison to other kubernetes objects. I would prefer to add a `Spec` field.
```
    providerSpec:
      value:
        apiVersion: openstackproviderconfig/v1alpha1
        kind: OpenstackClusterProviderConfig
        spec:
          externalNetworkId: 54258498-a513-47da-9369-1a644e4be692
          floatingIP: 10.0.0.1
          nodeCidr: 10.1.0.0/24
```

How do you think about it?

pkg/apis/openstackproviderconfig/v1alpha1/types.go
chaosaffe: What is it the FloatingIP/LoadBalancerIP for? I would be more specific, indicating that this is for access to the control plane (I think?)

chrigl: I will leave this out. Currently not sure how to implement it.

The rough idea is to create a floating IP for the loadbalancer. That then can be used for a HA setup. This IP must reach the first master anyways, because it also needs to include it in its SANs. To create the actual LB we might decide to create a Type-LB-Service for kube-apiserver. But I would leave this out of scope here.

pkg/apis/openstackproviderconfig/v1alpha1/types.go
chrigl: Leaving this out as well at this point.

I will re-introduce it with a broader description when I add the reconciling code for networks. Does not make much sense to have fields that are actually not used.

chrigl:
/hold cancel

@chaosaffe and @flaper87 could you please have another look at this. Thank you :)

PR #112, opened 2018-11-23
Renamed providerConfig to providerSpec

### LINKED ISSUE #109: Rename ProviderConfig to ProviderSpec
There have been changes upstream to rename `ProviderConfig` to `ProviderSpec`. We need to pull in the upstream changes and update our references and documentation.

xref: https://github.com/kubernetes-sigs/cluster-api/pull/548

chrigl:

Updated cluster-api to get `providerSpec` in favor of `providerConfig`.
Renamed `providerConfig` to `providerSpec`.
Also added two little railguards in the actuator, to not panic if there
is no `providerSpec.Value`. In theory with this, it is possible to
update the controller and edit existing manifests from `providerConfig`
to `providerSpec`.

xref: kubernetes-sigs/cluster-api#548

Fixes #109 

cluster-api changed the interface `machine.Actuator` to use a `context.Context` as its first parameter. I added one, but so far it is not used.
https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/controller/machine/actuator.go#L28

If you test:
During the minikube phase, the `k8scloudprovider/openstack-cluster-api-controller:latest` image is used. You must edit the deployment to use `chrigl/openstack-cluster-api-controller:latest`. Before managing this, potentially, the old controller removed the `providerSpec` and added an empty `providerConfig` of the cluster object. You should re-apply the `ubuntu/out/cluster.yaml`.

Same is true for the deployed cluster. The old controller will end in a crashloop. Again it should be enough to change the image.

**Release note**:

```release-note
This in fact is a breaking change. A currently managed cluster will not be migrated to `providerSpec`.
```


chrigl:
Working on rebasing to master and getting Gopkg.lock into shape.

chrigl:
Don't know what happend in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/74d848f232ef2b3e33ec487d257b71b9d56ce7a4#diff-bd247e83efc3c45ae9e8c47233249f18 . Maybe an old or dev version of dep running?

PR #115, opened 2018-11-29
Adding security group management to machine actuator

gdoctor:
**What this PR does / why we need it**: Adding logic to machine actuator that allows addition and removal of security groups by editing the Machine Spec. This PR is currently assuming that all added security groups are already created in the OpenStack project, and therefore does not handle the actual creation or deletion of a security group. It simply adds/removes a security group to a machine. 

It stores all security groups that it manages within an annotation "openstack-security-groups". This allows a machine to have additional security groups that are managed externally to cluster-api if a user desires this behavior.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/openstack/machine/actuator.go
+++ b/pkg/cloud/openstack/machine/actuator.go
@@ -18,17 +18,16 @@ package machine
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"net"
 	"os"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/golang/glog"
+	errorwrapper "github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	tokenapi "k8s.io/cluster-bootstrap/token/api"
@@ -225,7 +224,7 @@ func (oc *OpenstackClient) Create(cluster *clusterv1.Cluster, machine *clusterv1
 
 	}
 
-	return oc.updateAnnotation(machine, instance.ID)
+	return oc.updateAnnotation(machine, instance, providerConfig)
 }
 
 func (oc *OpenstackClient) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
@@ -249,7 +248,13 @@ func (oc *OpenstackClient) Delete(cluster *clusterv1.Cluster, machine *clusterv1
 	return nil
 }
 
+// TODO cluster-api is only watching for updates to individual Machines, not MachineSets.. why?
 func (oc *OpenstackClient) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+	machineConfig, err := openstackconfigv1.MachineConfigFromProviderConfig(machine.Spec.ProviderConfig)
+	if err != nil {
+		return err
+	}
+
 	status, err := oc.instanceStatus(machine)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


pkg/cloud/openstack/machine/actuator.go
```diff
@@ -249,7 +248,13 @@ func (oc *OpenstackClient) Delete(cluster *clusterv1.Cluster, machine *clusterv1
 	return nil
 }
 
+// TODO cluster-api is only watching for updates to individual Machines, not MachineSets.. why?
```
chaosaffe:
```suggestion
```
`MachineDeployment`s and `MachineSet`s are handled by `cluster-api`, not the providers. The provider only ever looks at the machines, and `cluster-api` is responsible for actuating the changes from the `MachineDeployment`s and `MachineSet`s to the `Machine`s

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -263,16 +268,22 @@ func (oc *OpenstackClient) Update(cluster *clusterv1.Cluster, machine *clusterv1
...
+	if err := oc.updateSecurityGroups(cluster, machine, machineConfig); err != nil {
+		return errorwrapper.Wrap(err, "failed to ensure security groups")
+	}
+
+	// TODO updatedSecurityGroups is a part of Spec.ProviderConfig so it will trigger machine re-build as is
 	if !oc.requiresUpdate(currentMachine, machine) {
 		return nil
 	}
 
+	// TODO not all changes require recreating machine from scratch, need to differentiate.
```
chaosaffe:
```suggestion
```
Please open an issue for this

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -18,17 +18,16 @@ package machine
...
 	"fmt"
 	"io/ioutil"
-	"net"
 	"os"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/golang/glog"
+	errorwrapper "github.com/pkg/errors"
```
chaosaffe:
```suggestion
```
Please open an issue to discuss adding a dependency on `github.com/pkg/errors` and using throughout `cluster-api` and `cluster-api-provider-openstack`

pkg/cloud/openstack/machine/actuator.go
gdoctor: What I actually meant here is that the core `cluster-api` is not propagating changes to the MachineSet to the Machine level. This is probably the wrong spot for the comment, but it's important to test/get this resolved.

gdoctor:
/hold @chaosaffe thanks for the comments. I am new to this whole thing so still trying to figure out your process :) Sounds like I should hold off on my changes, and join office hours to discuss my approach to the machine actuator as a whole going forward.

I have since realized I may have a different ideology towards the actuator since I would rather _update_ mutable fields, and declare other fields as immutable, for example maybe the network uuid. I just don't see a lot of value in letting fields such as that one get edited.

chaosaffe:
@gdoctor I would love to hear your ideas and see how we can incorporate them into the provider! If you don't have a chance to join the meetings probably the best option is to open some issues on the relevant repos so that we can have some conversations with the relevant groups on how we can proceed!

pkg/cloud/openstack/machine/actuator.go
chaosaffe: Changes should be applied from `MachineSet`->`Machine`. This appears to be broken right now as the most recent controller container version is from September, but we should have a build shortly that will hopefully resolve this

PR #116, opened 2018-11-30
Don't hide manager's CLI flags

flaper87:
We were overriding the default flags with an empty list of flags and by parsing an empty string array.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -21,7 +21,6 @@ import (
 	"log"
 
 	"github.com/golang/glog"
-	"github.com/spf13/pflag"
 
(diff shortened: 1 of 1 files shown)


flaper87:
nvm, this is broken upstream
/close

chaosaffe:
The symptoms you are seeing are related to https://github.com/kubernetes-sigs/cluster-api/issues/587#issuecomment-442699904 and https://github.com/kubernetes-sigs/cluster-api/issues/607. The flags are correct, it is issues upstream that are causing the issue
/close

flaper87:
@chaosaffe too late :P

PR #117, opened 2018-11-30
Fixup logs

flaper87:


wking:
Looks like there's an `openshift-master` branch pointing at 29f1ca7.  Is that branch doing anything?  If not, can we delete it (because `dep ensure` in openshift/install#792 WIP pulled that branch in instead of master, which has advanced a lot since 29f1ca7)?

PR #122, opened 2018-12-04
Update how generate-yaml.sh handles inplace sed for replacing image

stevesloka:

On OSX when I attempted to generate the files, the `sed` piece was failing with error but the file did exist on disk. I think the normal `sed -i` workflow should work fine on both Linux & Darwin. 

```
sed: can't read : No such file or directory
```

Signed-off-by: Steve Sloka <steves@heptio.com>

### DIFF
--- a/cmd/clusterctl/examples/openstack/generate-yaml.sh
+++ b/cmd/clusterctl/examples/openstack/generate-yaml.sh
@@ -222,10 +222,8 @@ cat "$PROVIDERCOMPONENT_TEMPLATE_FILE" \
   | sed -e "s/\$MACHINE_CONTROLLER_SSH_PRIVATE/$MACHINE_CONTROLLER_SSH_PRIVATE/" \
   >> "$PROVIDERCOMPONENT_GENERATED_FILE"
 
-if [[ "$OS" =~ "Linux" ]]; then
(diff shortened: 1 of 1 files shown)


chrigl:
What sed are you using? I use the default macOS sed and not gnu sed. For me `generate-yaml.sh` works fine, but `sed -i` fails.

```
$ sed -i "s#image: controller:latest#image: gcr.io/k8s-cluster-api/cluster-api-controller:latest#" tmp.yaml
sed: 1: "tmp.yaml": undefined label 'mp.yaml'

$ sed -i '' -e "s#image: controller:latest#image: gcr.io/k8s-cluster-api/cluster-api-controller:latest#" tmp.yaml
```

stevesloka:
Hey @chrigl yup that was it. I hit issues earlier since I had coreutils installed, but uninstalled and thought I was good.

Would you be open to switching to Kustomize to work around these issues?

chaosaffe:
@stevesloka I think we certainly would be open to getting away from the `bash`isms we currently have. Is that something you would be interested in helping with?
/close

stevesloka:
Yes I think I'd like to help out, let me open an issue and we can work from there. Thanks!

PR #135, opened 2018-12-07
Move to https://github.com/mikefarah/yq

### LINKED ISSUE #134: Move generate-yaml.sh to go version of yq
generate-yaml.sh currently uses the `yq` found at `https://github.com/kislyuk/yq` but the `yq` found at `https://github.com/mikefarah/yq` is more widely available. The script should use the syntax of the latter and detect if the other is installed instead.

ainmosni:

This patch changes the following:

 * Use the syntax of mikefarah's yq.
 * Check if the installed yq is a python script and error out if it is.

Fixes #134


m1093782566:
/lgtm

/approve

THANKS!

PR #136, opened 2018-12-07
Conf from secret

### LINKED ISSUE #130: Allow the actuator to get the openstack configs from a configmap
We currently depend on having `clouds.yaml` in one of the standard paths. In a more k8s fashion, it'd be great if the actuator could get the secret itself (assuming there's a service account token in the pod) and use the data in that secret to create the openstack client.

flaper87:
Allow for the actuator to gather its configs from a secret.

The config parsing logic is as follows:

- Check if the machine has `CloudsSecret` set. If yes, get the secret, load the cloud from it using the value of `CloudName` (default is "openstack"), and use the credentials of this cloud.
- If `CloudSecret` is not set, check for `clouds.yaml` files in the container and proceed using the standard flow.

Some things to note:
- As of now, if the `OS_CLOUD` env variable is set, and there's a `clouds.yaml` file in the container with a cloud corresponding to `OS_CLOUD`, it'll overwrite the one loaded from the secret.
- The current implementation requires setting `CloudsSecret` on every machine definition. Future patches PRs will allow for setting `CloudsSecret` in the cluster spec too.

Fixes #130 

Note to reviewers:

I'm sure there are lots of things that could be improved, changed, etc in this patch. Unfortunately, we're still cleaning up the actuator code and there's only so much I could implement without having to refactor most of it. More patches to come to improve this implementation and the rest of the actuator.

### DIFF
--- a/config/rbac/rbac_role.yaml
+++ b/config/rbac/rbac_role.yaml
@@ -42,3 +42,11 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
--- a/pkg/cloud/openstack/clients/machineservice.go
+++ b/pkg/cloud/openstack/clients/machineservice.go
@@ -20,6 +20,9 @@ import (
 	"encoding/base64"
 	"fmt"
 
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/yaml"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips"
@@ -29,9 +32,13 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/gophercloud/utils/openstack/clientconfig"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
+
+const CloudsSecretKey = "clouds.yaml"
 
 type InstanceService struct {
 	provider       *gophercloud.ProviderClient
@@ -71,12 +78,71 @@ type InstanceListOpts struct {
 	Name string `q:"name"`
 }
 
-func NewInstanceService() (*InstanceService, error) {
-	clientOpts := new(clientconfig.ClientOpts)
-	opts, err := clientconfig.AuthOptions(clientOpts)
+func GetCloudFromSecret(kubeClient kubernetes.Interface, namespace string, secretName string, cloudName string) (clientconfig.Cloud, error) {
+	emptyCloud := clientconfig.Cloud{}
+
+	if secretName == "" {
+		return emptyCloud, nil
+	}
+
+	if secretName != "" && cloudName == "" {
+		return emptyCloud, fmt.Errorf("Secret name set to %v but no cloud was specified. Please set cloud_name in your machine spec.", secretName)
+	}
+
+	secret, err := kubeClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
+	if err != nil {
+		return emptyCloud, err
+	}
+
+	content, ok := secret.Data[CloudsSecretKey]
+	if !ok {
+		return emptyCloud, fmt.Errorf("OpenStack credentials secret %v did not contain key %v",
+			secretName, CloudsSecretKey)
+	}
+	var clouds clientconfig.Clouds
+	err = yaml.Unmarshal(content, &clouds)
+	if err != nil {
+		return emptyCloud, fmt.Errorf("failed to unmarshal clouds credentials stored in secret %v: %v", secretName, err)
+	}
+
+	return clouds.Clouds[cloudName], nil
+}
+
+// TODO: Eventually we'll have a NewInstanceServiceFromCluster too
+func NewInstanceServiceFromMachine(kubeClient kubernetes.Interface, machine *clusterv1.Machine) (*InstanceService, error) {
+	machineSpec, err := openstackconfigv1.MachineSpecFromProviderSpec(machine.Spec.ProviderSpec)
 	if err != nil {
 		return nil, err
 	}
+	cloud, err := GetCloudFromSecret(kubeClient, machine.Namespace, machineSpec.CloudsSecret, machineSpec.CloudName)
+	if err != nil {
+		return nil, err
+	}
+	return NewInstanceServiceFromCloud(cloud)
+}
+
+func NewInstanceService() (*InstanceService, error) {
+	cloud := clientconfig.Cloud{}
+	return NewInstanceServiceFromCloud(cloud)
+}
+
+func NewInstanceServiceFromCloud(cloud clientconfig.Cloud) (*InstanceService, error) {
+	clientOpts := new(clientconfig.ClientOpts)
+	var opts *gophercloud.AuthOptions
+
+	if cloud.AuthInfo != nil {
+		clientOpts.AuthInfo = cloud.AuthInfo
+		clientOpts.AuthType = cloud.AuthType
+		clientOpts.Cloud = cloud.Cloud
+		clientOpts.RegionName = cloud.RegionName
+	}
+
+	opts, err := clientconfig.AuthOptions(clientOpts)
+
+	if err != nil {
+		return nil, err
+	}
+
 	provider, err := openstack.AuthenticatedClient(*opts)
 	if err != nil {
 		return nil, fmt.Errorf("Create providerClient err: %v", err)
--- a/pkg/cloud/openstack/cluster/actuator.go
+++ b/pkg/cloud/openstack/cluster/actuator.go
@@ -6,25 +6,19 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 	providerv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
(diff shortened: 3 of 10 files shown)


flaper87:
@chaosaffe @chrigl r? :)

config/rbac/rbac_role.yaml
```diff
@@ -42,3 +42,11 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
```
chaosaffe: Can we scope this to something less than _all_ secrets in _all_ namespaces?

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -71,12 +76,72 @@ type InstanceListOpts struct {
...
+
+	if cloudName == "" {
+		cloudName = "openstack"
+	}
+
+	if secretName == "" {
+		return emptyCloud, nil
+	}
+
+	// TODO(flaper87): Should we be able to read secrets from a different namespace?
```
chaosaffe: I would vote no
```suggestion
```

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -71,12 +76,72 @@ type InstanceListOpts struct {
...
+	var clouds clientconfig.Clouds
+	err = yaml.Unmarshal(content, &clouds)
+	if err != nil {
+		return emptyCloud, fmt.Errorf("failed to unmarshal clouds credentials stored in secret %v: %v", secretName, err)
+	}
+
+	return clouds.Clouds[cloudName], nil
+}
+
+// TODO: Eventually we'll have a NewInstanceServiceFromCluster too
```
chrigl: +1

config/rbac/rbac_role.yaml
chrigl: Also it is not safe to edit this file, because controller-tools (https://github.com/kubernetes-sigs/cluster-api/blob/master/Makefile#L59) would override this.

flaper87: What would be the right way to scope/implement this?

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -460,3 +428,50 @@ func (oc *OpenstackClient) validateMachine(machine *clusterv1.Machine, config *o
 	// TODO: other validate of openstackCloud
 	return nil
 }
+
+func (oc *OpenstackClient) createSshCreds(machine *clusterv1.Machine) error {
```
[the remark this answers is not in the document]
flaper87: whoops, I didn't mean to commit this function

pkg/cloud/openstack/clients/machineservice.go
flaper87: Agreed, thanks!

config/rbac/rbac_role.yaml
chrigl: Having access to all secrets sounds odd, but I also can't imagine how to solve it differently.

I would like to follow kubebuilders annotations ( `kubebuilder:rbac: ...`, see https://github.com/kubernetes-sigs/cluster-api/blob/299c318690e526b7d8c704ab40a813c00ab7b0b0/pkg/controller/node/node_controller.go#L80) and enable `controller-gen` in the Makefile. For sure, you must migrate the current rbac fules to the annotations as well.

Maybe this is a bit out of scope of this PR, and currently running `controller-gen` doesn't modify the `rbac_role.yaml` (as far as I see). Should we maybe just create an issue to migrate rbacs to kubebuilder?

flaper87: I agree this is out of scope for this PR. I'd rather get this in and improve it in a follow-up patch. I'd like this patch to focus on improving the UX/functionality and we can clean it up asap.

flaper87:
@chaosaffe @chrigl can we get this in? :smiley:

pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -6,25 +6,19 @@ import (
...
-	ClustersGetter client.ClustersGetter
-}
-
 // NewActuator creates a new Actuator
-func NewActuator(params ActuatorParams) (*Actuator, error) {
-	res := &Actuator{
-		clustersGetter: params.ClustersGetter,
-	}
-
+func NewActuator(params osactuators.ActuatorParams) (*Actuator, error) {
```
chaosaffe: This stutters (`actuator.actuator`) 
Why not keep the default import, and it becomes `openstack.ActuatorParams`

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -63,64 +62,20 @@ type SshCreds struct {
...
-func NewActuator(machineClient client.Client, scheme *runtime.Scheme) (*OpenstackClient, error) {
-	machineService, err := clients.NewInstanceService()
-	if err != nil {
-		return nil, err
-	}
-	var sshCred SshCreds
-	b, err := ioutil.ReadFile(SshKeyUserPath)
-	if err != nil {
-		return nil, err
+func NewActuator(params osactuators.ActuatorParams) (*OpenstackClient, error) {
```
chaosaffe: Stutter (`actuators.Actuator`)

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -35,6 +34,7 @@ import (
 	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
 	bootstrap "sigs.k8s.io/cluster-api-provider-openstack/pkg/bootstrap"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
+	osactuators "sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
```
chaosaffe: Duplicate import

pkg/controller/controller.go
```diff
@@ -33,3 +37,20 @@ func AddToManager(m manager.Manager) error {
 
 	return nil
 }
+
+func getActuatorParams(mgr manager.Manager) osactuators.ActuatorParams {
```
chaosaffe: Stutter (`actuators.Actuator`)
```suggestion
func getActuatorParams(mgr manager.Manager) openstack.ActuatorParams {
```

pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -6,25 +6,19 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 	providerv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+	osactuators "sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
```
chaosaffe: Why is this import name overridden?

pkg/controller/controller.go
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 package controller
 
 import (
+	"github.com/golang/glog"
```
chaosaffe:
```suggestion
	"k8s.io/klog"
```

pkg/controller/controller.go
```diff
@@ -33,3 +37,20 @@ func AddToManager(m manager.Manager) error {
 
 	return nil
 }
+
+func getActuatorParams(mgr manager.Manager) osactuators.ActuatorParams {
+	config := mgr.GetConfig()
+
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		glog.Fatalf("Could not create kubernetes client to talk to the apiserver: %v", err)
```
chaosaffe:
```suggestion
		glog.Fatalf("Could not create kubernetes client to talk to the apiserver: %v", err)
```
```suggestion
		klog.Fatalf("Could not create kubernetes client to talk to the apiserver: %v", err)
```

pkg/controller/controller.go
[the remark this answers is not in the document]
flaper87: done, thanks!

pkg/controller/controller.go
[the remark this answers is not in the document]
flaper87: done!

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -71,12 +76,71 @@ type InstanceListOpts struct {
...
+	if secretName == "" {
+		return emptyCloud, nil
+	}
+
+	secret, err := kubeClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
+	if err != nil {
+		return emptyCloud, err
+	}
+
+	content, ok := secret.Data["clouds.yaml"]
```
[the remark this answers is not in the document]
flaper87: done!

pkg/cloud/openstack/machine/actuator.go
flaper87: done!

pkg/controller/controller.go
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 package controller
 
 import (
+	"k8s.io/klog"
+
+	"k8s.io/client-go/kubernetes"
+	osactuators "sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
```
chaosaffe:
```suggestion
  "sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
```

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -71,12 +76,71 @@ type InstanceListOpts struct {
 	Name string `q:"name"`
 }
 
+func GetCloudFromSecret(kubeClient kubernetes.Interface, namespace string, secretName string, cloudName string) (clientconfig.Cloud, error) {
+	emptyCloud := clientconfig.Cloud{}
+
+	if cloudName == "" {
+		cloudName = "openstack"
```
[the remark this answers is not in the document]
chaosaffe: The issue is that if you don't have `openstack` as the name for your cloud and you are having issues because you didn't set the `cloudName`, you won't easily be able to trace or discover that issue. 
If you are completely against erroring here then there needs to be some logging that clearly states `cloudName` was unset and that we are defaulting it to `openstack`

pkg/controller/controller.go
```diff
@@ -33,3 +37,20 @@ func AddToManager(m manager.Manager) error {
...
+
+func getActuatorParams(mgr manager.Manager) osactuators.ActuatorParams {
+	config := mgr.GetConfig()
+
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		klog.Fatalf("Could not create kubernetes client to talk to the apiserver: %v", err)
+	}
+
+	return osactuators.ActuatorParams{
```
chaosaffe:
```suggestion
	return openstack.ActuatorParams{
```

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
flaper87: done

PR #137, opened 2018-12-07
Pass machine config into VM startup script template

YorikSar:

This PR adds `MachineConfig` (`OpenstackProviderConfig` object) of machine to the context of startup script template. This is useful if we need to adjust startup script for some specific cases. In my case, I need to know floating IP of the master node in startup script for it (to work around an issue with networking).

**Release note**:

```release-note
NONE
```


YorikSar:
Updated types and functions naming according to #107.

PR #139, opened 2018-12-10
Add ns into deployer consideration

jichenjc:
when the network created in VXlan mode, the ssh to won't work

when network created in VXLan mode, we still use ssh so that will fail, need ip netns exec xxx 
Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/openstack/deployer.go
+++ b/pkg/cloud/openstack/deployer.go
@@ -81,7 +81,18 @@ func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clu
 		"echo STARTFILE; sudo cat /etc/kubernetes/admin.conf"))
 	parts := strings.Split(result, "STARTFILE")
 	if len(parts) != 2 {
-		return "", nil
+		// can't ssh to the instance, try ip netns again with the first network given
+		result := strings.TrimSpace(util.ExecCommand(
+			"ip", "netns", "exec", fmt.Sprintf("qdhcp-%s", machineSpec.Networks[0].UUID),
+			"ssh", "-i", homeDir+"/.ssh/openstack_tmp",
+			"-o", "StrictHostKeyChecking no",
+			"-o", "UserKnownHostsFile /dev/null",
+			fmt.Sprintf("%s@%s", machineSpec.SshUserName, ip),
+			"echo STARTFILE; sudo cat /etc/kubernetes/admin.conf"))
+		parts := strings.Split(result, "STARTFILE")
+		if len(parts) != 2 {
+			return "", nil
+		}
 	}
 	return strings.TrimSpace(parts[1]), nil
 }


chrigl:
I'm afraid, I don't understand this. For me this looks like `clusterctl` is running on a particular OpenStack Network Node, because `qdhcp-$UUID` is usually a network namespace of OpenStack.
But… there are potentially more than one OpenStack Network Node, also some, where this NS doesn't exist.

Is this a misconception on my end?

jichenjc:
my openstack cloud is tenant based (which is normal and widely use, it's based on vxlan or GRE) and every tenant cloud will be put into namespace so you have to use ns in order to ssh to the new created instance

instead , you can create provider network which is only created by admin and you can use either use flat or vlan mode and you can of course ssh directly

so the deployer will fail if I am usnig tenant tenant work because ssh won't work here

jichenjc:
the network node you are talking about might be related to 'qrouter' which is L3 layer concept, you can ssh to the instance even you don't have a router ...

chrigl:
I think my assumption was correct. If you really want to use `clusterctl` to be able to reach the internal IP, just run `clusterctl` from within the network namespace.

But `cluster-api` has a different design. The idea is, that you can run `clusterctl` on your local computer. Except for getting the `admin.conf` (see https://github.com/kubernetes-sigs/cluster-api/issues/122), nothing depends on ssh. The controller of `cluster-api-provider-openstack` is responsible for talking to the OpenStack API and to create VMs. To be able to reach the created Kubernetes Control Plane, you need to generate a config, and configure it in an appropriate way.
```
$ cd cmd/clusterctl/examples/openstack
$ export OS_CLOUD=openstack
$ ./generate-yaml.sh -c ~/.config/cluster-api/clouds.yaml --provider-os ubuntu
Done generating /Users/chris/Projects/kube-gopath/src/sigs.k8s.io/cluster-api-provider-openstack/cmd/clusterctl/examples/openstack/ubuntu//out/provider-components.yaml /Users/chris/Projects/kube-gopath/src/sigs.k8s.io/cluster-api-provider-openstack/cmd/clusterctl/examples/openstack/ubuntu//out/machines.yaml /Users/chris/Projects/kube-gopath/src/sigs.k8s.io/cluster-api-provider-openstack/cmd/clusterctl/examples/openstack/ubuntu//out/cluster.yaml
You should now manually change your cluster configuration by editing the generated files.
``` 

Now you need to configure the `machines.yaml` and `provider-components.yaml`. In `machines.yaml` you especially need to change the VM-Images, the network.uuid (this is your internal network), and the floatingIP. The floatingIP of the master must be one, you already created in OpenStack, but didn't use. The worker nodes don't need floatingIPs. In the current status, you should add this floatingIP to `provider-components.yaml` as well as `controlPlaneEndpoint`:
```
        apiVersion: kubeadm.k8s.io/v1alpha3
        kind: ClusterConfiguration
        kubernetesVersion: v${CONTROL_PLANE_VERSION}
        networking:
          serviceSubnet: ${SERVICE_CIDR}
        clusterName: kubernetes
        controlPlaneEndpoint: "129.114.111.153" ### <===
```

jichenjc:
ok, you made a good point here... I didn't realize that we can do that through floating ip...
I may try this and get back here, probably add some document about this usage if success to build

jichenjc:
pkg/cloud/openstack/machine/actuator.go:getIPFromInstance
confirm the floating ip settings will be returned first

PR #140, opened 2018-12-10
Check jq first

jichenjc:
if jq not exist, it will complain after all cloud param input complete
this is not efficient, so check before generate cloud configuration files.

avoid find issue after input all param

Fixes #

**Release note**:
```release-note
NONE
```


chaosaffe:
Thanks for your patch, but the issue was resolved in a different way.
The newly referenced `yq` is a go binary without a dependency on `jq`, in addition to being portable
xref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/135
/close

PR #151, opened 2018-12-17
Add boot from volume support

jichenjc:
Support boot from volume when RootVolume openstack providerSpec

honor the RootVolume settings (RootVolume)

so following spec can work by requesting cinder volumes

items:
- apiVersion: "cluster.k8s.io/v1alpha1"
  kind: Machine
  metadata:
    generateName: openstack-master-
    labels:
      set: master
  spec:
    providerSpec:
     ...........
        root_volume:
          diskSize: 5

Fixes #

```release-note
Support boot from volume in openstack settings in some scenario that no local storage available
```


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -201,16 +202,50 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
 		SecurityGroups:   config.SecurityGroups,
 		ServiceClient:    is.computeClient,
 	}
+
+	if config.RootVolume.Size > 0 {
+		var blocks []bootfromvolume.BlockDevice
+		block := bootfromvolume.BlockDevice{
+			SourceType:          bootfromvolume.SourceImage,
+			BootIndex:           0,
+			UUID:                config.Image,
```
chrigl: This only works if `config.Image` is specified as UUID, while we allow the Image to be the name.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -201,16 +202,50 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
+			DeleteOnTermination: true,
+			DestinationType:     bootfromvolume.DestinationVolume,
+			VolumeSize:          config.RootVolume.Size,
+			DeviceType:          config.RootVolume.VolumeType,
+		}
+		blocks = append(blocks, block)
+
+		createOptswithbfv := bootfromvolume.CreateOptsExt{
```
chrigl: How about using `createOpts` as `var createOpts servers.CreateOptsBuilder`. This would save us the second place to call `servers.Create`
```
func (is *InstanceService) InstanceCreate(name string, [...]) {
        var createOpts servers.CreateOptsBuilder
        [...]


    	createOpts = servers.CreateOpts{
		Name:             name,
		ImageName:        image,
		FlavorName:       config.Flavor,
		AvailabilityZone: config.AvailabilityZone,
		Networks:         nets,
		UserData:         []byte(userData),
		SecurityGroups:   config.SecurityGroups,
		ServiceClient:    is.computeClient,
	}

        if config.RootVolume.Size > 0 {
                [...]

                 createOpts = bootfromvolume.CreateOptsExt{
			CreateOptsBuilder: createOpts,
			BlockDevice:       blocks,
		}
	}

	server, err := servers.Create(is.computeClient, keypairs.CreateOptsExt{
		CreateOptsBuilder: createOpts,
		KeyName:           keyName,
	}).Extract()
	if err != nil {
		return nil, fmt.Errorf("Create new server err: %v", err)
	}
	return serverToInstance(server), nil
}
```

pkg/cloud/openstack/clients/machineservice.go
jichenjc: exactly.. this was found during my local test as well
I will figure out how to translate it ... or maybe a doc indication?

pkg/cloud/openstack/clients/machineservice.go
jichenjc: ok, will do

jichenjc:
I seems didn't set volume type , will double check this 

and for the make generate, seems no impact... 

anyway, will test again

chrigl:
Sorry for bugging you so hard on this. I think I have a better idea of the problem now.

My assumption was that: `volumeType` related to OpenStack source_type. https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/bootfromvolume/requests.go#L47
For me that would make sense, because one might want to create a VM out of a snapshot.

I would like to rename the field `volumeType` field to `SourceType` (`json:"sourceType"`). As a consequence, you could remove the `getImageIDsByFilter` functionality, because now we would need a function to get the volume uuid, image uuid or snapshot uuid. I would prefer to also have a field `SourceUUID` on `RootVolume` and ignoring `config.ImageName`. I would definitely extend the comment on `ImageName`, that this will be ignored if `RootVolume` is in use.

jichenjc:
ok, that make sense , I will add a UUID field first then try to use it ,and yes , the source type should be added

config/crd/openstackproviderconfig_v1alpha1.yaml
```diff
@@ -37,7 +37,7 @@ spec:
           items:
             type: string
           type: array
-        root_volume:
+        rootVolume:
```
chrigl: Could you change and add the fields here as well?

In the future, I want us to finally migrate to `kubebuilder` autogeneration of this file. But we are not there yet.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -38,6 +39,7 @@ type InstanceService struct {
 	computeClient  *gophercloud.ServiceClient
 	identityClient *gophercloud.ServiceClient
 	networkClient  *gophercloud.ServiceClient
+	imageClient    *gophercloud.ServiceClient
```
chrigl: Not needed any more.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -100,12 +102,19 @@ func NewInstanceService() (*InstanceService, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Create networkingClient err: %v", err)
 	}
+	imageClient, err := openstack.NewImageServiceV2(provider, gophercloud.EndpointOpts{
```
chrigl: Same like above, not needed any more.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -100,12 +102,19 @@ func NewInstanceService() (*InstanceService, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Create imageClient err: %v", err)
+	}
 
 	return &InstanceService{
 		provider:       provider,
 		identityClient: identityClient,
 		computeClient:  serverClient,
 		networkClient:  networkingClient,
+		imageClient:    imageClient,
```
chrigl: Same.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -201,16 +202,44 @@ func (is *InstanceService) InstanceCreate(name string, config *openstackconfigv1
...
+		ImageName:        image,
 		FlavorName:       config.Flavor,
 		AvailabilityZone: config.AvailabilityZone,
 		Networks:         nets,
 		UserData:         []byte(userData),
 		SecurityGroups:   config.SecurityGroups,
 		ServiceClient:    is.computeClient,
 	}
+
+	if config.RootVolume.Size > 0 {
```
chaosaffe: I would prefer either a check to ensure that the `SourceUUID` is in the struct rather than checking the size:
```
	if config.RootVolume.SourceUUID != "" {
```

We could also check the struct as well:
```
	if config.RootVolume != nil {
```

chaosaffe:
Please update the documentation, including the applicable documentation and examples so they match the new feature

jichenjc:
for the sample file, I made a minor update

the applicable document might need be updated after the floating ip (config.md) PR merged.... otherwise it will be conflict ..

chaosaffe:
If there is a conflict it can be resolved at that time. Please ensure that
you enable maintainers to edit your PR so that we can resolve merge
conflicts on review 😊

pkg/cloud/openstack/clients/machineservice.go
jichenjc: ok, missed this :(, will update to config.RootVolume != nil {

jichenjc:
@chaosaffe can you take a look again ? Thanks

chrigl:
For me this looks good. @jichenjc could you rebase in order to get the crd -> crds stuff in, and run make generate. Thanks!

chaosaffe:
Do the changes with #214 impact this PR?

jichenjc:
did a quick rebase just now and re-generated the crds ..

docs/config.md
```diff
@@ -137,4 +137,28 @@ securityGroups:
...
+   kind: Machine
+   ...
+   spec:
+     providerSpec:
+       value:
+        ...
+        rootVolume:
+          diskSize: 0
+          sourceType: ""
+          SourceUUID: ""
```
shiftky: I think it would be better to unify the style to camelCase. (sourceUUID)

jichenjc:
@chrigl can you help to review this again? several rebase action already...thanks

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -408,8 +410,30 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 		SecurityGroups:   config.SecurityGroups,
 		ServiceClient:    is.computeClient,
 	}
+
+	// If the root volume Size is not 0, means boot from volume
+	if config.RootVolume.Size != 0 {
```
flaper87: Should this check be `if config.RootVolume != nil && config.RootVolume.Size != 0` ?

I believe, as is, this will fail if `rootVolume` is omitted from the `MachineSpec`

jichenjc: yes, you are right

jichenjc:
@chrigl @flaper87 can you help to take a look at this? stay for a while now...

chrigl:
/lgtm

@flaper87 Could you approve when this is ok form you point of view?

jichenjc:
@chaosaffe maybe you can take a look and see whether it's ok to merge? you add a change request before :)

chaosaffe:
Sorry about the hang on this one @jichenjc

PR #152, opened 2018-12-17
WIP Configure Calico default ippool to POD_CIDR

chrigl:

When changing the PodCIDR in the cluster spec, kubernetes will happily
use this, but networking won't work. This is because the default calico
deployment hard coded 192.168.0.0/24 in the default ip pool.

Decided to patch the created ippool over using sed on calico.yaml,
because of sed.

Fixes #

**Release note**:

```release-note

```


flaper87:
@chrigl nees rebase :(

chrigl:
/close

This is nothing I would get back to. Probably we just give a pointer in our documentation who to change the default-pool.

Also deploying the CNI plugin can be done by addons in favour of doing it in the setup script.

https://docs.projectcalico.org/v3.5/usage/changing-ip-pools

PR #153, opened 2018-12-17
Kustomize

### LINKED ISSUE #124: Use Kustomize in generate-yaml.sh
Currently the generate-yaml.sh script utilizes bash and sed to handle building out the yaml files to pass to clusterctl. There can be environmental issues (mac vs linux, tooling version) which makes that script flakey. 

I'd like to propose we move to Kustomize for the generation and reduce this need on bash. I'm happy to take this on and send a PR.

iamemilio:
**What this PR does / why we need it**: Converts everything to use kustomize in place of the bash script

Fixes #124 

**Special notes for your reviewer**: Still a wip, and not rebased against upstream yet. Check that the bash script changes check out. I made quite a few changes. I also resturctured a lot of the files to be more consistent with how kustomize should be set up. Readme updates yet to come..

**Release note**: WIP


stevesloka:
I had started some work on this (#124), but feel free to finish it up! =)

flaper87:
@chaosaffe you mentioned you may be interested in taking this on. Will you? Or should we try to find someone that can knock this off?

iamemilio:
@flaper87 me and imain are picking it back up

chrigl:
@iamemilio Because we are also discussing about to have real reference implementations, and not just examples (see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/159#issuecomment-454040715), and I think the `kustomize`-version of `generate-yaml.sh` will be almost written from scratch, I would start with everything in a separate folder. This way, you don't have to deal with conflicts. We can stick with the "old" version in our docs and learn how the `kustomize`-version works. When we are confident, we can change the docs (or move the folder around). What do you think?

iamemilio:
@chrigl There arent any conflicts from doing it this way that I see. That being said, I think the core difference is that Kustomize will expect you to have all the configs, like clouds.yaml and cloud.conf already generated and ready to go, while our current implementation does not, which is why I coupled them together. I think it would actually make sense to keep parts of the generate.yaml script around, specifically the parts that create a cloud.yaml and cloud.conf. Then we can seperately call the kustomize script. How does that architecture sound?

chrigl:
@iamemilio What ever works best for you :)

iamemilio:
@chrigl There aren't any conflicts from doing it this way that I see. That being said, I think the core difference is that Kustomize will expect you to have all the configs, like clouds.yaml and cloud.conf already generated and ready to go, while our current implementation does not, which is why I coupled them together. I think it would actually make sense to keep parts of the generate.yaml script around, specifically the parts that create a cloud.yaml and cloud.conf. Then we can seperately call the kustomize script. How does that architecture sound?

iamemilio:
I can squash

cmd/clusterctl/examples/openstack/generate-yaml.sh
```diff
@@ -197,91 +157,31 @@ tenant-id=\"$PROJECT_ID\"
...
-echo "Done generating $PROVIDERCOMPONENT_GENERATED_FILE $MACHINE_GENERATED_FILE $CLUSTER_GENERATED_FILE"
-echo "You should now manually change your cluster configuration by editing the generated files."
+cat "$MASTER_USER_DATA" \
+    | sed -e "s#\$OPENSTACK_CLOUD_PROVIDER_CONF#$OPENSTACK_CLOUD_PROVIDER_CONF#" \
+    > $USERDATA/$PROVIDER_OS/master-user-data.sh
+cat "$WORKER_USER_DATA" \
+  | sed -e "s#\$OPENSTACK_CLOUD_PROVIDER_CONF#$OPENSTACK_CLOUD_PROVIDER_CONF#" \
+  > $USERDATA/$PROVIDER_OS/worker-user-data.sh
+
+echo $CLOUD > $CONFIG_DIR/os_cloud.txt
```
chrigl: Could you please use `echo -n "$CLOUD"` here.. And maybe even  `echo -n "$CLOUD" | tr -d '\n'`, I guess you know why ;)

iamemilio: how about printf instead

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -258,15 +258,6 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 	if config == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace")
 	}
-	if config.Trunk == true {
```
chrigl: Why is this commit removing this code? /me kind of confused

iamemilio: must be another rebase bug. Sorry

PR #165, opened 2018-12-27
Disable ufw on ubuntu #149

### LINKED ISSUE #149: Disable ufw on ubuntu
We faced an interference with ufw running on ubuntu 18.04.
IMHO we should disable ufw by default, no matter the used Ubuntu version is 16.04 or 18.04.

On both versions, this in the startup script of master and node should do the trick
```
ufw disable
```

syjabri:
**What this PR does / why we need it**:  Disable ufw on ubuntu 

We don't have access from inside the pod because ufw block the traffic. 

We need to allow external clients such as kubectl and kubelet to access the Kubernets API, so we should disable the firewall or create new iptables rules. 

Fixes #149

### DIFF
--- a/cmd/clusterctl/examples/openstack/ubuntu/provider-components.yaml.template
+++ b/cmd/clusterctl/examples/openstack/ubuntu/provider-components.yaml.template
@@ -253,7 +253,6 @@ data:
         # Reset iptables config  
         cat > /etc/systemd/system/docker.service.d/10-iptables.conf <<EOF
         [Service]
-        # Yes, we need to clear it, because otherwise it is addative!
         ExecStart=
         ExecStart=/usr/bin/dockerd -H fd:// --iptables=false --ip-masq=false
         EOF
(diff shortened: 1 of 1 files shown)


cmd/clusterctl/examples/openstack/ubuntu/provider-components.yaml.template
```diff
@@ -249,6 +249,17 @@ data:
...
+        
+        # Reset iptables config  
+        cat > /etc/systemd/system/docker.service.d/iptables.conf <<EOF
+        [Service]
+        # Yes, we need to clear it, because otherwise it is addative!
+        ExecStart=
+        ExecStart=/usr/bin/dockerd -H fd:// --iptables=false --ip-masq=false
+        EOF
+        systemctl daemon-reload
+        systemctl disable ufw
```
chaosaffe: Please add some whitespace above and a comment referencing this PR/Issue  as to why we disable `ufw`

syjabri: We don't have access from inside the pod because ufw block the traffic.

We need to allow external clients such as kubectl and kubelet to access the Kubernets API, so we should disable the firewall or create new iptables rules.

Issue #166, opened 2019-01-02
Incorrect debug information after template generating


naumvd95:
**Steps to reproduce**

1. Download repo
2. Generate templates from openstack examples
`` env OS_CLOUD=cloud ./generate-yaml.sh --provider-os .....``

**expected result**
Files generated, paths descriptions viewed

**actual result**

incorrect path description. IMO in future it may corrupt scripting arch.
logs:
```
Done generating /Users/vladislavnaumov/go-projects-space/src/sigs.k8s.io/cluster-api-provider-openstack/cmd/clusterctl/examples/openstack/here-is-bad-slash//out/provider-components.yaml
```

PR #167, opened 2019-01-02
Drop unnecessary slash in TEMPLATE_PATH.

russellb:
Issue #166 reported an unnecessary extra slash at the end of the
default value of TEMPLATE_PATH in this script.  This removes it.


naumvd95:
I also tried to prepare PR, but stucked on CNF auth =(
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/168
so now we have 2 PR's what we should do with it?

russellb:
I'll just close mine and you can proceed with #168

PR #172, opened 2019-01-07
Remove MasterEndpoint string in favour of a function

### LINKED ISSUE #163: Don't fail if the apiEndpoint is not set
We currently depend on the `ApiEndpoint` to be set in the `Cluster` resource to be able to generate the setup script. Sadly, this assumes there's a `setupScript` for the machine and that the script requires the `ApiEndpoint` to work.

This is limiting for scenarios where the node does a self-bootstrap and discovery or that the user data passed to the node already has all the required configs. We should not depend on this piece of data for the bootstrap process.

Current code: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/machine/machineScript.go#L84

```go
	params := setupParams{
		Token:          token,
		Cluster:        cluster,
		Machine:        machine,
		MachineSpec:    machineSpec,
		PodCIDR:        getSubnet(cluster.Spec.ClusterNetwork.Pods),
		ServiceCIDR:    getSubnet(cluster.Spec.ClusterNetwork.Services),
		MasterEndpoint: getEndpoint(cluster.Status.APIEndpoints[0]),
}
```

tomassedovic:

Some startup scripts do not require the endpoint to be available (because they've got other means of bootstrapping for example) and calling it right now would fail unconditionally.
    
Instead, this introduces a `GetMasterEndpoint` function that can be called from the template via `"{{ call .GetMasterEndpoint }}` to query the endpoint. Scripts that don't need it will not call this function.

Fixes #163

**Release note**:

```release-note
The node startup script will now need to `{{ call .GetMasterEndpoint }}` in the template
to get the value, rather than querying it with `{{ .MasterEndpoint }}`. This lets the scripts
that don't need it be robust against it failing.
```

### DIFF
--- a/pkg/cloud/openstack/machine/machineScript.go
+++ b/pkg/cloud/openstack/machine/machineScript.go
@@ -33,9 +33,9 @@ type setupParams struct {
 	Machine     *clusterv1.Machine
 	MachineSpec *openstackconfigv1.OpenstackProviderSpec
 
-	PodCIDR        string
-	ServiceCIDR    string
-	MasterEndpoint string
+	PodCIDR           string
+	ServiceCIDR       string
+	GetMasterEndpoint func() (string, error)
 }
 
 func init() {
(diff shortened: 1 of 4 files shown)


tomassedovic:
@flaper87 this addresses an issue you've created, would appreciate if you had a look.

This just sets `MasterEndpoint` to an empty string. Another option is to not even run the templating code and just return an empty string in that case. Would you prefer that?

flaper87:
> @flaper87 this addresses an issue you've created, would appreciate if you had a look.
> 
> This just sets `MasterEndpoint` to an empty string. Another option is to not even run the templating code and just return an empty string in that case. Would you prefer that?

As I was reviewing this, it occurred to me that we could also move this into a function and pass that function to the templates. Let the template call it and fail only when the template tries to access it.

pkg/cloud/openstack/machine/machineScript.go
```diff
@@ -81,7 +81,7 @@ func nodeStartupScript(cluster *clusterv1.Cluster, machine *clusterv1.Machine, t
 		MachineSpec:    machineSpec,
 		PodCIDR:        getSubnet(cluster.Spec.ClusterNetwork.Pods),
 		ServiceCIDR:    getSubnet(cluster.Spec.ClusterNetwork.Services),
-		MasterEndpoint: getEndpoint(cluster.Status.APIEndpoints[0]),
+		MasterEndpoint: masterEndpoint,
```
flaper87: We can create a `GetMasterEndpoint` function, pass it to the template and then have the templates under [`examples/openstack`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/master/cmd/clusterctl/examples/openstack) access it with `{{ .GetMasterEndpoint }}`

flaper87:
Also, passing a function is a good solution for now. Eventually, we might not want to pass any of this to the templates.

tomassedovic:
@flaper87 I've pushed a new version that returns a closure instead.

It still returns an empty string when there are no endpoints available though. Do you want me to change it back to returning an error instead? Looking at the `text/template` docs, functions should be able to return a `value, error` tuple in the templates:

https://golang.org/pkg/text/template/

flaper87:
@tomassedovic I'm a bit torn on the empty string vs error question. A template that calls this function propably expects to receive a value. Receiving an empty string means there will have to be some logic in the template to handle the case, which is not ideal. We should try to not push logic into the templates. I'd say that the function should just error out.

tomassedovic:
Yep, makes sense to me. I'll make the change.

flaper87:
@tomassedovic one more thing. Could you also modify the examples under https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/master/cmd/clusterctl/examples/openstack that use the `MasterEndpoint` so that they will call the new closure instead?

tomassedovic:
Yeah, sorry! I meant to do it and then forgot. Thanks for the reminder!

PR #173, opened 2019-01-08
Get user-data/startupScripts from secrets

### LINKED ISSUE #158: Pass user data to machine in the form of secret
As of now, the userData we pass to the machine corresponds to the startUp script. However, it should be possible to override/pass the userData to the machine config in the form of secret. For example:

```
items:
- apiVersion: "cluster.k8s.io/v1alpha1"
  kind: Machine
  metadata:
    generateName: openstack-master-
    labels:
      set: master
  spec:
    providerSpec:
      value:
        apiVersion: "openstackproviderconfig/v1alpha1"
        kind: "OpenstackProviderSpec"
        flavor: m1.medium
        image: ...
        sshUserName: ...
        availabilityZone: nova
        networks:
        - uuid: ...
        floatingIP: 129.114.111.153
        securityGroups:
        - default
        userDataSecret: $secret-name
    versions:
      kubelet: 1.12.3
      controlPlane: 1.12.3
```
This way we can do away with serializing and mounting the startup script into the CAPO pod.

flaper87:
This commit adds the ability to pass custom userData/startupScripts to
the machines using kubernetes secrets. The previous workflow that used
the machine-setup ConfigMap has been removed in favor of the new one.

Fixes #158

```release-note
Startup scripts are now passed using kubernetes secrets. Create a secret in the Machine's namespace and set the `userDataSecret` key in the machine definition.
```


pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -59,6 +59,9 @@ type OpenstackProviderSpec struct {
 	// The names of the security groups to assign to the instance
 	SecurityGroups []string `json:"securityGroups,omitempty"`
 
+	// The name of the secret containing the user data (startup script in most cases)
+	UserDataSecret string `json:"userDataSecret,omitempty"`
```
chrigl: Should we use a `SecretReference` of `k8s.io/api/core/v1` here? This would in the future give us the flexibility of defining the user-data per namespace. So in theory each cluster can have its own userData.

cmd/clusterctl/examples/openstack/ubuntu/machines.yaml.template
```diff
@@ -19,6 +19,7 @@ items:
         floatingIP: 129.114.111.153
         securityGroups:
         - default
+        userDataSecret: master-user-data
```
chrigl: Might change if you decide to change type of userDataSecret to `SecretReference`. See other comment.

cmd/clusterctl/examples/openstack/ubuntu/machines.yaml.template
```diff
@@ -42,5 +43,6 @@ items:
         floatingIP: 129.114.111.153
         securityGroups:
         - default
+        userDataSecret: worker-user-data
```
chrigl: Might change if you decide to change type of userDataSecret to `SecretReference`. See other comment.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -121,21 +112,23 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
+		if err != nil {
+			return err
+		}
+
+		userData, ok = userDataSecret.Data[UserDataKey]
+		if !ok {
+			return fmt.Errorf("Machine's userdata secret %v did not contain key %v", providerSpec.UserDataSecret, UserDataKey)
+		}
 	}
+
```
chrigl: Would it make sense to error out here if no `UserDataSecret` is provided? I think rendering userdata does not make much sense without script data.

cmd/clusterctl/examples/openstack/centos/provider-components.yaml.template
```diff
@@ -19,204 +19,19 @@ data:
...
+  namespace: openstack-provider-system
+data:
+  userData: $WORKER_USER_DATA
+---
+apiVersion: v1
+kind: Secret
+type: Opaque
 metadata:
-  name: machine-setup
+  name: master-user-data
```
chrigl: This comment doesn't make much sense here... but you forgot to edit `examples/openstack/centos/machines.yaml.template` and add `userDataSecret` (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7ff2fcb1c095db35d2701a1fa12658869a249633/cmd/clusterctl/examples/openstack/centos/machines.yaml.template)

pkg/cloud/openstack/machine/actuator.go
flaper87: I'd prefer not to error but I agree we shouldn't render anything if userData is empty. One may decide to just not pass any userData and it should be a valid use case. I'll add a check in the if below. thanks

pkg/apis/openstackproviderconfig/v1alpha1/types.go
flaper87: I like the idea of using SecretReference. If `Namespace` is not set, we should default to the `machine.Namespace`, tho. thoughts?

cmd/clusterctl/examples/openstack/centos/machines.yaml.template
```diff
@@ -19,6 +19,8 @@ items:
         floatingIP: 129.114.111.153
         securityGroups:
         - default
+        userDataSecret:
+          name: master-user-data
```
chrigl: Just one more little complaint... Could you add the namespace in all the yaml files for now. Because you generate the secrets in `openstack-provider-system` rather than `machine.namespace`.

flaper87: done, good catch!

Issue #175, opened 2019-01-08
Don't mount ssh keys on the actuator POD


flaper87:
We shouldn't be uploading sshkeys into kubernetes and mounting them in the actuator POD. These keys are currently used to get the kubeconfig files from the master nodes (#174). Once #174 is solved, we should stop depending on these keys being present in the container and we should also not inject them into the machine nodes. Let the user specify what sshkeys should be injected/used.

Issue #174, opened 2019-01-08
Generate kubeconfig rather than getting it from the master nodes


flaper87:
We're currently using ssh to get the kubeconfig from the master nodes. We should find a better way to do it that doesn't depend on ssh and on having access to the master nodes' network.

```go
func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clusterv1.Machine) (string, error) {
	ip, err := d.GetIP(cluster, master)
	if err != nil {
		return "", err
	}

	homeDir, ok := os.LookupEnv("HOME")
	if !ok {
		return "", fmt.Errorf("unable to use HOME environment variable to find SSH key: %v", err)
	}

	machineSpec, err := openstackconfigv1.MachineSpecFromProviderSpec(master.Spec.ProviderSpec)
	if err != nil {
		return "", err
	}

	result := strings.TrimSpace(util.ExecCommand(
		"ssh", "-i", homeDir+"/.ssh/openstack_tmp",
		"-o", "StrictHostKeyChecking no",
		"-o", "UserKnownHostsFile /dev/null",
		fmt.Sprintf("%s@%s", machineSpec.SshUserName, ip),
		"echo STARTFILE; sudo cat /etc/kubernetes/admin.conf"))
	parts := strings.Split(result, "STARTFILE")
	if len(parts) != 2 {
		return "", nil
	}
	return strings.TrimSpace(parts[1]), nil
}
```

chrigl:
There are already efforts in https://github.com/kubernetes-sigs/cluster-api/issues/122 to get rid of SSH... while the initial thought was to postpone cert generation to the actual node-setup.

chrigl:
Just thinking out loud... We talked about this could probably work. My idea is:
In `clusterctl` create a `remote-bootstrap-secret.yaml` in form of a bootstrap-token with an additional group (see below). This should be only valid for lets say 20 minutes. The yaml data is passed as a shared secret within the userdata. But because it is only valid for a short period of time, we in the first place I would accept the fact, that it in theory could be intercepted. Even though unlikely.

Jump to the master node. Bootstrapping as usual with `kubeadm init`. After bootstrap creating a secret containing an `admin.conf`. Also creating the shipped `remove-bootstrap-secret`, and adding a Role+RoleBinding to give access to the `admin.conf`-secret.

[...]

PR #176, opened 2019-01-09
Don't fail on missing sshkeys

flaper87:
Not all the scenarios supported by CAPO require the manager to pull the kubeconfig configs from the nodes. Rather than failing due to missing ssh keys, we should just log the inability to pull the config file and continue with the execution.
    
This is an interim solution until #174 is resolved. Once #174 is fixed, we'll be able to stop using sshkeys for this entirely.

Related to #175

```release-note
Missing sshkeys won't prevent the actuator's manager from starting
```

### DIFF
--- a/pkg/cloud/openstack/machine/actuator.go
+++ b/pkg/cloud/openstack/machine/actuator.go
@@ -68,14 +68,6 @@ type OpenstackClient struct {
 }
 
 func NewActuator(params openstack.ActuatorParams) (*OpenstackClient, error) {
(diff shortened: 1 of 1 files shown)


flaper87:
(note that this commit is based on top of this PR: #173)

chrigl:
Looks ok to me. What does @chaosaffe think?

chaosaffe:
Why _don't_ we want to fail if we can't get the SSH key? We need it to get the kubeconfig so we can do the pivot

PR #178, opened 2019-01-10
Tagging of created network resources

### LINKED ISSUE #171: Tag the OpenStack resources created by CAPO
#150 added network, subnet, and router creation as part of the cluster reconciler. We'll likely be adding logic to create other resources like SecurityGroups (#170). We should tag all these resources in order to ease their management not only for the provider itself but also for operators.

Support for Neutron tags was recently added: https://github.com/gophercloud/gophercloud/pull/1259

chrigl:

Tagged resouces networks, subnet and router with
`cluster-api-provider-openstack` and the cluster name in form of
`<namespace>/<name>`.

Fixes #171

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/openstack/clients/networkservice.go
+++ b/pkg/cloud/openstack/clients/networkservice.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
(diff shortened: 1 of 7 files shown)


pkg/cloud/openstack/clients/networkservice.go
```diff
@@ -115,13 +116,22 @@ func (s *NetworkService) reconcileNetwork(networkName string, desired openstackc
 		return emptyNetwork, err
 	}
 
+	_, err = attributestags.ReplaceAll(s.client, "networks", network.ID, attributestags.ReplaceAllOpts{
+		Tags: []string{
+			"cluster-api-provider-openstack",
+			clusterName,
+		}}).Extract()
+	if err != nil {
+		return emptyNetwork, err
```
flaper87: I wonder if we should return the error or just ignore it. By returning it we're requiring OpenStack deployments to have this endpoint available (we'll be requiring newer versions of OpenStack).

FWIW, I prefer this but I wanted to make sure we're on the same page here.

chaosaffe:
Thanks @flaper87

PR #182, opened 2019-01-11
Don't require kubelet version to be set

flaper87:
In an isolated scenario where only the actuator is being used (no clusterctl or any of the CAPO scripts) the kubelet version may not be required. Pre-built images may come with a pre-installed kubelet or versions could be provided through different means.

This PR removes the check in favor of letting the templates manage/verify they have all the data they need.


chaosaffe:
It appears this code is a vestigial tail from when we were an aggregated API, now we are using the kube-apiserver's validation mechanisms and defer the check up

ref: https://github.com/kubernetes-sigs/cluster-api/blob/5a27d984f9af41aa4f952e7c36a11bfbfb042b83/config/crds/cluster_v1alpha1_machine.yaml#L54

/lgtm

PR #192, opened 2019-01-17
Remove need to mount ssh keys in the manager container

flaper87:
This pull request contains 3 commits that aim to clean the code around the ssh keys management and requirement in the machine actuator's code.

The first commit removes the `GetKubeConfig` function from the actuator as it's expected to exist only in the deployer's definition. The second commit removes the need to mount `sshkeys` on the manager's pod. The third commit removes the code related to managing keypairs from the machine service as those keypairs are expected to be managed by operators themselves.


ainmosni:
/lgtm

Less code is good code. :)

PR #197, opened 2019-01-21
Fix default cloud usage when cloudsSecret is absent

YorikSar:

When cloudsSecret is not specified in machine's provider config,currently provider segfaults because CloudsSecret is nil and we try to access Namespace field from it. Add a guard to ignore secret if it is nil or name is empty.

**Release note**:

```release-note
NONE
```


chrigl:
/lgtm

Works on my cluster

PR #198, opened 2019-01-21
Use # as separator for sed s command

YorikSar:

base64 output can contain `/` which breaks sed as it considers `/` to be end of command.

Also remove remnants of SSH key substitution in templates.

**Release note**:

```release-note
NONE
```


jichenjc:
looks good

chrigl:
/lgtm

Works. Thanks @YorikSar for all your great work!

PR #203, opened 2019-01-24
Add managed security groups to provider.

### LINKED ISSUE #170: Implement creation of SecurityGroups in cluster actuator
With #150 in, we have a basic reconciler to create Network, Subnet and Router.

We need to extend this to create appropriate SecurityGroups for control plane and worker nodes.

This ticket must be filled with needed Rules, but I already created it as a reminder for myself.

ainmosni:
This adds the option to have the provider manager OpenStack security
groups. If the `ManagedSecurityGroup` property in the providerSpec is
set to `true`, two security groups will be created, one specifically for
the controlplane machine and one for all machines. These security groups
are hardcoded for the current Cluster-API setup. The reason for this is
that this is a temporary solution that should be replaced once the
Cluster-API group decides on how they want to handle infrastructure.

Fixes #170 

**Release note**:
```release-note
Create the option to have SecurityGroups created by the provider. If providerSpec.ManagedSecurityGroups is set to `true`, SecurityGroups will be created. Note that these will not be deleted or automatically assigned to machines yet.
```


flaper87:
I'd like to take a good look at this

/hold

pkg/cloud/openstack/clients/secgroupservice.go
```diff
@@ -0,0 +1,358 @@
...
+func (s *SecGroupService) reconcileGroup(desired, observed *openstackconfigv1.SecurityGroup) (*openstackconfigv1.SecurityGroup, error) {
+	klog.V(6).Infof("Deleting all rules for group %s", observed.Name)
+	for _, rule := range observed.Rules {
+		klog.V(6).Infof("Deleting rule %s from group %s", rule.ID, observed.Name)
+		err := rules.Delete(s.client, rule.ID).ExtractErr()
+		if err != nil {
+			return &openstackconfigv1.SecurityGroup{}, err
+		}
+	}
+	recreatedRules := make([]openstackconfigv1.SecurityGroupRule, len(desired.Rules), len(desired.Rules))
```
chrigl: `make([]..., 3, 3)` is redundant. `make([]..., 3)`. Both create a slice with `cap = len = 3`. Which means, if you do `append`, you will increase the slice, and not fill from zero.

As the code looks like, I think what you wanted to make is `make([...], 0, len(x))`.
There are some more places with the same pattern in this file. See https://play.golang.org/p/dnmg_tce2U5

pkg/cloud/openstack/clients/secgroupservice.go
```diff
@@ -0,0 +1,358 @@
...
+			return &openstackconfigv1.SecurityGroup{}, err
+		}
+		rules = append(rules, newRule)
+	}
+	newGroup.Rules = rules
+
+	return newGroup, nil
+}
+
+func (s *SecGroupService) matchRule(r1, r2 openstackconfigv1.SecurityGroupRule) bool {
```
chrigl: Would it make sense to implement this as `Equal` method on `SecurityGroupRule`? The standard library uses this pattern.
* https://golang.org/pkg/time/#Time.Equal
* https://golang.org/pkg/net/#IP.Equal

pkg/cloud/openstack/clients/secgroupservice.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
+
+	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+)
+
+const (
+	secGroupPrefix      string = "k8s"
+	controlPlanePostfix string = "controlplane"
```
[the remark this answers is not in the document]
ainmosni: Indeed it is, my brain keeps refusing to remember that though.

ainmosni:
> While I do like the idea of being able to generate required _SecurityGroups_, I am hesitant to assume we know enough about the users environment to create default rules, specifically ones that allow ingress or egress to _ANY_. I think we should discuss how we can tackle this while ensuring we are not potentially compromising the users security.

The idea was that this will be the "standard" secgroup for a "standard" clusterapi cluster until the infrastructure plan from cluster-api is more materialised. It's also an optional operation.

Honestly, if there were no upstream plans to change how infra is done, I'd have proposed to have seperate resources for both secgroups and networks, but as all this code will probably redone in a few months, it didn't seem like it was worth the effort.

chaosaffe:
In that case can you please add explicit documentation in the README to cover the security rules created by default, and how to enable/disable them?

ainmosni:
Yeah of course, was actually thinking about that after I replied to you.

chaosaffe:
/approve

@flaper87 can you PTAL and drop the hold if you are happy?

pkg/cloud/openstack/clients/secgroupservice.go
```diff
@@ -0,0 +1,348 @@
...
+	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+)
+
+const (
+	secGroupPrefix     string = "k8s"
+	controlPlaneSuffix string = "controlplane"
+	globalSuffix       string = "all"
+)
+
+var defaultRules = []openstackconfigv1.SecurityGroupRule{
```
flaper87: nit: It would be nice to add a comment here explaining what the rules below do. Who/where they provide access for/to

pkg/cloud/openstack/clients/secgroupservice.go
```diff
@@ -0,0 +1,348 @@
...
+				{
+					Direction:      "ingress",
+					EtherType:      "IPv4",
+					PortRangeMin:   22,
+					PortRangeMax:   22,
+					Protocol:       "tcp",
+					RemoteIPPrefix: "0.0.0.0/0",
+				},
+			},
+			defaultRules...,
```
flaper87: Today I learned ^

thanks :smile:

pkg/cloud/openstack/clients/secgroupservice.go
ainmosni: Good point, I kinda assumed that anyone looking at this would know about OpenStack secgroups but that might be a bit much to ask for.

Issue #85, opened 2018-10-30
Leverage cloud-provider-openstack


chaosaffe:
Our implementations currently focus on using the Kubernetes OpenStack built-ins which are deprecated. We should make the switch to using [cloud-provider-openstack](https://github.com/kubernetes/cloud-provider-openstack) instead once [Use standard clouds.yaml file as config file](https://github.com/kubernetes/cloud-provider-openstack/issues/323) closes.

iamemilio:
/assign @iamemilio 
> https://github.com/kubernetes/cloud-provider-openstack/pull/350

PR #204, opened 2019-01-25
Implemented cloud-init userdata for Ubuntu and CentOS

### LINKED ISSUE #159: Rework OS support
Right now, our generation scripts only support "CentOS" and "Ubuntu". We should add an option for Container Linux and/or Flatcar to them so that people can easily spin up clusters based on that.

chrigl:

Converted the bootstrap scripts to a unified one. This also should be
usable on Container Linux.

Wrapped around cloud-init for Ubuntu and CentOS. There is just a
single distribution specific file, that adds yum or apt repo, and adding
some specific other files.

I took care of reuse as much as possible and have as little as possible
to be written into the distri specific templates.

Fixes #159

I didn't implement anything to get `cloud.conf` into the VMs. I think this is OK, because with this PR, I consider `userDataSecret` as a feature flag. The examples say one should use `userDataSecret`.

Also @ainmosni currently works on migrating to `cloud-provider-openstack` in #85 . So we would have to write a whole lot of code, just to throw it away for #85.

Tested with:
* Ubuntu 18.04
* Ubuntu 16.04
* CentOS 7

**Release note**:

```release-note

```


pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,51 +99,45 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
+	if err != nil {
+		return oc.handleMachineError(machine, apierrors.CreateMachine(
+			"error creating Openstack instance: %v", err))
+	}
 
-		userDataSecret, err := kubeClient.CoreV1().Secrets(namespace).Get(providerSpec.UserDataSecret.Name, metav1.GetOptions{})
+	isMaster := util.IsControlPlaneMachine(machine)
+	var userDataRendered string
+	if len(userData) > 0 {
+		klog.Info("")
```
jichenjc: what's this log for?

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,51 +99,45 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
-
-		userData, ok = userDataSecret.Data[UserDataKey]
-		if !ok {
-			return fmt.Errorf("Machine's userdata secret %v in namespace %v did not contain key %v", providerSpec.UserDataSecret.Name, namespace, UserDataKey)
+	} else {
+		var distri string
+		if strings.Contains(strings.ToLower(providerSpec.Image), "ubuntu") {
+			distri = "ubuntu"
+		} else if strings.Contains(strings.ToLower(providerSpec.Image), "centos") {
+			distri = "centos"
```
jichenjc: do we request the Image's name need to contain 'centos' keyword in order to work? this seems expect too much on user's input

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,51 +99,45 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
-	if len(userData) > 0 {
-		if util.IsControlPlaneMachine(machine) {
-			userDataRendered, err = masterStartupScript(cluster, machine, string(userData))
-			if err != nil {
-				return oc.handleMachineError(machine, apierrors.CreateMachine(
-					"error creating Openstack instance: %v", err))
-			}
+		var controlPlaneEndpoint string
+		if isMaster {
+			controlPlaneEndpoint = fmt.Sprintf("%s:443", providerSpec.FloatingIP)
```
jichenjc: understand 443 is https... but should make it configurable? or left a TODO here

pkg/cloud/openstack/machine/actuator.go
chrigl: :joy: ... if I ever find such things *before* creating a PR

pkg/cloud/openstack/machine/actuator.go
chrigl: Currently yes.

I was hoping, that this will create as little user's input as possible. Adding another field to be filled for a machine... like e.g. `distributionType: ...`. But IMHO this causes more possibility to do an user error. You could specify an Ubuntu image, but use `distributionType: centos`. I thought tying the Distri-Information to the image would prevent this. What does others think on this @ainmosni , @flaper87 ?

For sure, I need to change the documentation for this.

chrigl: But. It just came to my mind... as soon as we support volumes, we need to specify the distributionType... So I will add it.

pkg/cloud/openstack/machine/actuator.go
chaosaffe: This is hardcoded to 443 in the upstream cluster-api code. There is an issue to make it configurable, but currently we are stuck with the hard coded port

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -396,3 +400,75 @@ func (oc *OpenstackClient) validateMachine(machine *clusterv1.Machine, config *o
...
+		},
+		ScriptParams: userdata.ScriptParams{
+			Namespace: machine.ObjectMeta.Namespace,
+			Name:      machine.ObjectMeta.Name,
+		},
+		Token: token,
+	}
+
+	var buf bytes.Buffer
+	if err := userdata.WriteCloudConfig(&buf, distri, isMaster, params); err != nil {
```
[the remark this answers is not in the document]
chrigl: Hmm... In theory both can be set, even though it is not implemented yet.

When implementing a multi-master setup, only the first master will not have any bootstrap token, but other master will do have.

But if you want me to delete this flag, I will do.

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -40,6 +40,8 @@ type OpenstackProviderSpec struct {
 	Flavor string `json:"flavor"`
 	// The name of the image to use for your server instance.
 	Image string `json:"image"`
+	// The type of the distribution you use.
```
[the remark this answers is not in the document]
chrigl: Are you ok with this? I don't like to create too many places where we have to maintain supported distributions.

pkg/cloud/openstack/machine/actuator.go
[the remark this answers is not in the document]
ainmosni: I'm not against it on principle, it's just that it seems a bit redundant, if you think there will be a reason to have masters with a token defined in the future, keep it. If not, scrap it.

pkg/apis/openstackproviderconfig/v1alpha1/types.go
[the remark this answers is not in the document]
ainmosni: Yeah good point, shame we can't really return good errors, else that'd be a great place for it.

docs/userdata.md
```diff
@@ -0,0 +1,88 @@
+# Passing in user-data
+
+`user-data` is a way to ship config files and scripts into OpenStack VMs. Usually VMs come up and call `cloud-init` to get `user-data` from the OpenStack Metadata Service and execute it.
+
+For further documentations have a look at:
+* [OpenStack Metadata Service](https://docs.openstack.org/nova/rocky/user/metadata-service.html)
```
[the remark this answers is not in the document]
chrigl: Great catch!

flaper87:
/hold

I'd like to review this

ainmosni:
It looks good to me but I'll let @flaper87 give it the lgtm. :)

docs/userdata.md
```diff
@@ -0,0 +1,91 @@
+# Passing in user-data
+
+`user-data` is a way to ship config files and scripts into OpenStack VMs. Usually VMs come up and call `cloud-init` to get `user-data` from the OpenStack Metadata Service and execute it.
```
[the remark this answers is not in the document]
chrigl: ... is this better now?

I implemented the userdata package exactly for the case of extending ignition stuff.

chrigl:
> I feel like these configs (and the current templates) are examples, tests, and not necessarily production-ready scripts/configs. I'm not sure we want to get in the business of supporting distro specific configs.

I thought there was an agreement on having this "in-tree": https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/159#issuecomment-454011745

> Doing this requires making the actuator aware of other logic that may not belong in it. For example, we have to instruct what type of distro (I'd rather not have the distributionType attribute) is being deployed, we need to tell it what init system is running in the image, etc. For example, there could be a centos image that uses ignition.

I decided to wo with distributionType because it is hard to find the type automatically. Currently we depend on someone managing and passing in correct secrets. Which is IMHO more work to add them to the API and use them via `userDataSecret`, than it is to pass in a `distributionType`.

The implementation is exactly meant to be extended by ignition / container linux config. I'm also thinking about migrating `machineScript` to be one "userdata-Provider".

> We will never get these configs right. In these configs we're not just creating a node for the kubernetes cluster. We're also configuring the base OS so that it can host the kubelet and other services required for this cluster, which is a realm I don't think we should get into in any official form. I'd rather let operators decide what they want to do.

I think we would get the configs right for the vast majority of cases. If someone wants to pass in own userdata, we still support this.
Other provider even prepare VM images.

> Although not really required, this patch also adds a parser for the cloud-config and a way to generate the final configs. This is a bunch of code that I don't think belongs here. Perhaps we should have a separate repo for this? (thinking out loud)

The generation primarily exists to get the files into go-variable. We could remove the files, and only keep the generated ones. I just thought it is easier to have them editable with syntax highlighting outside of a 
```
var someThing = `#!/bin/bash
something
`
```

Sure. A generator could live outside this repo, but I thought for most deployments it should be fine with a sane set of defaults, and still providing the ability to use entirely customised userdata.

chrigl:
@flaper87 I see and understand your points. Maybe that was just a misconception on my side. I also had a talk to @chaosaffe .

IMHO I could do a couple of things with this code.

1. Generate the cloud-init-files into the example folder, and get rid of this code.
2. Moving this code to `cmd/clusterctl/cmd/generate/userdata` or something similar, and create an appropriate sub-command like `clusterctl generate phase userdata -c ubuntu`.

The code is meant to work with io.reader/io.writer. So this is none of a big deal. In this case, the generate.go will go away, because I would be able to read in files (which was my initial idea for this code, before the misconception ;)). This would spit out a cloud-init template as a starting point for users to customize. Which than can to in via `userDataSecret`.

I would prefer option 2 in favour of 1 (removing the code), because this would be one step to get away of `generate-yaml.sh`.

chrigl:
/close

I will open a new one when I am done.

Currently thinking about just replacing XX-user-data.sh by a XX-cloud-config.yaml. I think this makes the most sense for the project.

PR #207, opened 2019-01-28
Fix the log level of failing to access the master node

### LINKED ISSUE #202: Error reported before being able to acess via floating IP
There is a short period of time the master node can be accessed via floating IP after allocating floating IP.
During the period, clusterctl returns the following error.

```bash
E0124 01:34:57.770799   26526 util.go:150] error executing command "ssh -i /home/ubuntu/.ssh/openstack_tmp -o StrictHostKeyChecking no -o UserKnownHostsFile /dev/null -o BatchMode=yes ubuntu@10.0.0.206 echo STARTFILE; sudo cat /etc/kubernetes/admin.conf": exit status 1
```

This error message should be deleted.

hidekazuna:
When creating cluster, there is a short period of time the master node
can not be accessed via floating IP after allocating floating IP.
During the period, the error message logged even if creating cluster
ends successfully.

This patch updates the message to warning and added information when
the master node was being able to accessed via floating IP.

Fixes #202

### DIFF
--- a/pkg/cloud/openstack/deployer.go
+++ b/pkg/cloud/openstack/deployer.go
@@ -20,13 +20,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"k8s.io/klog"
 	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
 	clustercommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
-	"sigs.k8s.io/cluster-api/pkg/util"
 )
 
 const ProviderName = "openstack"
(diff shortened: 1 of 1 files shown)


pkg/cloud/openstack/deployer.go
```diff
@@ -72,7 +72,7 @@ func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clu
 		return "", err
 	}
 
-	result := strings.TrimSpace(util.ExecCommand(
+	result := strings.TrimSpace(execCommand(
```
jichenjc: I am not against this kind of change but seems we should leverage Utils function as much as we can
so is it possible to have a default log level (error) and use log.warn for error cases in this typical secenia?

pkg/cloud/openstack/deployer.go
```diff
@@ -85,3 +85,14 @@ func (d *DeploymentClient) GetKubeConfig(cluster *clusterv1.Cluster, master *clu
...
 	return strings.TrimSpace(parts[1]), nil
 }
+
+func execCommand(name string, args ...string) string {
+	cmdOut, err := exec.Command(name, args...).Output()
+	s := strings.Join(append([]string{name}, args...), " ")
+	if err != nil {
+		klog.Warningf("failed to execute command %q: %v", s, err)
+	} else {
+		klog.Infof("successfully executed command %q", s)
```
gyliu513: Can you please update the util function in `cluster-api`? I think it is OK to use `Errorf` if the command failed, but you can add the `else` logic for successful command, comments? @hidekazuna

hidekazuna:
Close this PR because I understand the purpose of This PR is tiny issue as of now.

Issue #205, opened 2019-01-25
Use created networks for machines.


ainmosni:
Right now, we have the option of having an OpenStack network created when a cluster is created, except there is no way to have the machines actually use this network. This makes the actual usage of this network a bit clumsy as you first have to create a cluster, wait for the network to be created, get the status of the cluster and then manually fill it in in the machine spec. We should add a `fromCluster` (or similar) option to machineSpec.providerSpec.networks, that will look at a specified cluster and uses the network that's defined in the status.

jichenjc:
I can work on that if no one already work on it

ainmosni:
Yeah, I was thinking to just check the status, if it's filled in, go ahead, else abort.

jichenjc:
basic idea will be Add a bool 'fromCluster' in NetworkParam (inside OpenstackProviderSpec) or a special value for FixedIp (such as 'fromcluster') so that when creating machine, we won't use the ID , instead, when doing InstanceCreate , we will query the info from apis/openstackproviderconfig/v1alpha1/types.go:Network , and use that info, of course , need double check whether the network is ready or not before use it....

PR #211, opened 2019-01-29
Add nodeCidr into template file

jichenjc:
nodeCidr is actually supported, just provide an empty value
here so user/admin is able to make changes accordingly

enhancement 

Fixes #

**Release note**:
```release-note

```


chrigl:
Hmm. I think we should leave this out, and just include it with #205 .There is not yet a way to create machines without a network configuration. What do you think @jichenjc ?

jichenjc:
ok, I can do that ,actually I found this because I am working on #205 
so I will close this and combine it within 205

Issue #219, opened 2019-01-30
Create FloatingIP for machines


posox:
I'd like to add an ability to create floating ip for machines automatically when `externalNetworkId` specified in `ClusterProviderSpec`.

PR #220, opened 2019-01-30
Create floatingip for machines

posox:
Automatically create floatingip for machines when externalNetworkId
specified in ClusterProviderSpec and floatingip is not specified in
machine spec

Closes: #219

### DIFF
--- a/pkg/cloud/openstack/clients/networkservice.go
+++ b/pkg/cloud/openstack/clients/networkservice.go
@@ -20,14 +20,20 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/client-go/kubernetes"
+
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
+	"github.com/gophercloud/utils/openstack/clientconfig"
 	"k8s.io/klog"
 	openstackconfigv1 "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,6 +104,17 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
 		return nil
 	}
 
+	// create floatingip if needed
+	if providerSpec.FloatingIP == "" && machineSpec.ExternalNetworkID != "" {
```
jichenjc: maybe add a log to indicate the creation here?

chrigl:
@posox I think, I understand your idea, but I'm not sure if it works out as expected. You are about to modify the state that is already stored... Meaning, you create the FloatingIP in OpenStack and add to the machine spec.

But, what if the controller crashes (due to machine error, oom or whatever),right after you created the FIP, but did not already create the VM. When the the controller comes up again, it acquires the next FIP. In the worst case up until there are no more available ones. At least you need a way to find potentially already created FloatingIPs. A quick look at gophercloud lets me guess, that the only sane way to do this is to (mis-)use the description field of a FIP.

Also, this might not align with further implementation of a multi-master-setup. A multi-master setup will bring a LB (probably in the cluster actuator, or some separate infra structure layer see https://github.com/kubernetes-sigs/cluster-api/issues/639

I think creating a FIP in the machine actuator is nothing we want in the long-term. But I am with you that we should help people to have an easier getting started.
In the first step, it could just modifying the `machine.yaml.template` in order to replace things the user must change with a clear placeholder. Image, availabilityZone, networks, floatingIP and SecurityGroups come to my mind.
The next step would be to fix the generate-stuff to be kind of better. I would love to integrate it into `clusterctrl`.

But I do want to have other opinions on this

/cc @flaper87 @chaosaffe 

/hold

chaosaffe:
I second what @chrigl said irt the implementation itself. That said, I would like to see the option to have a FIP assigned to a node, worker or master, but we also need to ensure that it would be completely idempotent to handle failure and retry cases. There are cases where I may want my worker nodes, or a subset of nodes to obtain FIP's, and I would like to bypass having to manually interact with OS to provision them

chrigl:
I could also imagine scenarios where whatever type nodes should get a floating ip. Then this should be a configuration flag on the individual machine.

Some random thoughts:

I think updating the floatingIP Field of a machine in the controller is OK, because the machine object in hand is updated in the kubernetes API anyways to bring new annotations.

Thinking a bit further. Maybe we do want to be able to attach floating ips to individual nodes at runtime. So just changing the floatingIP field for an existing machine. This would trigger `machine/actuator.Actuator.Update`... So we must observe the real VM in update and check if the elected floatingIP is already present on this particular node. If not, it should be attached. But this should not force a recreation of the node. This also means, that the selected floatingIP should go to the status, to be able to remove it, in case someone removes it again from the machine.

With this. Even attaching a FIP to the node for ssh-debugging purposes could be done entirely via kubectl and without any need to touch OpenStack.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -75,12 +75,22 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
+		return err
+	}
+
 	providerSpec, err := openstackconfigv1.MachineSpecFromProviderSpec(machine.Spec.ProviderSpec)
 	if err != nil {
 		return oc.handleMachineError(machine, apierrors.InvalidMachineConfiguration(
 			"Cannot unmarshal providerSpec field: %v", err))
 	}
 
+	machineSpec, err := openstackconfigv1.ClusterSpecFromProviderSpec(cluster.Spec.ProviderSpec)
```
chrigl: This is `clusterSpec`, not `machineSpec`.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,6 +104,17 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
...
 
+	// create floatingip if needed
+	if providerSpec.FloatingIP == "" && machineSpec.ExternalNetworkID != "" {
+		fip, err := networkService.CreateFloatingIP(machineSpec.ExternalNetworkID)
+		if err != nil {
+			return oc.handleMachineError(machine, apierrors.CreateMachine(
+				"Failed to create floating ip: %v", err))
+		}
+
+		providerSpec.FloatingIP = fip
```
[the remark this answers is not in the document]
chrigl: Yes. The machine will be updated in the APIServer for changing the annotations. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/machine/actuator.go#L329

pkg/cloud/openstack/machine/actuator.go
chrigl: Please make auto-creation of floating IPs configureable on Machine.

pkg/cloud/openstack/clients/networkservice.go
```diff
@@ -320,3 +382,15 @@ func (s *NetworkService) getNetworkByName(networkName string) (networks.Network,
 	}
 	return networks.Network{}, errors.New("too many resources")
 }
+
+func (s *NetworkService) CreateFloatingIP(floatingNetworkID string) (string, error) {
+	opts := floatingips.CreateOpts{
+		FloatingNetworkID: floatingNetworkID,
```
chrigl: As mentioned. Set a Description here with which you can find the floating ip in case of a restart of the controller. Maybe `machine.Meta.Uid` does work!? Otherwise `namespace/cluster.Name/machine.Name` would be fine.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -94,6 +104,17 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
 		return nil
 	}
 
+	// create floatingip if needed
+	if providerSpec.FloatingIP == "" && machineSpec.ExternalNetworkID != "" {
+		fip, err := networkService.CreateFloatingIP(machineSpec.ExternalNetworkID)
```
chrigl: Before creating the FIP, try to find an existing one with the description `machine.Meta.Uid` (?)

jichenjc:
are you still working on this ? or I can help to update this a little bit @posox

CamelCaseNotation:
@jichenjc are you planning to update this for v1alpha2? I think having a FIP be created would be useful for dev clusters where 1 master is sufficient

jichenjc:
low priority to me ... I think we need focus on v1alpha2 adoption now ..

PR #221, opened 2019-02-07
Revert "Add kuryr support"

chaosaffe:
Reverts kubernetes-sigs/cluster-api-provider-openstack#61

Code does not build


flaper87:
@chaosaffe thanks for taking care of this and I'm sorry for the breakage. 

Digging into what happened. This built locally for me and it was tested by @luis5tb prob some weird obscure case, dunno.

chaosaffe:
@flaper87 IIUC was related to the confused imports, as there was some
imports being done from `
github.com/kubernetes-sigs/cluster-api-provider-openstack` and some from `
sigs.k8s.io/cluster-api-provider-openstack`. @chrigl brought it to my
attention after there was an issue reported in Slack and he wsa able to
reproduce. I did _not_ investigate myself, but felt it was prudent to
roll-back, reassess and then apply the patch if all was well.

WRT the import issues, dep can work around it, but it is easy to get
dependencies out of sync. I think we need to make a choice on either using
GitHub or the vanity import (vanity is the standard to abstract in case the
codebase needs to be moved to other tooling afaik) and do a mass fix in the
codebase to set the standard, then we as a review team need to enforce that.

That being said, has this patch been tested in environments without Kuryr
deployed? I haven't had a chance to review the code beyond the initial
pass, and while I certainly want Kuryr support where it is available, I
don't want to break the controller for environments without it deployed. I
think @chrigl was working on testing the patch in our environment (without
Kuryr), but I am not sure how far he got with that.

flaper87:
I'm all for stepping stones. This patch doesn't break any of the supported scenarios so I'd be ok with merging it and then have a follow-up patch to improve it.

flaper87:
By merging it I meant merging #222

PR #222, opened 2019-02-08
Trunk revert

luis5tb:
Problem was cluster-api-provider-openstack vendor was wrongly added

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -107,11 +107,13 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:bf40c04c15471450e644627eb1317197cac533e60883cf44a6a8a2fdb5f3a016"
+  digest = "1:a83fdc9d6e0a39d0f4874b9c6021d77eaf62efdd5a868e6c9c500bdb00086196"
   name = "github.com/gophercloud/gophercloud"
   packages = [
     ".",
     "openstack",
+    "openstack/common/extensions",
+    "openstack/compute/v2/extensions/attachinterfaces",
     "openstack/compute/v2/extensions/floatingips",
     "openstack/compute/v2/extensions/keypairs",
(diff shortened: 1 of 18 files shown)


flaper87:
/approve

I'll let @chaosaffe merge this one!

chaosaffe:
/assign @chrigl
I will defer to Chris as I haven't touched this one in a while

@luis5tb sorry for the confusion, and I want to emphasize that we are very
grateful for your work! Thank you again

luis5tb:
@chaosaffe: Thanks! And don't worry, it was a newbie mistake! Seems my go env is wrongly configured and dep ensure added the wrong vendor stuff. In the env where I was testing it it was not the case and thus I thought it was working. Sorry for the breakage!

flaper87:
@chrigl I think @luis5tb already ran the tests you wanted. Quoting his comment on the revert PR below:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/221#issuecomment-461733127

> @chaosaffe: I tested the next 3 scenarios and they worked as expected:
> 
>     * Trunk support + trunks==true --> working
> 
>     * Trunk support + trunks==false --> working
> 
>     * No trunk support + trunks==false --> working
> 
> 
> I did not test though a kind of wrong configuration, where there is no support for trunks, but still we set it on the machine.yaml to true. In that case creation fail as expected as there is no trunk support. The problem I found is that is leaving a leftover port. I can add an extra checking in a follow up PR to avoid this situation. We have two options:
> 
>     * If there is no trunk support, fall back to trunks=false regardless of what machine.yaml says
> 
>     * if trunks=true but no trunk support, fail before creating the port, so that machine creation fail (with the right message, wrong configuration) but leaving no leftovers.
> 
> 
> I don't have strong opinions here, but I would go for the second. What do you think?

PR #228, opened 2019-02-12
Subnet Support

iamemilio:
**What this PR does / why we need it**: Allows users to specify a specific subnet to use for their machine

### DIFF
--- a/pkg/cloud/openstack/clients/machineservice.go
+++ b/pkg/cloud/openstack/clients/machineservice.go
@@ -37,6 +37,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/gophercloud/utils/openstack/clientconfig"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -67,6 +68,10 @@ type Instance struct {
 	servers.Server
 }
 
+type ProviderNetwork struct {
+	networkID string
+	subnetID  string
+}
 type InstanceListOpts struct {
 	// Name of the image in URL format.
 	Image string `q:"image"`
@@ -254,6 +259,22 @@ func getNetworkIDsByFilter(is *InstanceService, opts *networks.ListOpts) ([]stri
 	return uuids, nil
 }
 
+func CreatePort(is *InstanceService, name string, net ProviderNetwork, securityGroups *[]string) (ports.Port, error) {
+	portCreateOpts := ports.CreateOpts{
+		Name:           name,
+		NetworkID:      net.networkID,
+		SecurityGroups: securityGroups,
+	}
+	if net.subnetID != "" {
+		portCreateOpts.FixedIPs = []ports.IP{{SubnetID: net.subnetID}}
+	}
+	newPort, err := ports.Create(is.networkClient, portCreateOpts).Extract()
+	if err != nil {
+		return ports.Port{}, fmt.Errorf("Create port for server err: %v", err)
+	}
+	return *newPort, nil
+}
+
 func (is *InstanceService) InstanceCreate(clusterName string, name string, config *openstackconfigv1.OpenstackProviderSpec, cmd string, keyName string) (instance *Instance, err error) {
 	if config == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace")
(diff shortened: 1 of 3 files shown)


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -254,6 +258,22 @@ func getNetworkIDsByFilter(is *InstanceService, opts *networks.ListOpts) ([]stri
 	return uuids, nil
 }
 
+func CreatePort(is *InstanceService, name string, net ProviderNetwork, securityGroups *[]string) (ports.Port, error) {
+	portCreateOpts := ports.CreateOpts{
+		Name:           name,
+		NetworkID:      net.networkID,
```
flaper87: Can we make `NetworkID` optional?

iamemilio: Done :)

iamemilio: Here is the the way things shake out:
Passing both a uuid and a subnet_id: Expect the user knows what they are doing, and let it fail if not. attempts to create port in the subnet in the network specified. openstack will get mad at them if they did something wrong here
Passing Filters and subnet_id: Redundant lookup, uses subnet id to lookup network id. Not sure why anyone would do this, but if they do it will work fine as long as subnet_id is valid.
subnet_id: uses subnet_id to lookup network id
Multiple Networks: exactly what you think it does

flaper87:
> Here is the the way things shake out:
> Passing both a uuid and a subnet_id: Expect the user knows what they are doing, and let it fail if not. attempts to create port in the subnet in the network specified. openstack will get mad at them if they did something wrong here

We may need to revisit this in the future but I think it's fine for now.

> Passing Filters and subnet_id: Redundant lookup, uses subnet id to lookup network id. Not sure why anyone would do this, but if they do it will work fine as long as subnet_id is valid.does 

We should print a warning that filters are ignored in this case.

> subnet_id: uses subnet_id to lookup network id

+1

> Multiple Networks: exactly what you think it does

Nice! That means it creates new unicorns and sends them to find new rainbows in the universe

Thanks for the great, speedy, work on this

flaper87:
/approve

I'll give some time for @chrigl or @chaosaffe to take a look at this, otherwise I'll send it through :+1:

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -262,22 +283,33 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
...
+	var nets []ProviderNetwork
 	for _, net := range config.Networks {
-		if net.UUID == "" {
+		if net.UUID == "" && net.SubnetID == "" {
 			opts := networks.ListOpts(net.Filter)
 			ids, err := getNetworkIDsByFilter(is, &opts)
 			if err != nil {
 				return nil, err
 			}
 			for _, netID := range ids {
```
jichenjc: forgive me if I am wrong, but looks like NetworkParam is an array in OpenstackProviderSpec
so you have a look at line 288 then you loop again to query all the network IDs

do this means if we have 2 networks defined without netid we will create from network 2 times? guess a fail should be better..

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -67,6 +68,10 @@ type Instance struct {
 	servers.Server
 }
 
+type ProviderNetwork struct {
```
jichenjc: guess I missed some mail and I asked whether ProviderNetwork can be renamed because openstack provider network has special meaning...

pkg/cloud/openstack/clients/machineservice.go
jichenjc: @iamemilio maybe this is old behavior to previous code ...

pkg/cloud/openstack/clients/machineservice.go
iamemilio: Didnt see that, but no problem. Easy fix.

pkg/cloud/openstack/clients/machineservice.go
iamemilio: @jichenjc Are you talking about filters loop specifically? I'm not sure I understand the case you are suggesting.

iamemilio: If you use a filter to specify a network, it is possible it will find multiple networks. We decided that this should be a feature of the system. I dont think its possible for the openstack query to return duplicate network id's, so I'm not sure where this duplication case would be.

iamemilio: @jichenjc feel free to ping me on slack if you want to chat about it though. I'm: egarcia

jichenjc: ok, I might need create the environment to prove what my thought is correct or not ,I thought if we have 2 networks defined and both don't have uuid given whether we will actually append network creation twice given current code logic...  (line 288 and line 295 here) ,anyway, I think it's eligible to be merged and if I can find anything I can create a bug/PR later.. thanks

iamemilio: @jichenjc  I understand now. You are asking if you can add 2 different networks with the same uuid. I'm pretty sure you can. That may be worth raising an issue over.

flaper87:
@chrigl @chaosaffe can you please review this patch?

flaper87:
@iamemilio can you rebase this, please? 

@chaosaffe @chrigl u guys want to take a look at this patch?

flaper87:
@chaosaffe @chrigl This PR shouldn't introduce any breaking changes so I'm going to send it through.

/lgtm

PR #229, opened 2019-02-13
Ensure trunk is not enabled if there is no support

luis5tb:
This PR ensures trunk cannot be set if the system doesn't have trunk
support. This will avoid leaving orphan ports when failing to create
the trunk for it.

### DIFF
--- a/pkg/cloud/openstack/clients/machineservice.go
+++ b/pkg/cloud/openstack/clients/machineservice.go
@@ -258,6 +258,15 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 	if config == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace")
 	}
+	if config.Trunk == true {
+		trunkSupport, err := GetTrunkSupport(is)
+		if err != nil {
+			return nil, fmt.Errorf("There was an issue verifying whether trunk support is available, please disable it: %v", err)
+		}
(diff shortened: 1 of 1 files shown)


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -258,6 +258,13 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 	if config == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace")
 	}
+	trunkSupport, err := GetTrunkSupport(is)
+	if err != nil {
+		return nil, fmt.Errorf("Obtaining network extensions err: %v", err)
```
flaper87: Do we have to return an error in this case?

What about something like this:

```go
if (err != nil && config.Trunk == true)  || (trunkSupport == false && config.Trunk == true) {
     return nil, fmt.Errorf("There was an issue verifying whether trunk support is available, please disable it: {} {}", config.Trunk, err)
}
```

(of course I didn't test that code)

Basically, I don't think we should fail/return in cases where `config.Trunk` is set to false and there was a problem getting network extensions.

luis5tb: makes sense! thanks! I'll update it asap

flaper87:
@luis5tb nice, I like it!

tests are failing, tho :smile:

PR #231, opened 2019-02-15
Add information about security group UUID usage

luis5tb:
It is required to use security group UUIDs instead of their names at the machine yaml


flaper87:
Thanks, @luis5tb 

/ok-to-test

flaper87:
At some point, we should also add filters for security groups. It'll allow for creating more generic machine templates.

PR #239, opened 2019-02-26
Provide SecurityGroups attachment to machines

### LINKED ISSUE #235: Provide SecurityGroups attachments to machines
Right now, if clusterSpec contains `managedSecurityGroups:true` field, two security groups will be created (one for controlplane machine and one for all the nodes), but currently they don't attach to machines. 
I'd like to add automatic attachment of created controlplane and general SecurityGroups to machines.

eromanova:
Automatically attach created security groups
to machines if managedSecurityGroups is true
in clusterSpec.providerSpec.

Fixes #235

### DIFF
--- a/pkg/cloud/openstack/clients/machineservice.go
+++ b/pkg/cloud/openstack/clients/machineservice.go
@@ -357,7 +357,36 @@ func isDuplicate(list []string, name string) bool {
 	return false
 }
 
-func GetSecurityGroups(is *InstanceService, sg_param []openstackconfigv1.SecurityGroupParam) ([]string, error) {
+func GetSecurityGroups(is *InstanceService, sg_param []openstackconfigv1.SecurityGroupParam, cluster *clusterv1.Cluster, machine *clusterv1.Machine) ([]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -357,7 +357,35 @@ func isDuplicate(list []string, name string) bool {
...
+		return nil, fmt.Errorf("failed to load cluster provider spec: %v", err)
+	}
+
+	err = secGroupService.Reconcile(clusterName, *clusterSpec, clusterStatus)
+	if err != nil {
+		return nil, fmt.Errorf("failed to reconcile security groups: %v", err)
+	}
+
+	if clusterSpec.ManagedSecurityGroups {
+		if util.IsControlPlaneMachine(machine) {
```
jichenjc: I understand we want to return cluster defined sec and how about add a log in order to let operator understand we ignored the sec group we defined in machine spec to avoid confusion?

eromanova: @jichenjc Sorry for the delay, had a vacation. I agree about the fact that user should be aware of ignored security groups. I updated changes, please review it again.

PR #241, opened 2019-02-26
Delete SecurityGroups from cluster

### LINKED ISSUE #227: SecurityGroup automatic removal during cluster deletion
Currently, users are able to set `managedSecurityGroups: true` in cluster ProviderSpec. This will create two security groups: one for control-plane machine and another for all nodes. But created SecurityGroups keep alive after machine removal. It'll be good to add their automatic removal during cluster deletion.

eromanova:
This changes provide security groups removal
during cluster deletion if managedSecurityGroups
is true in clusterSpec.providerSpec

Fixes #227

### DIFF
--- a/pkg/cloud/openstack/cluster/actuator.go
+++ b/pkg/cloud/openstack/cluster/actuator.go
@@ -100,6 +100,11 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
 		return err
 	}
 
+	secGroupService, err := clients.NewSecGroupService(client)
+	if err != nil {
+		return err
+	}
+
 	// Load provider config.
 	_, err = providerv1.ClusterSpecFromProviderSpec(cluster.Spec.ProviderSpec)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -87,20 +87,41 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
...
 	// Load provider status.
-	_, err = providerv1.ClusterStatusFromProviderStatus(cluster.Status.ProviderStatus)
+	providerStatus, err := providerv1.ClusterStatusFromProviderStatus(cluster.Status.ProviderStatus)
 	if err != nil {
 		return errors.Errorf("failed to load cluster provider status: %v", err)
 	}
 
 	// Delete other things
 
+	if providerStatus.GlobalSecurityGroup != nil {
```
gyliu513: Do we need to check `managedSecurityGroups` to see if it is enabled or not when deleting the security groups?

eromanova: I suppose checking `managedSecurityGroups` parameter here is not really necessary, because `providerStatus.GlobalSecurityGroup` and `providerStatus.ControlPlaneSecurityGroup` values provide us information about security groups being created (this fields are filling up only if `managedSecurityGroups` is true). If such parameters are not `nil`, then security groups were created and we need to delete them.

gyliu513: make sense, thanks @eromanova 

/approve

/cc @jichenjc @chrigl

pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -87,20 +87,41 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
...
+		klog.Infof("Deleting global security group: %v", providerStatus.GlobalSecurityGroup.Name)
```
xunpan: If it is a name, please use `security group %q` instead of `security group: %v`. `%q` will add quotes automatically to the name and no need for `:`.

pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -87,20 +87,41 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
...
+		if err != nil {
+			return errors.Errorf("failed to delete security group %v", err)
+		}
+	}
+
+	if providerStatus.ControlPlaneSecurityGroup != nil {
+		klog.Infof("Deleting control plane security group: %v", providerStatus.ControlPlaneSecurityGroup.Name)
+		err := secGroupService.Delete(providerStatus.ControlPlaneSecurityGroup)
+		if err != nil {
+			return errors.Errorf("failed to delete security group %v", err)
```
xunpan: If it is error, please use `: %v` instead of `%v` to make error message clear displayed in its postion.

pkg/cloud/openstack/cluster/actuator.go
eromanova: @xunpan thanks for your comments, done.

PR #242, opened 2019-02-26
Allow for filtering subnets rather than passing an UUID

flaper87:
Requesting the subnetID is very limiting as it requires to create
subnets in advance and know the uuid. We allow for filtering netwoks and
we should allow for filtering subnets too. This will make the API more
consistent.


pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -306,21 +331,37 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 					networkID: netID,
 				})
 			}
-		} else if net.UUID == "" && net.SubnetID != "" {
-			subnet, err := subnets.Get(is.networkClient, net.SubnetID).Extract()
-			if err != nil {
-				return nil, err
+		} else if net.Subnets != nil {
+			for _, snet := range net.Subnets {
+				if snet.UUID != "" {
```
chrigl: You could leave out this if... see next comment.

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -306,21 +331,37 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
...
+					subnet, err := subnets.Get(is.networkClient, snet.UUID).Extract()
+					if err != nil {
+						return nil, err
+					}
+					nets = append(nets, ServerNetwork{
+						networkID: subnet.NetworkID,
+						subnetID:  snet.UUID,
+					})
+				} else {
+					sopts := subnets.ListOpts(snet.Filter)
```
chrigl: ... alse withouth this else for sure...
do
```
sopts.ID = snet.UUID
sopts.NetworkID = net.UUID
```

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -306,21 +331,37 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 					networkID: netID,
 				})
 			}
-		} else if net.UUID == "" && net.SubnetID != "" {
-			subnet, err := subnets.Get(is.networkClient, net.SubnetID).Extract()
-			if err != nil {
-				return nil, err
+		} else if net.Subnets != nil {
```
chrigl: This could become an else, or might go away. When net.Subnets are empty, the for loop just don't kick in.

pkg/cloud/openstack/clients/machineservice.go
flaper87: a-ha! Good catch. I like it :+1:

pkg/cloud/openstack/clients/machineservice.go
flaper87: I think we should keep this one, tho.

This prevents from adding the same network twice, one with a subnet and one by itself. For example:

```
networks:
- uuid: MYUUID
  subnets:
    - name: test
```

Without this clause, the above example could end up with 2 `ServerNetwork` objects. One obtained by filtering the subnet and another one because `uuid` is set.

flaper87: Actually, now that I read this again, I think this code should be re-organized.

PR #249, opened 2019-03-05
Tags and Metadata

iamemilio:
**What this PR does / why we need it**: Allows you to tag servers

**Special notes for your reviewer**: Looks like it should be working, but I don't have a cluster with the appropriate api to test on. Requires Openstack version Pike.

### DIFF
--- a/cmd/clusterctl/examples/openstack/machines.yaml.template
+++ b/cmd/clusterctl/examples/openstack/machines.yaml.template
@@ -26,7 +26,7 @@ items:
         trunk: false
         tags:
           - <Your Tag>
-        server_metadata:
+        serverMetadata:
           key: value
     versions:
       kubelet: 1.12.3
@@ -58,7 +58,7 @@ items:
         trunk: false
         tags:
           - <Your Tag>
-        server_metadata:
+        serverMetadata:
           key: value
     versions:
       kubelet: 1.12.3
--- a/cmd/clusterctl/examples/openstack/machines.yaml.template
+++ b/cmd/clusterctl/examples/openstack/machines.yaml.template
@@ -24,6 +24,10 @@ items:
           name: master-user-data
           namespace: openstack-provider-system
         trunk: false
+        tags:
+          - <Your Tag>
+        server_metadata:
+          key: value
     versions:
       kubelet: 1.12.3
       controlPlane: 1.12.3
@@ -52,5 +56,9 @@ items:
           name: worker-user-data
           namespace: openstack-provider-system
         trunk: false
+        tags:
+          - <Your Tag>
+        server_metadata:
+          key: value
     versions:
       kubelet: 1.12.3
--- a/config/crds/openstackproviderconfig_v1alpha1_openstackproviderspec.yaml
+++ b/config/crds/openstackproviderconfig_v1alpha1_openstackproviderspec.yaml
@@ -39,6 +39,8 @@ spec:
             properties:
               filter:
                 properties:
+                  description:
+                    type: string
                   id:
                     type: string
                   limit:
@@ -59,6 +61,22 @@ spec:
                 items:
                   properties:
                     filter:
+                      properties:
+                        cidr:
+                          type: string
+                        description:
+                          type: string
+                        id:
+                          type: string
+                        limit:
+                          format: int64
+                          type: integer
+                        marker:
+                          type: string
+                        name:
+                          type: string
+                        tags:
+                          type: string
                       type: object
                     uuid:
                       type: string
@@ -73,6 +91,10 @@ spec:
             properties:
               filter:
                 properties:
+                  description:
+                    type: string
+                  id:
+                    type: string
                   limit:
                     format: int64
                     type: integer
@@ -89,8 +111,14 @@ spec:
                 type: string
             type: object
           type: array
+        serverMetadata:
+          type: object
         sshUserName:
           type: string
+        tags:
+          items:
+            type: string
+          type: array
         trunk:
           type: boolean
         userDataSecret:
--- a/docs/config.md
+++ b/docs/config.md
@@ -176,3 +176,39 @@ You can specify multiple networks (or subnets) to connect your server to. To do 
           - uuid: your_network_id
           - subnet_id: your_subnet_id
 ```
+
+## Tagging Instances
+Tags can be added to instances on startup by populating the tags array where it says `<Your Tags>`. You can remove this if you do not want to tag your instances. Please note that your Nova api must be at least mivroversion 2.52 to use this api! Please refer to the following usage example:
+
+```yaml
+- apiVersion: "cluster.k8s.io/v1alpha1"
+  kind: Machine
+  metadata:
+    generateName: openstack-node-
+    labels:
+      set: node
+  spec:
+    providerSpec:
+      value:
+        tags:
+          - tag1
+          - tag2
+```
+
+## Metadata
+Instead of tagging, you also have the option to add metadata to instances. This functionality should be more commonly available than tagging. Here is a usage example:
+
+```yaml
+- apiVersion: "cluster.k8s.io/v1alpha1"
+  kind: Machine
+  metadata:
+    generateName: openstack-node-
+    labels:
+      set: node
+  spec:
+    providerSpec:
+      value:
+        server_metadata:
+          name: bob
+          nickname: bobbert
+```
--- a/docs/config.md
+++ b/docs/config.md
@@ -208,7 +208,7 @@ Instead of tagging, you also have the option to add metadata to instances. This 
   spec:
     providerSpec:
       value:
-        server_metadata:
+        serverMetadata:
           name: bob
           nickname: bobbert
 ```
--- a/pkg/apis/openstackproviderconfig/v1alpha1/types.go
+++ b/pkg/apis/openstackproviderconfig/v1alpha1/types.go
@@ -69,6 +69,13 @@ type OpenstackProviderSpec struct {
 	Trunk bool `json:"trunk,omitempty"`
 
 	RootVolume RootVolume `json:"root_volume,omitempty"`
+
+	// Server tags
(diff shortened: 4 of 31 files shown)


flaper87:
giving @chrigl and others more time before merging this

cmd/clusterctl/examples/openstack/machines.yaml.template
```diff
@@ -24,6 +24,8 @@ items:
           name: master-user-data
           namespace: openstack-provider-system
         trunk: false
+        tags:
```
flaper87: Should we also do this for metadata?

config/crds/openstackproviderconfig_v1alpha1_openstackproviderspec.yaml
```diff
@@ -91,6 +111,10 @@ spec:
           type: array
         sshUserName:
           type: string
+        tags:
```
flaper87: metadata seems missing

iamemilio: Yeah, I am not sure what the convention for this is to be honest. These have been auto generated up until this point.

cmd/clusterctl/examples/openstack/machines.yaml.template
iamemilio: Yeah we can do that

config/crds/openstackproviderconfig_v1alpha1_openstackproviderspec.yaml
flaper87: It's weird that the auto-generation didn't add metadata here. :thinking:

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -481,6 +481,8 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 		UserData:         []byte(userData),
 		SecurityGroups:   securityGroups,
 		ServiceClient:    is.computeClient,
+		Tags:             config.Tags,
```
flaper87: I was thinking about this a bit more and I think we should add the machine tags to all the resources that are created as part of the instance creation. Therefore, the ports created in this function should also have `config.Tags` added to them.

Also, what happens if you run this code against a Nova that doesn't have support for Tags? Can you check what OpenStack release added tags to Nova?

iamemilio: Yeah, I think it would makes sense to create a "Global Tags" field of tags that get applied to everything created by CAPO

iamemilio:
Its failing to vet over a naming error, I can patch tomorrow morning if you want

docs/config.md
```diff
@@ -231,3 +235,39 @@ You can specify multiple networks (or subnets) to connect your server to. To do
           - uuid: your_network_id
           - subnet_id: your_subnet_id
 ```
+
+## Tagging Instances
+Tags can be added to instances on startup by populating the tags array where it says `<Your Tags>`. You can remove this if you do not want to tag your instances. Please note that your Nova api must be at least mivroversion 2.52 to use this api! Please refer to the following usage example:
```
jichenjc: https://developer.openstack.org/api-ref/compute/#server-tags-servers-tags
actually you can set 2.26 version by first create a VM then tag it

pkg/cloud/openstack/clients/machineservice.go
jichenjc: https://developer.openstack.org/api-ref/compute/#server-tags-servers-tags 
2.26 added tag

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -134,6 +143,7 @@ type SubnetParam struct {
 
 type SubnetFilter struct {
 	Name            string `json:"name,omitempty"`
+	Description     string `json:"description,omitempty"`
```
jichenjc: what's this for??

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -432,7 +435,7 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
 		}
 
 		_, err = attributestags.ReplaceAll(is.networkClient, "ports", port.ID, attributestags.ReplaceAllOpts{
-			Tags: clusterTags}).Extract()
```
jichenjc: this if for network client? why we tag a server tag to a network ? for example ,one server tag as 't1' we have to tag all ports with 't1'?? IMHO we should not have those kind of logic unless we have very specified reason to do so

if user need to tag something on their own, they can tag in their 
- network
section...

pkg/apis/openstackproviderconfig/v1alpha1/types.go
flaper87: This is coming from a gophercloud update

docs/config.md
flaper87: A-ha, this is a good point. We should probably do this. Just set the version to 2.52 when `Tags` is present.

@iamemilio mind doing this? :)

pkg/cloud/openstack/clients/machineservice.go
flaper87: mmh, gophercloud says we should be using 2.52 :/

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -176,6 +176,10 @@ func NewInstanceServiceFromCloud(cloud clientconfig.Cloud) (*InstanceService, er
 	serverClient, err := openstack.NewComputeV2(provider, gophercloud.EndpointOpts{
 		Region: clientOpts.RegionName,
 	})
+
+	// NOTE(flaper87): This is the minimum required version
+	// to use tags.
+	serverClient.Microversion = "2.52"
```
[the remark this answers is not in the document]
flaper87: Moved it to 2.52. As for providing a better structure, I'd rather start with this and then engineer a more generic solution once we actually need it.

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -69,6 +69,13 @@ type OpenstackProviderSpec struct {
 	Trunk bool `json:"trunk,omitempty"`
 
 	RootVolume RootVolume `json:"root_volume,omitempty"`
+
+	// Server tags
+	// Requires Nova api 2.52 minimum!
+	Tags []string `json:"tags,omitempty"`
+
+	// Metadata mapping. Allows you to create a map of key value pairs to add to the server instance.
+	ServerMetadata map[string]string `json:"server_metadata,omitempty"`
```
chrigl: Rename to `json:"serverMetadata,omitempty"` to be consistent.

flaper87: yup, on it :D

docs/config.md
flaper87: Did it, a bit different but I think it's a good start.

pkg/cloud/openstack/clients/machineservice.go
flaper87: I actually think it's fine to propagate the server tags down to the network as it creates a relationship between the resources, it makes it easier to filter, and to clean up later on. To be clear, I think this is a good thing only within the context of the cluster-api-provider-openstack.

That said, I didn't want to touch this part of the code much as it's not really the main goal of this PR

pkg/apis/openstackproviderconfig/v1alpha1/types.go
flaper87: done

config/crds/openstackproviderconfig_v1alpha1_openstackproviderspec.yaml
flaper87: done!

pkg/cloud/openstack/clients/machineservice.go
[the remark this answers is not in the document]
jichenjc: ok, I can patch it within my patch
I am going to propose a add 'description' for the vm and it starts from v2.19...

pkg/cloud/openstack/clients/machineservice.go
jichenjc: I agree, I think we should use one PR for one thing, this add can be another PR and more opionion got

pkg/cloud/openstack/clients/machineservice.go
jichenjc: ok. 2.26 means support add tags and 2.52 means support add tags *during* boot command
 :)

cmd/clusterctl/examples/openstack/machines.yaml.template
```diff
@@ -24,6 +24,10 @@ items:
...
+          - <Your Tag>
+        server_metadata:
```
chrigl: Also `serverMetadata`

cmd/clusterctl/examples/openstack/machines.yaml.template
```diff
@@ -52,5 +56,9 @@ items:
           name: worker-user-data
           namespace: openstack-provider-system
         trunk: false
+        tags:
+          - <Your Tag>
+        server_metadata:
```
chrigl: `serverMetadata`

docs/config.md
```diff
@@ -176,3 +176,39 @@ You can specify multiple networks (or subnets) to connect your server to. To do
...
+- apiVersion: "cluster.k8s.io/v1alpha1"
+  kind: Machine
+  metadata:
+    generateName: openstack-node-
+    labels:
+      set: node
+  spec:
+    providerSpec:
+      value:
+        server_metadata:
```
chrigl: `serverMetadata`

iamemilio: Ah, I will fix all instances of this

PR #261, opened 2019-03-18
make generate-yaml.sh work from anywhere

jichenjc:
some folder not workable through relative path

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/openstack/generate-yaml.sh
+++ b/cmd/clusterctl/examples/openstack/generate-yaml.sh
@@ -122,9 +122,9 @@ MACHINE_CONTROLLER_SSH_PRIVATE_FILE=openstack_tmp
 MACHINE_CONTROLLER_SSH_HOME=${HOME}/.ssh/
 
 # Set up the output dir if it does not yet exist
-mkdir -p out
-cp -n cluster.yaml out/cluster.yaml
-cp -n machines.yaml.template out/machines.yaml
+mkdir -p $PWD/out
+cp -n $PWD/cluster.yaml $PWD/out/cluster.yaml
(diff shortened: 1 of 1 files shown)


cmd/clusterctl/examples/openstack/generate-yaml.sh
```diff
@@ -122,9 +122,9 @@ MACHINE_CONTROLLER_SSH_PRIVATE_FILE=openstack_tmp
 MACHINE_CONTROLLER_SSH_HOME=${HOME}/.ssh/
 
 # Set up the output dir if it does not yet exist
-mkdir -p out
-cp -n cluster.yaml out/cluster.yaml
-cp -n machines.yaml.template out/machines.yaml
+mkdir -p $PWD/out
```
gyliu513: One question for this: If run anywhere, then when we run `git status`, git may treat this as a new added file if this folder was in the source code of openstack provider. Currently we are adding `out` to `.gitignore` at https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/.gitignore#L1

jichenjc: you are right, I would say the original purpose of this file seems to be make the generateyaml file running anywhere and some stuffs missing so it can't 

it's not a big deal if we have this out file somewhere and manually delete/move it ? apparently it's not possible to add it into gitignore file

gyliu513: Yes, I'm ok with this proposal, we can get some comments from approvers ;-)

/lgtm

ainmosni:
I have no problems with this.

PR #262, opened 2019-03-18
Add configuration for create timeout

jichenjc:
timeout is hard code and sometimes it's might be need to wait
for more than 5 min for instance create in some cloud.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -280,3 +280,9 @@ Instead of tagging, you also have the option to add metadata to instances. This 
         securityGroups:
    ...
    ```
+
+## Timeout settings
+During some heavy workload cloud, the time for create and delete openstack instance might takes long time, by default it's 5 minute.
+you can set:
+`CLUSTER_API_OPENSTACK_INSTANCE_DELETE_TIMEOUT` for instance delete timeout value.
+`CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT` for instance create timeout value.
--- a/pkg/cloud/openstack/machine/actuator.go
+++ b/pkg/cloud/openstack/machine/actuator.go
@@ -21,7 +21,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"reflect"
+	"strconv"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -46,8 +48,8 @@ const (
 	DisableTemplatingKey = "disableTemplating"
 	PostprocessorKey     = "postprocessor"
 
-	TimeoutInstanceCreate       = 5 * time.Minute
-	TimeoutInstanceDelete       = 5 * time.Minute
+	TimeoutInstanceCreate       = 5
+	TimeoutInstanceDelete       = 5
 	RetryIntervalInstanceStatus = 10 * time.Second
 
 	TokenTTL = 60 * time.Minute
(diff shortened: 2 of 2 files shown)


pkg/cloud/openstack/machine/actuator.go
```diff
@@ -66,6 +68,16 @@ func NewActuator(params openstack.ActuatorParams) (*OpenstackClient, error) {
 	}, nil
 }
 
+func getEnv(name string, def int) time.Duration {
```
ainmosni: The function is called 'getEnv' but it's very timeout specific. It should at least be renamed to getTimeout or something like that.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -66,6 +68,16 @@ func NewActuator(params openstack.ActuatorParams) (*OpenstackClient, error) {
 	}, nil
 }
 
+func getTimeout(name string, def int) time.Duration {
```
gyliu513: s/def/timeout

docs/config.md
```diff
@@ -235,3 +235,9 @@ Instead of tagging, you also have the option to add metadata to instances. This
         securityGroups:
    ...
    ```
+
+## Timeout settings
+During some heavy workload cloud, the time for create and delete openstack instance might takes long time, by default it's 5 minute.
+you can set:
+``CLUSTER_API_OPENSTACK_DELETE_TIMEOUT`` for instance delete timeout value.
+``CLUSTER_API_OPENSTACK_CREATE_TIMEOUT`` for instance create timeout value.
```
gyliu513: How about 

- s/CLUSTER_API_OPENSTACK_DELETE_TIMEOUT/CLUSTER_API_OPENSTACK_INSTANCE_DELETE_TIMEOUT
- s/CLUSTER_API_OPENSTACK_CREATE_TIMEOUT/CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT

You only need one `

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -150,7 +162,8 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
 		return oc.handleMachineError(machine, apierrors.CreateMachine(
 			"error creating Openstack instance: %v", err))
 	}
-	// TODO: wait instance ready
+	timeout := getTimeout("CLUSTER_API_OPENSTACK_CREATE_TIMEOUT", TimeoutInstanceCreate)
```
gyliu513: s/timeout/instanceCreateTimeout

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -234,7 +247,9 @@ func (oc *OpenstackClient) Update(ctx context.Context, cluster *clusterv1.Cluste
 		if err != nil {
 			klog.Errorf("delete machine %s for update failed: %v", currentMachine.ObjectMeta.Name, err)
 		} else {
-			err = util.PollImmediate(RetryIntervalInstanceStatus, TimeoutInstanceDelete, func() (bool, error) {
+			timeout := getTimeout("CLUSTER_API_OPENSTACK_DELETE_TIMEOUT", TimeoutInstanceDelete)
```
gyliu513: s/timeout/instanceDeleteTimeOut

docs/config.md
jichenjc: ok

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -69,6 +71,16 @@ func NewActuator(params openstack.ActuatorParams) (*OpenstackClient, error) {
...
 }
 
+func getTimeout(name string, timeout int) time.Duration {
+	if v := os.Getenv(name); v != "" {
+		timeout, err := strconv.Atoi(v)
+		if err == nil {
+			return time.Duration(timeout)
+		}
+	}
+	return time.Duration(def)
```
gyliu513: s/def/timeout

docs/config.md
```diff
@@ -257,3 +257,9 @@ Instead of tagging, you also have the option to add metadata to instances. This
         securityGroups:
    ...
    ```
+
+## Timeout settings
+During some heavy workload cloud, the time for create and delete openstack instance might takes long time, by default it's 5 minute.
```
gyliu513: Since those parameters are used in controller, I think the doc here should also be updated by clarifying how to update the deployment yaml file by adding those variables?

jichenjc: Do we really need those? it's general stuff of k8s?

PR #274, opened 2019-03-25
Add providerID

jichenjc:
refer to aws implementation, add providerID in exists function

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/openstack/machine/actuator.go
+++ b/pkg/cloud/openstack/machine/actuator.go
@@ -232,7 +232,7 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
 
 	}
 
-	return oc.updateAnnotation(machine, instance.ID)
+	return oc.updateMachine(machine, instance.ID)
 }
 
 func (oc *OpenstackClient) Delete(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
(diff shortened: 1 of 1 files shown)


jichenjc:
need upgrade cluster-api.......

chrigl:
/hold

Just to not being merged accidentally. As @jichenjc mentioned... we need an update of cluster-api for this.

With having this, we might want to remove the annotation `openstack-resourceId`, because in fact this is the same.

jichenjc:
@chrigl can you share how to upgrade cluster-api? seems change the master branch to latest commit doesn't help, should we wait for a release of cluster-api?

chrigl:
@jichenjc We keep the master branch of `cluster-api`, since there is no release at the horizon. You could update it using `dep ensure --update sigs.k8s.io/cluster-api`... But the hard part is testing it. And please do it in a separate branch.

jichenjc:
@chrigl because we already updated the cluster-api to branch 0.1.0 , we can remove the /hold flag here? Thanks

frobware:
I plan to update https://github.com/kubernetes/autoscaler/pull/1866 so that mapping between nodes and machines relies solely on the ProviderID being set -- this PR would help that goal.

chrigl:
@jichenjc Could you fit that into the refactor of #383 (which is already in master)?

CamelCaseNotation:
What's the hold up on merging this?

jichenjc:
we should already have this provider ID in latest code already

PR #277, opened 2019-03-26
Added "new" node taints to tolerations

chrigl:

The old labels are removed.

added `node.kubernetes.io/not-ready` and
`node.kubernetes.io/unreachable`

See https://github.com/kubernetes/kubernetes/pull/73001

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -42,12 +42,6 @@ spec:
       - key: CriticalAddonsOnly
         operator: Exists
       - effect: NoExecute
-        key: node.alpha.kubernetes.io/notReady
-        operator: Exists
-      - effect: NoExecute
-        key: node.alpha.kubernetes.io/unreachable
-        operator: Exists
-      - effect: NoExecute
         key: node.kubernetes.io/not-ready
         operator: Exists
       - effect: NoExecute
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -47,6 +47,12 @@ spec:
       - effect: NoExecute
         key: node.alpha.kubernetes.io/unreachable
         operator: Exists
+      - effect: NoExecute
+        key: node.kubernetes.io/not-ready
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -47,6 +47,12 @@ spec:
       - effect: NoExecute
         key: node.alpha.kubernetes.io/unreachable
         operator: Exists
```
jichenjc: should we remove those? seems those are alpha and per existing taints this need to be cleaned?

gyliu513: I think we still need to keep this in order to make sure old k8s still works?

jichenjc:
the commit message say 'The old labels are removed.' ......
I agree it's reasonable to keep old things for backward compatible purpose

config/manager/manager.yaml
jichenjc: the commit message say 'The old labels are removed.' ......
I agree it's reasonable to keep old things for backward compatible purpose

chrigl: Hmm. The taints were changed in 1.9 (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#scheduling)

I think we safely could remove the old ones. What do you think @gyliu513 @jichenjc?

jichenjc:
it's removed in 1.9 so very unlikely someone will use it with cluster-api? I vote to remove since it's fairly old now..

config/manager/manager.yaml
jichenjc: it's removed in 1.9 so very unlikely someone will use it with cluster-api? I vote to remove since it's fairly old now..

ainmosni: I agree, I see no reason for why we should support unsupported versions of k8s. :)

gyliu513: +1 to remove as well

Issue #268, opened 2019-03-20
Get error 'Resource not found' when creating instance from machine service.


morvencao:
### Issue description:

After I workaround issue of [`unknown authority CA` issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/264), I encounter another issue when creating openstack instances from machine service.

See logs of cluster-api-controller:
```
I0320 07:24:36.575242       1 controller.go:114] Running reconcile Machine for openstack-master-5ptq7
I0320 07:24:39.329965       1 controller.go:184] Reconciling machine object openstack-master-5ptq7 triggers idempotent create.
E0320 07:24:52.032660       1 actuator.go:319] Machine error openstack-master-5ptq7: error creating Openstack instance: Create new server err: Resource not found
W0320 07:24:52.032882       1 controller.go:186] unable to create machine openstack-master-5ptq7: error creating Openstack instance: Create new server err: Resource not found
```
Here is my `clouds.yaml` and `machines.yaml`

```
clouds:
  openstack:
    auth:
      auth_url: https://xxxxx.com:5000/v3
      username: "xxxxxxx"
      password: "xxxxxxxxx"
      project_id: cfc34894651f4035a09f623a926af2d9
      project_name: "xxxxx"
      domain_name: "xxx"
      user_domain_name: "xxx"
    region_name: "RegionOne"
    interface: "public"
    identity_api_version: 3
```

```
items:
- apiVersion: "cluster.k8s.io/v1alpha1"
  kind: Machine
  metadata:
    generateName: openstack-master-
    labels:
      set: master
  spec:
    providerSpec:
      value:
        apiVersion: "openstackproviderconfig/v1alpha1"
        kind: "OpenstackProviderSpec"
        flavor: m1.large
        image: u16.04-2018-23-1
        sshUserName: ubuntu
        keyName: cluster-api-provider-openstack
        availabilityZone: nova
        networks:
        - uuid: b518e837-9c2a-4ea8-a9ef-a2bd4d5764de
        floatingIP: x.x.x.x
        securityGroups:
        - uuid: 08fd7463-bdc0-4f5e-922f-697332e35b0b
        - uuid: 0bd39afa-90f5-48fa-bc5b-d08f011e930f
        - uuid: 54a80908-9f42-4dbe-b0c1-623cd852ab4f
        - uuid: dd2abe1a-3c0e-4cdd-adec-e906a328f25a
        tags:
        - master
[...]

PR #279, opened 2019-03-27
Add container linux support.

### LINKED ISSUE #272: support CoreOS
coreos is important part of openshift IPI and UPI in the future..we should consider to enable it

ainmosni:
This adds Container Linux support. The way this is done is via Container
Linux configuration. To be able to make this possible, this patch adds:

* Postprocessor support to the userData secrets.
    This was needed so that we could add the container linux yaml file to the
    secret instead of an ignition file. Besides the yaml being much more
    readable and maintainable, having template variables in the ignition
    file proved to be much more difficult and fragile than wanted. To
    work around these issues, I've chosen to add postprocessor support
    to the userData injection.
* Create a coreos overlay to the manager kustomization.
    Container Linux has a read-only /usr, which made the mount of
    /usr/bin/kubeadm in the manager container problematic. This
    customization changes /usr/bin/kubeadm to /opt/bin/kubeadm. This is
    where kubeadm lives on Container Linux. Adding this overlay also
    required me to break up the manager kustomization into its seperate
    components. And move the kustomization files to base/ and
    overlays/ directories in config.
* Create common.yaml, master.yaml, and worker.yaml configs.
    This is the actual configuration for container linux. common.yaml
    contains the configuration that's common to both controlplane and worker
    nodes, master.yaml has the relevant configuration to controlplane
    nodes, and worker.yaml has the configuration relevant to worker
    nodes. generate-yaml.sh will use yq to merge them into the proper
    configuration files.
* Change generate-yaml.sh to support Container Linux.
* New dependency on the container linux ct config transpiler for
  postprocessing support.

**Which issue(s) this PR fixes** :
Fixes #272

**Release note**:

```release-note
* Adds Container Linux support to the OS examples.
* Adds postprocessors to userData templates, with the Container Linux config transpiler as the first postprocessor.
```


jichenjc:
went through code a little bit and hopefully I can try this soon

some comments , config/rbac/* is generated by makefile , so rename this need change Makefile as well
I missed the update? 
config/rbac/rbac_role.yaml → config/base/rbac/rbac_role.yaml

ainmosni:
@jichenjc Ah yes, thanks, I missed that, will fix later today.

ainmosni:
@jichenjc I can't see anything about it in the Makefile but I was informed that I'm breaking kubebuilder conventions with this. I'll look into it tomorrow morning and potentially move some stuff around.

jichenjc:
try make manifests then you will find the crds/rbac will be put into some directory
you may want to adjust the folder path @ainmosni

ainmosni:
@jichenjc After some looking around I was forced to move the overlays out of the config directory and into their own (overlays-config, because kustomize didn't like config-overlays). Not a pretty solution, but it works.

iamemilio:
Yeah, the file structure we are using is not the way Kustomize expects it to be. This was due to a couple of issues, as well as the way that the file structure of the project is formatted. It would be good in the future for us to try to conform to the formatting that Kustomize expects, especially since the newer version of Kustomize enforce it more strictly. That being said, for the sake of this pull request, the way that you have things set up is fine.

iamemilio:
Sorry I havn't had time to look at this for real yet. I will give you a review tomorrow.

iamemilio:
When testing this with just a ubuntu image, I got the following error:
```
I0405 18:07:57.620047       1 controller.go:184] Reconciling machine object openstack-master-dvwkd triggers idempotent create.
E0405 18:08:03.837588       1 actuator.go:319] Machine error openstack-master-dvwkd: error creating Openstack instance: Create new server err: Resource not found
W0405 18:08:03.838758       1 controller.go:186] unable to create machine openstack-master-dvwkd: error creating Openstack instance: Create new server err: Resource not found
```

This was an issue that we had run into before. Here is the tracking for it https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/268. I think that if you just rebase your changes against the master, it should be resolved. Other than that, just a nit pick or two :)

cmd/clusterctl/examples/openstack/provider-component/user-data/coreos/templates/common.yaml
```diff
@@ -0,0 +1,223 @@
...
+            # This is a file that "kubeadm init" and "kubeadm join" generates at runtime, populating the KUBELET_KUBEADM_ARGS variable dynamically
+            EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
+            # This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
+            # the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
+            EnvironmentFile=-/etc/default/kubelet
+            ExecStart=
+            ExecStart=/opt/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS
+locksmith:
+  # Set to off, install https://github.com/coreos/container-linux-update-operator once the cluster is up.
+  reboot_strategy: "off"
```
iamemilio: This happens a lot if you use a text editor like vscode, it tends to not put a newline at the end of the file. It doesn't effect anything code wise, it just makes github put those annoying red markers there. It would be much appreciated if you could add newlines to the end of any file lacking them.

ainmosni: Added the newlines. :)

ainmosni:
@iamemilio Rebased with master.

iamemilio:
I'll give it an end to end in a few minutes

ainmosni:
@iamemilio Thanks :)

PR #281, opened 2019-03-27
Clearer tagging UI and Bugfix

### LINKED ISSUE #280: Clarify Tagging UI
The tagging system is confusing as it is. Proposed changes:

```yaml
Tags:
  Cluster:
      Tags:
          - a tag applied to all resources created by capo
  Server:
    Tags:
      - a tag applied to servers only
    Disable: false // Default:false; for systems that don't support nova 2.52
  Network:
    Tags:
      - a tag for networking resources only
```

@chrisgl, @flaper87: thoughts?

iamemilio:
1. Fix tagging bug introduced by egarcia
2. Allow cluster scoped, or machine specific tagging of resources

Fixes #280

**Special notes for your reviewer**: None

**Release note**: 

```release-note
NONE
```

### DIFF
--- a/cmd/clusterctl/examples/openstack/cluster.yaml
+++ b/cmd/clusterctl/examples/openstack/cluster.yaml
@@ -13,4 +13,6 @@ spec:
       value:
         apiVersion: "openstackproviderconfig/v1alpha1"
         kind: "OpenstackProviderSpec"
+        tags:
+          - a_cluster_wide_tag
 
--- a/cmd/clusterctl/examples/openstack/machines.yaml.template
+++ b/cmd/clusterctl/examples/openstack/machines.yaml.template
@@ -29,7 +29,7 @@ items:
           namespace: openstack-provider-system
         trunk: false
         tags:
-          - <Your Tag>
+          - a-machine-specific-tag
         serverMetadata:
           key: value
     versions:
@@ -61,7 +61,7 @@ items:
           namespace: openstack-provider-system
         trunk: false
         tags:
-          - <Your Tag>
+          - a-machine-specific-tag
         serverMetadata:
           key: value
     versions:
--- a/docs/config.md
+++ b/docs/config.md
@@ -176,8 +176,31 @@ You can specify multiple networks (or subnets) to connect your server to. To do 
           - subnet_id: your_subnet_id
 ```
 
-## Tagging Instances
-Tags can be added to instances on startup by populating the tags array where it says `<Your Tags>`. You can remove this if you do not want to tag your instances. Please note that your Nova api must be at least mivroversion 2.52 to use this api! Please refer to the following usage example:
+## Tagging
+By default, all resources will be tagged with the values: `clusterName` and `cluster-api-provider-openstack`. The minimum microversion of the nova api that you need to support server tagging is 2.52. If your cluster does not support this, then disable tagging servers by setting `disableServerTags: true` in cluster.yaml. By default, this value is false, so there is no need so set it in machines.yaml. If your cluster supports tagging servers, you have the ability to tag all resources created by the cluster in the cluster.yaml script. Here is the example of the tagging options available in cluster.yaml.
+
+```yaml
+apiVersion: "cluster.k8s.io/v1alpha1"
+kind: Cluster
+metadata:
+  name: test1
+spec:
+    clusterNetwork:
+        services:
+            cidrBlocks: ["10.96.0.0/12"]
+        pods:
+            cidrBlocks: ["192.168.0.0/16"]
+        serviceDomain: "cluster.local"
+    providerSpec:
+      value:
+        apiVersion: "openstackproviderconfig/v1alpha1"
+        kind: "OpenstackProviderSpec"
+        disableServerTags: false
+        tags:
+          - cluster-tag
+```
+
+To tag resources specific to a machine, add a value to the tags field in machines.yaml like this.
 
 ```yaml
 - apiVersion: "cluster.k8s.io/v1alpha1"
@@ -190,8 +213,7 @@ Tags can be added to instances on startup by populating the tags array where it 
     providerSpec:
       value:
         tags:
-          - tag1
-          - tag2
+          - machine-tag
 ```
 
 ## Metadata
--- a/pkg/apis/openstackproviderconfig/v1alpha1/types.go
+++ b/pkg/apis/openstackproviderconfig/v1alpha1/types.go
@@ -40,6 +40,7 @@ type OpenstackProviderSpec struct {
 
 	// The flavor reference for the flavor for your server instance.
 	Flavor string `json:"flavor"`
+
 	// The name of the image to use for your server instance.
 	// If the RootVolume is specified, this will be ignored and use rootVolume directly.
 	Image string `json:"image"`
@@ -69,7 +70,7 @@ type OpenstackProviderSpec struct {
 	// Whether the server instance is created on a trunk port or not.
 	Trunk bool `json:"trunk,omitempty"`
 
-	// Server tags
+	// Machine tags
 	// Requires Nova api 2.52 minimum!
 	Tags []string `json:"tags,omitempty"`
 
@@ -201,6 +202,12 @@ type OpenstackClusterProviderSpec struct {
 	// and API access from everywhere, and another one that allows all traffic to/from
 	// machines belonging to that group. In the future, we could make this more flexible.
 	ManagedSecurityGroups bool `json:"managedSecurityGroups"`
+
+	// Tags for all resources in cluster
+	Tags []string `json:"tags,omitempty"`
+
+	// Default: True. In case of server tag errors, set to False
+	DisableServerTags bool `json:"disableServerTags,omitempty"`
 }
 
 // +genclient
--- a/pkg/cloud/openstack/clients/machineservice.go
+++ b/pkg/cloud/openstack/clients/machineservice.go
@@ -35,6 +35,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	netext "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions"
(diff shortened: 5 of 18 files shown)


chrigl:
I like it. The only thing that comes to my mind is: Do we want to define cluster tags on the `OpenstackClusterProviderSpec` rather than on the individual machine?

docs/config.md
```diff
@@ -177,8 +177,8 @@ You can specify multiple networks (or subnets) to connect your server to. To do
           - subnet_id: your_subnet_id
 ```
 
-## Tagging Instances
-Tags can be added to instances on startup by populating the tags array where it says `<Your Tags>`. You can remove this if you do not want to tag your instances. Please note that your Nova api must be at least mivroversion 2.52 to use this api! Please refer to the following usage example:
+## Tagging
+By default, all resources will be tagged with the values: `clusterName` and `cluster-api-provider-openstack`. The minimum microversion of the nova api that you need to support server tagging is 2.52. If your cluster does not support this, then disable tagging servers by setting `tags.server.disable: true`. By default, this value is false, so there is no need so set it in machines.yaml. If your cluster supports tagging servers, and you want to create tags that only get added to servers, then create an entry in the `tags.server.tags` list. To create tags that are only applied to networking resources, create an entry in the `tags.network.tags` list. Otherwise, to create a tag that gets added to every resource, then create an entry in the `tags.cluster.tags` list. Here is an example of this syntax.
```
jichenjc: s/no need so set it/no need to set it/

cmd/clusterctl/examples/openstack/machines.yaml.template
```diff
@@ -57,7 +66,15 @@ items:
...
-          - <Your Tag>
+          cluster:
+            tags:
+              - a_tag_applied_to_all_created_resources
+          server:
+            tags:
+              - a_tag_only_applied_to_servers
+          network:
+            tags:
+              - a_tag_only_applied_to_network_resources
```
jichenjc: sorry but I don't understand why cluster/network tag applied in this machine template? my understanding cluster/network should only apply to cluster ?

docs/config.md
jichenjc: I also would rename this to tags.server.tags to True so by default we enable it and someone can disable it because of this API version (just more clear to me   :) )

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -81,6 +80,37 @@ type OpenstackProviderSpec struct {
 	ConfigDrive *bool `json:"configDrive,omitempty"`
 }
 
+type Tags struct {
+	// Tags added to all resources created by cluster
+	// Default tags include: cluster-api-provider-openstack, clusterName
```
jichenjc: I think cluster-api-provider-openstack:clusterName should be enough?

pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -81,6 +80,37 @@ type OpenstackProviderSpec struct {
...
+	Cluster ClusterTags `json:"cluster,omitempty"`
+
+	// Server tags: tags servers only
+	Server ServerTags `json:"server,omitempty"`
+
+	// Port tags: tags only applied to networking resources created
```
jichenjc: Port tags seems confusing... only ports or other stuffs related to network(network, secgroup) will be taged?

cmd/clusterctl/examples/openstack/machines.yaml.template
iamemilio: cluster, server, and network are seperate. Cluster tags get applied to all created resources. Server tags get applied to servers, and network tags only get applied to networks.

docs/config.md
iamemilio: the problem is that tags.<resource>.tags is the syntax we used to specify our list of tags

jichenjc: sorry, I didn't get it , I am suggesting to make it by default to be enable then user can change it to disable, which makes me eaiser to understand...

cmd/clusterctl/examples/openstack/machines.yaml.template
jichenjc: understood, but this applied in machine.yaml.template makes me confusing, should cluster tags applied to cluster wide yaml or something like that?

docs/config.md
iamemilio: sure, that makes sense

pkg/apis/openstackproviderconfig/v1alpha1/types.go
iamemilio: That sould say network tags, and yeah, it is only tagging resources created by running cluster-api-provider-openstack

docs/config.md
iamemilio: Since the zero value of booleans in go is false, in order to make it so the user doesn't have to set this value if they don't need it, it should be disableServerTags: true | false

iamemilio: sorry about that, I think I will have to leave it how it is

pkg/apis/openstackproviderconfig/v1alpha1/types.go
iamemilio: These are the default tags that the project already had. I would prefer not to change them in this pull request, but can in another one.

cmd/clusterctl/examples/openstack/machines.yaml.template
iamemilio: my new revision will address this

iamemilio:
@chrigl @jichenjc Ready for review

iamemilio:
@chaosaffe please let me know if this would effect the cluster delete in any way

cmd/clusterctl/examples/openstack/cluster.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "cluster.k8s.io/v1alpha1"
 kind: Cluster
 metadata:
-  name: test1
+  name: emilio
```
chrigl: I think this was by accident ;)

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -342,7 +352,34 @@ func GetSecurityGroups(is *InstanceService, sg_param []openstackconfigv1.Securit
...
+		return "", fmt.Errorf("no image with the name %s could be found", imageName)
+	case 1:
+		return allImages[0].ID, nil
+	default:
+		return "", fmt.Errorf("too many images with the name, %s, were found", imageName)
+	}
+}
+
+// InstanceCreate creates a compute instance
+func (is *InstanceService) InstanceCreate(clusterName string, name string, config *openstackconfigv1.OpenstackProviderSpec, cmd string, keyName string, cluster *openstackconfigv1.OpenstackClusterProviderSpec) (instance *Instance, err error) {
```
chrigl: I would like to change to order or parameters. And rename param `cluster` to `clusterSpec`.
`clusterName string, name string, clusterSpec *openstackconfigv1.OpenstackClusterProviderSpec, config *openstackconfigv1.OpenstackProviderSpec,...`, just because of convenience. The `machine.Actuator` also references cluster before machine.

pkg/cloud/openstack/machine/actuator.go
```diff
@@ -145,7 +145,16 @@ func (oc *OpenstackClient) Create(ctx context.Context, cluster *clusterv1.Cluste
 		userDataRendered = string(userData)
 	}
 
-	instance, err = machineService.InstanceCreate(fmt.Sprintf("%s/%s", cluster.ObjectMeta.Namespace, cluster.Name), machine.Name, providerSpec, userDataRendered, providerSpec.KeyName)
+	var clusterSpec openstackconfigv1.OpenstackClusterProviderSpec
```
chrigl: there is already a function doing this for us. It also checks `Value` for being nil.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/apis/openstackproviderconfig/v1alpha1/register.go#L42

cmd/clusterctl/examples/openstack/cluster.yaml
iamemilio: haha yeah it is

pkg/cloud/openstack/machine/actuator.go
iamemilio: ah, good to know. Ill change that

chrigl:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/dd569ed5e4801cf69739a88361a7d0f060ce1215/pkg/cloud/openstack/clients/machineservice.go#L549

You forgot to remove this if-block. Currently the `config.Tags` are always set, no matter what you decided above.

chrigl:
/approve

@jichenjc could you add the lgtm when this is ok from you point of view?

pkg/cloud/openstack/clients/machineservice.go
```diff
@@ -479,28 +518,34 @@ func (is *InstanceService) InstanceCreate(clusterName string, name string, confi
...
+		// to use tags.
+		is.computeClient.Microversion = "2.52"
+	}
+
+	// Get image ID
+	imageID, err := getImageID(is, config.Image)
+	if err != nil {
+		return nil, fmt.Errorf("Create new server err: %v", err)
+	}
+
```
jichenjc: sorry for late reply ,but this seems different to the tag purpose ? 

https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail
imageRef (Optional) | body | string | The UUID of the image to use for your server instance. This is not required in case of boot from volume. In all other cases it is required and must be a valid UUID otherwise API will return 400.
-- | -- | -- | --

so this logic seems force Image to be existing, I am ok to this since our use case seems should comes from 
image but I just want to raise this up and get agreement @iamemilio @chrigl

jichenjc: other than this ,lgtm

chrigl: @jichenjc You are right, that was not on my radar. Not having an image configured is a valid case if there is a bootvolume configured.
@iamemilio We could return an empty string from getImageID if the image name is also an empty string.

We need the code to get the imageID from glance over passing in the imageName, because under the covers, gophercloud does the same thing... but when using microversion >= 2.36, the route in nova which gophercloud attempts to use, is gone. Thus it lead to #268.

jichenjc: ok thanks for the sharing, I don't know the root cause of #268 and this seems a reasonable change

PR #291, opened 2019-04-01
Enabled kubeadm debug log for better troubleshooting.

gyliu513:

Fixes #

**Release note**:

```release-note
Enabled debug log for kubeadm post install.
```

### DIFF
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
@@ -147,7 +147,7 @@ networking:
   serviceSubnet: ${SERVICE_CIDR}
 EOF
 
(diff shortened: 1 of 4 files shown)


xunpan:
/lgtm
It is example code and already set `-x` for shell. I think add this detailed log helps in trying the case.

gyliu513:
This can  also help me to debug the issue like https://github.com/kubernetes/kubeadm/issues/1474

jichenjc:
I think a better way is to provide a configuration option such as 'debug' so that we can use in clusterctl to pass the flag for whole debug purpose, I will open an issue in cluster-api

jichenjc:
ok, already a '-v' flag in clusterctl and seems this is another good example to discuss along with 
https://github.com/kubernetes-sigs/cluster-api/issues/866

a common configuration for those settings seems reasonable

Issue #292, opened 2019-04-01
disable swap by default


jichenjc:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

see this [ERROR Swap]: running with swap on is not supported. Please disable swap
during kubeadm start, it's not likely all distro will disable swap by default
we should disable in user-data.sh

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

jichenjc:
update user-data.sh should be enough...I will work on it

jichenjc:
close this per https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/293

PR #293, opened 2019-04-02
disable swap in master and node

jichenjc:
encounter issue when kubeadm start as swap is on
so disable swap during start of user-data execution.

Fixes Bug #292

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
@@ -14,6 +14,9 @@ POD_CIDR={{ .PodCIDR }}
 SERVICE_CIDR={{ .ServiceCIDR }}
 ARCH=amd64
 
+swapoff -a
+# disable swap in fstab
+sed -i.bak -r 's/(.+ swap .+)/#\1/' /etc/fstab
 # Getting master ip from the metadata of the node. By default we try the public-ipv4
 # If we don't get any, we fall back to local-ipv4 and in the worst case to localhost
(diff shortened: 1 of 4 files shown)


xunpan:
it works in my environment.

/lgtm

cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
```diff
@@ -14,6 +14,7 @@ POD_CIDR={{ .PodCIDR }}
 SERVICE_CIDR={{ .ServiceCIDR }}
 ARCH=amd64
 
+swapoff -a
```
chrigl: Hmm... is this enough in this context? If there is swap enabled in the image, is there a need to remove id from fstab as well?

jichenjc: maybe ok because kubeadm has some dependency? this user-data will only be executed once ...

root@kvm-017751:~/go/src/sigs.k8s.io/cluster-api/cmd# swapon -a
root@kvm-017751:~/go/src/sigs.k8s.io/cluster-api/cmd# kubectl get nodes
NAME       STATUS   ROLES    AGE    VERSION
minikube   Ready    master   120m   v1.12.7

chrigl: yeah. But they check it because having swap enabled is a bad thing for kubernetes, and after rebooting, the swap will be enabled again (if there are things left in fstab).

PR #297, opened 2019-04-03
Delete project_name from clouds.yaml.

gyliu513:

Fixes #

**Release note**:

```release-note
NONE
```


hidekazuna:
This is reasonable fix. project_name does not reflect on cloud.conf which is generated in generate-yaml.sh.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/generate-yaml.sh#L149-L155

PR #300, opened 2019-04-04
Add Kubeadm and Kubelet versions info in README.md

hidekazuna:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,9 @@ This provider's versions are able to install and manage the following versions o
 |-|-|-|
 |OpenStack Provider v1alpha1 (ea309e7f)|✓|✓|
 
+Kubernetes control plane and Kubelet versions are defined in `spec.versions.controlPlane` and `spec.versions.kubelet` of `cmd/clusterctl/examples/openstack/machines.yaml.template` respectively.
+You can generate `cmd/clusterctl/examples/openstack/out/machines.yaml` by running the `generate-yaml.sh` from the template and change the versions if you want.
+
 **NOTE**: Because the user is able to customize any `user-data`, it is also possible to deploy older versions.
 But we won't provide any examples or working templates. See [user-data in the examples](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/master/cmd/clusterctl/examples/openstack/provider-component/user-data).
 


README.md
```diff
@@ -31,6 +31,8 @@ This provider's versions are able to install and manage the following versions o
 |-|-|-|
 |OpenStack Provider v1alpha1 (ea309e7f)|✓|✓|
 
+Kubeadm and Kubelet versions are defined in the the file `cmd/clusterctl/examples/openstack/out/machines.yaml` which is generated by running the `generate-yaml.sh`.
```
gyliu513: Can we highlight the version is in `spec.versions.kubelet`?

hidekazuna: Thanks, I updated.

README.md
```diff
@@ -31,6 +31,8 @@ This provider's versions are able to install and manage the following versions o
 |-|-|-|
 |OpenStack Provider v1alpha1 (ea309e7f)|✓|✓|
 
+Kubeadm and Kubelet versions are defined in `spec.versions.kubelet` of `cmd/clusterctl/examples/openstack/out/machines.yaml` which is generated by running the `generate-yaml.sh`.
```
jichenjc: nit: maybe mention the template machines.yaml.template as well? and you may mention 
cmd/clusterctl/examples/openstack/out/machines.yaml is by default ,it depends where you run generate-yaml.sh

hidekazuna:
@jichenjc Thanks for your comment. Should we update machines.yaml.template rather than machines.yaml?

jichenjc:
I think we can generate the machines.yaml and change it , no need to touch the template
but provide some info in the README will be helpful as well since we anyway update the doc to provide more info

hidekazuna:
@jichenjc OK, I updated to refer to the template.

README.md
```diff
@@ -31,6 +31,9 @@ This provider's versions are able to install and manage the following versions o
 |-|-|-|
 |OpenStack Provider v1alpha1 (ea309e7f)|✓|✓|
 
+Kubeadm and Kubelet versions are defined in `spec.versions.controlPlane` and `spec.versions.kubelet` of `cmd/clusterctl/examples/openstack/machines.yaml.template` respectively.
+You generate `cmd/clusterctl/examples/openstack/out/machines.yaml` by running the `generate-yaml.sh` from the template and change the versions if you want.
```
gyliu513: s/You/You can

hidekazuna: @gyliu513 Thanks, Fixed.

gyliu513:
/lgtm

Thanks @hidekazuna for the update!

README.md
```diff
@@ -31,6 +31,9 @@ This provider's versions are able to install and manage the following versions o
 |-|-|-|
 |OpenStack Provider v1alpha1 (ea309e7f)|✓|✓|
 
+Kubeadm and Kubelet versions are defined in `spec.versions.controlPlane` and `spec.versions.kubelet` of `cmd/clusterctl/examples/openstack/machines.yaml.template` respectively.
```
chrigl: Maybe we should rephrase to "Kubernetes versions are defined in ...". I would not mention kubeadm here because it is an implementation detail. And in our particular case, the `versions.kubelet` is also used as the version for `kubeadm`.

jichenjc: seems we do have a control pane version and kubelet version ,maybe mention both of them ?

hidekazuna: How about this? "Kubernetes control plane and Kubelet versions are defined in `spec.versions.controlPlane` and `spec.versions.kubelet` of `cmd/clusterctl/examples/openstack/machines.yaml.template` respectively."

PR #312, opened 2019-04-19
WIP Remove KUBELET_VERSION from master user data.

gyliu513:
The KUBELET_VERSION was only used when installing worker node, and
there is no use to set this for master user data.

Fixes #

**Release note**:

```release-note
NONE
```

/cc @jichenjc @chrigl

### DIFF
--- a/cmd/clusterctl/examples/openstack/machines.yaml.template
+++ b/cmd/clusterctl/examples/openstack/machines.yaml.template
@@ -33,7 +33,6 @@ items:
         serverMetadata:
           key: value
     versions:
-      kubelet: 1.14.0
       controlPlane: 1.14.0
 - apiVersion: "cluster.k8s.io/v1alpha1"
   kind: Machine
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
@@ -2,8 +2,6 @@
 set -e
 set -x
 (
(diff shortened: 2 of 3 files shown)


jichenjc:
and to confirm, we only honor the spec.versions.controlPlane in our scripts to use kubeadm install control pane including kubelet, right?

gyliu513:
Seems centos and ubuntu has some different logic, let me consolidate the logic for those two platforms.

jichenjc:
also, seems coreos need to be double checked whether it's affected

jichenjc:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/provider-component/user-data/coreos/templates/common.yaml#L73

it's using Machine.Spec.Versions.Kubelet definition , this means, it doesn't honor the control plane info
I can add a PR to handle it ,before that , I give another thought, it's a little bit confusing now as we want to use `control plane` as kubeadm/kubectl/kubelet version for master node and `kubelet` as kubeadm/kubectl/kubelet version for worker node ? but anyway we still need kubeadm to be joined to master node and at least the naming convention is confusing here

gyliu513:
Thanks @jichenjc , let me check other cloud providers for this. Seems here the `control plane` should be the `Kubernetes installer` version and `kubelet` is `Kubernetes` version?

FYI @dims @chrigl any comments here? Thanks!

sbueringer:
@jichenjc I think we can close this one. Completly replaced in v1alpha2 through kubeadm bootstrapper

PR #320, opened 2019-04-25
Update keystone port number in README.md

hidekazuna:

Keystone port number 35357 is for admin user, cluster api user should use 5000.

Fixes #

**Release note**:

```release-note

```


jichenjc:
actually this is just an example, but you are right ,change to 5000 seems more fitful here

PR #321, opened 2019-04-25
Use apps/v1 for deployment apiVersion

jichenjc:
PR 319 fixed the deployment configuration file but failed to add kustomize update..

Error: failed to find an object with gvk.Gvk{Group:"apps", Version:"v1beta1", Kind:"Deployment"} to apply the patch

was caused by kustomize function, this is because the kustomize file didn't updated with the main manager file...
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #322, opened 2019-04-25
Add generate yaml test

### LINKED ISSUE #323: Add auto test for generate-yaml.sh
/kind feature

There is no test for generate-yaml.sh now, so sometimes the change will break this function
but developer won't pay attention to it

jichenjc:
Add test in Makefile (test) to include a test for generate-yaml.sh in order to avoid 
further change break the functions of yaml generation.

Fixes #323 

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,15 @@ HAS_MERCURIAL := $(shell command -v hg;)
 HAS_DEP := $(shell command -v dep;)
 HAS_LINT := $(shell command -v golint;)
 HAS_GOX := $(shell command -v gox;)
+HAS_YQ := $(shell command -v yq;)
+HAS_KUSTOMIZE := $(shell command -v kustomize;)
 GOX_PARALLEL ?= 3
 TARGETS ?= darwin/amd64 linux/amd64 linux/386 linux/arm linux/arm64 linux/ppc64le
 DIST_DIRS         = find * -type d -exec
+
+GENERATE_YAML_PATH=cmd/clusterctl/examples/openstack
+GENERATE_YAML_EXEC=generate-yaml.sh
+GENERATE_YAML_TEST_FOLDER=dummy-make-auto-test
 
 GOOS ?= $(shell go env GOOS)
 VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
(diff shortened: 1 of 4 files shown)


jichenjc:
this failure is valid, depends on #321 to fix the issue then rebase should work

cmd/clusterctl/examples/openstack/generate-yaml.sh
```diff
@@ -7,7 +7,7 @@ print_help()
   echo "$SCRIPT - generates a provider-configs.yaml file"
   echo ""
   echo "Usage:"
-  echo "$SCRIPT [options] <path/to/clouds.yaml> <cloud> <provider os: [centos,ubuntu,coreos]>"
+  echo "$SCRIPT [options] <path/to/clouds.yaml> <cloud> <provider os: [centos,ubuntu,coreos]> [output folder]"
   echo "options:"
   echo "-h, --help                    show brief help"
   echo "-f, --force-overwrite         if file to be generated already exists, force script to overwrite it"
```
gyliu513: Can you please update the help info as well for `output`?

Makefile
```diff
@@ -64,10 +70,30 @@ clusterctl:
...
 		cmd/clusterctl/main.go
 
-test: unit functional
+test: unit functional generate_yaml_test
 
 check: depend fmt vet lint
 
+generate_yaml_test:
+ifndef HAS_YQ
+		go get github.com/mikefarah/yq
```
gyliu513: One tab is good enough.

Ditto for others.

Makefile
```diff
@@ -64,10 +70,30 @@ clusterctl:
...
+ifndef HAS_KUSTOMIZE
+		# for now, higher version has some problem so we stick to 1.0.11
+		wget https://github.com/kubernetes-sigs/kustomize/releases/download/v1.0.11/kustomize_1.0.11_linux_amd64
+		mv kustomize_1.0.11_linux_amd64 /usr/local/bin/kustomize
+		chmod +x /usr/local/bin/kustomize
+endif
+		
+	# Create a dummy file for test only
+	echo 'clouds' > dummy-clouds-test.yaml
+	$(GENERATE_YAML_PATH)/$(GENERATE_YAML_EXEC) -f dummy-clouds-test.yaml openstack ubuntu $(GENERATE_YAML_TEST_FOLDER)
```
gyliu513: How to check if the command is succeed or not?

jichenjc: nothing to check ,just run the command to make sure we can generate yaml
for example , #321 actually can be found if we have this test

cmd/clusterctl/examples/openstack/generate-yaml.sh
jichenjc: um.. where to add those? I think we don't have the helpful info instead of the doc?

cmd/clusterctl/examples/openstack/generate-yaml.sh
```diff
@@ -7,7 +7,7 @@ print_help()
   echo "$SCRIPT - generates a provider-configs.yaml file"
   echo ""
   echo "Usage:"
-  echo "$SCRIPT [options] <path/to/clouds.yaml> <cloud> <provider os: [centos,ubuntu,coreos]>"
+  echo "$SCRIPT [options] <path/to/clouds.yaml> <cloud> <provider os: [centos,ubuntu,coreos]> [output folder]"
```
gyliu513: s/[output folder]/\<output folder\>

jichenjc: the output folder is optional.... in my memory<> means mandatory and [] means optional?

gyliu513: Ah, I see, thanks.

/lgtm

PR #324, opened 2019-04-25
add ignore file for manager

jichenjc:
the cmd/manager/manager is exec file and should not in git tree

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/.gitignore
+++ b/cmd/manager/.gitignore
@@ -1,0 +1,1 @@
+manager


cmd/manager/.gitignore
```diff
@@ -0,0 +1 @@
+manager
```
gyliu513: I think we already ignored this binary in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/.gitignore#L139 , it should be located in `bin/manager`?

jichenjc: I don't know, but I saw this manager in my cmd/manager folder now while definitely I didn't create the manger by running 'go build', so not sure where it comes from...

gyliu513: I usually use `make manager` in top dir.

jichenjc: ok, let's close this ,if I meet this again I will reopen it , thanks for the commetns

PR #335, opened 2019-04-28
init-cloud exit if the selinux did not enabled

hchenxa:

when selinux did not enabled, the init-cloud will be failed, so before run command setenforce, need to check if the selinux enabled or not.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
@@ -1,9 +1,8 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 set -x
 (
 KUBELET_VERSION={{ .Machine.Spec.Versions.Kubelet }}
-VERSION=v${KUBELET_VERSION}
 NAMESPACE={{ .Machine.ObjectMeta.Namespace }}
 MACHINE=$NAMESPACE
 MACHINE+="/"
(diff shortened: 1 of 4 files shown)


hchenxa:
cc @gyliu513

cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
```diff
@@ -58,25 +57,31 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cl
...
 }
 
 install_configure_docker
 
+# Get docker cgroup driver for kubelet configuration
+CG_DRIVER=$(docker info --format '{{.CgroupDriver}}')
+
 cat <<EOF > /etc/default/kubelet
-KUBELET_KUBEADM_EXTRA_ARGS=--cgroup-driver=systemd
+KUBELET_KUBEADM_EXTRA_ARGS=--cgroup-driver=$CG_DRIVER
```
gyliu513: @hchenxa would you please create another PR by adding this logic to ubuntu and coreos as well? Enable kubelet use same cgroup driver as docker.

cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/worker-user-data.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
gyliu513: Why updating here?

cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/master-user-data.sh
hchenxa: ok.

cmd/clusterctl/examples/openstack/provider-component/user-data/centos/templates/worker-user-data.sh
hchenxa: because for different OS or distruibution, the bash may under different path

PR #338, opened 2019-05-01
Remove CACertificate and CAPrivateKey

### LINKED ISSUE #333: the CAcert and CAkey from the cluster status was null
/kind bug

the cluster out have CAcertification and CAProvateKey, but the value seems be null always like below:
```
apiVersion: cluster.k8s.io/v1alpha1
kind: Cluster
metadata:
  creationTimestamp: "2019-04-28T05:47:04Z"
  finalizers:
  - cluster.cluster.k8s.io
  generation: 1
  name: test2
  namespace: default
  resourceVersion: "621"
  selfLink: /apis/cluster.k8s.io/v1alpha1/namespaces/default/clusters/test2
  uid: 0d2cf2d0-6979-11e9-840f-fa163eb27512
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
      - 10.96.0.0/12
  providerSpec:
    value:
      apiVersion: openstackproviderconfig/v1alpha1
      kind: OpenstackProviderSpec
      tags:
      - a_cluster_wide_tag
status:
  apiEndpoints:
  - host: 9.20.204.143
    port: 443
  providerStatus:
    CACertificate: null
    CAPrivateKey: null
    metadata:
      creationTimestamp: null
```

Does there need any specify configuration for this? 

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

jichenjc:
we are not providing any info about control pane CA key and pem
add those info will be confusing to end user
and seems add private key info into the status is not that secure..

Fixes #333 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/openstackproviderconfig/v1alpha1/types.go
+++ b/pkg/apis/openstackproviderconfig/v1alpha1/types.go
@@ -226,12 +226,6 @@ type OpenstackClusterProviderStatus struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	// CACertificate is a PEM encoded CA Certificate for the control plane nodes.
-	CACertificate []byte
-
-	// CAPrivateKey is a PEM encoded PKCS1 CA PrivateKey for the control plane nodes.
-	CAPrivateKey []byte
-
 	// Network contains all information about the created OpenStack Network.
 	// It includes Subnets and Router.
 	Network *Network `json:"network,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/apis/openstackproviderconfig/v1alpha1/types.go
```diff
@@ -226,11 +226,12 @@ type OpenstackClusterProviderStatus struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// TODO: create and provide this info later
```
chrigl: I think we should remove this entirely. Maybe this came only in due to an copy and paste error early on. And I am very opinionated in not having anything secret (CAPrivateKey) stored outside of a secret.

PR #339, opened 2019-05-01
Fix cluster nil lead to panic

### LINKED ISSUE #317: check nil cluster is provided
/kind bug

At the moment, cluster-api-provider-openstack panics with a nil pointer if there is no cluster
We might want to make sure to return an error if a nil cluster is provided.

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

jichenjc:
cluster variable might be be nil due to some configuration error
so this patch adds a check of cluster status before use it

Fixes #317 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/openstack/cluster/actuator.go
+++ b/pkg/cloud/openstack/cluster/actuator.go
@@ -41,6 +41,10 @@ func NewActuator(params providerv1openstack.ActuatorParams) (*Actuator, error) {
 
 // Reconcile creates or applies updates to the cluster.
 func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
+	if cluster == nil {
+		return fmt.Errorf("The cluster is nil, check your cluster configuration")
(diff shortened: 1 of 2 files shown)


pkg/cloud/openstack/cluster/actuator.go
```diff
@@ -42,6 +42,10 @@ func NewActuator(params providerv1openstack.ActuatorParams) (*Actuator, error) {
 // Reconcile creates or applies updates to the cluster.
 func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
 	klog.Infof("Reconciling cluster %v.", cluster.Name)
+	if cluster == nil {
```
gyliu513: This checking should right before the log at Line44?

jichenjc: right, I didn't pay attention, thanks for pointing out

PR #351, opened 2019-05-15
Add Record of instance action

### LINKED ISSUE #350: event creation during create / delete machine
/kind feature

we need something like event recording so that we can 'kubectl get events' to know whether the machine is really created/deleted

later on we can extend this to more places such as machine update etc

jichenjc:
instance create and delete can be have some events so we can record
them and for audit purpose.

Fixes #350 

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/klog"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/apis"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/controller"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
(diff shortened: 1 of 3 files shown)


hidekazuna:
Machine create and delete is defined in cluster-api. Why don't you implement in cluster-api?

jichenjc:
ok, let me try to post some question there..

jichenjc:
@hidekazuna aws implemented their own...
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/services/ec2/instances.go#L283

I created https://github.com/kubernetes-sigs/cluster-api/issues/949
for question about it..

chaosaffe:
Even if we do log out on the controller it would certainly be great to have the event recorded on the `Machine` as well. I like the idea of pulling this up to _cluster-api as well, but we can implement it on our own for now and surface it

PR #358, opened 2019-05-27
the generate.sh exited when the cluster.yaml exists

hchenxa:

below code will exited when the cluster.yaml exists.
```
# Set up the output dir if it does not yet exist
mkdir -p $PWD/$OUTPUT
cp -n $PWD/cluster.yaml.template $PWD/$OUTPUT/cluster.yaml  >>>>>>>> The return code was 1
cp -n $PWD/machines.yaml.template $PWD/$OUTPUT/machines.yaml 
```

so in order to avoid the file overwrite and keep the updated file of cluster.yaml and machines.yaml, create this PR.

Fixes #

**Release note**:

```release-note

```


hchenxa:
cc @gyliu513

Issue #382, opened 2019-06-30
Support for multi-node control plane


sbueringer:
/kind feature

Based on [ClusterAPI Provider Openstack: multi-node control plane](https://docs.google.com/document/d/11-xZ4_UpNYHYBNNkCPgsyEPKFKFS1DoJNfwM0nV_nn0/edit#) 

This is an umbrella issue for the implementation of this design.

It will be implemented via the following PRs:
* [x] Move services in individual packages for better separation of concerns like CAPA
  * PR: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/383 merged
* [x] Implement certificate handling based on Cluster CRD
  * PR open:  https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/394
* [x] Implement LoadBalancer (incl. FloatingIP) in front of the control plane nodes
  * See also https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/367
  * PR: open: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/401
* [ ] Generate kubeadm configs for multi-node control plane
  * WIP, but a preview is here: https://github.com/sbueringer/cluster-api-provider-openstack/pull/3

P.S. I have basically all the required code locally & working. It's just a matter of creating PRs, getting aggreement and merging it one after another :)

PR #394, opened 2019-07-11
Pr implement reconcile certs

sbueringer:
As the one before. This PR integrates code from CAPA. This PR will generate & store certificates 
in the Cluster CRD. With this we're now able to set certifcates on every control plane node (multi-node support will be added in a later PR). We're also able to generate a kubeconfig whenever we won't without ssh'ing on a control plane node

Another step towards multi-node control plane support. Without this we would have to copy certificates between control plane nodes

Implements part of #382


pkg/cloud/openstack/services/certificates/certificates.go
```diff
@@ -0,0 +1,334 @@
...
+type Config struct {
+	CommonName   string
+	Organization []string
+	AltNames     AltNames
+	Usages       []x509.ExtKeyUsage
+}
+
+// ReconcileCertificates generate certificates if none exists.
+func (s *Service) ReconcileCertificates(clusterName string, clusterProviderSpec *v1alpha1.OpenstackClusterProviderSpec) error {
+	if !clusterProviderSpec.CAKeyPair.HasCertAndKey() {
```
sbueringer: @jichenjc Your comment from the PR on the other repository:
> from code, looks like you are trying to generate those keys? if it's generated, then how do you store them and reuse? looks to me the Spec are the desired state but if you didn't give key at beginning, it's not desired state?? just curious ..

sbueringer: @jichenjc The logic is the following:

If no certificates are set in the Cluster CRD, ReconcileCertificates will generate them and add them to the Cluster CRD. But it's also possible (if wanted) to set the certificates when creating the Cluster CRD.

The main reasoning behind this is that we need a place to store certificates so we can give the same certifcates to multiple control plane nodes, without copying them from one host to another. Also it's now possible to generate a kubeconfig based on the Cluster CRD alone.

jichenjc:
I'll try this on my env and let you know :)

sbueringer:
> I'll try this on my env and let you know :)

Thx :). I could only verify it with CoreOS in my environment

sbueringer:
I'll rebase on top of master and also test it with clusterctl

jichenjc:
hi @sbueringer I tried this PR my understanding (from above comments) is this can be auto generated
but I got this error ,I guess I need more time to know how to add CA stuffs...

E0715 01:57:11.448579       1 actuator.go:323] Machine error openstack-master-8w7cr: error creating Openstack instance: CA cert material in the ClusterProviderSpec is missing cert/key
W0715 01:57:11.448610       1 machine_controller.go:229] Failed to create machine "openstack-master-8w7cr": error creating Openstack instance: CA cert material in the ClusterProviderSpec is missing cert/key

pkg/cloud/openstack/services/certificates/certificates.go
jichenjc: I understand the logic here :), the question I have originally is I didn't see some code like 'save' ' store' the Cert and key ...so I assume kubeconfig will be the place to hold those stuff..

sbueringer:
> hi @sbueringer I tried this PR my understanding (from above comments) is this can be auto generated
> but I got this error ,I guess I need more time to know how to add CA stuffs...
> 
> E0715 01:57:11.448579 1 actuator.go:323] Machine error openstack-master-8w7cr: error creating Openstack instance: CA cert material in the ClusterProviderSpec is missing cert/key
> W0715 01:57:11.448610 1 machine_controller.go:229] Failed to create machine "openstack-master-8w7cr": error creating Openstack instance: CA cert material in the ClusterProviderSpec is missing cert/key

Hi, That error is kind of expected. But only a few times. So the machine actuator should wait until the cluster actuator reconciled the certificates. After that is done the machine should be created. I tried the whole clusterctl flow with CoreOS and it worked.

pkg/cloud/openstack/services/certificates/certificates.go
sbueringer: Got it :). That's not really intuitive but the whole spec & status are saved here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/51c65e2000f3fa0a12931c8213ae1131a5878798/pkg/cloud/openstack/cluster/actuator.go#L107

But I see the `defer` statement is way to late. I moved it up in another PR. But I'll do it now in this PR.

sbueringer: It should only be a matter of cluster actuator reconciling the certs. As soon as that's done the machine actuator can use them. As long as the certs are not in the Cluster CRD yet the machine actuator has to wait (if not it will render empty certs in the userData)

jichenjc:
perfect, I will wait for longer time , I might be too rushy here.. thanks for the detail info~

sbueringer:
Yup. But it shouldn't take to long. Maybe the problem is that something fails in the Cluster actuator and that's why the defer is never triggered

sbueringer:
@jichenjc Commit regarding defer storeCluster is pushed

cmd/clusterctl/examples/openstack/machine-deployment.yaml.template
```diff
@@ -22,7 +22,6 @@ spec:
           kind: "OpenstackProviderSpec"
           flavor: m1.medium
           image: <Image Name>
-          sshUserName: <SSH Username>
```
jichenjc: I would keep those for debug purpose.... or at least show a way to user how to debug issues

and sshUsername and keyname below should be a group... remove or keep them both

jichenjc:
I tested this locally with Ubuntu and it works for me :)
so address some comments above then lgtm ~ thanks a lot

jichenjc:
I confirmed ubuntu test can pass test (though some other minor changes but not related) :)

so lgtm if above comments addressed... thanks~

cmd/clusterctl/examples/openstack/machine-deployment.yaml.template
sbueringer: I'm not sure. 

keyname is used to inject the given keypair. So that's absolutely necessary to jump on the host later, without it, it won't work. sshUserName would be only to document to the user inside the Machine resource what the name of his os user is. I'm not sure if he already knows that and if it's strictly necessary to put this inside the Machine resource

sbueringer: But I also don't have very strong opinions on keeping or deleting them both. I think deleting both is not really an option because without keyname we don't inject the key which makes it possible in the first place to ssh onto the host.

I guess I would go with documenting how to ssh onto the host.

jichenjc:
ok, I think a follow up document update will be needed. 
I can do that update accordingly

jichenjc:
give other folks a chance to take a look

sbueringer:
> ok, I think a follow up document update will be needed.
> I can do that update accordingly

That would be nice. Thx!

> give other folks a chance to take a look

Yup of course. I pinged @chrigl already. I cannot approve it anyway :)

jichenjc:
/approve

we do have chance to fix minor issues later on, for now, since both of us tested, I think we can go with it

pkg/deployer/deployer.go
```diff
@@ -35,31 +34,43 @@ func (d *Deployer) GetIP(cluster *clusterv1.Cluster, machine *clusterv1.Machine)
...
-		"ssh", "-i", homeDir+"/.ssh/openstack_tmp",
-		"-o", "StrictHostKeyChecking no",
-		"-o", "UserKnownHostsFile /dev/null",
-		"-o", "BatchMode=yes",
-		fmt.Sprintf("%s@%s", machineSpec.SshUserName, ip),
-		"echo STARTFILE; sudo cat /etc/kubernetes/admin.conf"))
-	parts := strings.Split(result, "STARTFILE")
-	if len(parts) != 2 {
-		return "", nil
+	server := fmt.Sprintf("https://%s:6443", ip)
```
jichenjc: this might be the root cause for https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/399

PR #424, opened 2019-08-05
Rename network prefix

jichenjc:
cloud provider and others might create this kind of resource
as well, we might consider to make it clusterapi specific

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/openstack/services/networking/service.go
+++ b/pkg/cloud/openstack/services/networking/service.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	networkPrefix   string = "k8s"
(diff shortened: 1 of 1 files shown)


sbueringer:
@jichenjc Are they really creating additional networks?

jichenjc:
they use k8s a lot, my assumption is to either make it clear created by clusterapi, or we make it configurable for resources created by us to avoid consume people ...anyway, just my own thought

sbueringer:
It's fine for me. Just wanted to ask :). In the current state of the ClusterAPI provider I don't really care about changes like this. After a release this would be a breaking change :)

/lgtm

Issue #428, opened 2019-08-08
support create LB with existing network/subnet


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
this is a use case at least in my env , you are not authorized to create network/subnet due to restriction on the system ,then you have to reuse existing network to create LB,
however, under current implementation we don't allow this use case, you have to create a new network by using the nodeCIDR setting.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

jichenjc:
to add more info here:
1) some times there is no floating ip needed, e.g someone is using provider network and we need make that as least works without floating ip introduction
2) with 1) above, you don't need create a nodecidr in order to work, some existing network can be specified in order to make cluster api to create..

sbueringer:
So we'll implement it like I proposed here?
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/422#issuecomment-519587486

sbueringer:
I would open a PR to fix this, if it's okay :)

jichenjc:
sure, I am too busy recently on other things. so please go ahead and push PR, thanks ~

sbueringer:
/assign @sbueringer 
PR is open here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/475

sbueringer:
Was fixed in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/477

/close

Issue #425, opened 2019-08-05
Failed to create machine with nodeCidr and without networks


hidekazuna:
/kind bug

**What steps did you take and what happened:**

When nodeCidr is set and networks is not set, server failed to be created.

```bash
W0805 13:01:57.560622       1 machine_controller.go:226] Failed to create machine "openstack-master-dvw8k": error creating Openstack ins
tance: create new server err: Bad request with: [POST http://controller.hidekazuna.test:8774/v2.1/servers], error message: {"badRequest"
: {"code": 400, "message": "Invalid input for field/attribute server. Value: {'availability_zone': 'nova', 'flavorRef': '1', 'imageRef':
 '3422e5fc-5c86-4cf4-8c8b-1716b8f3219d', 'key_name': 'cluster-api-provider-openstack', 'metadata': {'key': 'value'}, 'name': 'openstack-
master-dvw8k', 'security_groups': [{'name': '10adb7a1-5e5e-48da-aea2-d4e84a0b8ef3'}, {'name': '3b85495e-d51b-4f19-97fb-62f7ab645401'}], 
'tags': ['cluster-api-provider-openstack', 'default-test4010', 'a-machine-specific-tag', 'a_cluster_wide_tag'], 'user_data': 'IyEvdXNyL2
Jpbi9lbnYgYmFzaApzZXQgLWUKc2V0IC14CigKS1VCRUxFVF9WRVJTSU9OPTEuMTUuMApWRVJTSU9OPXYke0tVQkVMnCg=='}. 'networks' is a required property"}}
``` 

**What did you expect to happen:**

When nodeCidr is set and networks is not set, The network, subnet, and router are created, and server is created.

**Anything else you would like to add:**

**Environment:**

- cluster-api-provider-openstack: #401 
- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

sbueringer:
If I understand it correctly, we're currently trying to create Ports (or reuse existing ones) in the networks given in the machineSpec. If the networks parameter is not set, to what network should we default to?

hidekazuna:
@sbueringer This is the first time to create k8s cluster with nodeCidr so that I am sorry if I am wrong. I assumed if nodeCidr is given and networks is not given, network, subnet, and router are newly created and server uses the port of the network.

hidekazuna:
[...]

PR #421, opened 2019-08-02
finalize multi-node control plane implementation

sbueringer:
All kubeadm configs are now generated via go code. This allows us to
generate different configs for the first control plane node, every other
control plane node and normal worker nodes.

This also adds a flag to disablePortSecurity and it's now possible to
set the KUBECONFIG and KUBECONTEXT via environment variable.

**What this PR does / why we need it**:
This PR finalizes the multi-node control plane feature.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #382 

Notes:
* I verified it via a 3 master 3 node configuration with Neutron LBaaS and disabled port security
* Let's discuss what kind of documentation we need

PR #432, opened 2019-08-11
add apiserver lb doc

### LINKED ISSUE #422: doc for API loadbalancer
/kind feature

we merged API loadbalancer code and we need additional doc 

sbueringer:
Adds doc for APIServe rlb

Fixes #422

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,27 @@ policy may be made to more closely align with other providers in the Cluster API
    You can find some guidance on what needs to be edited, and how to create some of the
    required OpenStack resources in the [Configuration documentation](docs/config.md).
 
+
+   #### Notes on configuring an APIServer LoadBalancer
+   
+   It's possible to configure a loadbalancer, which is then used to loadbalance in front of the control 
+   plane nodes. This feature is optional for single-node clusters but highly recommmended for multi-node 
+   control plane clusters. The loadbalancer can be configured via the following properties in the Cluster CRD:
+   
+   ````
+     managedAPIServerLoadBalancer: true
+     apiServerLoadBalancerFloatingIP: <floating-ip>
+     apiServerLoadBalancerPort: 6443
+     apiServerLoadBalancerAdditionalPorts:
+       - 22
+   ````
+   
+   `6443` is usually used for the APIServer port. It's also possible to configure additional ports, e.g. for ssh.
+   
+   Note: This currently only works with `nodeCidr` set, as only then the created networks/ subnets are available in 
+   the cluster status. This will be fixed in [#428](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/428)
+   Note: Make sure the floating ip is not already used on a Machine.
+
    #### Notes on using custom CAs
    
    The Cluster CD has properties to specify the CAs of the Kubernetes cluster. Usually, they are not 


README.md
```diff
@@ -136,6 +136,27 @@ policy may be made to more closely align with other providers in the Cluster API
...
+   It's possible to configure a loadbalancer, which is then used to loadbalance in front of the control 
+   plane nodes. This feature is optional for single-node clusters but highly recommmended for multi-node 
+   control plane clusters. The loadbalancer can be configured via the following properties in the Cluster CRD:
+   
+   ````
+     managedAPIServerLoadBalancer: true
+     apiServerLoadBalancerFloatingIP: <floating-ip>
+     apiServerLoadBalancerPort: 6443
+     apiServerLoadBalancerAdditionalPorts:
+       - 22
```
jichenjc: sorry, but I am little bit curious on 22, why we need a ssh to a LB? should this be ssh to a host only? 

And, I think if possible, can we create a template for non-LB and LB cluster.yaml template? (non-LB can reuse previuos one)

sbueringer: We don't need it. But in my case I"m using it to ssh onto the nodes because I cannot access the Machines because I"m only setting a Floating IP on the LoadBalancer.

I would prefer adding these templates with multi-node control plane. I would say single-node control plane makes sense without lb and multi-node with lb.

To be honest I think we're loosing control over our permutations. We cannot really maintain & test all permutations of:
* networks are pre-created vs ClusterAPI creates them
* floating ips are pre-created vs ClusterAPI creates floating ips vs we're not using any
* using api server loadbalancer vs without api server loadbalancer
* single-node control plane vs multi-node control plane

I think we have to define our use cases from scratch and then document them cleanly. Right now I don't really now which permutations we support.

sbueringer: I'll give it a try :):

## Networks:

* Networks on cluster-level
  * Specify nodeCIDR and network/subnet/router will be created and added to the Cluster status
  * [to be implemented] Specify networks (network/subnet) filters to detect existing networks and add them Cluster status

> Result is always that we have network/subnet in the cluster status

* Networks on machine-level
  * Per default the networks from the Cluster are used
  * It's possible to overwrite them by filter (I saw in the doc that it's possible to add the Machine to multiple networks)

## Floating IPs

* LoadBalancer
  * [to be implemented] Can be set optionally on the APIServer loadbalancer

* Machines
  * Can be set optionally on every machine
    * [to be implemented] create them if they are specified and don't exist

* APIServer IP
  * After multi-node control plane pr: the controlPlaneEndpoint has to match either:
    * [with loadbalancer]: the load balancer floating ip
      * what do we do if none is set? 
    * [without loadbalancer]: the ip of the single control plane machine
      * no lb and multiple control plane machines doesn't make sense imho

## LoadBalancer

* Loadbalancer can be used with single & multi-node control plane (allows easier update from single to multi-node control plane)
* Is Loadbalancer mandatory for multi-node control plane? I think yes, because if it isn't what can we use as controlPlaneEndpoint?

jichenjc: > We don't need it. But in my case I"m using it to ssh onto the nodes because I cannot access the Machines because I"m only setting a Floating IP on the LoadBalancer.
> 
> I would prefer adding these templates with multi-node control plane. I would say single-node control plane makes sense without lb and multi-node with lb.
+1 to this, single node without lb and multi node with lb is the right way
> 
> To be honest I think we're loosing control over our permutations. We cannot really maintain & test all permutations of:
> 
> * networks are pre-created vs ClusterAPI creates them
this is because my env has some limitation, I can't use lb if it lead to mandatory nodecidr case, I think it's low priority in normal cases.....
> * floating ips are pre-created vs ClusterAPI creates floating ips vs we're not using any
currently the floating ips has to be created (either w/ or w/o LB, right) ? I think it's low priority too to support create floating ip on the fly
> * using api server loadbalancer vs without api server loadbalancer
I might lost some background, do we only have LB now ?without LB will make it single node case? 
> * single-node control plane vs multi-node control plane
we can't always request lb, so single node and multiple node support should be there
> * ...
> 
> I think we have to define our use cases from scratch and then document them cleanly. Right now I don't really now which permutations we support and how this dimensions interfere with each other.

sbueringer: See the 2nd post from me above :)

jichenjc: I totally agree non-lb case to sing-node ,this make sense :)

I can do the #428 thing, I am struggling with that now...

sbueringer: Okay. Tell me if I can help you

jichenjc: thanks a lot :), I am having 2 issues in my env:
1) I can't create floating ip ,I can only associate existing floating ip to LB 
2) the newly create network/subnet can't be used because of policy, I have to use existing network...

jichenjc: > 
> 
> I'll give it a try :):
> ## Networks:
> 
>     * Networks on cluster-level
>       
>       * Specify nodeCIDR and network/subnet/router will be created and added to the Cluster status
>       * [to be implemented] Specify networks (network/subnet) filters to detect existing networks and add them Cluster status
> 
> 
> > Result is always that we have network/subnet in the cluster status
> 
>     * Networks on machine-level
>       
>       * Per default the networks from the Cluster are used (#425)
>       * It's possible to overwrite them by filter (I saw in the doc that it's possible to add the Machine to multiple networks)
> 
> 
> ## Floating IPs
> 
>     * LoadBalancer
>       
>       * [to be implemented] Can be set optionally on the APIServer loadbalancer
> 
>     * Machines
>       
>       * Can be set optionally on every machine
>         
>         * [to be implemented] create them if they are specified and don't exist
> 
>     * APIServer IP
>       
>       * After multi-node control plane pr: the controlPlaneEndpoint has to match either:
>         
>         * [with loadbalancer]: the load balancer floating ip
>           
>           * what do we do if none is set?
>         * [without loadbalancer]: the ip of the single control plane machine
>           
>           * no lb and multiple control plane machines doesn't make sense imho
> 
> 
> ## LoadBalancer
> 
>     * Loadbalancer can be used with single & multi-node control plane (allows easier update from single to multi-node control plane)
> 
>     * Is Loadbalancer mandatory for multi-node control plane? I think yes, because if it isn't what can we use as controlPlaneEndpoint?
> 
> 
> P.S.: If we have consensus an what makes sense and what we can support we should document these cases. I think this would lead to far better documentation to just document each permutation dimension (floating ip/network/ load balancer/ single vs multi-node control plane) by itself.

exactly, we have limitted resource so maybe we can focus on some case we think important first... we can add more support if we have to our priority...
to my case, as I indicated, I have to use existing network because my env limitation..

anyway, let's merge https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/421 first..

PR #434, opened 2019-08-12
Add delete loadbalancer support

### LINKED ISSUE #433: LB not delete after cluster deletion
/kind bug

after fix #429 , the machines can be deleted correctly, however, the LB listener and LB not deleted

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

jichenjc:
during cluster deletion, the floating ip created and lb created are not deleted correctly

Fixes #433 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/openstack/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/openstack/services/loadbalancer/loadbalancer.go
@@ -264,6 +264,27 @@ func (s *Service) ReconcileLoadBalancerMember(clusterName string, machine *clust
 	return nil
 }
 
+func (s *Service) DeleteLoadBalancer(clusterName string, clusterProviderSpec *providerv1.OpenstackClusterProviderSpec, clusterProviderStatus *providerv1.OpenstackClusterProviderStatus) error {
+	loadBalancerName := fmt.Sprintf("%s-cluster-%s-%s", networkPrefix, clusterName, kubeapiLBSuffix)
+	lb, err := checkIfLbExists(s.loadbalancerClient, loadBalancerName)
+	if err != nil {
+		return err
+	}
+	deleteOpts := loadbalancers.DeleteOpts{
+		Cascade: true,
+	}
+	klog.Infof("Deleting loadbalancer %s", loadBalancerName)
+	err = loadbalancers.Delete(s.loadbalancerClient, lb.ID, deleteOpts).ExtractErr()
+	if err != nil {
+		return fmt.Errorf("error deleting loadbalancer: %s", err)
+	}
+
+	// floating ip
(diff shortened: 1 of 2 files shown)


pkg/cloud/openstack/services/loadbalancer/loadbalancer.go
```diff
@@ -264,6 +264,39 @@ func (s *Service) ReconcileLoadBalancerMember(clusterName string, machine *clust
...
+		Cascade: true,
+	}
+	klog.Infof("Deleting loadbalancer %s", loadBalancerName)
+	err = loadbalancers.Delete(s.loadbalancerClient, lb.ID, deleteOpts).ExtractErr()
+	if err != nil {
+		return fmt.Errorf("error deleting loadbalancer: %s", err)
+	}
+
+	// floating ip
+	if clusterProviderSpec.APIServerLoadBalancerFloatingIP != "" {
```
sbueringer: What should we do in cases a already existing floating ip is just associated? In my case that's not a problem because I can delete the floating ip. But what if that's forbidden?

jichenjc: that's a good question...can we distinguish those 2 cases? I don't know how to check it because it's already delete stage, should we create another flag?

sbueringer: I think we should avoid adding even more flags, because it get's problematic from a UX perspective.

I have 2 ideas:
1. [Not sure if that's possible] Ask Keystone if we're allowed to delete the floating ip and only do it if it's allowed
2. Try to delete the floating ip and ignore the "you're not allowed to" (not sure how it's phrased" error

In both variants we should probably for disassociate the floating ip (which is an update on the floating ip with an empty port_id, similar to what I did here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/435/commits/79cd5abd2f1fff965c5ea08f1e42eb71c082f933#diff-1144c99c6890d759d38d9549258a187aR84)

sbueringer:
That's related to this issue right? https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/433

jichenjc:
how about let's defer the floating ips to next PR and at least it works for current (deassociate the floating ip only during delete) approach? @sbueringer

sbueringer:
> how about let's defer the floating ips to next PR and at least it works for current (deassociate the floating ip only during delete) approach? @sbueringer

Fine for me

jichenjc:
@sbueringer made some changes (no touching floating ip)

sbueringer:
Okay

/lgtm

Issue #391, opened 2019-07-09
failed to create master and node


jichenjc:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
kubelet failed to start, the kubelet is saying 'node openstack-master-xxxx' can not be found
and this blocked kubelet start and in turn master and node failed to start

we have to workaround this first then check the k8s community `cloud-provider` expectation.

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

jichenjc:
I am seeing this in /var/log/messages


Jul 10 00:07:46 openstack-master-5wgrw kubelet: I0710 00:07:46.990277    4340 kubelet.go:1822] skipping pod synchronization - container runtime status check may not have completed yet
Jul 10 00:07:46 openstack-master-5wgrw kubelet: E0710 00:07:46.993032    4340 kubelet.go:2248] node "openstack-master-5wgrw" not found
Jul 10 00:07:47 openstack-master-5wgrw kubelet: E0710 00:07:47.093151    4340 kubelet.go:2248] node "openstack-master-5wgrw" not found
Jul 10 00:07:47 openstack-master-5wgrw kubelet: E0710 00:07:47.193338    4340 kubelet.go:2248] node "openstack-master-5wgrw" not found
Jul 10 00:07:47 openstack-master-5wgrw cloud-init: I0710 00:07:47.224406    3906 round_trippers.go:419] curl -k -v -XGET  -H "Accept: application/json, */*" -H "User-Agent: kubeadm/v1.15.0 (linux/amd64) kubernetes/e8462b5" 'https://9.20.206.79:443/healthz?timeout=32s'
Jul 10 00:07:47 openstack-master-5wgrw cloud-init: I0710 00:07:47.225438    3906 round_trippers.go:438] GET https://9.20.206.79:443/healthz?timeout=32s  in 1 milliseconds
Jul 10 00:07:47 openstack-master-5wgrw cloud-init: I0710 00:07:47.225448    3906 round_trippers.go:444] Response Headers:
Jul 10 00:07:47 openstack-master-5wgrw kubelet: E0710 00:07:47.293504    4340 kubelet.go:2248] node "openstack-master-5wgrw" not found
[...]

PR #438, opened 2019-08-15
follow up PR 421 to make ubuntu and centos work

jichenjc:

The PR #421  has some missing code change for ubuntu and centos, 
I created ubuntu correctly but centos has some issue (#391), so modify the code according to
ubuntu changes only 

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/worker-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/worker-user-data.sh
@@ -7,8 +7,6 @@ NAMESPACE={{ .Machine.ObjectMeta.Namespace }}
 MACHINE=$NAMESPACE
 MACHINE+="/"
 MACHINE+={{ .Machine.ObjectMeta.Name }}
-CLUSTER_DNS_DOMAIN={{ .Cluster.Spec.ClusterNetwork.ServiceDomain }}
-SERVICE_CIDR={{ .ServiceCIDR }}
 
 swapoff -a
 # disable swap in fstab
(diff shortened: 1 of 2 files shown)


jichenjc:
@sbueringer PTAL, thanks

cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/worker-user-data.sh
```diff
@@ -86,10 +83,10 @@ EOF
 
 # Override network args to use kubenet instead of cni, override Kubelet DNS args and
 # add cloud provider args.
```
sbueringer: Can you remove this instead of just commenting it out?

jichenjc: @sbueringer  done

sbueringer:
But if this PR here works for you it means we don't need the cacert for APIServer, right? :)

jichenjc:
@sbueringer 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/services/userdata/kubeadm.go#L71 already has it

sbueringer:
> @sbueringer
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/services/userdata/kubeadm.go#L71 already has it

Yup. But only for the controller-manager. The apiserver has the same cloudprovider openstack flags but no cacert

jichenjc:
ok, maybe I missed something, anyway, let's be it and if we find some issue later, we can do modifications

PR #443, opened 2019-08-16
use env for bash

hchenxa:
bash is not always in /bin, particularly on non-Linux systems. For example the OpenBSD system, it's in /usr/local/bin. so prefer to use #!/usr/bin/env bash.

refer https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

Fixes #

**Release note**:

```release-note

```


hchenxa:
cc @jichenjc @gyliu513

PR #444, opened 2019-08-16
fix empty default values if imageRepository is not set

sbueringer:
Fixes the problem with empty value in imageRepository. I didn't get this before because I never tried CoreOS with the default imageRepository. Also replaces the other default values with the corresponding constants


sbueringer:
@hidekazuna can you please take a look?

hidekazuna:
Cluster was created successfully.

/lgtm

Issue #436, opened 2019-08-13
Hostname missing in hosts file in openstack instance


prankul88:
/kind bug

**What steps did you take and what happened:**
While setting up cluster-api using clusterctl command, there was a portion in the clusterctl logs which had an error `exit status 255`  ~~while the command tried to ssh into the machine~~. This after a minute got resolved. After some searching, I found out it was related to the fact that hotsname was missing in /etc/hosts of the created openstack instance. 
Not sure if this is an issue of openstack or this repo. 

**What did you expect to happen:**
This error can be resolved by adding a command in the master-user-data.sh
Probably mapping loopback ip to the hostname will resolve the issue

**Anything else you would like to add:**
Once I get some feedback and if found valid, I can push a PR for the same.

**Environment:**

- Cluster-api version:
- Minikube/KIND version: v0.4.0
- Kubernetes version: (use `kubectl version`): v1.15
- OS (e.g. from `/etc/os-release`): Ubuntu16.04
-Openstack-release: Rocky

sbueringer:
The current master version shouldn't try to ssh into the machine. Can you post us the logs?

sbueringer:
Hostname is set here for CoreOS:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cmd/clusterctl/examples/openstack/provider-component/user-data/coreos/templates/common.yaml#L39-L43

Didn't find anything similar in Centos/Ubuntu user-data.sh files. But as I said I'm wondering why the controller tries to ssh onto the node

jichenjc:
yeah, we should never try to ssh, I guess maybe @prankul88 you might need update your code to latest?

prankul88:
@sbueringer @jichenjc Okay, thanks for clearing this doubt for me. I got a little confused. Just one more thing, I found something in the openstack logs 

> [   81.138455] cloud-init[1310]: + sudo apt-key add -
[   81.144172] cloud-init[1310]: + curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg
[   81.152234] cloud-init[1310]: **sudo: unable to resolve host openstack-master-ddwpq**
[   82.093848] cloud-init[1310]: OK

My master's hostname is not present in the /etc/hosts file.
Is it a concern for us?

jichenjc:
[...]

PR #446, opened 2019-08-18
[v1alpha2 Migration] add v1alpha2 controllers

sbueringer:
Initial migration to v1alpha2 controllers as described here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/380#issuecomment-522018403

(This also contains the commits from: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pulls)

PR #449, opened 2019-08-22
Hostname missing in /etc/hosts file in openstack instance

prankul88:
#436 

> What I felt is when any sudo commands run from master-user-data.sh, it hangs for a bit and then proceeds. If we can map loopback-ip to the hostname in /etc/hosts (can add command in master-user-data file), it would not get stuck.

### DIFF
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
@@ -10,6 +10,7 @@ MACHINE+="/"
 MACHINE+={{ .Machine.ObjectMeta.Name }}
 ARCH=amd64
 swapoff -a
+echo "127.0.0.1 $(hostname)"  >> /etc/hosts
 # disable swap in fstab
 sed -i.bak -r 's/(.+ swap .+)/#\1/' /etc/fstab
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
--- a/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/worker-user-data.sh
+++ b/cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/worker-user-data.sh
@@ -9,6 +9,7 @@ MACHINE+="/"
 MACHINE+={{ .Machine.ObjectMeta.Name }}
 
 swapoff -a
+echo "127.0.0.1 $(hostname)"  >> /etc/hosts
 # disable swap in fstab
 sed -i.bak -r 's/(.+ swap .+)/#\1/' /etc/fstab
 apt-get update


cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
```diff
@@ -11,6 +11,7 @@ MACHINE+={{ .Machine.ObjectMeta.Name }}
 ARCH=amd64
 swapoff -a
 # disable swap in fstab
+echo "127.0.1.1 $(hostname)"  >> /etc/hosts
```
sbueringer:
```suggestion
echo "127.0.0.1 $(hostname)"  >> /etc/hosts
```

sbueringer:
@prankul88 The master-user-data file moved because the v1alpha2 PR got merged on master it's now in `samples/templates/user-data/ubuntu`

prankul88:
@sbueringer Thanks I will update this patch to remove the conflict. 
Just one thing I saw that now we have master-user-data.sh in `samples/templates/user-data/` as well. (Before it was generated by ./generate.yaml). Can you help me understand why is it there? And probably I will need to update that file as well

sbueringer:
> @sbueringer Thanks I will update this patch to remove the conflict.
> Just one thing I saw that now we have master-user-data.sh in `samples/templates/user-data/` as well. (Before it was generated by ./generate.yaml). Can you help me understand why is it there? And probably I will need to update that file as well

The short answer is because the first parts of the migration to v1alpha2 are merged on master. (see this PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/446)

But there is not enough documentation yet how to setup a cluster like with clusterctl before. But clusterctl itself is not fully migrated to v1alpha2 as far as I know. So depending on the branch master or release-0.1 (v1alpha1) you have to modify different files.

Might make sense to merge this PR in release-0.1 so you can test and continue. I can create a cherry-pick for master if you want.

cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
jichenjc: +1

prankul88:
@sbueringer Sure I will move it to release-0.1 and then cherry-picking seems to be the right option. Thanks.

cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
```diff
@@ -11,6 +11,7 @@ MACHINE+={{ .Machine.ObjectMeta.Name }}
 ARCH=amd64
 swapoff -a
 # disable swap in fstab
+echo "127.0.0.1 $(hostname)"  >> /etc/hosts
```
sbueringer: Do you only need it on the master node?

sbueringer:
@prankul88 There's also an Issue in CABPK (which is used in CAPI v1alpha2). Probably better to solve the issue in v1alpha2 in the other issue: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/116

But the fix here for v1alpha1 makes sense

cmd/clusterctl/examples/openstack/provider-component/user-data/ubuntu/templates/master-user-data.sh
jichenjc: I believe we need for all nodes..

prankul88: @sbueringer @jichenjc I only did it for master as the error was there only. I too feel it must be there in other nodes as well but did not find any instance to justify it. If you say I can add.

jichenjc: yes, I think we can fix both as we anyway want to fix it:)

prankul88: Sure I will add it to the other user-data file as well.

sbueringer:
@jichenjc I think you have to approve because I'm not an approver yet on this branch :)

/lgtm

Issue #380, opened 2019-06-27
[v1alpha2]: prepare for v1alpha2 adoption


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

https://github.com/kubernetes-sigs/cluster-api/pull/997/ introduced some design doc
about v1alpha2, we should start to think about how to enable it
**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

sbueringer:
I would start working on this
/assign @sbueringer

jichenjc:
yes, let's joint work on this, I am studying and preparing for this as well
so maybe we can make sub-tasks :)

sbueringer:
Yup okay. I need a few days to catch up to the current status of v1alpha2

sbueringer:
It would be nice if we could get this done first: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/382

I have all the code here. Just have to merge the APIServerLB PR and one final on to implement kubeadm.

If we move to v1alpha2 first I fear I have to reimplement a lot of that stuff.

jichenjc:
we may need create another branch for v1alpha2 ?

sbueringer:
Definitely. I'm just looking at AWS and I guess it will be more or less half rewrite and the other half copy&pase of existing OpenStack-specific. That's why it would be nice to have the kubeadm change on master before start copying it on the v1alpha2 branch

CamelCaseNotation:
I'm interested in helping out with this effort. LMK how I can help

CamelCaseNotation:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/933 here's an issue with some guidance we could extrapolate from

CamelCaseNotation:
Consider either tagging this issue, or adding a `[v1alpha2]` prefix to the issue title so it may be more easily tracked, thanks!

sbueringer:
> I'm interested in helping out with this effort. LMK how I can help

That's nice to here, thx!

I'm not really sure how we should do the migration to v1alpha2. I think we should do it in the following order:

* Pre-Migration:
  * migrate to go.mod
  * Finalize https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/382

* Migration:
  * Migrate Machine Actuator/Controller, similar to: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/892
[...]

PR #452, opened 2019-08-26
Pr v1alpha2 kubeadm bootstrapper

sbueringer:
Migrated to kubeadm bootstrapper as described here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/380#issuecomment-522018403

Also rewrote the examples (which are probably still far from perfect) and moved them to the examples folder.

Only the operating systems supported by kubeadm bootstrapper are supported now (currently has a dependency to cloud-init >= 18.4 (see https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/41)


jichenjc:
@sbueringer I will try this but not sure how long will it take due to some other works in my daily job :) @hidekazuna if you have time , can you help test as well ? :)

examples/controlplane/controlplane.yaml
```diff
@@ -0,0 +1,239 @@
...
+spec:
+  version: ${KUBERNETES_VERSION}
+  bootstrap:
+    configRef:
+      apiVersion: bootstrap.cluster.x-k8s.io/v1alpha2
+      kind: KubeadmConfig
+      name: ${CLUSTER_NAME}-controlplane-0
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
+    kind: AWSMachine
```
hidekazuna: OpenStackMachine?

sbueringer: Thx fixed

sbueringer:
All rebased. I'll do some work later today and compare the examples with my own test files

sbueringer:
> All rebased. I'll do some work later today and compare the examples with my own test files

Done. I diffed it against my own test yamls at it's very similar. I think the biggest problem will be that you're maybe not using a OS image which already contains kubelet/kubeadm/docker (or containerd)

I think there are two options:
* Adjust the os image
* Add an install script to the KubeadmConfig and add a kubeadmPreCommand to execute it

sbueringer:
I updated the doc for the usage of clusterctl. I'm not aware if there are released images for the cluster-api-controller & cluster-api-bootstrap-provider-kubeadm, so the doc doesn't work out of the box. If somebody wants to find out if these images are already their, please do :) 

For now I will continue using Tilt for development. At the latest when the other providers are starting to use released images for those two controller we can also start using them

jichenjc:
as we are moving forward on this, let's quick merge and make changes later..

/lgtm
/approve

sbueringer:
> as we are moving forward on this, let's quick merge and make changes later..
> 
> /lgtm
> /approve

Thx. I'll update the v1alpha2 issue later.

I would also go over our existing issues & PRs and check if they are still valid.

I also have a few suggestions for followup work

PR #453, opened 2019-08-27
Update owners aliases 

roberthbailey:
... for shared groups to match what is currently defined in the aws provider and add security contacts (sig leads).

**Release note**:

```release-note
NONE
```


jichenjc:
I am not familiar with the procedure here.. I assume it's a right way to go, @chrigl comments?
/lgtm

PR #458, opened 2019-09-04
Remove gcloud shell usage

jichenjc:
not all folks has gcloud shell, use k8scloudprovider (original one)
instead

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 MOCKGEN := $(TOOLS_BIN_DIR)/mockgen
 
 # Define Docker related variables. Releases should modify and double check these vars.
-REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+REGISTRY ?= k8scloudprovider
 STAGING_REGISTRY := gcr.io/k8s-staging-cluster-api-openstack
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,7 +42,7 @@ GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 MOCKGEN := $(TOOLS_BIN_DIR)/mockgen
 
 # Define Docker related variables. Releases should modify and double check these vars.
-REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+REGISTRY ?= k8scloudprovider
```
sbueringer: The idea was to implement the regular staging process for cluster api images, for which I'm currently building another PR. I would really like to do it like every other provider and not push to dockerhub instead.

I'll open a PR for that shortly. If somebody wants to push to dockerhub meanly it can always be overwritten by using `export REGISTRY`

jichenjc: ok, I don't know the release process
 the reason I add this is because I am tring to build image and can't build it...
and I agree this might be a better way to be same to other providers

jichenjc:
REGISTRY=k8scloudprovider make images
fixed the problem

so let's hold this and update the doc when the settings are done

jichenjc:
right, thanks~

PR #459, opened 2019-09-04
Fix api port

jichenjc:
The kustomize will remove the # comments ... so no one knows the info provided
and even if we some one failed to set the failure, we should not let container panic ...
so I added a default value here

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -170,10 +170,17 @@ func (r *OpenStackClusterReconciler) reconcileCluster(logger logr.Logger, cluste
 			return reconcile.Result{}, errors.Errorf("failed to get control plane machine: %v", err)
 		}
 		if controlPlaneMachine != nil {
+			var apiPort int
+			if cluster.Spec.ClusterNetwork.APIServerPort == nil {
+				logger.Info("No API endpoint given, default to 6443")
+				apiPort = 6443
+			} else {
+				apiPort = int(*cluster.Spec.ClusterNetwork.APIServerPort)
+			}
 			openStackCluster.Status.APIEndpoints = []infrav1.APIEndpoint{
 				{
 					Host: controlPlaneMachine.Spec.FloatingIP,
-					Port: int(*cluster.Spec.ClusterNetwork.APIServerPort),
+					Port: apiPort,
 				},
 			}
 		} else {
--- a/examples/cluster/cluster.yaml
+++ b/examples/cluster/cluster.yaml
@@ -15,7 +15,7 @@ spec:
     # * Disable the apiServerPort property
     # single-node control-plane:
     # * Enable the apiServerPort property
-    #apiServerPort: 6443
+    apiServerPort: <disable when multi-node control-plane>
   infrastructureRef:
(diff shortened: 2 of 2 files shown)


controllers/openstackcluster_controller.go
```diff
@@ -170,10 +170,16 @@ func (r *OpenStackClusterReconciler) reconcileCluster(logger logr.Logger, cluste
 			return reconcile.Result{}, errors.Errorf("failed to get control plane machine: %v", err)
 		}
 		if controlPlaneMachine != nil {
+			var apiPort int
```
sbueringer: wouldn't this be place to default to 6443? If I read it correctly we're currently defaulting to `0` :)

examples/cluster/cluster.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     # * Disable the apiServerPort property
     # single-node control-plane:
     # * Enable the apiServerPort property
-    #apiServerPort: 6443
+    apiServerPort: <disable when multi-node control-plane>
```
sbueringer: The rest of the example is commented to support multi-node control plane per default. That's the only property the other way around after this PR. So i would suggest keeping the `#` at the start of the line

jichenjc: if you take a look at generated result...
the # will be removed by my kustomize and that's the reason why I have the nil setting and 
it lead to panic... (see my changes in the go file)

see my output. all the things marked as comments are removed, guess it's not my kustomize version issue? I am using 3.10 already
```
apiVersion: cluster.x-k8s.io/v1alpha2
kind: Cluster
metadata:
  name: test-cluster
  namespace: test-cluster
spec:
  clusterNetwork:
    apiServerPort: <disable when multi-node control-plane>
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
      - 10.96.0.0/12
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
    kind: OpenStackCluster
    name: test-cluster
    namespace: test-cluster
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
kind: OpenStackCluster
metadata:
  name: test-cluster
  namespace: test-cluster
spec:
  apiServerLoadBalancerAdditionalPorts:
  - 22
  apiServerLoadBalancerFloatingIP: <loadbalancer floating ip>
  apiServerLoadBalancerPort: 6443
  cloudName: openstack
  cloudsSecret:
    name: cloud-config
    namespace: test-cluster
  disablePortSecurity: true
  disableServerTags: true
  dnsNameservers: []
  externalNetworkId: <external-network-id>
```

controllers/openstackcluster_controller.go
jichenjc: I forgot to set 6443 here...

examples/cluster/cluster.yaml
sbueringer: Yup the comments are removed through kustomize that's right. So they are not that useful right now. We can add the property for now. I think I would open another PR sooner or later to create separate multi-node & single-node examples with aproprate properties

PR #462, opened 2019-09-07
Pr finalize v1alpha2

sbueringer:
Final fixes for v1alpha2. This has been tested with:
* [x] multi-node cluster with LBaaS
* [x] single-nocde cluster without LBaaS

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -114,6 +114,7 @@ CLUSTER_GENERATED_FILE=${OUTPUT_DIR}/cluster.yaml
 CONTROLPLANE_GENERATED_FILE=${OUTPUT_DIR}/controlplane.yaml
 MACHINEDEPLOYMENT_GENERATED_FILE=${OUTPUT_DIR}/machinedeployment.yaml
 MACHINES_GENERATED_FILE=${OUTPUT_DIR}/machines.yaml
+ADDONS_GENERATED_FILE=${OUTPUT_DIR}/addons.yaml
 
 rm -rf "${OUTPUT_DIR}"
 rm -rf "${CLOUDS_SECRETS_CONFIG_DIR}"
@@ -127,7 +128,7 @@ if [[ ! -f ${MACHINE_CONTROLLER_SSH_PRIVATE_FILE} ]]; then
   ssh-keygen -t rsa -f ${MACHINE_CONTROLLER_SSH_PRIVATE_FILE}  -N ""
 fi
 export MACHINE_CONTROLLER_SSH_PUBLIC_FILE_CONTENT
-MACHINE_CONTROLLER_SSH_PUBLIC_FILE_CONTENT=$(cat ${MACHINE_CONTROLLER_SSH_PRIVATE_FILE}.pub)
+MACHINE_CONTROLLER_SSH_PUBLIC_FILE_CONTENT="\"$(cat ${MACHINE_CONTROLLER_SSH_PRIVATE_FILE}.pub)\""
 
 CLOUDS_PATH=${CLOUDS_PATH:-""}
 OPENSTACK_CLOUD_CONFIG_PLAIN=$(cat "$CLOUDS_PATH")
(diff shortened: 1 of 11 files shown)


sbueringer:
/assign @jichenjc 
/assign @hidekazuna 

After this only the staging release job and docs should be left for v1alpha2

examples/generate.sh
```diff
@@ -144,24 +145,34 @@ if [[ "$DOMAIN_NAME" = "null" ]]; then
...
   OPENSTACK_CLOUD_PROVIDER_CONF="$OPENSTACK_CLOUD_PROVIDER_CONF
-          region=\"${REGION}\"
+region=\"${REGION}\"
   "
 fi
+
+export OPENSTACK_CLOUD_PROVIDER_CONF6
+OPENSTACK_CLOUD_PROVIDER_CONF6=$(echo "${OPENSTACK_CLOUD_PROVIDER_CONF}" | sed -e 's/^/      /')
+export OPENSTACK_CLOUD_PROVIDER_CONF10
+OPENSTACK_CLOUD_PROVIDER_CONF10=$(echo "${OPENSTACK_CLOUD_PROVIDER_CONF}" | sed -e 's/^/          /')
```
jichenjc: I didn't get this CONF6/CONF10 ... what's the meaning of this?

examples/generate.sh
```diff
@@ -199,25 +210,19 @@ echo "Generated ${CONTROLPLANE_GENERATED_FILE}"
...
-echo "---" >> ${MACHINES_GENERATED_FILE}
-cat ${MACHINEDEPLOYMENT_GENERATED_FILE} >> ${MACHINES_GENERATED_FILE}
-echo "---" >> ${MACHINES_GENERATED_FILE}
-echo "Generated ${MACHINES_GENERATED_FILE}"
+cp ${SOURCE_DIR}/addons.yaml "${ADDONS_GENERATED_FILE}"
+echo "Generated ${ADDONS_GENERATED_FILE}"
 
 # Generate Cluster API provider components file.
-kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" --reorder=none > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
```
jichenjc: nit: use / instead //

examples/generate.sh
sbueringer: The problem is that I sometimes need the config with indent 6 (KubeadmConfig) and sometimes with indent 10 (KubeadmConfigTemplate). Seems to be the easiest way to just have to vars for envsubst.

examples/generate.sh
sbueringer: Done

sbueringer:
@jichenjc ptal :)

examples/generate.sh
jichenjc: ok, got it thanks

CamelCaseNotation:
@sbueringer could you share your YAML for creating MM cluster with LBaaS, as well as single node cluster without LBaaS? It would be greatly appreciated!!

sbueringer:
> @sbueringer could you share your YAML for creating MM cluster with LBaaS, as well as single node cluster without LBaaS? It would be greatly appreciated!!

That's the one for single node control plane:
https://gist.github.com/sbueringer/35e21417670a6a0c4983bad2911ea538

But please just use the documentation (and examples/generate.sh), which should be up-to-date with this PR: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/474

There's also an issue to generate separate single-node and multi-node control plane examples: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/473

I also have my dev-env published here: https://github.com/sbueringer/capi-dev/tree/master/cluster

Which first generates examples with the scripts from the repo and then kustomizes them to my env. If you have further questions you can always ask :)

PR #467, opened 2019-09-16
Use Nova's Scheduler Hints

### LINKED ISSUE #118: Think about using OpenStack Server Groups
Currently, when the machine actuator creates VMs, it goes to OpenStack Compute and asks for a VM, for the next VM, again, and again and again. This might result in having all VMs on a single hypervisor.

Well. OpenStack provides a way to configure anti-affinity. See https://docs.openstack.org/python-openstackclient/pike/cli/command-objects/server-group.html.

For sure, this does not come without problems. As long as there are hypervisor without "our"-VMs running on, everything is fine. But if "our"-VMs are scheduled to all of the Hypervisors, we will get a "no host found" exception.
What we would need is kind of soft-anti-affinity. As long as it is possible, OpenStack should schedule VMs on different nodes, but as soon as it is not possible any more, OpenStack should accept that it needs do deploy VMs to an already used Hypervisor.

We probably can create a ServerGroup in the cluster actuator for master nodes. For worker nodes we might think about using its own group, but implement the fuzzy scheduling in the machine actuator. Like. Count the number of failures due to "no host found", if that increases $MAX, try to spin up a VM without servergroup.
But I want hear other opinions on this.

oadivtal:
A user can define Scheduler Hints in machines.yaml or
machine-deployment.yaml for individual machines.

This can help users in situations where nova scheduler
will create workers on the same hypervisor.

Information regarding scheduler hints:
https://docs.openstack.org/nova/latest/reference/scheduler-hints-vs-flavor-extra-specs.html

Fixes #118 

Once this is approved for release-0.1, I will also implement it for v1alpha2

**Release note**:
NONE


oadivtal:
/retest
(after signing the CLA)

oadivtal:
This PR relies heavily on @chrigl 's work in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/275/

oadivtal:
I signed it

jichenjc:
seems you need sign cncf or refresh this is needed... @oadivtal

sbueringer:
@oadivtal Why are there so many deleted files in the vendor folder? :)

sbueringer:
@oadivtal Just a hint. We will be releasing 0.2.0 based on v1alpha2 in a week or so based on the master branch. I would suggest implementing this feature based on master.

oadivtal:
Signed in cncf, hope it works now. Sorry for the confusion

jichenjc:
@oadivtal please refer to the comments added by @sboeuf 
I see no reason to remove vendor subfolder... are you missed rebase action? 

and please consider to add to master branch about this feature later on.. thanks

docs/config.md
```diff
@@ -259,6 +259,113 @@ Instead of tagging, you also have the option to add metadata to instances. This
...
+You must reference this Server Group in the definition of your `MachineDeployment`
+
+```yaml
+apiVersion: cluster.k8s.io/v1alpha1
+kind: MachineDeployment
+metadata:
+  labels:
+    set: node
+  name: test
+  namespace: default
```
sbueringer: Let's shorten this example a bit by only showing the relevant parts for scheduler hints

pkg/cloud/openstack/services/compute/instance.go
```diff
@@ -215,7 +217,9 @@ func (is *Service) InstanceCreate(clusterName string, name string, clusterSpec *
 		return nil, fmt.Errorf("create new server err: %v", err)
 	}
 
-	serverCreateOpts := servers.CreateOpts{
+	var serverCreateOpts servers.CreateOptsBuilder
+
+	serverOpts := servers.CreateOpts{
```
sbueringer: why is the separate variable necessary?

pkg/cloud/openstack/services/compute/instance.go
```diff
@@ -530,3 +538,45 @@ func (is *Service) UpdateToken() error {
...
+	if config.SchedulerHints == nil {
+		return opts, nil
+	}
+	hints := queryFixer{
+		SchedulerHints: schedulerhints.SchedulerHints{
+			Group:           config.SchedulerHints.Group,
+			DifferentHost:   config.SchedulerHints.DifferentHost,
+			SameHost:        config.SchedulerHints.SameHost,
+			TargetCell:      config.SchedulerHints.TargetCell,
+			BuildNearHostIP: config.SchedulerHints.BuildNearHostIP,
```
sbueringer: Why are we not using the Query field? https://github.com/gophercloud/gophercloud/blob/19abc56a8cd8d3630ab9735173cb92c0bf635f33/openstack/compute/v2/extensions/schedulerhints/requests.go#L28

vendor/sigs.k8s.io/cluster-api/BUILD.bazel
```diff
@@ -1,21 +0,0 @@
-load("@bazel_gazelle//:def.bzl", "gazelle")
```
sbueringer: Why are all this files deleted?

docs/config.md
```diff
@@ -259,6 +259,113 @@ Instead of tagging, you also have the option to add metadata to instances. This
           nickname: bobbert
 ```
 
+
+## Scheduler hints
+OpenStack provides a way to define scheduler hints, which is useful to implement affinity/anti affinity. See [Nova Filter Scheduler](https://docs.openstack.org/nova/rocky/user/filter-scheduler.html) for more information.
```
jichenjc: s/rocky/latest/

docs/config.md
jichenjc: or maybe a new file and create a link to that file..

pkg/cloud/openstack/services/compute/instance.go
jichenjc: seems the query is listed below?

sbueringer: Yup but it's build custom on top and (at least the current) struct directly has it. So seems like this workaround is not requrired anymore

PR #475, opened 2019-09-24
use existing networks, fallback to network/subnet defined in cluster …

### LINKED ISSUE #428: support create LB with existing network/subnet

sbueringer:
This PR enables the following:
* reuse existing networks/subnet if they cannot be created (by defining NodeCIDR in the OpenStackCluster)
* define network/subnet on the cluster level so it doesn't has to be defined in every Machine (only works for one net/subnet not multiple)

Fixes #428

### DIFF
--- a/api/v1alpha2/openstackcluster_types.go
+++ b/api/v1alpha2/openstackcluster_types.go
@@ -42,6 +42,13 @@ type OpenStackClusterSpec struct {
 	// network, a subnet with NodeCIDR, and a router connected to this subnet.
 	// If you leave this empty, no network will be created.
 	NodeCIDR string `json:"nodeCidr,omitempty"`
+
+	// If NodeCIDR cannot be set this can be used to detect an existing network.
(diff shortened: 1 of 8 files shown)


sbueringer:
/assign @jichenjc 
/assign @hidekazuna 
/hold

I didn't test it yet myself but I'm on it :)

sbueringer:
Okay I tested my existing multi-node setup and I tested it with remove networks filter. Both worked

/hold cancel

@jichenjc Can you test if you're now able to use loadbalancers by setting the network and subnet fields (filters) in the ClusterSpec?

sbueringer:
Was fixed in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/477

PR #476, opened 2019-10-01
Generate seperate multi-node and single-node yaml

### LINKED ISSUE #473: Generate separate multi-node and single-node control plane examples
/kind feature

We're currently just generating one example for single and multi-node. Because the configurations for both are different it would be nice if the examples/generate.sh shell script could be adjusted to take a parameter `single-node-controlplane` or `multi-node-controlplane` and generate examples accordingly (or just generate both)

prankul88:

Fixes #473 

1. As suggested by @sbueringer , I created separate folders for single-node/multi-node cluster and controlplane.
2. generate.yaml takes input "single-node" /"multi-node" and generates yaml accordingly 

**Release note**:

```release-note

```


examples/generate.sh
```diff
@@ -83,8 +83,23 @@ else
   exit 1
 fi
 
+# Check if correct controlplane preference is given
 if [[ -n "${3-}" ]] && [[ $3 != -* ]] && [[ $3 != --* ]]; then
-  OUTPUT_DIR=$(echo $3 | tr '[:upper:]' '[:lower:]')
+  if [[ "$3" == "single-node" || "$3" == "multi-node" ]]; then
+    export CONTROLPLANE=$3
+  else"$
```
sbueringer:
```suggestion
  else
```

I think you have to remove the "$

examples/generate.sh
```diff
@@ -190,12 +205,22 @@ else
 fi
 
 # Generate cluster resources.
-kustomize build "${SOURCE_DIR}/cluster" | envsubst > "${CLUSTER_GENERATED_FILE}"
-echo "Generated ${CLUSTER_GENERATED_FILE}"
+if [[ $CONTROLPLANE == "single-node" ]]; then
```
sbueringer: I would propose to call these folder cluster-single-node and cluster-multi-node then you can remove the if else and just use the $CONTROLPLANE var directly in the folder path

examples/generate.sh
```diff
@@ -190,12 +205,22 @@ else
...
+  echo "Generated ${CLUSTER_GENERATED_FILE}"
+else
+  kustomize build "${SOURCE_DIR}/cluster/cluster-multi" | envsubst > "${CLUSTER_GENERATED_FILE}"
+  echo "Generated ${CLUSTER_GENERATED_FILE}"
+fi
 
 # Generate controlplane resources.
-kustomize build "${SOURCE_DIR}/controlplane" | envsubst > "${CONTROLPLANE_GENERATED_FILE}"
-echo "Generated ${CONTROLPLANE_GENERATED_FILE}"
+if [[ $CONTROLPLANE == "single-node" ]]; then
```
sbueringer: same here

examples/cluster/cluster-single/cluster.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: ${CLUSTER_NAME}
+  namespace: ${CLUSTER_NAME}
+spec:
+  clusterNetwork:
+    services:
+      cidrBlocks: ["10.96.0.0/12"]
+    pods:
+      cidrBlocks: ["192.168.0.0/16"]
+    serviceDomain: "cluster.local"
+    apiServerPort: <disable when multi-node control-plane>
```
sbueringer: let's just use the value 6443 here

sbueringer:
@prankul88 
Sorry for the late review. I also want to test the new generated files sometime in the next week. Don't have time for that right now.

Can you also do the following changes in controlplane.yaml

* Remove these blocks:
  * multi
```
  # multi-node control-plane:
  # * Disable the floatingIP property
  # single-node control-plane:
  # * Enable the floatingIP property
  #floatingIP: <floating IP>
```
  * single 
```
  # multi-node control-plane:
  # * Disable the floatingIP property
  # single-node control-plane:
  # * Enable the floatingIP property
```

* Theres some documentation around: clusterConfiguration.controlPlaneEndpoint. Let's move that inside the value of controlPlaneEndpoint (which would be `<loadbalancer ip>` and `<floating ip of control plane node>` respectively

Let's use

examples/generate.sh
jichenjc: +1

examples/controlplane/controlplane-single/controlplane.yaml
```diff
@@ -0,0 +1,115 @@
+#####################################################
+# ${CLUSTER_NAME}-controlplane-0
+#####################################################
+apiVersion: cluster.x-k8s.io/v1alpha2
+kind: Machine
+metadata:
+  name: ${CLUSTER_NAME}-controlplane-0
```
jichenjc: not sure keep this -0 is still useful, ok to keep it

examples/controlplane/controlplane-single/controlplane.yaml
```diff
@@ -0,0 +1,115 @@
...
+  networks:
+  - filter:
+      name: k8s-clusterapi-cluster-${CLUSTER_NAME}-${CLUSTER_NAME}
+    subnets:
+    - filter:
+        name: k8s-clusterapi-cluster-${CLUSTER_NAME}-${CLUSTER_NAME}
+  # multi-node control-plane:
+  # * Disable the floatingIP property
+  # single-node control-plane:
+  # * Enable the floatingIP property
```
jichenjc: do we still want to keep those comments?

examples/controlplane/controlplane-single/controlplane.yaml
```diff
@@ -0,0 +1,115 @@
...
+      kubeletExtraArgs:
+        cloud-provider: openstack
+        cloud-config: /etc/kubernetes/cloud.conf
+  clusterConfiguration:
+    # multi-node control-plane:
+    # * Configure the floating ip of the loadbalancer
+    # single-node control-plane:
+    # * Configure the floating ip of the (only) control
+    #   plane node
+    controlPlaneEndpoint: "<floating ip>:6443"
```
jichenjc: ditto

examples/generate.sh
```diff
@@ -83,8 +83,23 @@ else
   exit 1
 fi
 
+# Check if correct controlplane preference is given
 if [[ -n "${3-}" ]] && [[ $3 != -* ]] && [[ $3 != --* ]]; then
-  OUTPUT_DIR=$(echo $3 | tr '[:upper:]' '[:lower:]')
+  if [[ "$3" == "single-node" || "$3" == "multi-node" ]]; then
```
jichenjc: not sure if we can have a default value such as single-node so that by default only one machine created?
the new comer might be esaier to get start from our provider instead of learning too much concepts..

examples/controlplane/controlplane-single/controlplane.yaml
sbueringer: It would make it more consistent if the cluster is Lager upgraded to multi-node. Also the diff is smaller to multi so its easier to compare

jichenjc: ok, works for me

examples/generate.sh
prankul88: Agreed, this will simplify things !!
Would it not be better to just name it "single-node" and "multi-node" instead. Because using naming "cluster-single-node" would not go well in control-plane folder. Just asking, will do it what fits best

examples/cluster/cluster-single/cluster.yaml
prankul88: Will update. Thanks

examples/controlplane/controlplane-single/controlplane.yaml
prankul88: Would delete it. Thanks

examples/generate.sh
prankul88: Sure. Can set the default to "single-node"

examples/generate.sh
prankul88: My bad. Will update

examples/generate.sh
sbueringer: What do you mean with go well? The folder would be relatively long but we definitely need separate folders for control plane an cluster (or at least it would be easier if it's not mixed)

examples/generate.sh
sbueringer: +1

sbueringer:
@prankul88 test failure is fixed on my pr. See the changes in the Makefile: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/477/files

This was caused by a change in the kustomize release process.

examples/generate.sh
sbueringer: Ah now I know what you mean. If we have a cluster and controlplane folder it would be better to not repeat cluster and controlplane. So yes you're right. For example cluster/multi-node would be better.

sbueringer:
@prankul88 Very nice overall :). Only the folder rename is left and it would make sense to use the test fix from my PR so you don't have to wait until mine is merged.

examples/generate.sh
prankul88: Ya sure I will. :)

prankul88:
@sbueringer Great. Thanks for bringing out the job failure reason. I will update the changes soon.

sbueringer:
@prankul88 You can also rebase on master for the test fix

examples/generate.sh
```diff
@@ -94,7 +93,10 @@ if [[ -n "${3-}" ]] && [[ $3 != -* ]] && [[ $3 != --* ]]; then
     exit 1
   fi
 else
-  echo "Generating yaml for \"single-node\" cluster by default"
```
prankul88: @jichenjc @sbueringer I had earlier set `single-node` as default but would leed to weird implementation. 

If we set "single-node" as default, it would mean that command like `./generate.sh clouds.yaml openstack <folder directory>` would be invalid. 
Default "single-node" would only work in case we dont give out folder directory Eg. "./generate.sh clouds.yaml openstack"

We can have 2 ways out:
1. Don't have a default value for control-plane preference and ask user to mandatory ask for preference. (update docs for that)
2. Switch order of user input. Eg. If we can take something like `./generate.yaml <clouds.yaml> <cloud> <out directory> <controlplane preference>`. (which would also leed to the same situation for out directory)

sbueringer: The problem is it's not straightforward to support multiple optional parameters this way. It would be possible to call it like this `./generate.yaml <clouds.yaml> <cloud> "" <out directory>` Then the default value would be used. But that's also not really convenient.

But for me it would be okay to change the order and declare the output folder as mandatory parameter. In a lot of other scripts/tools it is. It's not really intuitive that _out is a default :)

prankul88: sounds good !!

prankul88:
@sbueringer Sorry did not see your network patch got merged. Will have to rebase

sbueringer:
> @sbueringer Sorry did not see your network patch got merged. I guess l have to rebase and update the controlplane.yaml according to #477

You don't have to change anything because of #477 . It's just another variant is supported if it's not allowed to create networks and subnets in OpenStack. I would like to keep the default the same as it is right now

sbueringer:
@prankul88 so only the paramter change is left, right? Could you also update the script documentation in examples/README.md?

sbueringer:
@prankul88 It would be far easier if you just push individual commits :). By force pushing an individual commit it's hard to figure out what actually changed. Prow will squash on merge anyway.

prankul88:
@sbueringer sorry, I would push them individually from now on. Would like to clarify the latest commit I have pushed. 
1. Have updated the controlpane.yaml and machinedeployments.yaml ( according to #477 )
2. Updated generate.sh which makes `<output dir>` as mandatory input and `"single-node"` as the default parameter.
3. Updated README file accordingly

Kindly take a look.

sbueringer:
@prankul88 Of course. Just wanted to tell you that it's easier that way :)

sbueringer:
@prankul88 Looks good. Apparently they changed something in the upstream ClusterAPI repos. Can you replace l.223-229 in generate.sh with:

```
# Generate Cluster API provider components file.
CAPI_BRANCH=${CAPI_BRANCH:-"stable"}
if [[ ${CAPI_BRANCH} == "stable" ]]; then
  curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.2.4/cluster-api-components.yaml > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
  echo "Downloaded ${COMPONENTS_CLUSTER_API_GENERATED_FILE} from cluster-api stable branch - v0.2.4"
else
  kustomize build "github.com/kubernetes-sigs/cluster-api/config/default/?ref=${CAPI_BRANCH}" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
  echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE} from cluster-api - ${CAPI_BRANCH}"
fi

# Generate Kubeadm Bootstrap Provider components file.
CABPK_BRANCH=${CABPK_BRANCH:-"stable"}
if [[ ${CABPK_BRANCH} == "stable" ]]; then
  curl -L https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/releases/download/v0.1.2/bootstrap-components.yaml > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
  echo "Downloaded ${COMPONENTS_KUBEADM_GENERATED_FILE} from cluster-api-bootstrap-provider-kubeadm stable branch - v0.1.2"
else
  kustomize build "github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/config/default/?ref=${CABPK_BRANCH}" > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
  echo "Generated ${COMPONENTS_KUBEADM_GENERATED_FILE} from cluster-api-bootstrap-provider-kubeadm - ${CABPK_BRANCH}"
fi
```

Everything else lgtm :)

prankul88:
@sbueringer Have updated !! PTAL

sbueringer:
Perfect. Thx very much :-)

/lgtm
/approve

PR #477, opened 2019-10-06
Pr use existing networks

sbueringer:
Ups, deleted the old branch so the old PR was closed. So again:

Okay I tested my existing multi-node setup and I tested it with remove networks filter. Both worked

@jichenjc Can you test if you're now able to use loadbalancers by setting the network and subnet fields (filters) in the ClusterSpec?

Edit: Old PR was this one: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/475

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ ifndef HAS_YQ
 endif
 ifndef HAS_KUSTOMIZE
 	echo "installing kustomize"
-	go get sigs.k8s.io/kustomize/kustomize
+	GO111MODULE=off go get sigs.k8s.io/kustomize/kustomize
 endif
 ifndef HAS_ENVSUBST
 	echo "installing envsubst"
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ ifndef HAS_YQ
 endif
 ifndef HAS_KUSTOMIZE
 	echo "installing kustomize"
-	go get sigs.k8s.io/kustomize/v3/cmd/kustomize
+	go get sigs.k8s.io/kustomize/kustomize
 endif
 ifndef HAS_ENVSUBST
 	echo "installing envsubst"
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -18,6 +18,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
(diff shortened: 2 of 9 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -99,36 +97,51 @@ func (s *Service) InstanceCreate(clusterName string, machine *clusterv1.Machine,
...
+							networkID: subnetByFilter.NetworkID,
+							subnetID:  subnetByFilter.ID,
+						})
+					}
 				}
 			}
 		}
+	} else {
+		if openStackCluster.Status.Network == nil {
+			return nil, fmt.Errorf("no network was found in OpenStackCluster.Status.Network")
```
jichenjc: maybe we need add some info that because the machine doesn't have network spec and cluster network is also nil
so we return error..

controllers/openstackcluster_controller.go
```diff
@@ -129,7 +131,35 @@ func (r *OpenStackClusterReconciler) reconcileCluster(logger logr.Logger, cluste
 
 	logger.Info("Reconciling network components")
 	if openStackCluster.Spec.NodeCIDR == "" {
-		logger.V(4).Info("No need to reconcile network")
+		logger.V(4).Info("No need to reconcile network, searching network and subnet instead")
+
+		netOpts := networks.ListOpts(openStackCluster.Spec.Network)
+		networkList, err := networkingService.GetNetworksByFilter(&netOpts)
+		if err != nil && len(networkList) == 0 {
+			return reconcile.Result{}, errors.Errorf("failed to find network: %v", err)
```
jichenjc: do we allow the filter to be none? if so will this return error breaks the whole reconcile but actually we can continue?

pkg/cloud/services/compute/instance.go
sbueringer: Agree. Fixed

controllers/openstackcluster_controller.go
sbueringer: If neither the CIDR nor the Network Filter is configured it would break here. But I think this makes sense. It should be possible to fullfill all cases with those two properties. I think all other cases are easier to reason about if we can enforce here that we have a network and a subnet in the status. If not it just gets even more complicated.

Makefile
```diff
@@ -101,7 +101,7 @@ ifndef HAS_YQ
 endif
 ifndef HAS_KUSTOMIZE
 	echo "installing kustomize"
-	go get sigs.k8s.io/kustomize/v3/cmd/kustomize
+	GO111MODULE=off go get sigs.k8s.io/kustomize/kustomize
```
jichenjc: let me steal this change to my leader election PR :)

sbueringer: Of course :)

controllers/openstackcluster_controller.go
jichenjc: ok, let's get this in and if we have further question, we can work on that

Issue #480, opened 2019-10-09
CAPO Release v0.2.0


sbueringer:
Hi,

I would propose releasing a v0.2.0 version of CAPO after these 3 PRs are merged:
* [x] [LeaderElectionID should be unique](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/479)
* [x] [Pr use existing networks](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/477)
* [x] [Generate seperate multi-node and single-node yaml](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/476)
* [x] Release image (https://github.com/kubernetes/k8s.io/pull/433)
* [x] Finalize doc (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/488)
* [x] Add CAPO to cluster api book (https://github.com/kubernetes-sigs/cluster-api/pull/1661)

@jichenjc @hidekazuna @prankul88 
What do you think? We can always release follow-up patch releases

The process is more or less opening a PR to publish our staging images to the production buckets.

prankul88:
Sounds Good to me!

jichenjc:
the plan works for me  :)

hidekazuna:
Good to me, too.

PR #488, opened 2019-10-26
finalize doc & examples for v0.2.0

sbueringer:
Finalizes doc & examples for v0.2.0

See also https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/480

### DIFF
--- a/examples/cluster/multi-node/cluster.yaml
+++ b/examples/cluster/multi-node/cluster.yaml
@@ -9,7 +9,7 @@ spec:
     services:
       cidrBlocks: ["10.96.0.0/12"]
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks: ["192.168.0.0/16"] # CIDR block used by Calico.
     serviceDomain: "cluster.local"
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
@@ -40,7 +40,6 @@ spec:
   #   * creating floating ips
   #   * creating load balancer
   externalNetworkId: <external-network-id>
-  managedSecurityGroups: false
   disablePortSecurity: true
   disableServerTags: true
   useOctavia: true
--- a/examples/controlplane/multi-node/controlplane.yaml
+++ b/examples/controlplane/multi-node/controlplane.yaml
@@ -31,7 +31,6 @@ metadata:
 spec:
   flavor: m1.medium
   image: <Image Name>
-  sshKeyName: cluster-api-provider-openstack
   availabilityZone: nova
   cloudName: $CLOUD
   cloudsSecret:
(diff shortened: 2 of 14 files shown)


examples/cluster/multi-node/cluster.yaml
```diff
@@ -40,7 +40,6 @@ spec:
   #   * creating floating ips
   #   * creating load balancer
   externalNetworkId: <external-network-id>
-  managedSecurityGroups: false
```
prankul88: Hi, would probably need to update [config file](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/config.md#security-groups) as well if we remove this.

sbueringer: Why? I just removed it because it's the default value. So it shouldn't change anything

jichenjc:
generally looks good, don't have to time to go through this recently
so maybe let @prankul88 provide some feed back and let's merge this and evolute from there

examples/controlplane/multi-node/controlplane.yaml
```diff
@@ -59,21 +58,24 @@ spec:
...
     - "$MACHINE_CONTROLLER_SSH_PUBLIC_FILE_CONTENT"
+  # For more information about these values,
+  # refer to the Kubeadm Bootstrap Provider documentation.
   initConfiguration:
     localAPIEndpoint:
       advertiseAddress: '{{ ds.ec2_metadata.local_ipv4 }}'
       bindPort: 6443
     nodeRegistration:
       name: '{{ local_hostname }}'
+      criSocket: "/var/run/containerd/containerd.sock"
```
detiber: You shouldn't need this here (unless you are attempting to support k8s v1.13), since kubeadm should autodetect the criSocket for you by default.

examples/controlplane/single-node/controlplane.yaml
```diff
@@ -60,21 +59,24 @@ spec:
...
     - "$MACHINE_CONTROLLER_SSH_PUBLIC_FILE_CONTENT"
+  # For more information about these values,
+  # refer to the Kubeadm Bootstrap Provider documentation.
   initConfiguration:
     localAPIEndpoint:
       advertiseAddress: '{{ ds.ec2_metadata.local_ipv4 }}'
       bindPort: 6443
     nodeRegistration:
       name: '{{ local_hostname }}'
+      criSocket: "/var/run/containerd/containerd.sock"
```
detiber: You shouldn't need this here (unless you are attempting to support k8s v1.13), since kubeadm should autodetect the criSocket for you by default.

examples/machinedeployment/machinedeployment.yaml
```diff
@@ -73,13 +72,15 @@ spec:
         permissions: "0600"
       joinConfiguration:
         nodeRegistration:
+          name: '{{ local_hostname }}'
+          criSocket: "/var/run/containerd/containerd.sock"
```
detiber: You shouldn't need this here (unless you are attempting to support k8s v1.13), since kubeadm should autodetect the criSocket for you by default.

examples/controlplane/multi-node/controlplane.yaml
sbueringer: Thx. I think you're right. And I think it's okay to not trying to support 1.13 :)

sbueringer:
@prankul88 can you take another look and tell us if it's okay for you. So somebody else, e.g. @jichenjc can lgtm & approve for you

examples/cluster/multi-node/cluster.yaml
prankul88: Ohh okay, I thought maybe there is no need to mention that in config file. But looks good now

prankul88:
@sbueringer looks good to me.

sbueringer:
> @sbueringer looks good to me.

Thx :)

@jichenjc Can you approve/lgtm the PR?

PR #497, opened 2020-01-21
Add additional description for DNS

jichenjc:
I am working on my env and noticed that the new VM is not able to
start because my VM can't talk to openstack URL (which is web URI
instead of IP), so we need guide user to create this DNS explicility.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/api/v1alpha2/openstackcluster_types.go
+++ b/api/v1alpha2/openstackcluster_types.go
@@ -50,6 +50,8 @@ type OpenStackClusterSpec struct {
 	Subnet SubnetFilter `json:"subnet,omitempty"`
 
 	// DNSNameservers is the list of nameservers for OpenStack Subnet being created.
+	// Set this value when you need create a new network/subnet while the access
+	// through DNS is required.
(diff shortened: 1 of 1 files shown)


sbueringer:
Hm I'm a little bit confused when exactly this property is needed. I wasn't even aware the property exists until now and I did not set it in any of my clusters. (I'm also accessing OpenStack via URL not IP)

jichenjc:
it's needed. here's the reason :)

the network is created by our logic so if we pass the DNS param, the /etc/resolv.conf of openstack deployed VM will be updated to include the DNS location , my openstack URL is abc.com (different, but like this), not IP ,so when the kubeadm start, because it's openstack cloud provider, it's trying to reach the URL to get additional info, if the DNS is not set, then the URL might not be recognized so that the kubeadm/kubelet is not able to correctly start

sbueringer:
@jichenjc Never mind. I'm actually setting this property and forgot that I did :)

That explains why it works in my env

/lgtm

PR #500, opened 2020-01-25
🐛 don't require cacert in cloud.yaml

sbueringer:

Before this PR it was mandatory to set a cacert. Without it the cloud.yaml was not parsed correctly.

Bug was identified here: https://kubernetes.slack.com/archives/CFKJB65G9/p1579624001004800?thread_ts=1579502983.003200&cid=CFKJB65G9

### DIFF
--- a/pkg/cloud/services/provider/provider.go
+++ b/pkg/cloud/services/provider/provider.go
@@ -84,7 +84,9 @@ func newClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 	if cloud.Verify != nil {
 		config.InsecureSkipVerify = !*cloud.Verify
 	}
-	config.RootCAs.AppendCertsFromPEM(caCert)
+	if caCert != nil {
+		config.RootCAs.AppendCertsFromPEM(caCert)
+	}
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/provider/provider.go
```diff
@@ -84,7 +84,9 @@ func newClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 	if cloud.Verify != nil {
 		config.InsecureSkipVerify = !*cloud.Verify
 	}
-	config.RootCAs.AppendCertsFromPEM(caCert)
+	if caCert != nil {
+		config.RootCAs.AppendCertsFromPEM(caCert)
+	}
```
prankul88: @sbueringer What would happen if in this case caCert is nil?

sbueringer: The cacert is not added? :)

So either you set a cacert then it's also used to verify the OpenStack Certs or you don't and it isn't used (only the rootCAs)

prankul88: @sbueringer Got it thanks :+1:

jichenjc: that's my bad, actually I had a mandatory cacert requirement but agree it's optiponal

Issue #499, opened 2020-01-22
migration of cluster-api v1alpha2 ==> v1alpha3


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

add v1alpha3 support due to clusterctl will be deprecated later, I tried 
https://cluster-api.sigs.k8s.io/user/quick-start.html able to create control pane but failed to create machines due to 
```
E0122 04:27:05.167452       1 controller.go:218] controller-runtime/controller "msg"="Reconciler error" "error"="failed to retrieve Status.APIEndpoints from infrastructure provider for Cluster \"test-cluster\" in namespace \"test-cluster\": field not found"  "controller"="cluster" "request"={"Namespace":"test-cluster","Name":"test-cluster"
```
this is listed https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/providers/v1alpha2-to-v1alpha3.md

so the to do is
- [x] add v1alpha3 types
- [x] modify code to fit for v1alpha3 types
- [x] modify doc to remove clusterctl and a new method to create cluster
- [x] a new release of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/releases/download/v0.2.0/infrastructure-components.yaml , currently something is missing such as namespace, secret, machines CRD etc

   

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

jichenjc:
@sbueringer @prankul88 let me you know any comments on this ... thanks

sbueringer:
@jichenjc I'm just starting to read up on v1alpha3. Do you have time to work on this issue?

jichenjc:
yes, I have some time in upcoming 10-15 days so I will work on it recently


https://master.cluster-api.sigs.k8s.io/ is the book for v1alpha3 and 
https://cluster-api.sigs.k8s.io/ is for v1alpha2

so my plan is to make use my image (created locally) and switch code to v1alpha3 with master CAPI code ...

prankul88:
@jichenjc +1 for the migration. Should there be a new branch or the current master branch would do?

>Are you sure upstream ClusterAPI Book is already based on v1alpha3?

Don't think it is completely in line with the v1Alpha3. Although it is frequently updated based on master

jichenjc:
[...]

PR #516, opened 2020-03-11
🐛 fixup Makefile, should fix image build on master

sbueringer:
**What this PR does / why we need it**:

Syncs the Makefile with CAPA and should fix our build on master

Also: 
As we're not really (imho) supporting migrations from v1alpha2 to v1alpha3 and the v1alpha2 package was unused in our implementation I would suggest removing it. I'm happy when we have v1alpha3 finalized I don't think we have the bandwidth to support migrations explicitly right now

See also: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/499

PR #517, opened 2020-03-11
⚠️  remove unused v1alpha2 package and update doc

sbueringer:

As we're not really (imho) supporting migrations from v1alpha2 to v1alpha3 and the v1alpha2 package was unused in our implementation I would suggest removing it. I'm happy when we have v1alpha3 finalized I don't think we have the bandwidth to support migrations explicitly right now

See also: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/499


hidekazuna:
@sbueringer I agree we remove v1alpha2, I am not sure Cluster API policy though.
This PR seems conflict with #516 which drops v1alpha2 partially.

sbueringer:
> @sbueringer I agree we remove v1alpha2, I am not sure Cluster API policy though.
> This PR seems conflict with #516 which drops v1alpha2 partially.

Shouldn't conflict. The goal of the other one was mainly to fix the makefile so our images after merge on master are build again, because that's broken for a while now. It's just that the regeneration with the new makefile removed a lot of v1alpha2 occurences.

Afaik there is no official policy. Or at least it doesn't make sense to keep orphaned packages when we're not using them

sbueringer:
I'll close this PR here and use just the other one to make it not even more complicated. Sorry :/

PR #521, opened 2020-03-22
⚠️ fixup namespace handling in 0.3.0

sbueringer:
I just did a final manual test and saw that the quickstart with 0.3.0 only worked when --target-namespace was equal to the cluster-name. This PRs fixes our templates so that the namespace where all resources are deployed is equal to our secret references in the CAPO resources

P.S. I already replaced the uploaded cluster-templates in the published 0.3.0 release to avoid that somebody who is trying CAPO out right now directly fails with this error, which would be bad :)


jichenjc:
looks good, 
provide an opportunity for other folks take a look :)

hidekazuna:
@sbueringer @jichenjc This worked fine in my environment.

jichenjc:
/lgtm 
@hidekazuna thanks :)

PR #522, opened 2020-03-24
✨ Delete router/network/subnet

hidekazuna:

This PR adds to delete router, network, and subnet feature. Currently how to delete cluster is not provided explicitly in my understanding, but we need to delete OpenStack resources easily even for testing. By this PR, we can delete OpenStack resources created with `cluster-template-without-lb.yaml` by the following.

```
kubectl delete machinedeployment capi-quickstart-md-0 -n capi-quickstart
kubectl delete kubeadmcontrolplane capi-quickstart-control-plane -n capi-quickstart
kubectl delete cluster capi-quickstart -n capi-quickstart
```

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -150,7 +150,23 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, log lo
 		}
 	}
 
-	// TODO(sbueringer) Delete network/subnet/router/... if created by CAPO
+	if openStackCluster.Status.Network.Router != nil {
+		log.Info("Deleting router", "name", openStackCluster.Status.Network.Router.Name)
+		if err := networkingService.DeleteRouter(openStackCluster.Status.Network); err != nil {
+			return ctrl.Result{}, errors.Errorf("failed to delete router: %v", err)
+		}
+		log.Info("OpenStack router deleted successfully")
+	}
+
+	if openStackCluster.Status.Network != nil {
+		log.Info("Deleting network", "name", openStackCluster.Status.Network.Name)
+		if err := networkingService.DeleteNetwork(openStackCluster.Status.Network); err != nil {
+			return ctrl.Result{}, errors.Errorf("failed to delete network: %v", err)
+		}
+		log.Info("OpenStack network deleted successfully")
+	}
+
+	log.Info("OpenStack cluster deleted successfully")
 
 	// Cluster is deleted so remove the finalizer.
 	controllerutil.RemoveFinalizer(openStackCluster, infrav1.ClusterFinalizer)
@@ -297,6 +313,7 @@ func (r *OpenStackClusterReconciler) reconcileNormal(ctx context.Context, log lo
 	}
 
 	openStackCluster.Status.Ready = true
+	log.Info("Reconciled Cluster create successfully")
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -91,6 +91,22 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	return nil
 }
 
+func (s *Service) DeleteNetwork(network *infrav1.Network) error {
+	if network == nil || network.ID == "" {
+		s.logger.V(4).Info("No need to delete network since no network exists.")
+		return nil
+	}
+	exists, err := s.existsNetwork(network.ID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		s.logger.Info("Skipping network deletion because network doesn't exist", "network", network.ID)
+		return nil
+	}
+	return networks.Delete(s.client, network.ID).ExtractErr()
+}
+
 func (s *Service) ReconcileSubnet(clusterName string, openStackCluster *infrav1.OpenStackCluster) error {
 
 	if openStackCluster.Status.Network == nil || openStackCluster.Status.Network.ID == "" {
@@ -187,4 +203,18 @@ func (s *Service) getNetworkByName(networkName string) (networks.Network, error)
 		return allNetworks[0], nil
 	}
 	return networks.Network{}, errors.New("too many resources")
+}
+
+func (s *Service) existsNetwork(networkID string) (bool, error) {
+	if networkID == "" {
+		return false, nil
+	}
+	network, err := networks.Get(s.client, networkID).Extract()
+	if err != nil {
+		return false, err
+	}
+	if network == nil {
+		return false, nil
+	}
+	return true, nil
 }
--- a/pkg/cloud/services/networking/router.go
+++ b/pkg/cloud/services/networking/router.go
@@ -165,6 +165,33 @@ INTERFACE_LOOP:
 	return nil
 }
 
+func (s *Service) DeleteRouter(network *infrav1.Network) error {
+	if network.Router == nil || network.Router.ID == "" {
+		s.logger.V(4).Info("No need to delete router since no router exists.")
+		return nil
+	}
+	exists, err := s.existsRouter(network.Router.ID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		s.logger.Info("Skipping router deletion because router doesn't exist", "router", network.Router.ID)
+		return nil
+	}
+	if network.Subnet == nil || network.Subnet.ID == "" {
+		s.logger.V(4).Info("Skipping removing router interface since no subnet exists.")
+	} else {
+		_, err = routers.RemoveInterface(s.client, network.Router.ID, routers.RemoveInterfaceOpts{
+			SubnetID: network.Subnet.ID,
+		}).Extract()
+		if err != nil {
+			return fmt.Errorf("unable to remove router interface: %v", err)
+		}
+		s.logger.V(4).Info("Removed RouterInterface of Router", "id", network.Router.ID)
+	}
+	return routers.Delete(s.client, network.Router.ID).ExtractErr()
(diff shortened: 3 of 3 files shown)


sbueringer:
looks good to me. I'll test it later. Don't we have to delete the subnet?

hidekazuna:
@sbueringer When we delete a network, The subnet is deleted as well.

sbueringer:
> @sbueringer When we delete a network, The subnet is deleted as well.

okay thx good to know

controllers/openstackcluster_controller.go
```diff
@@ -150,7 +150,24 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, log lo
...
+	if openStackCluster.Status.Network.Router != nil {
+		log.Info("Deleting router", "name", openStackCluster.Status.Network.Router.Name)
+		err := networkingService.DeleteRouter(openStackCluster.Status.Network)
+		if err != nil {
+			return ctrl.Result{}, errors.Errorf("failed to delete router: %v", err)
+		}
+		log.Info("OpenStack router deleted successfully")
+	}
+
+	if openStackCluster.Status.Network != nil {
```
jichenjc: I don't remember clearly, do we have a chance that the network/router is not created by CPAO? 
others LGTM

pkg/cloud/services/networking/network.go
```diff
@@ -91,6 +91,15 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	return nil
 }
 
+func (s *Service) DeleteNetwork(network *infrav1.Network) error {
+
```
sbueringer: Can we change that code so it doesn't fail when the network already has been deleted?

Problem might be that we delete the network are somehow not able to patch the openstackcluster and the next reconcile fails because the network is already gone.

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,27 @@ INTERFACE_LOOP:
...
+	} else {
+		_, err := routers.RemoveInterface(s.client, network.Router.ID, routers.RemoveInterfaceOpts{
+			SubnetID: network.Subnet.ID,
+		}).Extract()
+		if err != nil {
+			return fmt.Errorf("unable to remove router interface: %v", err)
+		}
+		s.logger.V(4).Info("Removed RouterInterface of Router", "id", network.Router.ID)
+	}
+	return routers.Delete(s.client, network.Router.ID).ExtractErr()
```
sbueringer: Can we change that code so it doesn't fail when the router already has been deleted?

Problem might be that we delete the router are somehow not able to patch the openstackcluster and the next reconcile fails because the network is already gone.

controllers/openstackcluster_controller.go
sbueringer: I think it's okay as the cluster network/subnet are afaik always created by us

sbueringer:
@hidekazuna two small nits, apart from that everything worked :)

controllers/openstackcluster_controller.go
jichenjc: ok, I didn't remember clearly, will try this later on and submit PR if need :)

hidekazuna:
@sbueringer Thanks, fixed.

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,30 @@ INTERFACE_LOOP:
 	return nil
 }
 
+func (s *Service) DeleteRouter(network *infrav1.Network) error {
+	exists, err := s.existsRouter(network.Router.ID)
+	if err != nil {
+		return err
+	}
+	if exists {
```
sbueringer: I think a guard close would be more idiomatic in go, i.e.:

```golang
if !exist {
    s.logger.Info("No router", "router", network.Router.ID)
    return nil
}
if network.Subnet == nil || network.Subnet.ID == "" {
    s.logger.V(4).Info("No need to remove router interface since no subnet exists.")
    return nil
}
_, err := routers.RemoveInterface(s.client, network.Router.ID, routers.RemoveInterfaceOpts{
    SubnetID: network.Subnet.ID,
}).Extract()
if err != nil {
    return fmt.Errorf("unable to remove router interface: %v", err)
}
s.logger.V(4).Info("Removed RouterInterface of Router", "id", network.Router.ID)

err = routers.Delete(s.client, network.Router.ID).ExtractErr(

```
(same for DeleteNetwork)

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,30 @@ INTERFACE_LOOP:
 	return nil
 }
 
+func (s *Service) DeleteRouter(network *infrav1.Network) error {
+	exists, err := s.existsRouter(network.Router.ID)
```
sbueringer: I think you still have to check for empty routerID as in you're previous commit. The problem is if the ID is an empty string
and you list all routers in existsRouter without an ID you get all existing routers (same for network)

controllers/openstackcluster_controller.go
```diff
@@ -150,7 +150,24 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, log lo
 		}
 	}
 
-	// TODO(sbueringer) Delete network/subnet/router/... if created by CAPO
+	if openStackCluster.Status.Network.Router != nil {
+		log.Info("Deleting router", "name", openStackCluster.Status.Network.Router.Name)
+		err := networkingService.DeleteRouter(openStackCluster.Status.Network)
+		if err != nil {
```
sbueringer: when you're already on it :). You can inline this to
```golang
if err := networkingService.DeleteRouter(openStackCluster.Status.Network); err != nil {
```

hidekazuna:
@sbueringer Fixed.

pkg/cloud/services/networking/network.go
```diff
@@ -91,6 +91,23 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	return nil
 }
 
+func (s *Service) DeleteNetwork(network *infrav1.Network) error {
+	if network == nil || network.ID == "" {
+		s.logger.V(4).Info("No need to delete network since no networrk exists.")
```
sbueringer:
```suggestion
		s.logger.V(4).Info("No need to delete network since no network exists.")
```

pkg/cloud/services/networking/network.go
```diff
@@ -91,6 +91,23 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
...
+		return nil
+	}
+	exists, err := s.existsNetwork(network.ID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		s.logger.Info("No network", "network", network.ID)
```
sbueringer:
```suggestion
		s.logger.Info("Skipping network deletion because network doesn't exist", "network", network.ID)
```

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,34 @@ INTERFACE_LOOP:
...
+	if network.Router == nil || network.Router.ID == "" {
+		s.logger.V(4).Info("No need to delete router since no router exists.")
+		return nil
+	}
+	exists, err := s.existsRouter(network.Router.ID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		s.logger.Info("No router", "router", network.Router.ID)
```
sbueringer:
```suggestion
		s.logger.Info("Skipping router deletion because router doesn't exist", "router", network.Router.ID)
```

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,34 @@ INTERFACE_LOOP:
...
+		return nil
+	}
+	if network.Subnet == nil || network.Subnet.ID == "" {
+		s.logger.V(4).Info("No need to remove router interface since no subnet exists.")
```
sbueringer: Why do we skip the whole router deletion if subnet doesn't exist? I think we only have to skip `RemoveInterface` ?

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,34 @@ INTERFACE_LOOP:
...
+	}
+	_, err = routers.RemoveInterface(s.client, network.Router.ID, routers.RemoveInterfaceOpts{
+		SubnetID: network.Subnet.ID,
+	}).Extract()
+	if err != nil {
+		return fmt.Errorf("unable to remove router interface: %v", err)
+	}
+	s.logger.V(4).Info("Removed RouterInterface of Router", "id", network.Router.ID)
+	err = routers.Delete(s.client, network.Router.ID).ExtractErr()
+	return err
```
sbueringer:
```suggestion
	return routers.Delete(s.client, network.Router.ID).ExtractErr()
```

pkg/cloud/services/networking/router.go
hidekazuna: @sbueringer Right, I will fix.

pkg/cloud/services/networking/router.go
```diff
@@ -165,6 +165,33 @@ INTERFACE_LOOP:
...
+	if network.Router == nil || network.Router.ID == "" {
+		s.logger.V(4).Info("No need to delete router since no router exists.")
+		return nil
+	}
+	exists, err := s.existsRouter(network.Router.ID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		s.logger.Info("No router", "router", network.Router.ID)
```
sbueringer:
```suggestion
		s.logger.Info("Skipping router deletion because router doesn't exist", "router", network.Router.ID)
```

PR #523, opened 2020-03-25
🐛  use sec groups from the current project, wait for lb from last reconcile

sbueringer:

This PR fixes to issues:
* When using an admin user it wasn't possible to assign security groups to machines because the security group list call returned sec groups from all projects, now it's filtered to the current one
* When reconciling load balancer there was an issue when the lb was still "pending create" from the last reconcile. Now the controller also waits even when the lb already exists from before.


sbueringer:
@jichenjc can you also take a look at this one?

sbueringer:
> @jichenjc can you also take a look at this one? (as I wanted to merge it first :))

*push* @jichenjc

jichenjc:
I didn't do deep check on this but generally looks good, let's merge this and see whether we need further enhancement :)

PR #530, opened 2020-03-27
✨ Add availability zone to be used by controller plane

### LINKED ISSUE #525: AZ usage of v1alpha3
/kind bug

my cloud actually have more than 1 az, and I noticed that current behavior is to create vm on random az , I checked cluster-api code and seems their az is try to find in a set of domains then pick fewest one

but openstack az is not such concept as far as I can tell... 
are we sure the cluster-api failuredomain concept is same to az in openstack? 
as cluster api has scale up and down ,I am really not convieneced it's suitable for openstack

and , where can I set failuredomain on control pane? I knew we can set on worker node but I am not that far 

```

func (r *KubeadmControlPlaneReconciler) failureDomainForScaleDown(cluster *clusterv1.Cluster, machines internal.FilterableMachineCollection) *string {
        // See if there are any Machines that are not in currently defined failure domains first.
        notInFailureDomains := machines.Filter(
                machinefilters.Not(machinefilters.InFailureDomains(cluster.Status.FailureDomains.FilterControlPlane().GetIDs()...)),
        )
        if len(notInFailureDomains) > 0 {
                // return the failure domain for the oldest Machine not in the current list of failure domains
                // this could be either nil (no failure domain defined) or a failure domain that is no longer defined
                // in the cluster status.
                return notInFailureDomains.Oldest().Spec.FailureDomain
        }

        // Otherwise pick the currently known failure domain with the most Machines
        return internal.PickMost(cluster.Status.FailureDomains.FilterControlPlane(), machines)
}

func (r *KubeadmControlPlaneReconciler) failureDomainForScaleUp(cluster *clusterv1.Cluster, machines internal.FilterableMachineCollection) *string {
        // Don't do anything if there are no failure domains defined on the cluster.
        if len(cluster.Status.FailureDomains.FilterControlPlane()) == 0 {
                return nil
        }
        return internal.PickFewest(cluster.Status.FailureDomains.FilterControlPlane(), machines)
}
```

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

jichenjc:

Fixes #525 

follow up will add doc for this 

**Release note**:

```release-note

```


api/v1alpha3/openstackcluster_types.go
```diff
@@ -115,6 +115,9 @@ type OpenStackClusterSpec struct {
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
 	ControlPlaneEndpoint clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
+
+	// AvailabilityZone is the az to deploy controller to
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
```
sbueringer: Why do we need an additional property? Why not just set the az on the OpenStackMachineTemplate of KubeadmControlPlane like for MachineDeployments? I think that would be much more consistent. The property is also already there you just have to use it

sbueringer: Maybe a compromise would be to make it an AZ array. So this means if you have e.g. 5 AZs and you only want to distribute the control plane to 3 of them you can choose which (of course in your case it would be just one)

jichenjc: I had some analysis at #525
I think the az are returned by openstack side and cluster-api side seems don't have spec for this
maybe you can point me where should I add? The machine is created by cluster-api side automatically with the param found and I don't know it will honor the OpenStackMachineTemplate settings we give...

I think the az will be a good suggestion, thanks

sbueringer: Oh sorry of course you're right. I removed the az property in favor of the failureDomain property in the CAPI Machine resource :) Ups

So yes you're solution is the only way to restrict the AZs. Hm maybe we should open an issue in the main ClusterAPI repo to ask if they want to provide a way to restrict the AZs via a property in the KubeAdmControplane, might be an oversight that this cannot be specified there.

jichenjc: let's ask there (I will do it soon) before we do this and see whether they have better suggestions :)

sbueringer: Okay perfect :). If they don't want to do anything there, I'm fine with solving it in CAPO.

jichenjc:
https://github.com/kubernetes-sigs/cluster-api/issues/2800 is the issue for further info

controllers/openstackcluster_controller.go
```diff
@@ -291,6 +291,21 @@ func (r *OpenStackClusterReconciler) reconcileNormal(ctx context.Context, log lo
 		if az.ZoneName == "internal" {
 			continue
 		}
+		// If Az given, then check whether it's in the allow list
+		// If no Az given, then by default put into allow list
+		if len(openStackCluster.Spec.AvailabilityZones) > 0 {
```
sbueringer: Looks good. I personally would have moved the code to a contains function (`contains(strs []string, str string)`)

sbueringer: EDIT: Looks like gocyclo agrees with me :)

jichenjc: let's hold this so that we can use .ControlPlane flag

sbueringer: Okay so referring what I said in the issue. Let's first move the contains stuff in a separate function. Depending on if the az can be found in the cluster az list you then can set the ControlPlane boolean in line 310 to true or false

jichenjc:
my env is broken again ... so have to resubmit PR
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/537

PR #541, opened 2020-04-03
Add additional info when error occur

jichenjc:
it took me long time to find when I define the wrong cloud different to clouds.yaml 
in make file... I got err like `'auth_url' input can't be empty` which is really confusing
add this will be much helpful ..
Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/services/provider/provider.go
+++ b/pkg/cloud/services/provider/provider.go
@@ -85,7 +85,7 @@ func newClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 
 	opts, err := clientconfig.AuthOptions(clientOpts)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("auth option failed for cloud %v: %v", cloud.Cloud, err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/provider/provider.go
```diff
@@ -85,7 +85,7 @@ func newClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 
 	opts, err := clientconfig.AuthOptions(clientOpts)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("auth option failed: %v for cloud: %v", err, cloud)
```
sbueringer, line 88: afaik it's best practise to append the error at the end
```suggestion
		return nil, nil, fmt.Errorf("auth option failed for cloud %s: %v", cloud, err)
```

sbueringer: I would also suggest just printing the cloud name as I think we would be leaking credentials in the log otherwise

jichenjc: done :)

PR #546, opened 2020-04-17
add CI_VERSION into makefile

jichenjc:
Add CI_VERSION into Make file

when using `make create-cluster`, I am suffering the controller run CI stuffs which is weird
turn out I need set CI_VERSION , guess that's easier to store in Makefile to turn off this
(still open by default)

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,6 +306,7 @@ OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR ?= "m1.medium"
 CLUSTER_NAME ?= "capi-quickstart"
 OPENSTACK_CLUSTER_TEMPLATE ?= "./templates/cluster-template-without-lb.yaml"
 KUBERNETES_VERSION ?= "v1.17.3"
+CI_VERSION ?= $(KUBERNETES_VERSION)
 CONTROL_PLANE_MACHINE_COUNT ?= "1"
 WORKER_MACHINE_COUNT ?= "3"
 LOAD_IMAGE=$(CONTROLLER_IMG)-$(ARCH):$(TAG)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -386,7 +387,8 @@ create-cluster: $(CLUSTERCTL) $(KUSTOMIZE) $(ENVSUBST) ## Create a development K
 	  sed "s|\$${OPENSTACK_CLOUD_PROVIDER_CONF_B64}|$(OPENSTACK_CLOUD_PROVIDER_CONF_B64)|" | \
 	  sed "s|\$${OPENSTACK_CLOUD_CACERT_B64}|$(OPENSTACK_CLOUD_CACERT_B64)|" | \
 	  sed "s|\$${KUBERNETES_VERSION}|$(KUBERNETES_VERSION)|" | \
-	  sed "s|\$${CLUSTER_NAME}|$(CLUSTER_NAME)|"  \
+	  sed "s|\$${CLUSTER_NAME}|$(CLUSTER_NAME)|"  | \
+	  sed "s|\$${CI_VERSION}|$(CI_VERSION)|" \
```
sbueringer, line 391: This change breaks our e2e tests as stuff like `${CI_VERSION//+/%2B}` is not properly replaced. I don't unterstand why you have to set the CI_VERSION locally?

sbueringer: I also wouldn't recommend create-cluster for usage outside of our e2e tests. As it always patches the node bootstrapping to use artifacts from Google Cloud storage (CI). Which is slower then just installing CAPO the regular way

sbueringer: I guess this PR fixes it because it probably breaks the node bootstrapping? (I guess the if in the e2e-conformance_patch.yaml.tpl is not executed anymore)

jichenjc: ok, I will manually patch my system
the reason is my bootstrap is blocked by some CI scripts locally
so I think it's worthy to avoid use it

and I think the logic is same , I still keep the CI_VERSION as its original value..

sbueringer: @jichenjc is this PR still active?

jichenjc:
nope, let's close this

PR #551, opened 2020-04-30
🐛 [OCCM] make sure to requeue in case of paused Cluster

gdetal:

After pivoting, I observed that my Cluster/OpenstackCluster object stayed forever in Provisioning state while they were Provisioned before the move.

After looking at the log of the OCCM, I found:

> controllers/OpenStackCluster "msg"="OpenStackCluster or linked Cluster is marked as paused. Won't reconcile"

And after, I found no further logs regarding the cluster.

If I am not mistaken, the OCCM only watches for OpenStackCluster objects and therefore never Reconcile based on Cluster objects changes which will happen when unpaused. For that we need to make sure that the controller Requeue in case of Cluster marked as paused.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -40,7 +40,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // OpenStackClusterReconciler reconciles a OpenStackCluster object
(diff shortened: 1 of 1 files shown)


sbueringer:
@gdetal I don't think we need to requeue if the cluster is paused as a requeue won't change anything as long as it's paused. If the pause is removed we will get a new update event.

I think the actual problem is that the cluster is never unpaused. Imho the problem should be in clusterctl as it is responsible for pausing/unpausing the cluster.


P.S. looks like other providers like CAPA do it the same way as CAPO: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/controllers/awsmachine_controller.go

gdetal:
@sbueringer thanks for the feedback. I've checked and can confirm that the Cluster is unpaused correctly but the occm does not continue. 

The AWS Cluster controller seems to watch for Cluster objects in https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/controllers/awscluster_controller.go#L213 which cause a requeue due to paused Cluster. 

Note that this is not done in the OpenStack Cluster Controller but is in the Machine Controller.
I'll update my patch to implement the same mechanism as done in AWS controller.

controllers/openstackcluster_controller.go
```diff
@@ -353,11 +357,60 @@ func (r *OpenStackClusterReconciler) reconcileNetworkComponents(log logr.Logger,
 }
 
 func (r *OpenStackClusterReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
-	return ctrl.NewControllerManagedBy(mgr).
+	controller, err := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options).
 		For(&infrav1.OpenStackCluster{}).
```
sbueringer, line 362: Hm what do you think, should we just use the same code as CAPA (with more log statements :))? (also in the machine controller)

controllers/openstackcluster_controller.go
```diff
@@ -353,11 +357,60 @@ func (r *OpenStackClusterReconciler) reconcileNetworkComponents(log logr.Logger,
...
 		WithEventFilter(pausePredicates).
-		Complete(r)
+		Build(r)
+
+	if err != nil {
+		return errors.Wrap(err, "error creating controller")
+	}
+
+	return controller.Watch(
```
sbueringer, line 370: Hm what do you think, should we just use the same code as CAPA (with more log statements :))? (also in the machine controller)

controllers/openstackcluster_controller.go
```diff
@@ -353,11 +357,60 @@ func (r *OpenStackClusterReconciler) reconcileNetworkComponents(log logr.Logger,
...
+		},
+		predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
+				newCluster := e.ObjectNew.(*clusterv1.Cluster)
+				return oldCluster.Spec.Paused && !newCluster.Spec.Paused
+			},
+			CreateFunc: func(e event.CreateEvent) bool {
+				cluster := e.Object.(*clusterv1.Cluster)
+				return !cluster.Spec.Paused
```
jichenjc, line 383: https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha2-to-v1alpha3.md seems recommend to use if util.IsPaused(cluster, <object>) { ?

controllers/openstackcluster_controller.go
jichenjc: +1 to reuse CAPA code for more logs as debug info

PR #555, opened 2020-05-25
🐛 Fix auto generate security groups

### LINKED ISSUE #553: Can not access applications via LoadBalancer
/kind bug

I deployed a web application using LoadBalancer type service on kubernetes cluster deployed by cluster-api-provider-openstack with Octavia. The application can not be accessed via loadbalancer.

The application can be accessed via loadbalancer.

The application could be acccessed by adding security group rule which accept NodePort services(port range 30000-32767) to worker nodes.
I think we should update security group rules according to [Check required ports](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#check-required-ports)

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

hidekazuna:

This PR updates generate security groups to follow kubeadm install documentation.

Fixes #553 

**Release note**:

```release-note
Fixed generate security groups to follow kubeadm install documentation for using NodePort services.

```

### DIFF
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -26,7 +26,7 @@ import (
 const (
 	secGroupPrefix     string = "k8s"
 	controlPlaneSuffix string = "controlplane"
-	globalSuffix       string = "all"
+	workerSuffix       string = "worker"
 	remoteGroupIDSelf  string = "self"
 )
 
@@ -56,14 +56,27 @@ func (s *Service) ReconcileSecurityGroups(clusterName string, openStackCluster *
 		s.logger.V(4).Info("No need to reconcile security groups", "cluster", clusterName)
 		return nil
 	}
-	desiredSecGroups := map[string]infrav1.SecurityGroup{
-		"controlplane": generateControlPlaneGroup(clusterName),
-		"global":       generateGlobalGroup(clusterName),
-	}
-	observedSecGroups := make(map[string]*infrav1.SecurityGroup)
 
+	secControlPlaneGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, controlPlaneSuffix)
+	secWorkerGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, workerSuffix)
+	secGroupNames := map[string]string{
+		controlPlaneSuffix: secControlPlaneGroupName,
+		workerSuffix:       secWorkerGroupName,
+	}
+	//create security groups first, because desired rules use group ids.
+	for _, v := range secGroupNames {
+		if err := s.createSecurityGroupIfNotExists(v); err != nil {
+			return err
+		}
+	}
+	// create desired security groups
+	desiredSecGroups, err := s.generateDesiredSecGroups(secGroupNames)
+	if err != nil {
+		return err
+	}
+
+	observedSecGroups := make(map[string]*infrav1.SecurityGroup)
 	for k, desiredSecGroup := range desiredSecGroups {
-		s.logger.Info("Reconciling security group", "name", desiredSecGroup.Name)
 
 		var err error
 		observedSecGroups[k], err = s.getSecurityGroupByName(desiredSecGroup.Name)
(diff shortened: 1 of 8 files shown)


pkg/cloud/services/networking/securitygroups.go
```diff
@@ -147,21 +153,35 @@ func generateControlPlaneGroup(clusterName string) infrav1.SecurityGroup {
 					RemoteIPPrefix: "0.0.0.0/0",
 				},
 				{
-					Direction:      "ingress",
-					EtherType:      "IPv4",
-					PortRangeMin:   22,
-					PortRangeMax:   22,
```
jichenjc, line 153: maybe I missed something ,does this means not able to ssh anymore?

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -147,21 +153,35 @@ func generateControlPlaneGroup(clusterName string) infrav1.SecurityGroup {
...
 			},
 			defaultRules...,
 		),
 	}
 }
 
-func generateGlobalGroup(clusterName string) infrav1.SecurityGroup {
-	secGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, globalSuffix)
+func (s *Service) generateWorkerGroup(clusterName string) infrav1.SecurityGroup {
+	secGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, workerSuffix)
```
jichenjc, line 178: nit: better us secWorkerGroupName as variable``

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -147,21 +153,35 @@ func generateControlPlaneGroup(clusterName string) infrav1.SecurityGroup {
...
+	secControlPlaneGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, controlPlaneSuffix)
+
+	secControlPlaneGroup, err := s.getSecurityGroupByName(secControlPlaneGroupName)
+	if err != nil || secControlPlaneGroup == nil || secControlPlaneGroup.ID == "" {
+		return infrav1.SecurityGroup{}
```
jichenjc, line 183: so we have to have control plane secure group in order to create network ?
is there some edge case e.g controller doesn't use it while worker use it?
I knew it's rare so just ask in case we know it

pkg/cloud/services/networking/securitygroups.go
hidekazuna: We can not use ssh access anymore. Because I really implemented [Check required ports](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#check-required-ports).

pkg/cloud/services/networking/securitygroups.go
hidekazuna: OK, I will fix.

pkg/cloud/services/networking/securitygroups.go
jichenjc: I am ok with it ,however, for debug, without ssh it will be really not easy
we at least need provide a way to enable ssh

or at worst case, create a doc guide user how to enable it by modifying the code...

pkg/cloud/services/networking/securitygroups.go
hidekazuna: If ManagedSecurityGroups is true, control plane security group and worker security group will be created by following  [Check required ports](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#check-required-ports). If we do not like these rules, set ManagedSecurityGroups to false and specify existing security groups.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: We can open ssh port by changing security groups dynamically after VM has been booted if some problems occurred( Actually I was doing so 😅 ).
I totally agree adding documentation.

hidekazuna: It is much easier to add existing security group to your yaml than modifying the code like the following.

```yaml
      securityGroups:
      - name: k8s-cluster-default-capi-quickstart-secgroup-controlplane
      - name: default
```
this default security group enables ssh port.

hidekazuna: @jichenjc Updated docs.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/555/files#diff-76e731333fb756df3bff5ddb3b731c46

pkg/cloud/services/networking/securitygroups.go
hidekazuna: fixed.

pkg/cloud/services/networking/securitygroups.go
jichenjc: thanks for the update

pkg/cloud/services/networking/securitygroups.go
jichenjc: thanks, the info works for me

prankul88:
@hidekazuna Thanks for the PR! Will take a look at it later today.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -56,14 +56,26 @@ func (s *Service) ReconcileSecurityGroups(clusterName string, openStackCluster *
 		s.logger.V(4).Info("No need to reconcile security groups", "cluster", clusterName)
 		return nil
 	}
+
+	secControlPlaneGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, controlPlaneSuffix)
+	var err error
+	err = s.createSecurityGroupIfNotExists(secControlPlaneGroupName)
+	if err != nil {
```
sbueringer, line 63:
```suggestion
	if err := s.createSecurityGroupIfNotExists(secControlPlaneGroupName); err != nil {
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -56,14 +56,26 @@ func (s *Service) ReconcileSecurityGroups(clusterName string, openStackCluster *
...
+		return nil
+	}
+	secWorkerGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, workerSuffix)
+	err = s.createSecurityGroupIfNotExists(secWorkerGroupName)
+	if err != nil {
```
sbueringer, line 68: same here

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -147,51 +153,57 @@ func generateControlPlaneGroup(clusterName string) infrav1.SecurityGroup {
...
 			defaultRules...,
 		),
 	}
 }
 
-func generateGlobalGroup(clusterName string) infrav1.SecurityGroup {
-	secGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, globalSuffix)
+func (s *Service) generateWorkerGroup(clusterName string) infrav1.SecurityGroup {
+	secWorkerGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, workerSuffix)
+	secControlPlaneGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, controlPlaneSuffix)
```
sbueringer, line 179: what is the dependency from the workersecgroup to the control plane sec group?

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -147,51 +153,57 @@ func generateControlPlaneGroup(clusterName string) infrav1.SecurityGroup {
...
+	}
 
 	// As above, hardcoded rules.
 
 	return infrav1.SecurityGroup{
-		Name: secGroupName,
+		Name: secWorkerGroupName,
 		Rules: append(
 			[]infrav1.SecurityGroupRule{
 				{
```
sbueringer, line 192: Looks to me like the workers cannot communicate with each other anymore? If I'm right, isn't that a problem? 
Also how does the communication between control plane and workers and back works now?

pkg/cloud/services/networking/securitygroups.go
hidekazuna: Thanks, will fix.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: The worker security group accept 10250/tcp from control plane security group. Please see line 206.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: I understand I am missing considering the cni plugin. Let's add the rule to allow workers to communicate each other.
In terms of the communication between controller and worker, controller accept 6443/tcp(kube-apiserver) from all including worker nodes, and worker accepts 10250/tcp(kubelet API) from controller. This should work fine.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: will fix.

pkg/cloud/services/networking/securitygroups.go
jichenjc: we might have more workers after the init cluster deploy, will this also considered?
maybe need some detail thought on what we expect the cluster network to be provisioned by default and guide user to add advanced by their own..

sbueringer: @hidekazuna worker<=> master sounds good. So you also add a rule that the worker nodes can communicate with each other, right?

pkg/cloud/services/networking/securitygroups.go
sbueringer: Seems a bit implicit to just return an empty security group if the control plane group doesn't exist. I don't get the logic behind that. What do we want to achieve with that?

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -56,14 +56,26 @@ func (s *Service) ReconcileSecurityGroups(clusterName string, openStackCluster *
...
 	desiredSecGroups := map[string]infrav1.SecurityGroup{
-		"controlplane": generateControlPlaneGroup(clusterName),
-		"global":       generateGlobalGroup(clusterName),
+		"controlplane": s.generateControlPlaneGroup(clusterName),
+		"worker":       s.generateWorkerGroup(clusterName),
 	}
 	observedSecGroups := make(map[string]*infrav1.SecurityGroup)
 
 	for k, desiredSecGroup := range desiredSecGroups {
-		s.logger.Info("Reconciling security group", "name", desiredSecGroup.Name)
```
prankul88, line 66: @hidekazuna Just curious. Won't it be better to keep this? Just to log which security group is being reconciled.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -229,7 +241,7 @@ func matchGroups(desired, observed infrav1.SecurityGroup) bool {
 
 // reconcileGroup reconciles an already existing observed group by essentially emptying out all the rules and
```
prankul88, line 242: Nit: Instead of reconcileGroup it must be `reconcileGroupRules`

pkg/cloud/services/networking/securitygroups.go
hidekazuna: @sbueringer Yes, I will also add the rule you mentioned.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: I will fix to return error if the control plane security group does not exist.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: fixed

pkg/cloud/services/networking/securitygroups.go
hidekazuna: fixed.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: @prankul88 fixed.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: @sbueringer Ther rule was added here:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/555/files#diff-981e50f3b13da070546823dd24e18b32R197-R211

pkg/cloud/services/networking/securitygroups.go
hidekazuna: @sbueringer I fixed this as below:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/555/files#diff-981e50f3b13da070546823dd24e18b32R182-R188

pkg/cloud/services/networking/securitygroups.go
prankul88: @hidekazuna Got it. Thanks

prankul88:
LGTM

hidekazuna:
@jichenjc @sbueringer Could you review again?

pkg/cloud/services/networking/securitygroups.go
hidekazuna: @sbueringer @jichenjc I studied calico and skeptical about adding the rule for worker communication now. Calico default networking requirement is bidirectional tcp 179 and IPinIP. I created new security group including the rule allowing bidirectional tcp 179 and IPinIP, and added the group to both controllers and workers. It is working fine.
We should leave the issue due to depending on the CNI implementation.

jichenjc: I am not fully understand the comments above
>>>I created new security group including the rule allowing bidirectional tcp 179 and IPinIP, and added the group to both controllers and workers. 

you mean after your change, the cluster still not work and you have to create the rules manually? as we are provide solution to `create a k8s cluster` so we should consider how to enhance it , if it's a generic requirements , we need talk to cluster-api folks?

hidekazuna: All nodes are ready status, but pod to pod communication is not good. I think it is not good before my change. Kubespray doc says  [in OpenStack you must allow ipip traffic in your security groups ](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/calico.md#cloud-providers-configuration).

hidekazuna: Calico is just an example for Cluster API. So we should leave the issue for users.

hidekazuna: @sbueringer @jichenjc I found [cluster-api-provider-aws implementation](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/ec2/securitygroups.go#L368-L488)! I will follow this implementaion 😃

jichenjc: seems that's for calico only :)

hidekazuna: now we understand cluster api is calico only :)

jichenjc: https://github.com/kubernetes-sigs/cluster-api/issues/3174 opened

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -56,14 +56,27 @@ func (s *Service) ReconcileSecurityGroups(clusterName string, openStackCluster *
 		s.logger.V(4).Info("No need to reconcile security groups", "cluster", clusterName)
 		return nil
 	}
+
+	secControlPlaneGroupName := fmt.Sprintf("%s-cluster-%s-secgroup-%s", secGroupPrefix, clusterName, controlPlaneSuffix)
```
sbueringer, line 60: this name is calculated here and in generateControlPlaneGroup and in generateWorkerGroup similar holds for the worker group name. Can we make sure we have only one place where we generate the sec group name and then hand it over?

pkg/cloud/services/networking/securitygroups.go
sbueringer: Can we already return the securityGroup in line 61 (call to createSecurityGroupIfNotExists) and then just hand it over to generateWorkerGroup so we don't need the additional lookup?

pkg/cloud/services/networking/securitygroups.go
sbueringer: Okay so if calico works now, it's fine for me to resolve this conversation.

hidekazuna: Calico is not working as of now. Pod to pod communication does not work. I will fix.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: I made generateControlPlaneSecuriyGroup and generateWorkerGroup into generateDesiredSecGroups, and build security group names only once.

pkg/cloud/services/networking/securitygroups.go
hidekazuna: Generating the sec group names is now only once at the beginning of ReconcileSecurityGroups.

sbueringer:
/approve

@hidekazuna  I assume it's tested in your environment and it works, I'm not using the feature :)

PR #560, opened 2020-06-11
Make DisablePortSecurity compatible with platforms not using the port…

### LINKED ISSUE #559: DisablePortSecurity implementation causes incompatitibily issues
/kind bug

I'm trying to provision Kubernetes Cluster on Top of our OpenStack platform, which does not have the port security extension enabled. However the provider fails to provision the network due to the missing extension.

```
E0611 08:10:18.281288       1 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile network: Bad request with: [POST https://<redacted>:9696/v2.0/networks], error message: {\"NeutronError\": {\"message\": \"Unrecognized attribute(s) 'port_security_enabled'\", \"type\": \"HTTPBadRequest\", \"detail\": \"\"}}"  "controller"="openstackcluster" "request"={"Namespace":"default","Name":"capi-quickstart"}
```

Successfully provisioning a working Kubernetes cluster running on our OpenStack platform.

We currently run Kubernetes clusters on top of our OpenStack platform using Calico IP in IP encapsulation without the port security extension. Forcing the use of an OpenStack extension that is not required for operation seems odd imho.
Simply commenting the code that forces the extension to be included results in a working cluster. This issue seems to fall in the same category as #543.

**Environment:**

- Cluster-api version: v0.3.6-90-7b9762100a1d48-dirty
- Minikube/KIND version: v0.9.0-alpha+b3e81f69654a36

qvicksilver:
… security extension

🐛 (:bug:, patch and bugfixes)

Improves compatibility with OpenStack without the port security extension
Fixes #559 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -58,15 +58,19 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	}
 
 	var portSecurityEnabled gophercloud.EnabledState
+	var opts createOpts
 	if openStackCluster.Spec.DisablePortSecurity {
 		portSecurityEnabled = gophercloud.Disabled
+		opts = createOpts{
+			AdminStateUp:        gophercloud.Enabled,
+			Name:                networkName,
+			PortSecurityEnabled: portSecurityEnabled,
+		}
 	} else {
-		portSecurityEnabled = gophercloud.Enabled
-	}
-	opts := createOpts{
-		AdminStateUp:        gophercloud.Enabled,
-		Name:                networkName,
-		PortSecurityEnabled: portSecurityEnabled,
+		opts = createOpts{
+			AdminStateUp: gophercloud.Enabled,
+			Name:         networkName,
+		}
 	}
 	network, err := networks.Create(s.client, opts).Extract()
 	if err != nil {
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -57,7 +57,6 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 		return nil
 	}
 
-	var portSecurityEnabled gophercloud.EnabledState
 	var opts createOpts
 	if openStackCluster.Spec.DisablePortSecurity {
 		opts = createOpts{
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -60,11 +60,10 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	var portSecurityEnabled gophercloud.EnabledState
 	var opts createOpts
 	if openStackCluster.Spec.DisablePortSecurity {
-		portSecurityEnabled = gophercloud.Disabled
 		opts = createOpts{
 			AdminStateUp:        gophercloud.Enabled,
 			Name:                networkName,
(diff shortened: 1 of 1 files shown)


qvicksilver:
That is my understanding at least, it is the default without the extension and the extension enables you to disable it.

pkg/cloud/services/networking/network.go
```diff
@@ -58,15 +58,19 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
...
 	} else {
-		portSecurityEnabled = gophercloud.Enabled
-	}
-	opts := createOpts{
-		AdminStateUp:        gophercloud.Enabled,
-		Name:                networkName,
-		PortSecurityEnabled: portSecurityEnabled,
+		opts = createOpts{
+			AdminStateUp: gophercloud.Enabled,
+			Name:         networkName,
```
jichenjc, line 72: if I understand correctly, the logic is if we should disable port security
then we should not provide this param to gophercloud and in turn not to opentsack neutron?

so we need 2 flag one is to indicate whether the port security will be used and the other is for whether the network itself need port security or not 

so better to rename the variable to `SecurityGroupEnabled`? so that we only have 2 options
if enbale security group, then mandatory to make the network enable security group and if not enable security group extension, then not put param here?

qvicksilver: The other way around, if we want to disable port security we pass that parameter to gophercloud and neutron. If we don't want to disable port security we don't pass the parameter.As far as I can tell from the documentation I can find networks are by default created with port security enabled.

https://wiki.openstack.org/wiki/Neutron/ML2PortSecurityExtensionDriver

sbueringer: Not sure what's going on without the extension. But when the extension is enabled the default is true: https://docs.openstack.org/api-ref/network/v2/ (search for `port_security_enabled`)

So imho the PR is fine. If we want to disable it disable it, if not, just don't set the attribute and the default value is used and port security is enabled

pkg/cloud/services/networking/network.go
```diff
@@ -58,15 +58,19 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 	}
 
 	var portSecurityEnabled gophercloud.EnabledState
+	var opts createOpts
 	if openStackCluster.Spec.DisablePortSecurity {
 		portSecurityEnabled = gophercloud.Disabled
```
sbueringer, line 63:
```suggestion
```

pkg/cloud/services/networking/network.go
```diff
@@ -58,15 +58,19 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
...
+		opts = createOpts{
+			AdminStateUp:        gophercloud.Enabled,
+			Name:                networkName,
+			PortSecurityEnabled: portSecurityEnabled,
```
sbueringer, line 67:
```suggestion
			PortSecurityEnabled: gophercloud.Disabled,
```

pkg/cloud/services/networking/network.go
sbueringer: @qvicksilver one nit otherwise /approve from me

qvicksilver: @sbueringer Looks sane to me!

jichenjc:
ok, works for me 
/lgtm

PR #561, opened 2020-06-16
Remove APIServerLoadBalancerAdditionalPorts

hidekazuna:

It is not good practice that ssh to controller nodes via API Server load balancer.
This PR removes APILoadBalancerAdditionalPorts.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -29,8 +29,6 @@ spec:
   managedAPIServerLoadBalancer: true
   apiServerLoadBalancerFloatingIP: ${OPENSTACK_CONTROLPLANE_IP}
   apiServerLoadBalancerPort: 6443
-  apiServerLoadBalancerAdditionalPorts:
-  - 22
   managedSecurityGroups: true
   nodeCidr: 10.6.0.0/24
   dnsNameservers:


sbueringer:
@hidekazuna This feature could not only be used for ssh but also for adding other ports to the Kubernetes API LoadBalancer if you e.g. host a Prometheus on the master nodes you want to expose. The ccm afaik doesn't handle the Kubernetes API LoadBalancer.

If we make it impossible to jump via the API LoadBalancer, what alternatives do we provide?

hidekazuna:
@sbueringer Ok, let's drop the default value from yaml, leaving users using this feature or not.

> If we make it impossible to jump via the API LoadBalancer, what alternatives do we provide?

We already closed the ssh port and wrote the solution.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/configuration.md#ssh-authorized-key

FYI: AWS provided the bastion node.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/accessing-instances.md#accessing-nodes-via-ssh

sbueringer:
> @sbueringer Ok, let's drop the default value from yaml, leaving users using this feature or not.
> 
> > If we make it impossible to jump via the API LoadBalancer, what alternatives do we provide?
> 
> We already closed the ssh port and wrote the solution.
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/configuration.md#ssh-authorized-key
> 
> FYI: AWS provided the bastion node.
> https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/accessing-instances.md#accessing-nodes-via-ssh

In my experience adding the ssh key is not enough. You also need network connectivity to the nodes. If the nodes don't have an external IP, you cannot access them from the outside. 

Imho the right solution is providing a bastion host, until we have that we can only use either the port or VNC console.

hidekazuna:
> > @sbueringer Ok, let's drop the default value from yaml, leaving users using this feature or not.
> > > If we make it impossible to jump via the API LoadBalancer, what alternatives do we provide?
> > 
> > 
> > We already closed the ssh port and wrote the solution.
> > https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/configuration.md#ssh-authorized-key
> > FYI: AWS provided the bastion node.
> > https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/accessing-instances.md#accessing-nodes-via-ssh
> 
> In my experience adding the ssh key and security groups is not enough. You also need network connectivity to the nodes. If the nodes don't have an external IP, you cannot access them from the outside.

Yes, adding a floating ip to the node is necessary. Or specify namespace on the network node which is technical.

> Imho the right solution is providing a bastion host, until we have that we can only use either the port or VNC console.

I agree, created issue #562

sbueringer:
@hidekazuna Might be a good idea to also document the additional steps (floating ip or network namespace). But that's independent of this PR

/approve

PR #563, opened 2020-06-17
Add generated security groups automatically

### LINKED ISSUE #556: Security Group Rules are not updated
/kind bug
Cluster API is not able to reach control plane end point of the openstack.

I have set up devstack on a VM running in public cloud.
I followed the quick start guide to initialize the and configure my management cluster running on kind.
I used cluster-template-without-lb.yaml to generate the workload manifest.
I had to manually add managedSecurityGroups attribute and set it to true. 
I enabled port security by setting disablePortSecurity: false
Cluster API creates 2 security groups. However there's is no rule to allow communication on port 6443.

In the controller manager logs, the error message appears as below
"name"="k8s-cluster-default-demo1cluster-secgroup-all"
I0608 18:14:36.939765       1 openstackcluster_controller.go:277] controllers/OpenStackCluster "msg"="No control plane node found yet, could not write OpenStackCluster.Spec.ControlPlaneEndpoint" "cluster"="demo1cluster" "namespace"="default" "openStackCluster"="demo1cluster"

I think this could be because the port is not open on the security groups.
The control plane machine is not attached to the control plane security group.

Be able to pass a flag in the clusterctl config command to enable port security and set managedsecuritygroups to true rather than editing the manifest manually.
Controlplane machine is attached to the security group.
Controlplane is reachable and controller is ready to communicate with the endpoint.

I have devstack installed in a VM (4 vCPUs and 15 GB RAM).
Cluster API and Kind are in same VM. 
I am using https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/configuration.md as a configuration guide.
I am using ubuntu-1804-kube-v1.16.2 qcow2 image as an image to bring up k8s servers
I have set the ssh_key as part of the environment variables. But I don't see the ssh_key attached to the controlplane instance in horizon dashboard.

**Environment:**

- Cluster-api version:
clusterctl version
clusterctl version: &version.Info{Major:"0", Minor:"3", GitVersion:"v0.3.2-dirty", GitCommit:"867b2afed0e6a4509a74bb123b7c6f2ab0f7ffd7", GitTreeState:"dirty", BuildDate:"2020-03-19T19:29:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
cratnoc@openstack:~$

- Minikube/KIND version:
kind --version
kind version 0.7.0

1.16.2

hidekazuna:
Currently generate security groups are not added to the instances.
This commit makes it automatically.

Currently generate security groups are not added to the instances if managedSecurityGroups: true.
This PR makes generate security groups to be added automatically.

Fixes #556 

**Release note**:

```release-note
Makes generate security groups to be added automatically if managedSecurityGroups: true
```

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -98,6 +98,14 @@ func (s *Service) InstanceCreate(clusterName string, machine *clusterv1.Machine,
 	if err != nil {
 		return nil, err
 	}
+	if openStackCluster.Spec.ManagedSecurityGroups {
+		if util.IsControlPlaneMachine(machine) {
+			securityGroups = append(securityGroups, openStackCluster.Status.ControlPlaneSecurityGroup.ID)
+		} else {
+			securityGroups = append(securityGroups, openStackCluster.Status.WorkerSecurityGroup.ID)
+		}
+	}
+
 	// Get all network UUIDs
 	var nets []ServerNetwork
 	if len(openStackMachine.Spec.Networks) > 0 {
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -121,8 +121,6 @@ spec:
       cloudsSecret:
         name: ${CLUSTER_NAME}-cloud-config
         namespace: ${NAMESPACE}
-      securityGroups:
-      - name: k8s-cluster-${NAMESPACE}-${CLUSTER_NAME}-secgroup-controlplane
 ---
 apiVersion: cluster.x-k8s.io/v1alpha3
 kind: MachineDeployment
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -98,6 +98,14 @@ func (s *Service) InstanceCreate(clusterName string, machine *clusterv1.Machine,
 	if err != nil {
 		return nil, err
 	}
+	if openStackCluster.Spec.ManagedSecurityGroups {
+		if util.IsControlPlaneMachine(machine) {
+			securityGroups = append(securityGroups, openStackCluster.Status.ControlPlaneSecurityGroup.ID)
+		} else {
+			securityGroups = append(securityGroups, openStackCluster.Status.WorkerSecurityGroup.ID)
```
jichenjc, line 105: so we are allowing both sec group from Machine spec and managed security groups created
maybe follow PR update doc will also talk about this to make it more clearer to end user

hidekazuna: Yes, we are. Still we can add more security groups if we need.

sbueringer: This PR also leads to a behavior where managesecuritygroups: true makes it impossible to not use the created sec groups. But I guess that's okay as all that managesecuritygroups does is to create these groups. Wouldn't make much sense to set the flag if they shouldn't be used.

PR #595, opened 2020-07-21
✨ Use external cloud provider

### LINKED ISSUE #489: Replace in-tree cloud-provider by cloud-provider-openstack
/kind feature

The current examples bring up a cluster with `cloud-provider: openstack`. This uses the kubernetes in-tree cloud-provider, which will be removed in the future.

We should be able to deploy a cluster with `cloud-provider: external` and deploy https://github.com/kubernetes/cloud-provider-openstack as an addon.

This might simplify stuff in the deployment, because there is no need to inject `/etc/kubernetes/cloud.conf` to VMs files any more.

In this step, we could also thing about moving to https://github.com/kubernetes/cloud-provider-openstack/tree/master/examples/cinder-csi-plugin for storage.

hidekazuna:

This PR provides template, tool, and document for using external cloud provider.

Fixes #489 

**Release note**:

```release-note
Use external cloud provider.
```

### DIFF
--- a/docs/external-cloud-provider.md
+++ b/docs/external-cloud-provider.md
@@ -1,0 +1,60 @@
+# External Cloud Provider
+
+To deploy a cluster using [external cloud provider](https://github.com/kubernetes-sigs/cloud-provider-openstack), create a cluster configuration with the [external cloud provider template](../../templates/cluster-template-external-cloud-provider.yaml).
+
+## Steps
+
+- After control plane is up and running, retrieve the workload cluster Kubeconfig:
+
+    ```shell
+    kubectl --namespace=default get secret/${CLUSTER_NAME}-kubeconfig -o jsonpath={.data.value} \
+    | base64 --decode \
+    > ./${CLUSTER_NAME}.kubeconfig
+    ```
+
+- Deploy a CNI solution
+
+    ```shell
+    curl https://docs.projectcalico.org/v3.12/manifests/calico.yaml | sed "s/veth_mtu:.*/veth_mtu: \"1430\"/g" | kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig apply -f -
+    ```
+
+- Create a secret containing the cloud configuration
+
+    ```shell
+    templates/create_cloud_conf.sh <path/to/clouds.yaml> <cloud> > /tmp/cloud.conf
+    ```
+
+    ```shell
+    kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig create secret -n kube-system generic cloud-config --from-file=/tmp/cloud.conf
+    ```
+
+    ```shell
+    rm /tmp/cloud.conf
+    ```
+
+- Create RBAC resources and openstack-cloud-controller-manager deamonset
+
+    ```shell
+    kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig apply -f https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/cluster/addons/rbac/cloud-controller-manager-roles.yaml
+    kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig apply -f https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/cluster/addons/rbac/cloud-controller-manager-role-bindings.yaml
+    kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig apply -f https://raw.githubusercontent.com/kubernetes/cloud-provider-openstack/master/manifests/controller-manager/openstack-cloud-controller-manager-ds.yaml
+    ```
+
+- Waiting for all the pods in kube-system namespace up and running
+
+    ```shell
+    $ kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig get pod -n kube-system
+    NAME                                                    READY   STATUS    RESTARTS   AGE
+    calico-kube-controllers-5569bdd565-ncrff                1/1     Running   0          20m
+    calico-node-g5qqq                                       1/1     Running   0          20m
+    calico-node-hdgxs                                       1/1     Running   0          20m
+    coredns-864fccfb95-8qgp2                                1/1     Running   0          109m
+    coredns-864fccfb95-b4zsf                                1/1     Running   0          109m
+    etcd-mycluster-control-plane-cp2zw                      1/1     Running   0          108m
+    kube-apiserver-mycluster-control-plane-cp2zw            1/1     Running   0          110m
+    kube-controller-manager-mycluster-control-plane-cp2zw   1/1     Running   0          109m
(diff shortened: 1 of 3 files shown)


templates/cluster-template-external-cloud-provider.yaml
```diff
@@ -0,0 +1,158 @@
...
+  apiServerLoadBalancerPort: 6443
+  managedSecurityGroups: true
+  nodeCidr: 10.6.0.0/24
+  dnsNameservers:
+  - ${OPENSTACK_DNS_NAMESERVERS}
+  externalNetworkId: ${OPENSTACK_EXTERNAL_NETWORK_ID}
+  disablePortSecurity: false
+  disableServerTags: true
+  useOctavia: true
+---
```
jichenjc, line 40: I believe the only difference is the kubeadm template for master and worker between internal/external?
so later on we may consider how to change it to avoid maintain too many templates

docs/external-cloud-provider.md
```diff
@@ -0,0 +1,60 @@
...
+    ```shell
+    kubectl --namespace=default get secret/${CLUSTER_NAME}-kubeconfig -o jsonpath={.data.value} \
+    | base64 --decode \
+    > ./${CLUSTER_NAME}.kubeconfig
+    ```
+
+- Deploy a CNI solution
+
+    ```shell
+    curl https://docs.projectcalico.org/v3.12/manifests/calico.yaml | sed "s/veth_mtu:.*/veth_mtu: \"1430\"/g" | kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig apply -f -
```
jichenjc, line 18: what's the 1430 for? internal cloud provider we don't have such requirements?

hidekazuna: MTU 1430 is [Calico MTU with IP-in-IP (IPv4)](https://docs.projectcalico.org/networking/mtu#determine-mtu-size). This size should be applied for internal cloud provider,too.

templates/cluster-template-external-cloud-provider.yaml
hidekazuna: Partially Yes. The difference between internal and external for creating cluster is only the template. But there is a huge difference after creating cluster. If we use external cloud provider, we have to create a secret and apply three yamls to create openstack-cloud-controller-manager.

docs/external-cloud-provider.md
jichenjc: yes, so this is not mentioned in internal cloud provider...

docs/external-cloud-provider.md
```diff
@@ -0,0 +1,60 @@
...
+    ```
+
+- Create a secret containing the cloud configuration
```
jichenjc, line 21: maybe you want to share some info about why we need perform following .e.g
 the node is Ready but not schedulable due to taint
`node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule`

templates/create_cloud_conf.sh
```diff
@@ -0,0 +1,110 @@
...
+fi
+
+if [[ "$CAPO_CACERT_ORIGINAL" != "" && "$CAPO_CACERT_ORIGINAL" != "null" ]]; then
+  echo "ca-file=\"${CAPO_CACERT_ORIGINAL}\"" >> ${CAPO_CLOUD_PROVIDER_CONF_TMP}
+fi
+if [[ "$CAPO_REGION" != "" && "$CAPO_REGION" != "null" ]]; then
+  echo "region=\"${CAPO_REGION}\"" >> ${CAPO_CLOUD_PROVIDER_CONF_TMP}
+fi
+
+cat ${CAPO_CLOUD_PROVIDER_CONF_TMP} > /tmp/cloud.conf
```
jichenjc, line 110: maybe you want to make `/tmp/cloud.conf` a configurable option..

templates/cluster-template-external-cloud-provider.yaml
jichenjc: oh... understand your point now,maybe we can keep this one and overtime the in tree provider will be removed
so how about put all those things into `outtree` folder and put original one into `intree` folder?

hidekazuna: As far as seeing docs like [Cloud Providers](https://github.com/kubernetes/kubernetes/tree/master/pkg/cloudprovider) and [Deprecation Notice](https://github.com/kubernetes/kubernetes/tree/master/pkg/cloudprovider), `external` is common these days.
I think `external` is more operator's point of view and better.

docs/external-cloud-provider.md
hidekazuna: I am sorry can not understand you want to point out.
I followed [cloud-provider-openstack doc](Deploy a Kubernetes cluster with openstack-cloud-controller-manager using kubeadm), I did not find `node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule` .

templates/create_cloud_conf.sh
hidekazuna: Should I update env.rc to add create `/tmp/cloud.conf` option?

jichenjc: yes, at least, accept a param instead of hard code /tmp/cloud.conf

templates/cluster-template-external-cloud-provider.yaml
jichenjc: so you mean by default to out of tree? I am ok to that and I think we need keep intree for a while (reason such as stable, someone might use it )

docs/external-cloud-provider.md
jichenjc: yes, I knew, what I mean is at this stage ,if you describe the node
the node is `Ready` state, but pods are pending due to this `taint`

instead of let people go to other doc, we might notify them that at this stage what happend and why the pods is pending etc (so people who as developer might know more background )

jichenjc:
I tested this locally and totally fine to me 
so after those nits are fixed ,we are good to go

templates/cluster-template-external-cloud-provider.yaml
hidekazuna: We do not change directory structure. This means default is intree because `cluster-template.yaml` is used by default.

jichenjc: both fine to me ..

templates/create_cloud_conf.sh
hidekazuna: updated like https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/595/files#diff-8058b1c0be9bf0e46ec0001c6e7c1308R24

docs/external-cloud-provider.md
hidekazuna: > yes, so this is not mentioned in internal cloud provider...

submitted #598 for internal cloud provider

PR #635, opened 2020-09-16
Refactor InstanceCreate

hidekazuna:

This PR is only refactoring, deletes network and subnet check logic from InstanceCreate.

Since openstackmachine controller checks whether infrastructure ready or not first, network and subnet are created before InstanceCreate.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/common/extensions"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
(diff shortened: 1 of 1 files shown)


hidekazuna:
recheck

hidekazuna:
recheck

hidekazuna:
recheck

pkg/cloud/services/compute/instance.go
```diff
@@ -102,30 +100,10 @@ func (s *Service) InstanceCreate(clusterName string, machine *clusterv1.Machine,
 		}
 	}
 
-	// Get all network UUIDs
-	var nets []ServerNetwork
-	if len(openStackMachine.Spec.Networks) > 0 {
-		var err error
```
jichenjc, line 108: I assume this change will ignore the OpenstackMachine Spec networks and only honor the cluster status network
then what's the reason to keep OpenstackMachine Spec networks?

hidekazuna: Aha, I understand openStackMachine.Spec.Networks is used the instance is connected to multi networks.  I wonder if there is really useful though.

jichenjc: @hidekazuna I think multiple network is valid ...

PR #663, opened 2020-11-02
adjust sequence of eventf for creation of resources

jichenjc:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -129,7 +129,6 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 		record.Warnf(openStackCluster, "FailedCreateNetwork", "Failed to create network %s: %v", networkName, err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


jichenjc:
recheck

pkg/cloud/services/networking/network.go
```diff
@@ -139,6 +138,7 @@ func (s *Service) ReconcileNetwork(clusterName string, openStackCluster *infrav1
 		}
 	}
 
+	record.Eventf(openStackCluster, "SuccessfulCreateNetwork", "Created network %s with id %s", networkName, network.ID)
```
hidekazuna, line 141: If adding tags fails, network is created but no event recorded.
Current event recording is better.

sbueringer: I agree. If it's really important that we have another event after the tagging we could consider producing two events. One after network creation and one after tagging. But I'm not sure what the use case is for this and thus what we can achieve with it.

sbueringer:
@jichenjc do we need this PR *and* https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/828?

jichenjc:
I think we can use that PR now

PR #673, opened 2020-11-13
Move log to creating place

jichenjc:

keep seeing 
```
I1113 00:24:14.043296       1 openstackmachine_controller.go:291] controllers/OpenStackMachine "msg"="Creating Machine" "cluster"="capi-openstack-3" "machine"="capi-openstack-control-plane-d2fnl" "namespace"="default" "openStackCluster"="capi-openstack-3" "openStackMachine"="capi-openstack-control-plane-rthx6"

```

but actually it's not creating .. it's getOrCreate, so the log is confusing
Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -287,8 +287,7 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-
-	logger.Info("Creating Machine")
+	logger.Info("Reconciling Machine create started")
 
 	clusterName := fmt.Sprintf("%s-%s", cluster.ObjectMeta.Namespace, cluster.Name)
 
@@ -307,7 +306,7 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 		return ctrl.Result{}, err
 	}
 
-	instance, err := r.getOrCreate(computeService, machine, openStackMachine, cluster, openStackCluster, userData)
+	instance, err := r.getOrCreate(computeService, machine, openStackMachine, cluster, openStackCluster, userData, logger)
 	if err != nil {
 		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("OpenStack instance cannot be created: %v", err))
 		return ctrl.Result{}, err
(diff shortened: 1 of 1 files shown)


controllers/openstackmachine_controller.go
```diff
@@ -288,8 +288,6 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 		return ctrl.Result{}, err
 	}
 
-	logger.Info("Creating Machine")
```
hidekazuna, line 291: Not good message is true. I suggest to update to 'Reconciling Machine create started'
in contrast to L370 logger.Info("Reconciled Machine create successfully").

hidekazuna:
It is best to see event. because SuccessfulCreateServer is just after calling createInstance.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L145

jichenjc:
> 
> 
> It is best to see event. because SuccessfulCreateServer is just after calling createInstance.
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L145

yes, but reading log makes me very confusing .. so better update log

hidekazuna:
> 
> 
> > It is best to see event. because SuccessfulCreateServer is just after calling createInstance.
> > https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L145
> 
> yes, but reading log makes me very confusing .. so better update log

Currently create or delete resource start message is not logged. You can refer to network resources creation. We need logging policy.

jichenjc:
> > > It is best to see event. because SuccessfulCreateServer is just after calling createInstance.
> > > https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L145
> > 
> > 
> > yes, but reading log makes me very confusing .. so better update log
> 
> Currently create or delete resource start message is not logged. You can refer to network resources creation. We need logging policy.

agree

so are you suggesting should I add in this PR for all those logs ?

hidekazuna:
> > Currently create or delete resource start message is not logged. You can refer to network resources creation. We need logging policy.
> 
> agree
> 
> so are you suggesting should I add in this PR for all those logs ?

Raised another issue.

jichenjc:
recheck

jichenjc:
@hidekazuna so what about this one? maybe merge first? or anything need change ? thanks

jichenjc:
recheck

PR #702, opened 2020-12-07
Add tip for docker limit

jichenjc:

Fixes #

**Release note**:

```release-note

```


hidekazuna:
This is not cluster-api-provider-openstack specific. I think propose to Cluster API is more valuable.

jichenjc:
um.. I understand, but if we can provide some info here , we avoid people to jumping around ..

sbueringer:
@jichenjc @hidekazuna I'm not sure if we should add it or not. On the one hand it's useful if somebody is hitting it on the other hand it's really pretty general advice which is independent of CAPI. I lean into the direction of not adding it as we should not start giving general troubleshooting advice (also this is all documentation we have to maintain and keep up-to-date)

jichenjc:
give another thought, I think we should not add it , it should be general stuff for k8s ecosystem

PR #703, opened 2020-12-07
📖 Add instructions to use cluster template with load-balancer

prankul88:
Currently the capi quick-start only refers to generating config for default template. This PR adds instructions on how to use `without-lb` cluster-template.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -34,6 +34,26 @@ This command actually uses [the template file](../templates/cluster-template.yam
 
 Note: You can use [the template file](../templates/cluster-template.yaml) by manually replacing values.
 
+**Note:** By default the command creates highly available control plane with internal OpenStack cloud provider. If you wish to create highly available control plane with external OpenStack cloud provider or single control plane without load balancer, use external-cloud-provider or without-lb [flavor](https://cluster-api.sigs.k8s.io/clusterctl/commands/config-cluster.html#flavors) respectively. For example,
+
+```bash
+# Using 'external-cloud-provider' flavor
+clusterctl config cluster capi-quickstart \
+  --flavor external-cloud-provider
+  --kubernetes-version v1.18.2 \
+  --control-plane-machine-count=3 \
+  --worker-machine-count=1 \
+  > capi-quickstart.yaml
+
+# Using 'without-lb' flavor
+clusterctl config cluster capi-quickstart \
+  --flavor without-lb
+  --kubernetes-version v1.18.2 \
+  --control-plane-machine-count=1 \
+  --worker-machine-count=1 \
+  > capi-quickstart.yaml
+```
+
 ## Operating system image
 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -34,6 +34,18 @@ This command actually uses [the template file](../templates/cluster-template.yam
 
 Note: You can use [the template file](../templates/cluster-template.yaml) by manually replacing values.
 
+Note: If you wish to use cluster template without load-balancer, use command
```
hidekazuna, line 37: I want to care external cloud provider as well.

Suggestion:
By default the command creates highly available control plane with internal OpenStack cloud provider. If you wish to create highly available control plane with external OpenStack cloud provider or single control plane without load balancer, use external-cloud-provider or without-lb [flavor](https://cluster-api.sigs.k8s.io/clusterctl/commands/config-cluster.html#flavors) respectively.

hidekazuna (changes requested):

docs/configuration.md
jichenjc: I think we should say following is a sample, not mandatory
they should adjust the param such as `18.2` and `3` below

hidekazuna: One thing tends to misconfigure is control plane count is not 1 for non ha template.

prankul88: @hidekazuna Great, Adding external provider template reference too.

prankul88: @jichenjc Sorry I didnt get you.
You would not want example to explicitly mention k8s version, count ...?

prankul88:
@hidekazuna Hi, PTAL. Would be happy to update if anything else is required.

prankul88:
recheck

docs/configuration.md
```diff
@@ -34,6 +34,26 @@ This command actually uses [the template file](../templates/cluster-template.yam
...
 Note: You can use [the template file](../templates/cluster-template.yaml) by manually replacing values.
 
+**Note:** By default the command creates highly available control plane with internal OpenStack cloud provider. If you wish to create highly available control plane with external OpenStack cloud provider or single control plane without load balancer, use external-cloud-provider or without-lb [flavor](https://cluster-api.sigs.k8s.io/clusterctl/commands/config-cluster.html#flavors) respectively.
+
+```bash
+# Using 'external-cloud-provider' flavor
+clusterctl config cluster capi-quickstart \
+  --flavor external-cloud-provider
+  --kubernetes-version v1.18.2 \
+  --control-plane-machine-count=1 \
```
hidekazuna, line 44: control-plane-machine-count should be 3.
because we say "create highly available control plane with external OpenStack cloud provider" as above, and actually user would expect so.

docs/configuration.md
jichenjc: no , I am only saying we need emphasis it's a `example` and they need change it according to their env

prankul88: @jichenjc Have mentioned the commands to be as examples in the new commit.

docs/configuration.md
prankul88: Updated

prankul88:
@hidekazuna @jichenjc PTAL. Have updated upon the comments. Please let me know if any other changes are required. Thanks

PR #706, opened 2020-12-08
check if compute AZ is available

EmilienM:

Before creating a machine in a given AZ, it'll check that the AZ
actually exists and is available.

None

Updating vendoring for gophercloud/utils + dependencies.

**Release note**:

```release-note
When a machine is created, the Compute Availability Zone will be checked before and an error
message will be displayed if the zone doesn't exist or is not available.
```


EmilienM:
wrong PR, I need to re-do it.

PR #714, opened 2020-12-15
:bug: Fix domain_name and domain_id usage in env.rc

### LINKED ISSUE #713: env.rc not reading domain_name and domain_id
/kind bug

When running _source ./env.rc /path/to/clouds.yaml <cloud>_, if the domain name in clouds.yaml's auth section is set in _domain_name_ instead of _user_domain_name_, it is ignored due to yq returning an empty string and not "null".

_env.rc_ should pick the domain name from _domain_name_ in case _user_domain_name_ is not provided.

The failure happens when starting the kubelet during provisioning with:

`failed to run Kubelet: could not init cloud provider "openstack": You must provide exactly one of DomainID or DomainName.
`
**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.3.11
- Cluster-API version: v0.3.11
- OpenStack version: Ussuri
- Minikube/KIND version: v0.9.0
- Kubernetes version (use `kubectl version`): v1.19.1
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.4 LTS

alexpilotti:
Currently, _domain_name_ and _domain_id_ are being ignored when getting the OpenStack credentials from clouds.yaml.

Fixes #713 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/templates/env.rc
+++ b/templates/env.rc
@@ -72,11 +72,11 @@ CAPO_DOMAIN_NAME=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yq r - clouds.${C
 CAPO_APPLICATION_CREDENTIAL_NAME=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yq r - clouds.${CAPO_CLOUD}.auth.application_credential_name)
 CAPO_APPLICATION_CREDENTIAL_ID=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yq r - clouds.${CAPO_CLOUD}.auth.application_credential_id)
 CAPO_APPLICATION_CREDENTIAL_SECRET=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yq r - clouds.${CAPO_CLOUD}.auth.application_credential_secret)
-if [[ "$CAPO_DOMAIN_NAME" = "null" ]]; then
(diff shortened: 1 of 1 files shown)


jichenjc:
>>>Currently, domain_name and domain_id are being ignored when getting the OpenStack credentials from clouds.yaml.

I think from code the proposal is `when the user domain is not defined` (which means it is "") then `domain` will be used ? just for understanding the real error case perspective..

alexpilotti:
> > > > Currently, domain_name and domain_id are being ignored when getting the OpenStack credentials from clouds.yaml.
> 
> I think from code the proposal is `when the user domain is not defined` (which means it is "") then `domain` will be used ? just for understanding the real error case perspective..

When _user_domain_name_ is not specified in _clouds.yaml_, env.rc looks for  _domain_name_, which at the moment fails, because _$CAPO_DOMAIN_NAME_ is "" and not "null".

Similarly, for _user_domain_id_, if not specified, it should then look for _domain_id_, where the same "" vs "null" comparison issue applies.

alexpilotti:
As an additional real world example, here's my clouds.yaml. Without this patch, provisioning fails because _domain_name_ gets ignored.

```
clouds:
  demo:
    auth:
      auth_url: http://x.y.z.w:5000/v3
      project_name: admin
      username: admin
      password: ...
      domain_name: admin_domain
    identity_api_version: 3
    volume_api_version: 2
    region_name: RegionOne
    verify: false
```

jichenjc:
ok, I  checked my test env and it's `user_domain_name` so I didn't have this problem ...thanks for the info

/approve

PR #736, opened 2021-02-08
Update sync logic

hidekazuna:

Add sync logic for OpenStackCluster and OpenStackMachine.

Fixes #

**Release note**:

```release-note

```


hidekazuna:
recheck

hidekazuna:
recheck

hidekazuna:
recheck

controllers/openstackcluster_controller.go
```diff
@@ -438,9 +438,109 @@ func (r *OpenStackClusterReconciler) reconcileNetworkComponents(log logr.Logger,
...
+			ToRequests: handler.ToRequestsFunc(r.requeueOpenStackClusterForUnpausedCluster),
+		},
+		predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
+				newCluster := e.ObjectNew.(*clusterv1.Cluster)
+				log := r.Log.WithValues("predicate", "updateEvent", "namespace", newCluster.Namespace, "cluster", newCluster.Name)
+				switch {
+				// return true if Cluster.Spec.Paused has changed from true to false
+				case oldCluster.Spec.Paused && !newCluster.Spec.Paused:
```
sbueringer, line 484: I don't get why this code can be correct. Looks like we're now only reconciling unpause events? Any other update is just ignored? Do you know how this can be correct?

sbueringer: Could it be that we avoid hitting this code via `requeueOpenStackClusterForUnpausedCluster`?

sbueringer (commented):
Only one question, apart from that /approve 
Sorry for the late review

controllers/openstackcluster_controller.go
hidekazuna: Other update is still done by `WithEventFilter(pausedPredicates(r.Log))`  in L444.

hidekazuna: I am afraid I do not fully understand your second question, but Watch signature is the following so that this code is good to me.
```
Watch(src source.Source, eventhandler handler.EventHandler, predicates ...predicate.Predicate) error
```

hidekazuna:
> Only one question, apart from that /approve
> Sorry for the late review

Do not worry. We are not in a hurry 😄

controllers/openstackcluster_controller.go
sbueringer: Ah so this means the watch in l.472 is an additional watch? So we reconcile both on the stuff which matches the filters in l.444 & l.445 and independently on the events that match our filteres in l.477?

I though all those filters are applied to the same event stream

hidekazuna: Yes, I understood as you said.

sbueringer: Okay perfect :)

sbueringer:
@hidekazuna For me it would be okay to merge, so I can also give you a lgtm. Wdyt?

hidekazuna:
> @hidekazuna For me it would be okay to merge, so I can also give you a lgtm. Wdyt?

It is good to me.

jichenjc:
I will verify this (the long time move action issue) then we can consider about 0.3.4 capi

PR #743, opened 2021-02-16
🐛 fix rbac aggregation manager role

seanschneeweiss:
The manager aggregation patch for kustomization does not match the correct ClusterRole. This fix ensures to use the correct ClusterRole definition, instead of overwriting it with the broken patch.

We've seen the following errors:
```sh
[manager] E0211 09:08:55.893564       8 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.9/tools/cache/reflector.go:105: Failed to list *v1alpha3.OpenStackCluster: openstackclusters.infrastructure.cluster.x-k8s.io is forbidden: User "system:serviceaccount:capo-system:default" cannot list resource "openstackclusters" in API group "infrastructure.cluster.x-k8s.io" at the cluster scope: RBAC: clusterrole.rbac.authorization.k8s.io "capi-system-capi-aggregated-manager-role" not found
[manager] E0211 09:08:55.893677       8 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.9/tools/cache/reflector.go:105: Failed to list *v1alpha3.OpenStackMachine: openstackmachines.infrastructure.cluster.x-k8s.io is forbidden: User "system:serviceaccount:capo-system:default" cannot list resource "openstackmachines" in API group "infrastructure.cluster.x-k8s.io" at the cluster scope: RBAC: clusterrole.rbac.authorization.k8s.io "capi-system-capi-aggregated-manager-role" not found
```

Fixes #

Removing the `config/default/manager_role_aggregation_patch.yaml` patch should fix this problem.

**Release note**:

```release-note
NONE
```
<sub>Sean Schneeweiss <sean.schneeweiss@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>


sbueringer:
recheck

sbueringer:
@hidekazuna @jichenjc 
Some additional info:
* we compared the CAPO implementation with the one in CAPA:
  * the general pattern seems to be to just define the role and use it inside of CAPA/CAPO
  * our current master implementation has a ClusterRolebinding which depends on a ClusterRole of the ClusterAPI main deployment, which in turn refers back to ours
  * but because this ClusterRoleBinding has the same name as our own it is usually overwritten.
  * Afaik this somehow still works when using clusterctl but not when deploying the regular YAML files
  * CAPA still has the kustomize patch file we're deleting here, but they are not using it (and thus it's not contained in the release yamls)

Slightly off-topic:
* we're slowly starting to use CAPO in production right now, so just fyi, there will be a lot more PRs and contributions from my team in the coming weeks/ months :)

PR #755, opened 2021-02-23
🐛 use APIServerFloatingIP instead of ControlPlaneEndpoint.Host for LB

### LINKED ISSUE #754: CAPO trying to create floating IP with FQDN
/kind bug

CAPO is trying to create floating IP with value of  `openStackCluster.Spec.ControlPlaneEndpoint.Host`, this is wrong when `Host` is a DNS name not an IP address

CAPO trying to create floating IP from `openStackCluster.Spec.APIServerFloatingIP`

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `7c08f2442ee9ede21a74bd7f3c3ee1081b721fc5`
- Cluster-API version: - OpenStack version: `not important`
- Minikube/KIND version: `not important`
- Kubernetes version (use `kubectl version`): `not important`
- OS (e.g. from `/etc/os-release`): `not important`

jsen-:

Fixes #754

I'm not sure this is the correct way to solve this, but fixed the problem for us

**Release note**:
```release-note
You can use a FQDN in `openStackCluster.Spec.ControlPlaneEndpoint.Host`
```

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -71,7 +71,7 @@ func (s *Service) ReconcileLoadBalancer(clusterName string, openStackCluster *in
 		}
 	}
 
-	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster.Spec.ControlPlaneEndpoint.Host)
+	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster.Spec.APIServerFloatingIP)
 	if err != nil {
 		return err
 	}
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -71,7 +71,11 @@ func (s *Service) ReconcileLoadBalancer(clusterName string, openStackCluster *in
 		}
 	}
 
-	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster.Spec.APIServerFloatingIP)
+	fip := openStackCluster.Spec.ControlPlaneEndpoint.Host
+	if openStackCluster.Spec.APIServerFloatingIP != "" {
+		fip = openStackCluster.Spec.APIServerFloatingIP
+	}
+	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, fip)
 	if err != nil {
 		return err
 	}


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -71,7 +71,7 @@ func (s *Service) ReconcileLoadBalancer(clusterName string, openStackCluster *in
 		}
 	}
 
-	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster.Spec.ControlPlaneEndpoint.Host)
+	fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster.Spec.APIServerFloatingIP)
```
jichenjc, line 74: commit 4d3d98c1f77393d776c0278ec97dc877dfb775b6
Author: Hidekazu Nakamura <hidekazuna@gmail.com>
Date:   Fri Dec 11 07:32:54 2020 +0000

    Refactor loadbalancer package

    This commit removes floating IP related code from loadbalancer package.


@hidekazuna can you help to check this proposed change?

jichenjc:
recheck

pkg/cloud/services/loadbalancer/loadbalancer.go
hidekazuna: We need to consider the case `openStackCluster.Spec.APIServerFloatingIP` is not set.

Current flow:
If `openStackCluster.Spec.APIServerFloatingIP` is not set, create floating IP and set `openStackCluster.Spec.ControlPlaneEndpoint.Host`. After that we create load balancer using `openStackCluster.Spec.ControlPlaneEndpoint.Host`.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/controllers/openstackcluster_controller.go#L407-L435

hidekazuna: Is it time to introduce openStackCluster.Status.APIServerFloatingIP?

sbueringer (approved):

sbueringer:
/retitle 🐛 use APIServerFloatingIP instead of ControlPlaneEndpoint.Host for LB

pkg/cloud/services/loadbalancer/loadbalancer.go
jsen-: I added a new commit with fallback to original behavior.

> introduce openStackCluster.Status.APIServerFloatingIP

I'm not sure what you mean, I'm not super familiar with the codebase, so a little hesitant with doing bigger changes. However, if guided, I should be able to do a proper update.

sbueringer:
recheck

@jsen- just fyi our OpenLab based CI at the moment is very flaky :/. One of the next things I'll implement will be a stable e2e test based on Prow and a devstack on a GCP VM

sbueringer:
recheck

pkg/cloud/services/loadbalancer/loadbalancer.go
sbueringer: @jsen- I think you're current version is a fine solution for now. We can always adjust the inner workings in a follow-up PR

@hidekazuna I would probably want to take a look how other providers are handling this. Although I think we shouldn't change a lot here. There is a loadbalancer provider upcoming with v1alpha4 which imho we should implement. This will change the code quite a bit (although I assume the behaviour should stay the same)

sbueringer: For reference:
* current google doc version of the PR: https://docs.google.com/document/d/1wJrtd3hgVrUnZsdHDXQLXmZE3cbXVB5KChqmNusBGpE/edit
* but it was mentioned today in Slack that it will be converted to a PR in the cluster-api repo asap (If I got it correctly there's a deadline on Monday for v1alpha4 changes)

hidekazuna: @sbueringer Other providers does not support non lb case. We will re-visit this issue after v1alpha4 implementation merged.

hidekazuna:
recheck

jichenjc:
recheck

hidekazuna:
/lgtm

As temporary solution.

pkg/cloud/services/loadbalancer/loadbalancer.go
sbueringer: @hidekazuna Yup we will have to figure that out with v1alpha4 & the lb provider. But should be possibel as the lb provider itself is optional. So we should be able to have `cluster-tempate` with the lb provider and a `cluster-template-without-lb` without the lb provider

PR #756, opened 2021-02-24
:sparkles: AccessSubnetUUID: we can specify source subnet for access IP address

jsen-:
Currently, when we want to use multiple networks, the IP address is assigned basically randomly, this PR adds the ability to specify from which subnet the IP address will be obtained.

none

This is the minimal amount of changes I found to achieve the goal of getting IP address from the correct CIDR.

**Release note**:
```release-note
Added the ability to specify source subnet of compute instance access IP address
```

### DIFF
--- a/api/v1alpha3/openstackmachine_types.go
+++ b/api/v1alpha3/openstackmachine_types.go
@@ -57,6 +57,10 @@ type OpenStackMachineSpec struct {
 	// A networks object. Required parameter when there are multiple networks defined for the tenant.
 	// When you do not specify the networks parameter, the server attaches to the only network created for the current tenant.
 	Networks []NetworkParam `json:"networks,omitempty"`
+
+	// IP address of a port from this subnet will be marked as AccessIPv4 on the created compute instance
+	AccessSubnetUUID string `json:"accessSubnetUuid,omitempty"`
+
 	// The floatingIP which will be associated to the machine, only used for master.
 	// The floatingIP should have been created and haven't been associated.
 	FloatingIP string `json:"floatingIP,omitempty"`
--- a/api/v1alpha3/openstackmachine_types.go
+++ b/api/v1alpha3/openstackmachine_types.go
@@ -58,8 +58,8 @@ type OpenStackMachineSpec struct {
 	// When you do not specify the networks parameter, the server attaches to the only network created for the current tenant.
 	Networks []NetworkParam `json:"networks,omitempty"`
 
-	// IP address of a port from this subnet will be marked as AccessIPv4 on the created compute instance
-	AccessSubnetUUID string `json:"accessSubnetUuid,omitempty"`
+	// UUID, IP address of a port from this subnet will be marked as AccessIPv4 on the created compute instance
+	Subnet string `json:"subnet,omitempty"`
 
 	// The floatingIP which will be associated to the machine, only used for master.
 	// The floatingIP should have been created and haven't been associated.
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -132,6 +132,7 @@ type Instance struct {
 	FailureDomain  string            `json:"failureDomain,omitempty"`
 	SecurityGroups *[]string         `json:"securigyGroups,omitempty"`
 	Networks       *[]Network        `json:"networks,omitempty"`
+	AccessSubnet   string            `json:"accessSubnet,omitempty"`
 	Tags           []string          `json:"tags,omitempty"`
 	Image          string            `json:"image,omitempty"`
 	Flavor         string            `json:"flavor,omitempty"`
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -132,7 +132,7 @@ type Instance struct {
 	FailureDomain  string            `json:"failureDomain,omitempty"`
 	SecurityGroups *[]string         `json:"securigyGroups,omitempty"`
 	Networks       *[]Network        `json:"networks,omitempty"`
-	AccessSubnet   string            `json:"accessSubnet,omitempty"`
+	Subnet         string            `json:"subnet,omitempty"`
 	Tags           []string          `json:"tags,omitempty"`
 	Image          string            `json:"image,omitempty"`
 	Flavor         string            `json:"flavor,omitempty"`
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -229,6 +229,10 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
 			}
 		}
 	}
+	if i.AccessSubnet != "" && accessIPv4 == "" {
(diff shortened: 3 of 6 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -182,6 +184,12 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
 			port = portList[0]
 		}
 
+		for _, fip := range port.FixedIPs {
```
sbueringer, line 187: Just to confirm some assumptions:
* if AccessSubnet is set we will get a hit here and have a value for `accessIPv4`
* otherwise accessIPv4 and we have no change to before this PR

sbueringer (approved):

pkg/cloud/services/compute/instance.go
jsen-: yes, exactly

sbueringer: Okay, thx :)

api/v1alpha3/openstackmachine_types.go
```diff
@@ -57,6 +57,10 @@ type OpenStackMachineSpec struct {
 	// A networks object. Required parameter when there are multiple networks defined for the tenant.
 	// When you do not specify the networks parameter, the server attaches to the only network created for the current tenant.
 	Networks []NetworkParam `json:"networks,omitempty"`
+
+	// IP address of a port from this subnet will be marked as AccessIPv4 on the created compute instance
```
hidekazuna, line 61: I wonder what happens if networks and subnet which is not part of networks are given.
If we do not accept the case comment it here. Or add the check logic.

api/v1alpha3/openstackmachine_types.go
```diff
@@ -57,6 +57,10 @@ type OpenStackMachineSpec struct {
...
+	AccessSubnetUUID string `json:"accessSubnetUuid,omitempty"`
```
hidekazuna, line 62: Other variables does not start with Access. `Subnet` is easy to understand.

api/v1alpha3/openstackmachine_types.go
jichenjc: I guess below discussion covered this? maybe need some comments update?

    if AccessSubnet is set we will get a hit here and have a value for accessIPv4
    otherwise accessIPv4 and we have no change to before this PR

hidekazuna: Ok, got it.

api/v1alpha3/openstackmachine_types.go
jsen-: renamed AccessSubnetUUID to Subnet

api/v1alpha3/openstackmachine_types.go
jsen-: I added f5ff1bf7272b5a7e34eac665dd7e8e5182d4c86a.
The instance creation will fail if no ports with fixed IP are found in the specified subnet. This also covers the case when this subnet does not exist or is not in `.Networks`.
I didn't notice the ...
> Ok, got it.

... until now, so please let me know if you'd like to keep this change. I can drop this commit if not needed.

api/v1alpha3/types.go
```diff
@@ -132,6 +132,7 @@ type Instance struct {
 	FailureDomain  string            `json:"failureDomain,omitempty"`
 	SecurityGroups *[]string         `json:"securigyGroups,omitempty"`
 	Networks       *[]Network        `json:"networks,omitempty"`
+	AccessSubnet   string            `json:"accessSubnet,omitempty"`
```
sbueringer, line 135: hm does it make sense to call the property in instance differently than the one in the spec?

sbueringer: Hm maybe I missed something / or read it incorrectly. Could it be that this adds the Subnet property to the OpenStackMachine and the AccessSubnet to the Instance struct which seems to be only used for the status of the Bastion host?

jsen-: `Instance` struct seems to be used as a status of bastion and as a holder of state of a created VM, but at the same time, it is used as an input for `createInstance`. I don't know why this is so, but I wanted to minimize the impact of this PR.
There are basically two possibilities how to forward the subnet UUID into the `createInstance` function:
 - add this property to this `Instance` struct (what I did)
 - add another (optional) parameter to `createInstance` function
The second option seemed inappropriate, taking into consideration the current signature:
```go
func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error)
```
Adding `accessSubnetUUID string` does not feel right, but now that I'm thinking about it, the advantage would be that it would not change the public contract in bastion host status.

Just let me know, I can add another commit and forward the subnet UUID in an additional argument.

It also might make sense to keep it as it is and add the same possibility to specify subnet UUID for bastion host as well (adding similar `Subnet` property to `Bastion` struct https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/437e82faec1f9b77493e8e2bc387ea704c9d2373/api/v1alpha3/types.go#L251-L266)
Then having this property in Bastion status starts making sense.

api/v1alpha3/openstackmachine_types.go
hidekazuna: @jsen- Thanks for adding checking logic which is better to be added. Keep the commit, please.

api/v1alpha3/openstackmachine_types.go
hidekazuna: @jsen- Thanks updating.

api/v1alpha3/types.go
hidekazuna, line 135: `AccessSubnet` is also better to be renamed to `Subnet`.

pkg/cloud/services/compute/instance.go
```diff
@@ -221,6 +229,10 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
 			}
 		}
 	}
+	if i.AccessSubnet != "" && accessIPv4 == "" {
+		return nil, fmt.Errorf("no ports with fixed IPs found on AccessSubnet \"%s\"", i.AccessSubnet)
```
hidekazuna, line 233: If AccessSubnet is renamed to Subnet, This comment would be updated.

api/v1alpha3/types.go
jsen-: done

pkg/cloud/services/compute/instance.go
jsen-: :heavy_check_mark:

hidekazuna:
Looks good to me now.

sbueringer:
@jichenjc @hidekazuna I would have preferred to be able to at least take a quick look at the final implementation before merge.
It's not a big problem in this case, but what do you think about defaulting to /hold on PRs to avoid instant merge on-lgtm? https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/758

PR #758, opened 2021-02-26
🏃 Update pull request template to use /hold per default

sbueringer:
Adjust the pull request template to use /hold per default. Goal is just to avoid accidental instant merges on-/hold. We can always deliberately remove the hold if we really want to merge instantly :).

/hold


jichenjc:
/lgtm

I am ok for this

hidekazuna:
Ok to me.

/hold cancel

PR #765, opened 2021-03-03
WIP:no CI for irrevelant files

jichenjc:

Fixes #

**Release note**:

```release-note

```

/hold

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -2,7 +2,15 @@
     name: kubernetes-sigs/cluster-api-provider-openstack
     check:
       jobs:
-        - cluster-api-provider-openstack-current-acceptance-test-v1.18
+        - cluster-api-provider-openstack-current-acceptance-test-v1.18:
+            irrelevant-files:
+              - ^README.md$
+              - ^docs/.*$
+              - ^.*\.md$
+              - ^OWNERS$
+              - ^SECURITY_CONTACTS$
+              - ^.gitignore$
+              - ^.*\.sh$
 
 - job:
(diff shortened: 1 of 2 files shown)


.zuul.yaml
```diff
@@ -2,7 +2,14 @@
     name: kubernetes-sigs/cluster-api-provider-openstack
     check:
       jobs:
-        - cluster-api-provider-openstack-current-acceptance-test-v1.18
+        - cluster-api-provider-openstack-current-acceptance-test-v1.18:
```
sbueringer, line 5: I would really like to replace OpenLab with Prow. 
As soon as I have [this PR](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/759) green and we have the v1alpha4 and this PR merged there is no need to keep OpenLab:

jichenjc: it's different purpose, I want to avoid run CI for non-related test cases..

sbueringer: I know, but I will delete this file in my PR because this file is for OpenLab, it doesn't affect our Prow builds.

jichenjc: ok, I got it , so I can wait for your PR ready then work on that part

sbueringer: Yeah you basically have to do that change in the Prow test-infra repo (where I added the new jobs).
Problem is that Prow does not allow something like this as far as I know. You can say "only run this job when those files change" but you cannot say "don't run this job when only those files change).

sbueringer:
As the .zuul.yaml has been deleted with OpenLab I'll close this PR
/close

PR #773, opened 2021-03-08
✨ pkg/cloud/services/networking/securitygroups.go reimplement reconcilation

### LINKED ISSUE #771: Improve security group reconcilation
/kind feature

Currently the security groups get reconciled by deleting all existing rules and creating all wanted rules if only a single rule does not match (see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/networking/securitygroups.go#L111).

If in future the hardcoded security groups get changed and capo reconciles it will delete and recreate all security group rules which may result in connection drops.

It is possible to create two lists:
* `rulesToCreate` (desired - observed)
* `rulesToDelete` (observed - desired)

And reconcile using them by
* deleting the rules in `rulesToDelete`
* creating the rules in `rulesToCreate`
* return `rulesToCreate + (rules which are in desired and observed)` 

-

<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH, 
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Imprint</a>
</sup>

chrischdi:
Reimplements the security group reconcilation by only creating and/or
deleting the rules desired / not desired instead of deleting all and creating all rules.

Fixes #771 

**Release note**:

```release-note

```

/hold

### DIFF
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -453,7 +453,7 @@ func (s *Service) reconcileGroupRules(desired, observed infrav1.SecurityGroup) (
 		}
 	}
 
-	s.logger.V(6).Info("Creating new rules needed for group", "name", observed.Name, "amount", len(rulesToCreate))
+	s.logger.V(4).Info("Creating new rules needed for group", "name", observed.Name, "amount", len(rulesToCreate))
 	for _, rule := range rulesToCreate {
 		r := rule
 		r.SecurityGroupID = observed.ID
@@ -468,7 +468,7 @@ func (s *Service) reconcileGroupRules(desired, observed infrav1.SecurityGroup) (
 	}
 	observed.Rules = reconciledRules
 
-	s.logger.V(6).Info("Deleting rules not needed anymore for group", "name", observed.Name, "amount", len(rulesToDelete))
+	s.logger.V(4).Info("Deleting rules not needed anymore for group", "name", observed.Name, "amount", len(rulesToDelete))
 	for _, rule := range rulesToDelete {
 		s.logger.V(6).Info("Deleting rule", "ruleID", rule.ID, "groupName", observed.Name)
 		err := rules.Delete(s.client, rule.ID).ExtractErr()
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -448,6 +448,15 @@ func (s *Service) reconcileGroupRules(desired, observed infrav1.SecurityGroup) (
 		}
 	}
 
+	s.logger.V(4).Info("Deleting rules not needed anymore for group", "name", observed.Name, "amount", len(rulesToDelete))
+	for _, rule := range rulesToDelete {
+		s.logger.V(6).Info("Deleting rule", "ruleID", rule.ID, "groupName", observed.Name)
+		err := rules.Delete(s.client, rule.ID).ExtractErr()
+		if err != nil {
+			return infrav1.SecurityGroup{}, err
+		}
+	}
+
 	s.logger.V(4).Info("Creating new rules needed for group", "name", observed.Name, "amount", len(rulesToCreate))
 	for _, rule := range rulesToCreate {
 		r := rule
@@ -462,15 +471,6 @@ func (s *Service) reconcileGroupRules(desired, observed infrav1.SecurityGroup) (
 		reconciledRules = append(reconciledRules, newRule)
 	}
 	observed.Rules = reconciledRules
-
-	s.logger.V(4).Info("Deleting rules not needed anymore for group", "name", observed.Name, "amount", len(rulesToDelete))
-	for _, rule := range rulesToDelete {
-		s.logger.V(6).Info("Deleting rule", "ruleID", rule.ID, "groupName", observed.Name)
-		err := rules.Delete(s.client, rule.ID).ExtractErr()
-		if err != nil {
-			return infrav1.SecurityGroup{}, err
-		}
-	}
 
 	return observed, nil
 }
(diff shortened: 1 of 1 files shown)


chrischdi:
I had to remove the `RemoteIPPrefix: "0.0.0.0/0",` which at the end has the same result as setting nothing: allow all.

Otherwise the reconcilation is getting a `409` because it would try to create another security group rule which has the same spec.

jichenjc:
recheck

chrischdi:
> recheck

I'll debug it in a local devstack setup. From our environment I can see that the environment does not save the description.

chrischdi:
I was able to fix the issue. By default there are already security group rules inside the security groups which have another description.

Because of that I changed the order to first delete not needed rules and create the new ones afterwards

jichenjc:
reasonable change just curious about following statement, why are we creating an exactly same rule if we know 
it's same, why we will create it? 

```
I had to remove the RemoteIPPrefix: "0.0.0.0/0", which at the end has the same result as setting nothing: allow all.

Otherwise the reconcilation is getting a 409 because it would try to create another security group rule which has the same spec.
```

sbueringer:
/retitle ✨ pkg/cloud/services/networking/securitygroups.go reimplement reconcilation

sbueringer:
> reasonable change just curious about following statement, why are we creating an exactly same rule if we know
> it's same, why we will create it?
> 
> ```
> I had to remove the RemoteIPPrefix: "0.0.0.0/0", which at the end has the same result as setting nothing: allow all.
> 
> Otherwise the reconcilation is getting a 409 because it would try to create another security group rule which has the same spec.
> ```

I think the problem is that the deep equal does not detect that it's the same because the server will return something else compared to what we created before.

sbueringer:
@chrischdi please rebase so that the tests are run against the new v1alpha4 master version

chrischdi:
> > reasonable change just curious about following statement, why are we creating an exactly same rule if we know
> > it's same, why we will create it?
> > ```
> > I had to remove the RemoteIPPrefix: "0.0.0.0/0", which at the end has the same result as setting nothing: allow all.
> > 
> > Otherwise the reconcilation is getting a 409 because it would try to create another security group rule which has the same spec.
> > ```
> 
> I think the problem is that the deep equal does not detect that it's the same because the server will return something else compared to what we created before.

Exactly: on our OpenStack (which is "VMware Integrated OpenStack") the following happens:
* security groups get reconciled
  * rules get created
* security groups get reconciled again later on
  * rules previously created having `RemoteIPPrefix: "0.0.0.0/0"` will get deleted and created again, because the API returns `nil` instead of `0.0.0.0/0`.

I can see the same result when using the openstack cli and checking the output of `openstack security group rule list --debug <secgroup ID>`
  
However: I think the response of the API is still okay, because `nil` will default to `0.0.0.0/0` in the backend according the cli docs:
```
$ openstack help security group rule create | grep -e ' --remote-ip' -A 3
  --remote-ip <ip-address>
                        Remote IP address block (may use CIDR notation;
                        default for IPv4 rule: 0.0.0.0/0, default for IPv6
                        rule: ::/0)
```

It is also ok for me to revert the removal of `0.0.0.0/0` on this PR if you like me to do so :-)

chrischdi:
> @chrischdi please rebase so that the tests are run against the new v1alpha4 master version

rebased to master :-)

chrischdi:
recheck

chrischdi:
Looks like we're hitting the same issues as in #779 , maybe I should wait for #759

chrischdi:
rebased to master again to have gcp based tests

sbueringer (approved):
conformance test green now :)

jichenjc:
recheck

jichenjc:
> 
> 
> > > reasonable change just curious about following statement, why are we creating an exactly same rule if we know
> > > it's same, why we will create it?
> > > ```
> > > I had to remove the RemoteIPPrefix: "0.0.0.0/0", which at the end has the same result as setting nothing: allow all.
> > > 
> > > Otherwise the reconcilation is getting a 409 because it would try to create another security group rule which has the same spec.
> > > ```
> > 
> > 
> > I think the problem is that the deep equal does not detect that it's the same because the server will return something else compared to what we created before.
> 
> Exactly: on our OpenStack (which is "VMware Integrated OpenStack") the following happens:
> 
>     * security groups get reconciled
>       
>       * rules get created
> 
>     * security groups get reconciled again later on
>       
>       * rules previously created having `RemoteIPPrefix: "0.0.0.0/0"` will get deleted and created again, because the API returns `nil` instead of `0.0.0.0/0`.
> 
> 
> I can see the same result when using the openstack cli and checking the output of `openstack security group rule list --debug <secgroup ID>`
> 
> However: I think the response of the API is still okay, because `nil` will default to `0.0.0.0/0` in the backend according the cli docs:
> 
> ```
> $ openstack help security group rule create | grep -e ' --remote-ip' -A 3
>   --remote-ip <ip-address>
>                         Remote IP address block (may use CIDR notation;
>                         default for IPv4 rule: 0.0.0.0/0, default for IPv6
>                         rule: ::/0)
> ```
> 
> It is also ok for me to revert the removal of `0.0.0.0/0` on this PR if you like me to do so :-)

no need, just want to understand the logic here, all good :_)

jichenjc:
wait for zuul job to merge

jichenjc:
@sbueringer looks like we don't want to run openstack/check anymore as the playbook is missing
so we should go ahead merge this and overwrite the failure..

sbueringer:
@jichenjc Yes we have to ignore the OpenLab failures for now. I'll try to remove the OpenLab ASAP after we have the ProwJob on release-0.3

PR #778, opened 2021-03-10
✨ create ports

### LINKED ISSUE #788: add the possibility to create custom ports
/kind feature

Some of the workloads in our clusters require network acceleration (SR-IOV). In order to support this requirement, we'd like to be able to create and attach custom ports to worker nodes.

We tried attaching the ports to compute instances after they were created, unfortunately this does not work for some reason, the VM needs to be created with the ports already attached.

We were also playing with a possibility to pre-create the ports and only let `capo` attach them during VM creation, but this does not fit the the ClusterAPI design - we only have access to `OpenstackMachineTemplate` and one port can be only attached to one instance.

It seems most natural to let `capo` also create the ports based on a [user-defined template](https://github.com/gophercloud/gophercloud/blob/bcd179f2c98de61525a99bf9310d698fd7786a17/openstack/networking/v2/ports/requests.go#L112-L125).
The most important settings for us is actually [vnic_type](https://github.com/gophercloud/gophercloud/blob/bcd179f2c98de61525a99bf9310d698fd7786a17/openstack/networking/v2/extensions/portsbinding/requests.go#L18) which needs to be set to direct.

jsen-:
We need to assign additional ports to the created instances with `vnic_type=direct`.
Attempting to assign this port to an existing VM will fail, so it needs to be done during the VM creation.

Fixes #788

This is a draft to start a conversation, maybe the goal can be achieved in a better way.

**Release note**:
```release-note
Added the ability to create and assign additional ports
```

/hold


jichenjc:
@jsen- I think I understand most of the code and purpose of this PR
can we create an issue to track the reason/purpose of the issue so that we can know more background ? Thanks

pkg/cloud/services/compute/instance.go
```diff
@@ -233,6 +235,16 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
 		return nil, fmt.Errorf("no ports with fixed IPs found on Subnet \"%s\"", i.Subnet)
 	}
 
+	for _, portCreateOpts := range i.Ports {
+		port, err := getOrCreatePort(is, i.Name+"-"+portCreateOpts.NameSuffix, portCreateOpts)
+		if err != nil {
+			return nil, err
```
jichenjc, line 241: should we delete those ports when delete cluster? I assume those ports are independent to VM lifecylce?

sbueringer:
> @jsen- I think I understand most of the code and purpose of this PR
> can we create an issue to track the reason/purpose of the issue so that we can know more background ? Thanks

Yup, I would also like to understand why we need additional ports.

fyi. The v1alpha4 PR has been merged onto master. So if this feature is targeted towards master/v1alpha4 it has to 
be implemented against the v1alpha4 CRDs now.

jsen-:
Sorry for the delay, I'm actually out-of-office, but will try to be responsive.

> can we create an issue to track the reason/purpose of the issue so that we can know more background ? Thanks
Done (#788)

> should we delete those ports when delete cluster? I assume those ports are independent to VM lifecylce?
Definitely, I will try to implement cleanup of the ports and update the PR

> So if this feature is targeted towards master/v1alpha4 it has to be implemented against the v1alpha4 CRDs now.
I'm assuming, rebase should be enough, am I correct?

sbueringer:
@jsen- More or less. You have to rebase onto master and make your changes in the api/v1alpha4 folder instead of api/v1alpha3 but the structs are almost the same so there shouldn't be a lot to adjust

api/v1alpha4/openstackmachine_types.go
```diff
@@ -61,6 +61,9 @@ type OpenStackMachineSpec struct {
 	// UUID, IP address of a port from this subnet will be marked as AccessIPv4 on the created compute instance
 	Subnet string `json:"subnet,omitempty"`
 
+	// create and assign additional ports to instances
+	Ports []PortOpts `json:"ports,omitempty"`
```
hidekazuna, line 65: We only need VNICType actually. I try to find a light solution.
In my understanding, SR-IOV setting is done per compute node, but real use case is that create network for SR-IOV and create port for it. How about adding VNICType to NetworkParam? At least we can use existing network with SR-IOV.



we can add only VNICType. Because only VNICType is needed for SR-IOV.

In my understanding, OpenStack can change SR-IOV setting per compute node, but Cluster API can change the setting per MachineDeployment in availability zone.




Since we need only VNICType for SR-IOV, it seems

pkg/cloud/services/compute/instance.go
sbueringer: I implemented a cleanup orphaned port logic in one of my PRs before the sec group deletion. But it would be good if we try to delete all ports during machine deletion which we created for this machine before.

iamemilio:
Hi! We are actually working on something similar for RedHat OpenShift. Here is our work based off of a fork of the v1aplha1 API: https://github.com/openshift/cluster-api-provider-openstack/pull/170. The only difference in functionality between these two is that your patch does not account for allowing portSecurity to be set. In our platform, this would be a blocker. Other than that, where our implementation diverges the most  is that we chose to automate the creation of the additional ports in the network and subnet param, but I do like your changes to allow for additional ports to be added explicitly.

api/v1alpha4/openstackmachine_types.go
iamemilio: +1 Since Network param and Subnet Param are no longer nested in v1aplha4 api, it would also make sense to be able to set the vnicType in the subnet, in case that is a user's preferred way to set an interface.

jichenjc:
> 
> I am hoping that we could align our work on this feature, since we are hoping to convert to v1alpha4, or potentially v1beta at some point in the future.

+1 to see this happen , thanks @iamemilio

api/v1alpha4/openstackmachine_types.go
sbueringer: I'm basically fine with whatever option you get consensus on :)
But just asking. We didn't change NetworkParam and SubnetParam from v1alpha4 and as far as I can see they are still nested. Did I miss something that they are not nested anymore ? :)

sbueringer:
> > I am hoping that we could align our work on this feature, since we are hoping to convert to v1alpha4, or potentially v1beta at some point in the future.
> 
> +1 to see this happen , thanks @iamemilio

+1 for me too. I think this should be possible

api/v1alpha4/openstackmachine_types.go
iamemilio: I could be misunderstanding, but: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e1a13286736fd151106ddadd730a250713efde71/api/v1alpha4/openstackmachine_types.go#L57-L62

It looks like they are their own isolated api entry. Its very possible that I am incorrect though :)

api/v1alpha4/types.go
```diff
@@ -116,6 +120,28 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type PortOpts struct {
+	NetworkID           string              `json:"networkId" required:"true"`
```
adduarte, line 124: is there any value to adding a SubnetID field? 
I would seem overkill, but there always seems to be a case where it would be nice to easily lookup the subnetID based on port.

iamemilio: Also, if your goal is to support SR-IOV, then you will need to be able to set portSecurity on the ports.

iamemilio:
:space_invader: Our team discussed things, and due to our timeline, we need to merge the patch in our fork for now so that we can unblock our quality assurance team. However, we are still going to collaborate with you on this feature and will also be filing a bug against our platform to meet feature and api parity with the results of this pull request. :smiley:

api/v1alpha4/openstackmachine_types.go
sbueringer: Ah now I get what lead to this assumption. This Subnet is only for the AccessIPv4 "selection". (it was introduced in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/756)

SubnetParam is still nested.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e1a13286736fd151106ddadd730a250713efde71/api/v1alpha4/types.go#L61-L70

As mentioned elsewhere we have to rethink our design / CRD sooner or later as it's not that thought out / consistent.

iamemilio: ah ok, that makes sense now. thanks for explaining!

api/v1alpha4/types.go
```diff
@@ -116,6 +120,28 @@ type SubnetFilter struct {
...
+	FixedIPs            []ports.IP          `json:"fixedIPs,omitempty"`
+	DeviceID            string              `json:"deviceID,omitempty"`
+	DeviceOwner         string              `json:"deviceOwner,omitempty"`
+	TenantID            string              `json:"tenantID,omitempty"`
+	ProjectID           string              `json:"projectID,omitempty"`
+	SecurityGroups      *[]string           `json:"securityGroups,omitempty"`
+	AllowedAddressPairs []ports.AddressPair `json:"allowedAddressPairs,omitempty"`
+
+	// The ID of the host where the port is allocated
+	HostID string `json:"binding:host_id,omitempty"`
```
iamemilio, line 138:
```suggestion
	HostID string `json:"binding:hostID,omitempty"`
```
I would prefer to stick to camel case

api/v1alpha4/types.go
```diff
@@ -116,6 +120,28 @@ type SubnetFilter struct {
...
+
+	// The virtual network interface card (vNIC) type that is bound to the
+	// neutron port.
+	VNICType string `json:"binding:vnic_type,omitempty"`
```
iamemilio, line 142: ^

api/v1alpha4/types.go
sbueringer, line 124:
```suggestion
	NetworkID           string              `json:"networkID" required:"true"`
```

pkg/cloud/services/compute/instance.go
```diff
@@ -486,6 +497,62 @@ func isDuplicate(list []string, name string) bool {
 	return false
 }
 
+func getOrCreatePort(is *Service, name string, portOpts infrav1.PortOpts) (*ports.Port, error) {
+	existingPorts, err := listPorts(is, ports.ListOpts{
```
iamemilio, line 501: IMO these infra re-use logic blocks are problematic:
- what happens if the port you find with the same name does not meet the other requirements specified by the user?
- what happens if a port with this name exists in 2 different networks/subnets?
- what happens if another service takes this port before we call instance create?

Feel free to ignore this, since I realize that this behavior is consistent across the api. I may file a bug about this to cover all the parts of the API that do this.

sbueringer:
@jsen- in case you don't have time to address the review findings, just let us know :)

pkg/cloud/services/compute/instance.go
```diff
@@ -486,6 +497,62 @@ func isDuplicate(list []string, name string) bool {
...
+	}
+	if len(existingPorts) == 0 {
+		createOpts := ports.CreateOpts{
+			Name:                name,
+			NetworkID:           portOpts.NetworkID,
+			Description:         portOpts.Description,
+			AdminStateUp:        portOpts.AdminStateUp,
+			MACAddress:          portOpts.MACAddress,
+			DeviceID:            portOpts.DeviceID,
+			DeviceOwner:         portOpts.DeviceOwner,
```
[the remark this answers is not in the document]
jsen-: I included all available port options to widen the range of usefulness of this PR
now I removed both properties as it doesn't probably make sense to include them

pkg/cloud/services/compute/instance.go
jsen-: > this behavior is consistent across the api
this is where I was coming from and just matching based on `name` is "good enough"™ for us
also didn't want to complicate the PR too much

jsen-:
@sbueringer I'm back, sorry for the delay :) trying to address all the points above
Currently I'm implementing the port deletion logic

Can somebody help with the failing tests? There are several occurrences of `undeclared name: Convert_v1alpha4_Instance_To_v1alpha3_Instance` and I have no idea what this is or how to fix it

sbueringer:
> @sbueringer I'm back, sorry for the delay :) trying to address all the points above
> ~Currently I'm implementing the port deletion logic~ added in [002e4e4](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/002e4e4ef63ab7a7c7324d22b40341e030ec3ba3)
> 
> Can somebody help with the failing tests? There are several occurrences of `undeclared name: Convert_v1alpha4_Instance_To_v1alpha3_Instance` and I have no idea what this is or how to fix it

No worries, just wanted to ask :)

The issue is that we're generation conversion code. Conversion code is used so that v1alpha4 resources can be converted to v1alpha3 and vice-versa by our conversion webhook.

Most of the conversion code is just generated. In case the generator doesn't know how to convert some structs, the conversion logic has to be implemented by us.

The conversion logic for v1alpha3<=>v1alpha4 is generated into `./api/v1alpha3`. Our custom implemenations are in `./api/v1alpha3/conversion.go`, e.g.:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/api/v1alpha3/conversion.go#L112-L114

tl;dr you have to add the funcs which are missing in `./api/v1alpha3/conversion.go` and implement the conversion. In some cases thos funcs are basically empty because there are new structs/fields and then there's nothing to do. The signature of the func can be infered by the way it's called in `./api/v1alpha3/zz_generated.conversion.go`

pkg/cloud/services/compute/instance.go
iamemilio: Don't worry about it, I will open an issue for it

api/v1alpha4/types.go
```diff
@@ -116,6 +120,26 @@ type SubnetFilter struct {
...
+	AdminStateUp        *bool               `json:"adminStateUp,omitempty"`
+	MACAddress          string              `json:"macAddress,omitempty"`
+	FixedIPs            []ports.IP          `json:"fixedIPs,omitempty"`
+	TenantID            string              `json:"tenantID,omitempty"`
+	ProjectID           string              `json:"projectID,omitempty"`
+	SecurityGroups      *[]string           `json:"securityGroups,omitempty"`
+	AllowedAddressPairs []ports.AddressPair `json:"allowedAddressPairs,omitempty"`
+
+	// The ID of the host where the port is allocated
+	HostID string `json:"binding:hostID,omitempty"`
```
iamemilio, line 136:
```suggestion
	HostID string `json:"hostID,omitempty"`
```

api/v1alpha4/types.go
```diff
@@ -116,6 +120,26 @@ type SubnetFilter struct {
...
+
+	// The virtual network interface card (vNIC) type that is bound to the
+	// neutron port.
+	VNICType string `json:"binding:vnicType,omitempty"`
```
iamemilio, line 140:
```suggestion
	VNICType string `json:"vnicType,omitempty"`
```

api/v1alpha4/types.go
```diff
@@ -116,6 +120,26 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type PortOpts struct {
+	NetworkID           string              `json:"networkID" required:"true"`
+	NameSuffix          string              `json:"nameSuffix" required:"true"`
+	Description         string              `json:"description,omitempty"`
+	AdminStateUp        *bool               `json:"adminStateUp,omitempty"`
+	MACAddress          string              `json:"macAddress,omitempty"`
+	FixedIPs            []ports.IP          `json:"fixedIPs,omitempty"`
```
iamemilio, line 129: also the values for ports.IP are all camel case, we should probably create our own version and pass it in

sbueringer: Agree

api/v1alpha4/types.go
```diff
@@ -116,6 +120,26 @@ type SubnetFilter struct {
...
+	TenantID            string              `json:"tenantID,omitempty"`
+	ProjectID           string              `json:"projectID,omitempty"`
+	SecurityGroups      *[]string           `json:"securityGroups,omitempty"`
+	AllowedAddressPairs []ports.AddressPair `json:"allowedAddressPairs,omitempty"`
```
iamemilio, line 133: ^ same as Fixed IPs :)

Sorry I keep hitting you with these so late. I missed these when reviewing and am bumping into them now that I am testing.

api/v1alpha4/types.go
```diff
@@ -116,6 +120,26 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type PortOpts struct {
+	NetworkID           string              `json:"networkID" required:"true"`
```
macaptain, line 124: Since `NetworkID` is a mandatory field, would it make sense if `ports:` were children of the list items of `networks:`?

sbueringer:
@jsen- Please let me know if you are able to address this comments. I think we have a few other folks which would be happy to pick this up and geht it merged :)

jsen-:
@sbueringer apologies for the delays, we're still interested to get this into master, but I will be busy with other stuff for couple more week :( so don't mind if someone is willing to pick this up.

sbueringer:
> @sbueringer apologies for the delays, we're still interested to get this into master, but I will be busy with other stuff for couple more weeks :( so don't mind if someone is willing to pick this up.

Okay, no problem. Thx for the information. @macaptain as you expressed interest (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/788#issuecomment-839823511), do you want to pick this up?

macaptain:
> > @sbueringer apologies for the delays, we're still interested to get this into master, but I will be busy with other stuff for couple more weeks :( so don't mind if someone is willing to pick this up.
> 
> Okay, no problem. Thx for the information. @macaptain as you expressed interest ([#788 (comment)](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/788#issuecomment-839823511)), do you want to pick this up?

Hey @sbueringer and @jsen- . Thanks for letting me know. Yes, I'd be happy to pick this up and try and get this feature into master.

sbueringer:
@macaptain Great! It probably makes sense to wait a bit until: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/862 is merged as it refactors related code quite a bit. But this shouldn't take too long

sbueringer:
Can be closed as it has been implemented in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/876
/close

(correct me if I'm wrong :))

PR #779, opened 2021-03-11
✨ Update klog dependency to v2

stmcginnis:

The klog v1 dependency used is quite old now, and the k/k repo has
switched to more recent v2 versions for quite some time now.

This updates klog usage to use the v2 package version.

This is part of an overall effort along with updating the k-sigs/cluster-api repo here:

https://github.com/kubernetes-sigs/cluster-api/pull/4284

**Release note**:

```release-note
NONE
```

/hold


sbueringer:
recheck

sbueringer:
/approve cancel
Might have been a bit to optimistic :)

@stmcginnis  Can you rebase onto current master. I'm not sure why but Openlab seems to build with a pretty old configuration (Kubernetes 1.18 test with go 1.13)
OpenLab fails with an error I also had on another PR before upgrading to 1.16 in OpenLab.

stmcginnis:
This is current with master, so not sure what's up with OpenLab. Has that been stable lately? It looks like a config issue there.

Something definitely is not right though. The .zuul.yaml file locally doesn't even list `cluster-api-provider-openstack-current-acceptance-test-v1.18`. Though I suppose zuul could be picking that up from elsewhere.

sbueringer:
Okay. Yeah pretty strange. I have a pr open since today to drop openlab/zuul in favor of prow/gcp/devstack

stmcginnis:
I think that's a very good plan. I used to be involved in OpenLab. It hasn't taken off like originally planned. ;)

sbueringer:
Yeah unfortunately. It is stable but only when we hit a cluster with the right CPUs. That's why we run 3 tests at the same time since my pr today and fail instantly when we detect the wrong CPUs.

stmcginnis:
There should be some way to configure things on the OpenLab side to actually make sure the right kinds of nodes are scheduled. But unfortunately I can't recall how all that is done.

sbueringer:
> There should be some way to configure things on the OpenLab side to actually make sure the right kinds of nodes are scheduled. But unfortunately I can't recall how all that is done.

I have the right nodes (sizes?) but was trying to get it configured that we avoid certain clusters/cpus but without success.  But as we were using devstack there anyway (because I couldnt get a real octavia) gcp should be the better choice.

jichenjc:
recheck

hidekazuna:
@stmcginnis I wonder why this PR does not have go.sum. Did you run make modules?

stmcginnis:
> Did you run make modules?

Yep, I ran it initially on the first commit. I added the second commit when I realized I originally ran it with go 1.15, but there were additional changes once I ran it with 1.16.

hidekazuna:
> > Did you run make modules?
> 
> Yep, I ran it initially on the first commit. I added the second commit when I realized I originally ran it with go 1.15, but there were additional changes once I ran it with 1.16.

Thanks, understand.

jichenjc:
recheck

sbueringer:
@jichenjc please let's wait a bit until either: #780 or #759 is merged and we have green tests here. I think the problem here is OpenLab and not the change, but I really would like to not merge with a failed conformance test. 

/hold

jichenjc:
@sbueringer thanks for the reminder, I agree we need wait for green test first

prankul88:
> Yep, I ran it initially on the first commit. I added the second commit when I realized I originally ran it with go 1.15, but there were additional changes once I ran it with 1.16.

@stmcginnis I tried updating to klog v2 in my local with go1.16.2 and ran `make modules`, which brought changes to go.sum file as well. Wanted to understand this.

stmcginnis:
@prankul88 - that's odd - and a little concerning. I'm running go 1.16.1, but I would hope there wouldn't be any changes due to that.

I ran `make modules` again to see if maybe there were changes in the dependencies themselves that would cause those differences, but it comes back clean for me. So it does indeed seem to be the difference in bugfix version of go itself.

stmcginnis:
I upgraded locally to go 1.16.2 and ran `make modules` again, but it comes back clean for me. So not really sure what's going on there.

prankul88:
@stmcginnis Not sure either. Will double check again in my system.

sbueringer:
@stmcginnis looks like Prow is happy, at least :)

/approve

sbueringer:
Let's merge it. The code already had enough reviews and now the build is also green.
/lgtm

PR #780, opened 2021-03-11
🏃 Bump zuul jobs to k8s 1.20

stmcginnis:

1.20 has been out for some time now, good to raise our tested version on
OpenLab.

**Release note**:

```release-note
NONE
```

/hold


stmcginnis:
recheck

sbueringer:
Fine for me when tests are green. We will drop openlab and test against 1.20 aftet this pr https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/759

sbueringer:
@stmcginnis  I think you also have to change the version in the e2e-conformance.sh.

stmcginnis:
Ah, OK. I wasn't sure if we would want that updated too. Will get that updated to the latest for a default now.

stmcginnis:
recheck

sbueringer:
@stmcginnis I think the test failed here: https://logs.openlabtesting.org/logs/80/780/88d50b56cb14a054f8bb694b1134cc988c44ed37/check/cluster-api-provider-openstack-current-acceptance-test-v1.20-1/08b0f29/
because we're not using the right kubectl version in the tests and the 1.20 conformance test detects that. I think that's something I fixed here when migrating the tests to 1.20: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/759/files 
I'm not entirely sure what it was but I think we're using the kubectl from `ensure-kubectl` and not the one we built ourselves. 

But I"m hopeful we get the final reviews for PR #759 soon so we don't need #780 and can unblock your klog PR.

stmcginnis:
Don't think this is necessary now with the switch to prow.

sbueringer:
@stmcginnis I think we don't need this PR anymore as we're not using OpenLab on master anymore. I also opened an issue to drop it everywhere else: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/785

EDIT: ah you were faster :)

PR #797, opened 2021-03-19
🏃 Update golangci-lint and fix most of the linter issues

sbueringer:
**What this PR does / why we need it**:

After the v1alpha4 update I had to pin golangci-lint because there were just too many findings
and I didn't want to also fix them in the same PR. This PR now fixes all findings except some 
complexity findings (nestif & cyclop). Those can be fixed in other follow-up PRs.

Notes:
* this PR can also be reviewed but won't be merged before the test framework PR and also contains the test framework commit/PR

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

/hold

Issue #794, opened 2021-03-19
Provide qemu OS images via the capo staging bucket


sbueringer:
/kind feature

**Describe the solution you'd like**

We have a pretty fragile dependency on qcow2 images on Github releases at the moment.
Let's add a job which periodically builds and pushes Ubuntu 20 images with the help of the
image-builder repo. Probably the best "target" would be our staging bucket (but I would compare 
that how it's done with cluster-api nightly builds at the moment)

xref:
* https://kubernetes.slack.com/archives/C8TSNPY4T/p1616157439092100

/assign

jichenjc:
I checked that slack info and seems still lost some background
just for education, you want to create new images time to time (ubuntu 20) so that we can use latest 
image with kubeadm etc to make sure we won't suffer any not-up-to-date issue? @sbueringer

sbueringer:
@jichenjc
Yes. Problem right now is that we depend on an old Ubuntu image I uploaded on a previous release in our e2e tests. 
With my current PRs it gets a bit bitter (imho), as I'll now depend on images from [here](https://github.com/sbueringer/cluster-api-provider-openstack-images/releases). Ideally we don't depend on images of some of my repos. But I think that's a long-term discussion and I have to figure out with who I should talk about it

jichenjc:
@sbueringer thanks for the info, make sense to me :)

sbueringer:
My current plan is to add documentation and build scripts to the CAPO repo. With this documentation and sufficient rights to our bucket the images can then be manually build and uploaded.

PR #705, opened 2020-12-08
🐛 Fix SecurityGroupInUse error

hidekazuna:
**What this PR does / why we need it**:

This PR deletes port when an instance creation is not executed such as cpu quota error.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #701 

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #798, opened 2021-03-20
🏃 Add e2e tests for all flavors

### LINKED ISSUE #577: E2E test using Cluster API test framework
/kind feature

We can update E2E test to use Cluster API test framework.

Cluster API test framework is guided in [The Cluster API Book](https://cluster-api.sigs.k8s.io/developer/e2e.html).

### LINKED ISSUE #723: OpenStack 'port' does not get deleted if VM failed to create
/kind bug

* I created a cluster but had forgotten to create the openstack `keypair` beforehand.
* This led to VM not being created with error: `Invalid key_name provided` from openstack.
* Thus, deleted the cluster with `kubectl delete cluster <my-cluster>`.
* This resulted in errors in the controller-manager pod, complaining that the security group could not be deleted because it is in use.
* The reason for `security group in use` was that the `port` had been created for the VM. But since the VM was never created, the `port` was never attached and thus never got deleted.

If VM is not created for any reason (maybe invalid image name provided, invalid flavor provided, invalid keypair name provided etc.) then the openstack `port` which was created for VM but was never used, should be deleted when trying to delete the cluster.

Ref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/controllers/openstackmachine_controller.go#L221-L226
There should be a check here for existing ports too at the time of VM deletion. Maybe a `InstancePortExists`, just like the `InstanceExists` function?

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

### LINKED ISSUE #741: Can not delete cluster due to SecurityGroupInUse
/kind bug

1. Create a workload cluster
2. Delete the workload cluster

Deleting the cluster is not completed.
OpenStack instances is deleted.
But the following resources are not deleted.

 * network
 * security groups

A cause of the trouble is failure of security group deletion.
See the following logs:

```
$ k -n capo-system logs capo-controller-manager-7595db778f-9rl8n manager

--- snip ---

E0215 03:34:29.362070       1 controller.go:257] controller-runtime/controller "msg"="Reconciler error" "error"="failed to delete security group: Expected HTTP response code [] when accessing [DELETE https://172.16.71.80:9696/v2.0/security-groups/b9d2ad4d-dfc3-4f27-8a33-7bbba21b3cb3], but got 409 instead\n{\"NeutronError\": {\"message\": \"Security Group b9d2ad4d-dfc3-4f27-8a33-7bbba21b3cb3 in use.\", \"type\": \"SecurityGroupInUse\", \"detail\": \"\"}}" "controller"="openstackcluster" "name"="capi-quickstart" "namespace"="default"
```

Deleting the security group for worker nodes was failed.
The security group was applied to ports.
So the ports must be deleted before deleting the security groups.

Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `v0.3.3`
- Cluster-API version: `v0.3.13`
- OpenStack version: `Queens`
- Minikube/KIND version: `kind v0.9.0 go1.15.2 linux/amd64`
- Kubernetes version (use `kubectl version`): `v1.18.15`

```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.3 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.3 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

sbueringer:
This PR adds e2e tests for all our flavors (and some additional ones).

Fixes #577
Fixes #723
Fixes #741

Notes:

This PR can also be reviewed but won't be merged before the test framework and the linter PR and also contains those commits/PRs

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -180,7 +180,6 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 		}
 	}
 
-	// Delete other things
 	if workerSecGroup := openStackCluster.Status.WorkerSecurityGroup; workerSecGroup != nil {
 		log.Info("Deleting worker security group", "name", workerSecGroup.Name)
 		if err = networkingService.DeleteSecurityGroups(workerSecGroup); err != nil {
--- a/hack/ci/devstack-cloud-init.yaml.tpl
+++ b/hack/ci/devstack-cloud-init.yaml.tpl
@@ -1,8 +1,8 @@
 #cloud-config
-hostname: openstack
-password: ubuntu
-chpasswd: { expire: False }
-ssh_pwauth: True
+hostname: localhost
+users:
+- name: ubuntu
+  lock_passwd: true
 write_files:
   - content: |
       net.ipv4.ip_forward=1
@@ -22,8 +22,11 @@ write_files:
       # from https://raw.githubusercontent.com/openstack/octavia/master/devstack/contrib/new-octavia-devstack.sh
       git clone -b stable/${OPENSTACK_RELEASE} https://github.com/openstack/devstack.git /tmp/devstack
 
-      cat <<EOF > /tmp/devstack/localrc
+      cat <<EOF > /tmp/devstack/local.conf
+
+      [[local|localrc]]
       GIT_BASE=https://github.com
+      HOST_IP=10.0.2.15
 
       # Neutron
       enable_plugin neutron https://github.com/openstack/neutron stable/${OPENSTACK_RELEASE}
@@ -78,12 +81,26 @@ write_files:
 
       # Don't download default images, just our test images
       DOWNLOAD_DEFAULT_IMAGES=False
-      IMAGE_URLS="https://github.com/sbueringer/image-builder/releases/download/v1.18.15-01/ubuntu-2004-kube-v1.18.15.qcow2,"
-      IMAGE_URLS+="http://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img"
+      # We upload the Amphora image so it doesn't have to be build
+      IMAGE_URLS="https://github.com/sbueringer/cluster-api-provider-openstack-images/releases/download/amphora-victoria-1/amphora-x64-haproxy.qcow2"
+
+      # See: https://docs.openstack.org/nova/victoria/configuration/sample-config.html
+      # Helpful commands (on the devstack VM):
+      # * openstack resource provider list
+      # * openstack resource provider inventory list 4aa55af2-d50a-4a53-b225-f6b22dd01044
+      # * openstack resource provider usage show 4aa55af2-d50a-4a53-b225-f6b22dd01044
+      # * openstack hypervisor stats show
+      # * openstack hypervisor list
+      # * openstack hypervisor show openstack
+      # A CPU allocation ratio von 32 gives us 32 vCPUs in devstack
+      # This should be enough to run multiple e2e tests at the same time
+      [[post-config|\$NOVA_CONF]]
+      [DEFAULT]
+      cpu_allocation_ratio = 32.0
       EOF
 
       # Create the stack user
-      /tmp/devstack/tools/create-stack-user.sh
+      HOST_IP=10.0.2.15 /tmp/devstack/tools/create-stack-user.sh
 
       # Move everything into place (/opt/stack is the $HOME folder of the stack user)
       mv /tmp/devstack /opt/stack/
@@ -94,6 +111,10 @@ write_files:
 
       # Add environment variables for auth/endpoints
       echo 'source /opt/stack/devstack/openrc admin admin' >> /opt/stack/.bashrc
+
+      # Upload the images so we don't have to upload them from prow
+      su - stack -c "source /opt/stack/devstack/openrc admin admin && /opt/stack/devstack/tools/upload_image.sh https://github.com/sbueringer/cluster-api-provider-openstack-images/releases/download/ubuntu-2004-v1.18.15-0/ubuntu-2004-kube-v1.18.15.qcow2"
+      su - stack -c "source /opt/stack/devstack/openrc admin admin && /opt/stack/devstack/tools/upload_image.sh http://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img"
 
       sudo iptables -t nat -I POSTROUTING -o ens4 -s 172.24.4.0/24 -j MASQUERADE
       sudo iptables -I FORWARD -s 172.24.4.0/24 -j ACCEPT
--- a/hack/ci/devstack-on-gce-project-install.sh
+++ b/hack/ci/devstack-on-gce-project-install.sh
@@ -56,7 +56,9 @@ function init_networks() {
   if [[ ${GCP_NETWORK_NAME} != "default" ]]; then
     if ! gcloud compute networks describe "${GCP_NETWORK_NAME}" --project "${GCP_PROJECT}" > /dev/null;
     then
-      gcloud compute networks create --project "$GCP_PROJECT" "${GCP_NETWORK_NAME}" --subnet-mode auto --quiet
+      gcloud compute networks create --project "$GCP_PROJECT" "${GCP_NETWORK_NAME}" --subnet-mode custom
+      gcloud compute networks subnets create "${GCP_NETWORK_NAME}" --project "$GCP_PROJECT" --network="${GCP_NETWORK_NAME}" --range="10.0.0.0/20" --region "${GCP_REGION}"
+
       gcloud compute firewall-rules create "${GCP_NETWORK_NAME}"-allow-http --project "$GCP_PROJECT" \
         --allow tcp:80 --network "${GCP_NETWORK_NAME}" --quiet
       gcloud compute firewall-rules create "${GCP_NETWORK_NAME}"-allow-https --project "$GCP_PROJECT" \
@@ -124,13 +126,13 @@ main() {
       --project "${GCP_PROJECT}" \
       --zone "${GCP_ZONE}" \
       --image ubuntu-2004-nested \
-      --boot-disk-size 100G \
+      --boot-disk-size 300G \
       --boot-disk-type pd-ssd \
       --can-ip-forward \
       --tags http-server,https-server,novnc,openstack-apis \
       --min-cpu-platform "${GCP_MACHINE_MIN_CPU_PLATFORM}" \
       --machine-type "${GCP_MACHINE_TYPE}" \
-      --network-interface=network="${CLUSTER_NAME}-mynetwork,subnet=${CLUSTER_NAME}-mynetwork,aliases=/24" \
+      --network-interface="private-network-ip=10.0.2.15,network=${CLUSTER_NAME}-mynetwork,subnet=${CLUSTER_NAME}-mynetwork" \
       --metadata-from-file user-data=./hack/ci/devstack-cloud-init.yaml
   fi
 
@@ -192,12 +194,17 @@ main() {
   openstack availability zone list
   openstack domain list
 
+  # the flavors are created in a way that we can execute at least 2 e2e tests in parallel (overall we have 32 vCPUs)
   openstack flavor delete m1.tiny
   openstack flavor create --ram 512 --disk 1 --vcpus 1 --public --id 1 m1.tiny --property hw_rng:allowed='True'
   openstack flavor delete m1.small
   openstack flavor create --ram 4192 --disk 10 --vcpus 2 --public --id 2 m1.small --property hw_rng:allowed='True'
   openstack flavor delete m1.medium
   openstack flavor create --ram 6144 --disk 10 --vcpus 4 --public --id 3 m1.medium --property hw_rng:allowed='True'
+
+  # Adjust the CPU quota
+  openstack quota set --cores 32 demo
+  openstack quota set --secgroups 50 demo
 
   export OS_TENANT_NAME=demo
   export OS_USERNAME=demo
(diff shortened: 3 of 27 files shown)


sbueringer:
@jichenjc @hidekazuna @prankul88 Pull request can be reviewed now, please take a look :)

jichenjc:
several commits exist, should we merge from first PR?

hack/ci/devstack-cloud-init.yaml.tpl
```diff
@@ -78,8 +80,22 @@ write_files:
 
       # Don't download default images, just our test images
       DOWNLOAD_DEFAULT_IMAGES=False
-      IMAGE_URLS="https://github.com/sbueringer/image-builder/releases/download/v1.18.15-01/ubuntu-2004-kube-v1.18.15.qcow2,"
-      IMAGE_URLS+="http://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img"
+      # We upload the Amphora image so it doesn't have to be build
```
jichenjc, line 83: maybe a TODO here

sbueringer:
> several commits exist, should we merge from first PR?
> 
> do you have a way to extract the changes _only_ for this PR?

In this order:
* https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/782
* https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/797
* https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/798

I'll drop the duplicate commits here after #782 and #797 are merged. Just wanted to build on top and test the combination of this stuff.

For review there's a selection box in the files tab where you can select the commits of only this PR:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/798/files/97ff55feb541d83c9a33b808c49601cb6a58859c..55dd934763a30193525f27d33ec5c8912bc492b8

hack/ci/devstack-cloud-init.yaml.tpl
sbueringer: I would keep it for now. This adds a dependency but we already have the dependency because of the ubuntu image but this issue will try to fix it: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/794

(it just takes a long time to build the amphora image during the devstack installation on each test)

sbueringer:
/test pull-cluster-api-provider-openstack-make-conformance

Also dropped the commit here, so ready for review/merge now.

@jichenjc @hidekazuna @prankul88 PTAL, if you have some time :)

controllers/openstackcluster_controller.go
```diff
@@ -181,6 +181,10 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 	}
 
 	// Delete other things
+	if err = networkingService.DeleteOrphanedPorts(); err != nil {
+		return reconcile.Result{}, errors.Errorf("failed to delete orophaned ports: %v", err)
+	}
+
 	if workerSecGroup := openStackCluster.Status.WorkerSecurityGroup; workerSecGroup != nil {
```
sbueringer, line 188: This should fix: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/723

controllers/openstackcluster_controller.go
```diff
@@ -181,6 +181,10 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 	}
 
 	// Delete other things
+	if err = networkingService.DeleteOrphanedPorts(); err != nil {
```
sbueringer, line 184: This should fix: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/723

sbueringer:
Some final polishing (and subsequent linter fixes...), but now I'm done.
/test pull-cluster-api-provider-openstack-conformance-test

pkg/cloud/services/networking/network.go
```diff
@@ -369,3 +370,27 @@ func GetSubnetsByFilter(networkClient *gophercloud.ServiceClient, opts subnets.L
...
+	orphanedPorts, err := ports.ExtractPorts(portList)
+	if err != nil {
+		return fmt.Errorf("error extracting ports: %v", err)
+	}
+
+	for _, orphanedPort := range orphanedPorts {
+		if orphanedPort.DeviceOwner != "" {
+			continue
+		}
+
```
jichenjc, line 389: I'd think we at least log this behavior

pkg/cloud/services/networking/service.go
```diff
@@ -44,8 +47,21 @@ func NewService(client *gophercloud.ProviderClient, clientOpts *clientconfig.Cli
 	if err != nil {
 		return nil, fmt.Errorf("failed to create networking service client: %v", err)
 	}
+
+	var projectID string
+	if clientOpts.AuthInfo != nil {
```
jichenjc, line 52: is it possible clientOpts.AuthInfo is nil? 
if so the projectID will be "" , will it impact follow up features?

test/e2e/shared/exec_test.go
```diff
@@ -46,8 +47,8 @@ func Test_commandsForMachine(t *testing.T) {
 		{
 			name: "",
 			args: args{
-				machineIP: "10.6.0.230",
-				bastionIP: "172.24.4.58",
+				machineIP: "10.6.0.209",
+				bastionIP: "172.24.4.10",
```
jichenjc, line 51: I am not sure why we need such change? do we have hard dependency on IP?

jichenjc (commented):
overall quick , some nit question

pkg/cloud/services/networking/network.go
sbueringer: Yup I"ll add log statements where we do (i.e. delete) something

pkg/cloud/services/networking/service.go
sbueringer: I think it cannot be nil, but it has very high impact if it is. With an empty project id we would just list and delete orphaned ports depending on the rights of the user.

This means when you have some kind of admin / domain admin credentials in the cloud.yaml it will delete stuff all over the OpenStack.

I think we should fail hard here if we're unable to determine a project id

test/e2e/shared/exec_test.go
sbueringer: This was just for debugging locally. The test is not executed (t.Skip() above)
It's just way easier to develop the exec method locally when you don't have to  execute the whole e2e tests to do it :)

I'll rollback the IPs, I just changed them when debugging the func

hack/ci/devstack-on-gce-project-install.sh
```diff
@@ -56,7 +56,9 @@ function init_networks() {
   if [[ ${GCP_NETWORK_NAME} != "default" ]]; then
     if ! gcloud compute networks describe "${GCP_NETWORK_NAME}" --project "${GCP_PROJECT}" > /dev/null;
     then
-      gcloud compute networks create --project "$GCP_PROJECT" "${GCP_NETWORK_NAME}" --subnet-mode auto --quiet
+      gcloud compute networks create --project "$GCP_PROJECT" "${GCP_NETWORK_NAME}" --subnet-mode custom
```
sbueringer, line 59: Ah I got this from testing locally. I would like to keep it anyway. The idea is to not automatically create a lot of subnets as GCP is doing it with subnet mode auto

test/e2e/shared/exec_test.go
jichenjc: I am just curious as, not a big deal

pkg/cloud/services/networking/service.go
jichenjc: ok, thanks for the info

sbueringer:
/test pull-cluster-api-provider-openstack-conformance-test

@jichenjc thx for the review, good findings :) Should be all fixed now ptal

test/e2e/shared/exec_test.go
sbueringer: Absolutely fine :)

sbueringer:
/test pull-cluster-api-provider-openstack-conformance-test

(small fixup, was necessary because of the change to the subnet)

jichenjc:
@sbueringer please help to squash commit then I think we might merge this

sbueringer:
> @sbueringer please help to squash commit then I think we might merge this

@jichenjc done :)

pkg/cloud/services/networking/network.go
```diff
@@ -369,3 +370,28 @@ func GetSubnetsByFilter(networkClient *gophercloud.ServiceClient, opts subnets.L
...
+		return fmt.Errorf("error listing ports: %v", err)
+	}
+
+	orphanedPorts, err := ports.ExtractPorts(portList)
+	if err != nil {
+		return fmt.Errorf("error extracting ports: %v", err)
+	}
+
+	for _, orphanedPort := range orphanedPorts {
+		if orphanedPort.DeviceOwner != "" {
```
rustycl0ck, line 386: We shouldn't delete anything which is not created by ClusterAPI. This statement would lead to deletion of any ports which were not created by ClusterAPI but were created for other purposes (either manually or via some other automation).

One common use case for this: In cases where a loadbalancer is not available (or the user does not use if for whatever reasons), and instead a VIP (Virtual IP) along with `keepalived` is used for load balancing, then an openstack port is created (to reserve the IP and to create proper routes), but is not attached to any VM. Since this openstack port is not attached to any VM, it has `device_owner == ""`.

The `DeleteOrphanedPorts` function would delete this port unnecessarily and would cause service disruption elsewhere.

rustycl0ck (changes requested):

pkg/cloud/services/networking/network.go
sbueringer: @rustycl0ck Understood. So I assume there is no real way to cleanup those ports safely. I'll drop this for now from the e2e tests. Not sure how we can get rid of the problems during cluster deletion though. I'll try if I can get the e2e tests clean without this change but I'm not sure if it's possible

sbueringer: I'm not entirely sure about the general statement "we shouldn't delete anything which is not created by ClusterAPI". In my opinion there are cases where it's totally valid. One case is that the cluster as a whole should be deleted and somebody has the clean up the loadbalancers which has been created by the cloud provider openstack. This is also done in CAPA and the ccm itself is unable to delete them as it doesn't exist anymore.

Are those service interruptions really an issue when the whole cluster is deleted?

rustycl0ck: My use case was referring to a setup that I worked on, where a single openstack network (IPv6) was being used by multiple teams/products. While I was trying to use CAPI to setup a k8s cluster there, a few other teams had used the VIP approach for making their applications HA (since a loadbalancer wasn't available). Just FYI, the other products weren't k8s based, but legacy VM based systems/appliccations.

#723 was caused at the time of cluster creation where the port is created first, but if the image (or the ssh key or the flavor) does not exist, server creation would error out and the port would be left behind without any cleanup on server creation failure. This issue seems to have been addressed in #705 correctly (where a rollback is performed if the server creation failed). Similarly, we should try to fix the root cause instead of performing a general cleanup.

I think the correct way to fix this would be to figure out why this error occurred in the first place.
From the following code block: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L631-L648, I don't see how/why the port would have been left behind while the server got deleted successfully. If we can find the root cause, we can fix the problem in a cleaner way.

sbueringer: @rustycl0ck Thx providing some more details :). 

I think overall you're right. The better approach would be to cleanup the ports wherever they orphaned. I'm currently testing if this PR also works without the `DeleteOrphanedPorts` func: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/816
I assume this will be successful and then I'll drop the func on the current PR.

Out of curiosity (I'm not sure what is possible with OpenStack). Did you have:
* a separate project and where using the same project and network as the other teams or
* have you been using the same network but separate projects?

rustycl0ck: It was a single project and same network, because all the applications (k8s hosted as well as legacy VM based) had to interact with each other (on a private network internally).

sbueringer: As I feared the tests are now failing because they are timing out because the cluster deletion does not work: https://prow.k8s.io/log?container=test&id=1376784008490258432&job=pull-cluster-api-provider-openstack-e2e-test

I would suggest merging the PR anyway without the DeleteOrphanedPort func. I would rather have those tests with a known issue on master and unblock a lot of other PRs instead of having to rely only on the conformance tests.Those tests are not mandatory yet and I try to find the bug in a follow-up PR.

What do you think?
(@jichenjc )

sbueringer: Got the exact results. Looks like the error only happens when server are not successfully created right now. I'll check if I can just fix this...

sbueringer: Fixed it here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/798/commits/2e012c6e8fbbf51a844f9f514c62544c5ff998b1
Local tests were green. I'll squash and re-run the tests on this PR

jichenjc:
/lgtm 

thanks :)

sbueringer:
tests are green, so:
/hold cancel

PR #799, opened 2021-03-23
🐛 InstanceExists should not do substring search on name

mdbooth:
The server list api defines the name parameter to be a regular
expression:

https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detail#list-servers

This meant that the usage of it in InstanceExists was a
substring lookup rather than the intended lookup by name.

This is an 'upstream port' of the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1747270


sbueringer:
sgtm
/ok-to-test

/test pull-cluster-api-provider-openstack-make-conformance

jichenjc:
/approve

thanks~

PR #800, opened 2021-03-23
🏃 Optimize devstack setup on GCP, add a script to setup a devstack on AWS

sbueringer:

This PR extends our existing GCP setup script by a few new variants. After this PR the following is possible:
* GCP:
  * install a devstack from scratch
  * use a pre-installed devstack and save some time for our PR tests (it doesn't make sense to install devstack over and over again, we don't want to test the devstack installation :))
* AWS:
  * install a devstack from scratch

Regarding our PR tests this doesn't change that much, it only avoids installing devstack on every commit. 

However, I think it's important to provide a few options how our PR tests can be reproduced and developed manually. In my case I only have a AWS account from my company which I can use. One of the major issues I had with OpenLab was that the usual feedback loop was to push another commit and wait for an hour, without any access to the test system. I think we're now in a much better place when anybody can execute the same tests as Prow with either a GCP or AWS account (of course it's also still possible to just directly test against an OpenStack).

I think we should support the scripts which are not frequently run in CI only on a best-effort basis.

I also moved our OS image dependencies in the CAPO GCP bucket. This has the following advantages:
* it's probably cheaper and faster to just pull the image from GCS when our tests are run in GCP
* we don't depend on release artifacts of one of my private repositories

I would add some documentation how the images are build and uploaded to our GCS bucket in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/794

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

/hold


test/e2e/data/ci-artifacts-platform-kustomization.yaml
```diff
@@ -9,3 +9,25 @@ metadata:
   name: ${CLUSTER_NAME}-control-plane
 spec:
   version: ${KUBERNETES_VERSION}
+# Comment this out for debbuging in case there are problems with the ssky key
```
prankul88, line 12: Is it `ssh` key?

prankul88 (commented):
Otherwise LGTM

test/e2e/data/ci-artifacts-platform-kustomization.yaml
sbueringer: The idea is to use the commented out section if there is a problem with the OpenStack keypair which is otherwise used in the e2e tests. Because if there is a problem without it's very hard to debug because you then cannot ssh onto the nodes.

prankul88: > "Problems with the ssky key"

I meant is it "ssh key" instead of ssky key

sbueringer: Ups, yes that's a typo :)

sbueringer, line 12:
```suggestion
# Comment this out for debbuging in case there are problems with the OpenStack key pair mechanism
```

sbueringer: I changed it a bit to mention the OpenStack keypair mechanism explicitly

prankul88: Thanks !!

PR #831, opened 2021-04-06
🏃 Delete mockgen and defaulter-gen

hidekazuna:

We do not use mockgen and defaulter-gen so deleted them.

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


sbueringer:
I would prefer keeping them as it's only a matter of time until we need them and it keeps the diff smaller with the "upstream" (CAPI/CAPA) Makefiles, so it's easier to keep them in sync.

hidekazuna:
> I would prefer keeping them as it's only a matter of time until we need them and it keeps the diff smaller with the "upstream" (CAPI/CAPA) Makefiles, so it's easier to keep them in sync.

OK, I got diff with CAPI/CAPA/CAPZ Makefiles :)

PR #844, opened 2021-04-09
WIP: ✨ load balancer: delete orphaned load balancers from CCM witin project

### LINKED ISSUE #842: Cleanup loadbalancers created by ccm
/kind feature

Currently, we only delete the kubeapi loadbalancer created by ClusterAPI. Usually, there are additional loadbalancers created by the ccm. I would also like to delete those on cluster deletion. They cannot be deleted by the ccm, because the ccm is not running anymore. The cluster deletion get's blocked because network etc. cannot be deleted if there are ccm orphan lbs left.

Notes:
* we should also add a e2e test for it. I think we should implement it similarly to CAPA: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/test/e2e/suites/unmanaged/unmanaged_test.go#L178-L210
* we have to ensure we're not deleting unrelated LBs in the same OpenStack project
  * I'm not sure if there is a way to identify if the loadbalancers are related to the cluster we are deleting
  * worst case is that we need a flag on the CAPO controller to opt-out of this behaviour
* I assume https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/839 will not significatly change the problem, so let's ignore it for now.

seanschneeweiss:
There are additional load balancers created by the CCM which should be deleted when the cluster is deleted. Otherwise, the deletion is blocked as e.g. the network is used by the orphaned load balancers.

Draft implementation for #842.

This is a first draft and can be extended by the ideas of #842.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold
<sub>Sean Schneeweiss <sean.schneeweiss@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -318,13 +318,18 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 	return nil
 }
 
+// DeleteLoadBalancer deletes a load balancer identified by its name and all child objects (cascade).
 func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
 	loadBalancerName := getLoadBalancerName(clusterName)
 	lb, err := s.checkIfLbExists(loadBalancerName)
 	if err != nil {
 		return err
 	}
+	return s.deleteLoadBalancer(openStackCluster, lb)
+}
 
+// deleteLoadBalancer deletes the load balancer and all child objects (cascade).
+func (s *Service) deleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, lb *loadbalancers.LoadBalancer) error {
 	if lb == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -324,6 +330,28 @@ func (s *Service) DeleteLoadBalancerMember(openStackCluster *infrav1.OpenStackCl
 	return nil
 }
 
+// DeleteOrphanedLoadBalancers deletes all remaining load balancers and all child objects within an OpenStack project.
+// Orphaned load balancers are expected when deleting the cluster because the CCM manages additional load balancers.
+func (s *Service) DeleteOrphanedLoadBalancers(openStackCluster *infrav1.OpenStackCluster) error {
+	lbList, err := loadbalancers.List(s.loadbalancerClient, loadbalancers.ListOpts{ProjectID: s.projectID}).AllPages()
```
jichenjc, line 336: are we here list all LB from this project then delete all of them?

seanschneeweiss: Not sure if I understood the question correctly, but the intention here is to list all LBs from the project and delete all of them - so basically yes would be the answer.

seanschneeweiss: I think that the decision which load balancer should be deleted is part of the ongoing discussion in the linked issue. I've just added our implementation as a starting point.

seanschneeweiss:
Closing in favor of #990 and the [suggestion](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/842#issuecomment-901246398) to delete remaining load balancers by network id.

PR #845, opened 2021-04-12
Wait for instance delete

jichenjc:

remove a TODO (wait for instance delete complete in reconcile logic)

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/bastion.go
+++ b/pkg/cloud/services/compute/bastion.go
@@ -19,8 +19,11 @@ package compute
 import (
 	"fmt"
 
+	"sigs.k8s.io/cluster-api/util"
+
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha4"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
+	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
 )
 
 func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, serverID string) error {
@@ -34,6 +37,21 @@ func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, serv
 	if err = deleteInstance(s, instance.ID); err != nil {
 		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
 		return err
+	}
+
+	err = util.PollImmediate(RetryIntervalInstanceStatus, TimeoutInstanceDelete, func() (bool, error) {
+		_, err = s.GetInstance(instance.ID)
+		if err != nil {
+			if capoerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+	if err != nil {
+		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
+		return fmt.Errorf("error deleting Openstack instance %s, %v", instance.ID, err)
 	}
 
 	record.Eventf(openStackCluster, "SuccessfulDeleteServer", "Deleted server %s with id %s", instance.Name, instance.ID)
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -60,6 +60,8 @@ const (
 
 	TimeoutPortDelete       = 3 * time.Minute
 	RetryIntervalPortDelete = 5 * time.Second
+
+	TimeoutInstanceDelete = 5 * time.Minute
 )
 
 // InstanceCreate creates a compute instance.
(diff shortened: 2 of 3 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -580,6 +582,20 @@ func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *i
...
+		if err != nil {
+			if capoerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("error deleting Openstack instance %s, %v", parsed.ID(), err)
```
hidekazuna, line 596: better to update like Line 579-580:

```go
record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s: %v", parsed.ID(), err)
return err
```

hidekazuna: We need to add the same logic for deleting bastion.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/bastion.go#L34-L38

sbueringer: I think we should phrase the message a little bit differently to indicate the wait timeout so we now which one of the cases we hit.

pkg/cloud/services/compute/instance.go
```diff
@@ -580,6 +582,20 @@ func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *i
 		return err
 	}
 
+	err = util.PollImmediate(RetryIntervalInstanceStatus, TimeoutInstanceDelete, func() (bool, error) {
+		_, err = s.GetInstance(parsed.ID())
+		if err != nil {
+			if capoerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
```
sbueringer, line 591: It's fine for me but just to be sure. Do we want to stop polling here? Because returning the error will stop the PollImmediate func. Alternatively we could log the error and return nil so the PollImmedate tries again

pkg/cloud/services/compute/instance.go
jichenjc: @sbueringer does the error report already contains in Poll logic? I checked all other Poll method does not have special handle

pkg/cloud/services/compute/instance.go
jichenjc: if it's error and not NOTFOUND error, this means something wrong happened...
I think we can stop polling otherwise we will always get timeout error

pkg/cloud/services/compute/instance.go
sbueringer: Ah probably there is something in the poll timeout error, you're right this should be good enough

pkg/cloud/services/compute/instance.go
sbueringer: Yup fine for me. I might want to adjust that later on a bit if we see other retryable errors in production, but absolutely fine for now

pkg/cloud/services/compute/bastion.go
```diff
@@ -36,6 +39,21 @@ func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, serv
...
+		if err != nil {
+			if capoerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+	if err != nil {
+		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to deleted server %s: %v", instance.ID, err)
```
sbueringer, line 53: looks like we also add the name in l.38, maybe good to also add it here for consistency

pkg/cloud/services/compute/instance.go
```diff
@@ -580,6 +582,21 @@ func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *i
...
+		}
+		return false, nil
+	})
+	if err != nil {
+		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s: %v", parsed.ID(), err)
```
sbueringer, line 596: If we want to keep this consistent with the even in deleteBastion we should add the server name here and in l.581

pkg/cloud/services/compute/instance.go
```diff
@@ -576,10 +578,25 @@ func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *i
 		return err
 	}
 	if err = deleteInstance(s, parsed.ID()); err != nil {
-		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s: %v", parsed.ID(), err)
+		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s with ID %s: %v", openStackMachine.Name, parsed.ID(), err)
```
sbueringer, line 581:
```suggestion
		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s with id %s: %v", openStackMachine.Name, parsed.ID(), err)
```

pkg/cloud/services/compute/instance.go
```diff
@@ -576,10 +578,25 @@ func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *i
...
+		if err != nil {
+			if capoerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+	if err != nil {
+		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s with ID %s: %v", openStackMachine.Name, parsed.ID(), err)
```
sbueringer, line 596:
```suggestion
		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s with id %s: %v", openStackMachine.Name, parsed.ID(), err)
```

PR #846, opened 2021-04-12
Delete watching Status when deleting

### LINKED ISSUE #841: OpenStack network resources are not deleted just after created
/kind bug

- Create cluster
- Delete cluster just after OpenStack network resources(Load balancer, network, subnet, router, security groups) are created.
- Cluster was deleted successfully, but OpenStack network resources are not deleted at all.

OpenStack network resources are deleted successfully just after they are created, cluster is still provisioning phase.

I think we should not see openStackCluster.Status when deleting cluster.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): fa64ad4694324ffbdd3054aeb62c7946afccf8d5
- Cluster-API version: v1alpha4
- OpenStack version: Stein
- Minikube/KIND version: v0.10.0
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-21T01:11:42Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}

hidekazuna:

This PR fixes deleting cluster and machines.
- We can not delete OpenStack network resources just after resources are created.
- We can not delete machine before assigning provider ID.

Fixes #841 

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -197,7 +197,7 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
 
 	if instance != nil && instance.FloatingIP != "" {
 		if err = networkingService.DisassociateFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
-			return errors.Errorf("failed to delete floating IP: %v", err)
+			return errors.Errorf("failed to disassociate floating IP: %v", err)
 		}
 		if err = networkingService.DeleteFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
 			return errors.Errorf("failed to delete floating IP: %v", err)
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -110,14 +110,14 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// Handle deleted clusters
 	if !openStackCluster.DeletionTimestamp.IsZero() {
-		return reconcileDelete(ctx, log, r.Client, patchHelper, openStackCluster)
+		return reconcileDelete(ctx, log, r.Client, patchHelper, cluster, openStackCluster)
 	}
 
 	// Handle non-deleted clusters
 	return reconcileNormal(ctx, log, r.Client, patchHelper, cluster, openStackCluster)
 }
 
-func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client, patchHelper *patch.Helper, openStackCluster *infrav1.OpenStackCluster) (ctrl.Result, error) {
+func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client, patchHelper *patch.Helper, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster) (ctrl.Result, error) {
 	log.Info("Reconciling Cluster delete")
 
 	osProviderClient, clientOpts, err := provider.NewClientFromCluster(client, openStackCluster)
@@ -125,7 +125,7 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 		return reconcile.Result{}, err
 	}
 
-	if err = deleteBastion(log, osProviderClient, clientOpts, openStackCluster); err != nil {
+	if err = deleteBastion(log, osProviderClient, clientOpts, cluster, openStackCluster); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -139,41 +139,25 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 		return reconcile.Result{}, err
 	}
 
-	if openStackCluster.Spec.ManagedAPIServerLoadBalancer && openStackCluster.Status.Network != nil {
-		if apiLb := openStackCluster.Status.Network.APIServerLoadBalancer; apiLb != nil {
-			if err = loadBalancerService.DeleteLoadBalancer(openStackCluster, apiLb.Name); err != nil {
-				return reconcile.Result{}, errors.Errorf("failed to delete load balancer: %v", err)
-			}
+	clusterName := fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)
 
-			if openStackCluster.Spec.APIServerFloatingIP == "" {
-				if err = networkingService.DeleteFloatingIP(openStackCluster, apiLb.IP); err != nil {
-					return reconcile.Result{}, errors.Errorf("failed to delete floating IP: %v", err)
-				}
-			}
+	if openStackCluster.Spec.ManagedAPIServerLoadBalancer {
+		if err = loadBalancerService.DeleteLoadBalancer(openStackCluster, clusterName); err != nil {
+			return reconcile.Result{}, errors.Errorf("failed to delete load balancer: %v", err)
 		}
 	}
 
-	if workerSecGroup := openStackCluster.Status.WorkerSecurityGroup; workerSecGroup != nil {
-		if err = networkingService.DeleteSecurityGroups(openStackCluster, workerSecGroup); err != nil {
-			return reconcile.Result{}, errors.Errorf("failed to delete security group: %v", err)
-		}
-	}
-
-	if controlPlaneSecGroup := openStackCluster.Status.ControlPlaneSecurityGroup; controlPlaneSecGroup != nil {
-		if err = networkingService.DeleteSecurityGroups(openStackCluster, controlPlaneSecGroup); err != nil {
-			return reconcile.Result{}, errors.Errorf("failed to delete security group: %v", err)
-		}
+	if err = networkingService.DeleteSecurityGroups(openStackCluster, clusterName); err != nil {
+		return reconcile.Result{}, errors.Errorf("failed to delete security groups: %v", err)
 	}
 
 	// if NodeCIDR was not set, no network was created.
-	if openStackCluster.Status.Network != nil && openStackCluster.Spec.NodeCIDR != "" {
-		if openStackCluster.Status.Network.Router != nil {
-			if err = networkingService.DeleteRouter(openStackCluster, openStackCluster.Status.Network); err != nil {
-				return ctrl.Result{}, errors.Errorf("failed to delete router: %v", err)
-			}
+	if openStackCluster.Spec.NodeCIDR != "" {
+		if err = networkingService.DeleteRouter(openStackCluster, clusterName); err != nil {
+			return ctrl.Result{}, errors.Errorf("failed to delete router: %v", err)
 		}
 
-		if err = networkingService.DeleteNetwork(openStackCluster, openStackCluster.Status.Network); err != nil {
+		if err = networkingService.DeleteNetwork(openStackCluster, clusterName); err != nil {
 			return ctrl.Result{}, errors.Errorf("failed to delete network: %v", err)
 		}
 	}
@@ -196,7 +180,7 @@ func contains(arr []string, target string) bool {
 	return false
 }
 
-func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts, openStackCluster *infrav1.OpenStackCluster) error {
+func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster) error {
 	computeService, err := compute.NewService(osProviderClient, clientOpts, log)
 	if err != nil {
 		return err
@@ -206,22 +190,29 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
 		return err
 	}
 
-	if openStackCluster.Status.Bastion != nil {
-		if err = computeService.DeleteBastion(openStackCluster, openStackCluster.Status.Bastion.ID); err != nil {
-			return errors.Errorf("failed to delete bastion: %v", err)
-		}
-		if err = networkingService.DeleteFloatingIP(openStackCluster, openStackCluster.Status.Bastion.FloatingIP); err != nil {
-			return errors.Errorf("failed to delete floating IP: %v", err)
-		}
-		openStackCluster.Status.Bastion = nil
+	instance, err := computeService.InstanceExists(fmt.Sprintf("%s-bastion", cluster.Name))
+	if err != nil {
+		return err
 	}
 
-	if openStackCluster.Status.BastionSecurityGroup != nil {
-		if err = networkingService.DeleteSecurityGroups(openStackCluster, openStackCluster.Status.BastionSecurityGroup); err != nil {
-			return errors.Errorf("failed to delete security group: %v", err)
+	if instance != nil && instance.FloatingIP != "" {
+		if err = networkingService.DisassociateFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
+			return errors.Errorf("failed to delete floating IP: %v", err)
 		}
-		openStackCluster.Status.BastionSecurityGroup = nil
+		if err = networkingService.DeleteFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
+			return errors.Errorf("failed to delete floating IP: %v", err)
+		}
 	}
+
+	if err = computeService.DeleteBastion(openStackCluster, cluster.Name); err != nil {
+		return errors.Errorf("failed to delete bastion: %v", err)
+	}
+	openStackCluster.Status.Bastion = nil
+
+	if err = networkingService.DeleteBastionSecurityGroup(openStackCluster, fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)); err != nil {
+		return errors.Errorf("failed to delete bastion security group: %v", err)
+	}
+	openStackCluster.Status.BastionSecurityGroup = nil
 
 	return nil
 }
@@ -294,7 +285,7 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
 	log.Info("Reconciling Bastion")
 
 	if openStackCluster.Spec.Bastion == nil || !openStackCluster.Spec.Bastion.Enabled {
-		return deleteBastion(log, osProviderClient, clientOpts, openStackCluster)
+		return deleteBastion(log, osProviderClient, clientOpts, cluster, openStackCluster)
 	}
 
 	computeService, err := compute.NewService(osProviderClient, clientOpts, log)
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -204,9 +204,12 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
 		}
 	}
 
-	if err = computeService.DeleteBastion(openStackCluster, cluster.Name); err != nil {
-		return errors.Errorf("failed to delete bastion: %v", err)
+	if instance != nil && instance.Name != "" {
+		if err = computeService.DeleteInstance(openStackCluster, instance.Name); err != nil {
+			return errors.Errorf("failed to delete bastion: %v", err)
+		}
 	}
+
 	openStackCluster.Status.Bastion = nil
 
 	if err = networkingService.DeleteBastionSecurityGroup(openStackCluster, fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)); err != nil {
--- a/pkg/cloud/services/compute/bastion.go
+++ b/pkg/cloud/services/compute/bastion.go
@@ -26,35 +26,35 @@ import (
 	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
 )
 
-func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, serverID string) error {
-	instance, err := s.GetInstance(serverID)
+func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
+	instance, err := s.InstanceExists(fmt.Sprintf("%s-bastion", clusterName))
 	if err != nil {
 		return err
 	}
-	if instance == nil {
-		return nil
-	}
-	if err = deleteInstance(s, instance.ID); err != nil {
-		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
-		return err
-	}
 
-	err = util.PollImmediate(RetryIntervalInstanceStatus, TimeoutInstanceDelete, func() (bool, error) {
-		_, err = s.GetInstance(instance.ID)
-		if err != nil {
-			if capoerrors.IsNotFound(err) {
-				return true, nil
-			}
-			return false, err
+	if instance != nil && instance.ID != "" {
+		if err = deleteInstance(s, instance.ID); err != nil {
+			record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
+			return err
 		}
-		return false, nil
-	})
-	if err != nil {
-		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
-		return fmt.Errorf("error deleting Openstack instance %s, %v", instance.ID, err)
-	}
 
-	record.Eventf(openStackCluster, "SuccessfulDeleteServer", "Deleted server %s with id %s", instance.Name, instance.ID)
+		err = util.PollImmediate(RetryIntervalInstanceStatus, TimeoutInstanceDelete, func() (bool, error) {
+			_, err = s.GetInstance(instance.ID)
+			if err != nil {
+				if capoerrors.IsNotFound(err) {
+					return true, nil
+				}
+				return false, err
+			}
+			return false, nil
+		})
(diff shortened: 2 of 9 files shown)


controllers/openstackcluster_controller.go
```diff
@@ -206,22 +190,30 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
...
 
-	if openStackCluster.Status.BastionSecurityGroup != nil {
-		if err = networkingService.DeleteSecurityGroups(openStackCluster, openStackCluster.Status.BastionSecurityGroup); err != nil {
-			return errors.Errorf("failed to delete security group: %v", err)
-		}
-		openStackCluster.Status.BastionSecurityGroup = nil
+	if err = computeService.DeleteBastion(openStackCluster, clusterName); err != nil {
+		return errors.Errorf("failed to delete bastion: %v", err)
+	}
+	if err = networkingService.DeleteFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
```
sbueringer, line 208: we will have problems with reentrancy with the current impl, e.g.:
* DeleteBastion works
* DeleteFloatingIP fails
* retry will return in l.201
* security group will never get deleted

=> I think we can only skip the bastion / floating ip deletion when we are 100% sure they are gone. Only the ensure the server is gone is not good enough.

controllers/openstackcluster_controller.go
```diff
@@ -294,15 +286,17 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
...
 
 	computeService, err := compute.NewService(osProviderClient, clientOpts, log)
 	if err != nil {
 		return err
 	}
 
-	instance, err := computeService.InstanceExists(fmt.Sprintf("%s-bastion", cluster.Name))
+	clusterName := fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)
+
+	instance, err := computeService.InstanceExists(getBastionName(clusterName))
```
sbueringer, line 299: This changes the name of the bastion host right? I think that's okay for v1alpha3 => v1alpha4, but we should document this. Can you maybe start a MD to collect "breaking" changes for v1alpha4 so we can post them with our release notes when we release?

pkg/cloud/services/compute/bastion.go
```diff
@@ -23,12 +23,18 @@ import (
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
 )
 
-func (s *Service) DeleteBastion(openStackCluster *infrav1.OpenStackCluster, serverID string) error {
-	instance, err := s.GetInstance(serverID)
+func getBastionName(clusterName string) string {
```
sbueringer, line 26: Can we ensure we don't need two places where we calculate the bastionname? I.e. just call Delete/CreateBasiton and the compute package handles it from there?

pkg/cloud/services/compute/instance.go
```diff
@@ -566,21 +565,25 @@ func (s *Service) AssociateFloatingIP(instanceID, floatingIP string) error {
...
 	}
-
-	parsed, err := noderefutil.NewProviderID(*machine.Spec.ProviderID)
+	instance, err := s.InstanceExists(openStackMachine.Name)
 	if err != nil {
 		return err
 	}
-	if err = deleteInstance(s, parsed.ID()); err != nil {
-		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s: %v", parsed.ID(), err)
+	if instance == nil {
```
sbueringer, line 576: I'm not sure if that's safe. Do we know the deleteInstance has 100% completed when the server is gone in OpenStack?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -271,7 +272,16 @@ func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster,
 	}
 
 	record.Eventf(openStackCluster, "SuccessfulDeleteLoadBalancer", "Deleted load balancer %s with id %s", lb.Name, lb.ID)
-	return nil
+
```
sbueringer, line 275: /cc @seanschneeweiss @chrischdi can you please review this part if it breaks something for us

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -333,38 +333,42 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
 
-func (s *Service) DeleteSecurityGroups(openStackCluster *infrav1.OpenStackCluster, group *infrav1.SecurityGroup) error {
-	exists, err := s.exists(group.ID)
-	if err != nil {
-		return err
+func (s *Service) DeleteSecurityGroups(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
+	secGroupNames := []string{
+		getSecControlPlaneGroupName(clusterName),
+		getSecWorkerGroupName(clusterName),
+		getSecBastionGroupName(clusterName),
```
sbueringer, line 340: do we have to delete the sec groups here when wee have a separate func below?

sbueringer:
@hidekazuna I think I found all places where reentrancy could be a problem, but please also check if I'm correct ;)

controllers/openstackcluster_controller.go
hidekazuna: Thanks, I tried the rule, but totally wrong. I will fully update.

controllers/openstackcluster_controller.go
hidekazuna: Yes, but I should have focused on fixing deleting. I will update the name back.

pkg/cloud/services/compute/bastion.go
hidekazuna: I will update the name back, but existing code has the issue. The another PR might be solve the issue.

pkg/cloud/services/compute/instance.go
hidekazuna: #845 would fix :)

pkg/cloud/services/networking/securitygroups.go
hidekazuna: Do you mean it is enough to use separate function for each group?

controllers/openstackcluster_controller.go
sbueringer: Okay thx

pkg/cloud/services/compute/bastion.go
sbueringer: Okay fine for me

pkg/cloud/services/networking/securitygroups.go
sbueringer: No I was thinking we don't have to also delete the bastion group security group here because it's already deleted in deleteBastion (which calls DeleteBastionSecurityGroup)

pkg/cloud/services/compute/instance.go
sbueringer: Ah yes, and I also though incorrectly. deleteInstance *only* deletes the server and either it works or it doesn't. So the `instance == null` check verifies completely that the deleteInstance did what it should do. (It would be different if we would have to do something else after the server deletion in deleteInstance)

so tl;dr, fine for me as is!

pkg/cloud/services/loadbalancer/loadbalancer.go
sbueringer: Let's resolve this for now

chrischdi: However: I think this is ok :+1:

sbueringer: Thx! :)

hidekazuna:
PR updated. Before deleting instance, Floating IP is disassociated and deleted.

sbueringer:
@hidekazuna can you rebase? I'll try to take a look today or tomorrow

controllers/openstackcluster_controller.go
```diff
@@ -206,22 +190,29 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
...
-		if err = computeService.DeleteBastion(openStackCluster, openStackCluster.Status.Bastion.ID); err != nil {
-			return errors.Errorf("failed to delete bastion: %v", err)
+	instance, err := computeService.InstanceExists(fmt.Sprintf("%s-bastion", cluster.Name))
+	if err != nil {
+		return err
+	}
+
+	if instance != nil && instance.FloatingIP != "" {
+		if err = networkingService.DisassociateFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
+			return errors.Errorf("failed to delete floating IP: %v", err)
```
sbueringer, line 200:
```suggestion
			return errors.Errorf("failed to disassociate floating IP: %v", err)
```

pkg/cloud/services/compute/bastion.go
```diff
@@ -26,35 +26,35 @@ import (
...
 	if err != nil {
 		return err
 	}
-	if instance == nil {
-		return nil
-	}
-	if err = deleteInstance(s, instance.ID); err != nil {
-		record.Warnf(openStackCluster, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
-		return err
-	}
```
sbueringer, line 34: I think we should use a guard clause here:
```go
if instance == nil || instance.ID == "" {
  return nil
}
```

pkg/cloud/services/compute/instance.go
```diff
@@ -568,36 +567,38 @@ func (s *Service) AssociateFloatingIP(instanceID, floatingIP string) error {
...
-	parsed, err := noderefutil.NewProviderID(*machine.Spec.ProviderID)
+	instance, err := s.InstanceExists(openStackMachine.Name)
 	if err != nil {
 		return err
 	}
-	if err = deleteInstance(s, parsed.ID()); err != nil {
-		record.Warnf(openStackMachine, "FailedDeleteServer", "Failed to deleted server %s with id %s: %v", openStackMachine.Name, parsed.ID(), err)
-		return err
-	}
+	if instance != nil && instance.ID != "" {
```
sbueringer, line 579: I think we should use a guard clause here:
```go
if instance == nil || instance.ID == "" {
  return nil
}
```

pkg/cloud/services/compute/instance.go
```diff
@@ -568,36 +567,38 @@ func (s *Service) AssociateFloatingIP(instanceID, floatingIP string) error {
...
 
 func (s *Service) InstanceDelete(machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine) error {
-	if machine.Spec.ProviderID == nil {
+	if openStackMachine == nil {
 		// nothing to do
 		return nil
 	}
 
-	parsed, err := noderefutil.NewProviderID(*machine.Spec.ProviderID)
+	instance, err := s.InstanceExists(openStackMachine.Name)
```
sbueringer, line 575: This whole block looks like the same as in bastion.go. Can we move this in a func and reuse it?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -250,27 +250,40 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
...
-	deleteOpts := loadbalancers.DeleteOpts{
-		Cascade: true,
-	}
-	s.logger.Info("Deleting load balancer", "name", loadBalancerName, "cascade", deleteOpts.Cascade)
-	err = loadbalancers.Delete(s.loadbalancerClient, lb.ID, deleteOpts).ExtractErr()
-	if err != nil {
-		record.Warnf(openStackCluster, "FailedDeleteLoadBalancer", "Failed to delete load balancer %s with id %s: %v", lb.Name, lb.ID, err)
-		return err
-	}
+	if lb != nil && lb.VipPortID != "" {
```
sbueringer, line 260: is this really correct?
I get that we only want to disassociate the floating ip and delete it if lb.VipPortID != "", but I think we want to delete the lb even if the floating ip is already gone

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -250,27 +250,40 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 	return nil
 }
 
-func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, loadBalancerName string) error {
+func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
+	loadBalancerName := getLoadBalancerName(clusterName)
 	lb, err := checkIfLbExists(s.loadbalancerClient, loadBalancerName)
 	if err != nil {
 		return err
 	}
```
sbueringer, line 258: I would use a guard clause for lb == nil here:
```go
if lb == nil {
  return nil
}
```

pkg/cloud/services/networking/network.go
```diff
@@ -148,19 +147,17 @@ func (s *Service) ReconcileNetwork(openStackCluster *infrav1.OpenStackCluster, c
...
-	exists, err := s.existsNetwork(network.ID)
+func (s *Service) DeleteNetwork(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
+	networkName := getNetworkName(clusterName)
+	network, err := s.getNetworkByName(networkName)
 	if err != nil {
 		return err
 	}
-	if !exists {
-		s.logger.Info("Skipping network deletion because network doesn't exist", "network", network.ID)
+	if network.ID == "" {
```
sbueringer, line 156:
```suggestion
	if network == nil || network.ID == "" {
```

sbueringer: Can we return a Network pointer in getNetworkByName? Would be more consistent with our other code

pkg/cloud/services/networking/router.go
```diff
@@ -185,41 +186,41 @@ func setRouterExternalIPs(client *gophercloud.ServiceClient, openStackCluster *i
...
-		s.logger.V(4).Info("Skipping removing router interface since no subnet exists.")
-	} else {
-		_, err = routers.RemoveInterface(s.client, network.Router.ID, routers.RemoveInterfaceOpts{
-			SubnetID: network.Subnet.ID,
-		}).Extract()
-		if err != nil {
-			if errors.IsNotFound(err) {
-				s.logger.V(4).Info("Router Interface already removed, no actions", "id", network.Router.ID)
-				return nil
+
```
sbueringer, line 201: please also use a guard clause here and also check for nil

pkg/cloud/services/networking/router.go
```diff
@@ -185,41 +186,41 @@ func setRouterExternalIPs(client *gophercloud.ServiceClient, openStackCluster *i
...
+	router, err := s.getRouterByName(routerName)
 	if err != nil {
 		return err
 	}
-	if !exists {
-		s.logger.Info("Skipping router deletion because router doesn't exist", "router", network.Router.ID)
-		return nil
+
+	subnetName := getSubnetName(clusterName)
+	subnet, err := s.getSubnetByName(subnetName)
```
sbueringer, line 197: can we also return a pointer in getSubnetByName (please also check similar funcs in other places)

pkg/cloud/services/networking/router.go
```diff
@@ -185,41 +186,41 @@ func setRouterExternalIPs(client *gophercloud.ServiceClient, openStackCluster *i
...
+
+	if router.ID != "" {
+		if subnet.ID != "" {
+			_, err = routers.RemoveInterface(s.client, router.ID, routers.RemoveInterfaceOpts{
+				SubnetID: subnet.ID,
+			}).Extract()
+			if err != nil {
+				if capoerrors.IsNotFound(err) {
+					s.logger.V(4).Info("Router Interface already removed, no actions", "id", router.ID)
+					return nil
```
sbueringer, line 210: when we return here we skip the router deletion. I don't think that's intended

pkg/cloud/services/networking/router.go
```diff
@@ -239,19 +240,56 @@ func (s *Service) getRouterInterfaces(routerID string) ([]ports.Port, error) {
...
-		return false, nil
+
+	switch len(subnetList) {
+	case 0:
+		return subnets.Subnet{}, nil
+	case 1:
+		return subnetList[0], nil
 	}
-	return true, nil
+	return subnets.Subnet{}, errors.New("too many resources")
```
sbueringer, line 290: can we we write something about routers instead of the generic resources? not sure if we have similar errors already which we should stay consistent to

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -333,38 +333,41 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
-		ID: groupID,
-	}
-	allPages, err := groups.List(s.client, opts).AllPages()
+func (s *Service) DeleteBastionSecurityGroup(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
+	secBastionGroupName := getSecBastionGroupName(clusterName)
+	return s.deleteSecurityGroup(openStackCluster, secBastionGroupName)
+}
+
+func (s *Service) deleteSecurityGroup(openStackCluster *infrav1.OpenStackCluster, name string) error {
+	group, err := s.getSecurityGroupByName(name)
```
sbueringer, line 356: i would also return a pointer here and then check also for nil below

sbueringer (commented):
Next round :)

Please push the fixes as new commits for now. It's way easier for me to incrementally review the PR with the changes separated compared to reviewing the whole PR all over again.

Let's squash before merge.

pkg/cloud/services/networking/network.go
hidekazuna: Our code is not consistent as you expect. functions in networking package do not return pointer. functions in compute and loadbalancer do return pointer.  If we update getNetworkByName, it is inconsistent in networking package.

controllers/openstackcluster_controller.go
hidekazuna: Fixed.

pkg/cloud/services/compute/bastion.go
hidekazuna: Fixed.

pkg/cloud/services/compute/instance.go
hidekazuna: Fixed.

pkg/cloud/services/loadbalancer/loadbalancer.go
hidekazuna: Updated, review again, please.

pkg/cloud/services/networking/router.go
hidekazuna: Fixed deleting `return nil`

pkg/cloud/services/networking/router.go
hidekazuna: Introduced Router type.

hidekazuna:
@sbueringer I want to leave returning pointer issue, because functions in networking package do not return pointer.

pkg/cloud/services/compute/instance.go
sbueringer: I think you can use `object runtime.Object`

pkg/cloud/services/networking/network.go
sbueringer: Okay let's leave it like it is for now. Although I want to fix this sooner or later, but we don't have to do it here.

sbueringer:
> @sbueringer I want to leave returning pointer issue, because functions in networking package do not return pointer.

Yup it's fine for now. I would like to fix this sooner or later but we don't have to do this in this PR (and we should double check what the correct way is before we do that)

pkg/cloud/services/networking/router.go
```diff
@@ -185,41 +186,36 @@ func setRouterExternalIPs(client *gophercloud.ServiceClient, openStackCluster *i
...
 		}).Extract()
 		if err != nil {
-			if errors.IsNotFound(err) {
-				s.logger.V(4).Info("Router Interface already removed, no actions", "id", network.Router.ID)
-				return nil
+			if capoerrors.IsNotFound(err) {
+				s.logger.V(4).Info("Router Interface already removed, no actions", "id", router.router.ID)
+				// nothing to do
 			}
 			return fmt.Errorf("unable to remove router interface: %v", err)
```
sbueringer, line 208:
```suggestion
			}
			else {
				return fmt.Errorf("unable to remove router interface: %v", err)
			}
```

sbueringer, line 208: I think we have to do it this way:
```go
		if err != nil {
			if !capoerrors.IsNotFound(err) {
				return fmt.Errorf("unable to remove router interface: %v", err)
			}
			s.logger.V(4).Info("Router Interface already removed, nothing to do", "id", router.router.ID)
		} else {
			s.logger.V(4).Info("Removed RouterInterface of Router", "id", router.router.ID)
		}
```
I think that's the best way, with the current implementation we would return an error after `// nothing to do`

pkg/cloud/services/networking/router.go
```diff
@@ -239,19 +235,76 @@ func (s *Service) getRouterInterfaces(routerID string) ([]ports.Port, error) {
...
 
-func (s *Service) existsRouter(routerID string) (bool, error) {
-	if routerID == "" {
-		return false, nil
+type Router struct {
+	router routers.Router
+	subnet subnets.Subnet
+}
+
+func (s *Service) getRouter(clusterName string) (*Router, error) {
```
sbueringer, line 243: please return the Router and Subnet in individual vars. In Go imho usually there are no struct used to just wrap return values:
```go
(routers.Router, subnet.Subnet, error)
```

pkg/cloud/services/networking/router.go
```diff
@@ -239,19 +235,76 @@ func (s *Service) getRouterInterfaces(routerID string) ([]ports.Port, error) {
...
+		return routers.Router{}, fmt.Errorf("found %d router with the name %s, which should not happen", len(routerList), routerName)
+	}
+
+	switch len(routerList) {
+	case 0:
+		return routers.Router{}, nil
+	case 1:
+		return routerList[0], nil
+	}
+	return routers.Router{}, errors.New("too many resources")
```
sbueringer, line 281: Can you move line 272 here? Because that's the same case and so you don't have to handle this case twice and check the list length
```suggestion
	return routers.Router{}, fmt.Errorf("found %d router with the name %s, which should not happen", len(routerList), routerName)
```

pkg/cloud/services/networking/router.go
sbueringer: I think I meant it differently. I added new suggestions.

pkg/cloud/services/networking/router.go
```diff
@@ -239,19 +235,76 @@ func (s *Service) getRouterInterfaces(routerID string) ([]ports.Port, error) {
...
-		return false, nil
+
+	switch len(subnetList) {
+	case 0:
+		return subnets.Subnet{}, nil
+	case 1:
+		return subnetList[0], nil
 	}
-	return true, nil
+	return subnets.Subnet{}, errors.New("too many resources")
```
sbueringer, line 305:
```suggestion
	return subnets.Subnet{}, fmt.Errorf("found %d subnets with the name %s, which should not happen", len(subnetList), subnetName)
```

sbueringer:
New round of comments. I resolved the ones that are done.

sbueringer:
/approve
from me, thx, great work! :)

@jichenjc please also review this PR before merge as it's a bit tricky to get these things right without introducing bugs

jichenjc:
@sbueringer thanks :) on my to do list now ~

sbueringer:
> @sbueringer thanks :) on my to do list now ~

Great, thx :)

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -333,38 +333,41 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
 	if err != nil {
-		return false, err
+		return err
 	}
-	allGroups, err := groups.ExtractGroups(allPages)
-	if err != nil {
-		return false, err
+	if group.ID == "" {
```
jichenjc, line 360: just curious, why we can enter here? the sec group is deleted?

pkg/cloud/services/networking/network.go
```diff
@@ -375,3 +353,11 @@ func GetSubnetsByFilter(networkClient *gophercloud.ServiceClient, opts subnets.L
 	}
 	return snets, nil
 }
+
+func getSubnetName(clusterName string) string {
```
jichenjc, line 357: those 2 function are identical , on purpose?

pkg/cloud/services/networking/securitygroups.go
hidekazuna: Yes, if security group is already deleted, we go into this line. Because getSecurityGroupByName returns &infrav1.SecurityGroup{}, nil when security group is not found.

pkg/cloud/services/networking/network.go
hidekazuna: Both getSubnetName and getNetworkName return the same string.  This is intentional. getSubnetName is used for getting subnet name. getNetworkName is used for getting network name.

jichenjc:
/lgtm
/hold cancel

thanks~

PR #850, opened 2021-04-23
Add reference in README to downloadable images

berendt:

It makes sense to point to directly usable images for this provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,11 @@ Reference images can be found in [kubernetes-sigs/image-builder](https://github.
  prerequisites in the image, you can always deploy and execute some custom scripts 
  through the [KubeadmConfig](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm).  
 
+Images provided by the community for the community built
+with [kubernetes-sigs/image-builder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi)
+based on Ubuntu are provided via [23technologies/kubernetes-capi-images](https://github.com/23technologies/kubernetes-capi-images).
+URLs to download the images can be found there in the [README](https://github.com/23technologies/kubernetes-capi-images#openstack) file.
+
 ------
 
 ## Documentation


jichenjc:
looks like we have the info here
https://image-builder.sigs.k8s.io/capi/providers/openstack.html

I didn't go to detail but I think it make sense 

@hidekazuna @sbueringer FYI

README.md
```diff
@@ -92,6 +92,10 @@ Reference images can be found in [kubernetes-sigs/image-builder](https://github.
  prerequisites in the image, you can always deploy and execute some custom scripts 
  through the [KubeadmConfig](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm).  
 
+Images built with [kubernetes-sigs/image-builder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi)
+based on Ubuntu are provided via [23technologies/kubernetes-capi-images](https://github.com/23technologies/kubernetes-capi-images).
+URLs to download the images can be found there in the [README](https://github.com/23technologies/kubernetes-capi-images#openstack) file.
+
```
hidekazuna, line 98: [23technologies/kubernetes-capi-images](https://github.com/23technologies/kubernetes-capi-images) is not "official" image. If you want "official" downloadable image, consult with [image-builder](https://github.com/kubernetes-sigs/image-builder) project, please.

berendt: Is not intended to be an official image. We just have them available and I think a reference to it makes sense that there are no prebuilt images anywhere else.

sbueringer: I think it would be good to add a link, but please with a disclaimer that these are not images officially maintainted by CAPO/ CAPI sig-cluster-lifecycle, just to avoid misunderstandings.

berendt: Added "by the community for the community". Is this sufficient?

sbueringer: I'm not sure this is clear enough, because it's not really clear what the difference between the "official" (kubernetes-sigs) CAPO/CAPI community and CAPO "user" community is.

PR #862, opened 2021-05-07
🏃 Add some events and refactor related to instance

hidekazuna:

Add events
- Delete attachinterface, trunk, and port
- Create port and trunk
- Replace all attributestags

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -64,8 +64,7 @@ const (
 	TimeoutInstanceDelete = 5 * time.Minute
 )
 
-// InstanceCreate creates a compute instance.
-func (s *Service) InstanceCreate(openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, clusterName string, userData string) (instance *infrav1.Instance, err error) {
+func (s *Service) CreateInstance(openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, clusterName string, userData string) (instance *infrav1.Instance, err error) {
 	if openStackMachine == nil {
 		return nil, fmt.Errorf("create Options need be specified to create instace")
 	}
@@ -142,50 +141,36 @@ func (s *Service) InstanceCreate(openStackCluster *infrav1.OpenStackCluster, mac
 	}
 	input.Networks = &nets
 
-	out, err := createInstance(s, clusterName, input)
+	out, err := createInstance(s, openStackMachine, clusterName, input)
 	if err != nil {
-		record.Warnf(openStackMachine, "FailedCreateServer", "Failed to create server %s: %v", input.Name, err)
 		return nil, err
 	}
-	record.Eventf(openStackMachine, "SuccessfulCreateServer", "Created server %s with id %s", out.Name, out.ID)
 	return out, nil
 }
 
-func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error) {
-	// Get image ID
-	imageID, err := getImageID(is, i.Image)
-	if err != nil {
-		return nil, fmt.Errorf("create new server err: %v", err)
-	}
-
+func createInstance(is *Service, eventObject runtime.Object, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error) {
 	accessIPv4 := ""
-	networkList := i.Networks
-	portsList := []servers.Network{}
-	for _, network := range *networkList {
-		network := network
+	portList := []servers.Network{}
+
+	for _, network := range *i.Networks {
 		if network.ID == "" {
 			return nil, fmt.Errorf("no network was found or provided. Please check your machine configuration and try again")
 		}
-		allPages, err := ports.List(is.networkClient, ports.ListOpts{
-			Name:      i.Name,
-			NetworkID: network.ID,
-		}).AllPages()
+
+		port, err := getOrCreatePort(is, eventObject, clusterName, i.Name, network, i.SecurityGroups)
 		if err != nil {
-			return nil, fmt.Errorf("searching for existing port for server: %v", err)
+			return nil, err
 		}
-		portList, err := ports.ExtractPorts(allPages)
-		if err != nil {
-			return nil, fmt.Errorf("searching for existing port for server err: %v", err)
-		}
-		var port ports.Port
-		if len(portList) == 0 {
-			// create server port
-			port, err = createPort(is, clusterName, i.Name, &network, i.SecurityGroups)
+
+		if i.Trunk {
+			trunk, err := getOrCreateTrunk(is, eventObject, i.Name, port.ID)
 			if err != nil {
-				return nil, fmt.Errorf("failed to create port err: %v", err)
+				return nil, err
 			}
-		} else {
-			port = portList[0]
+
+			if err = replaceAllAttributesTags(is, eventObject, trunk.ID, i.Tags); err != nil {
+				return nil, err
+			}
 		}
 
 		for _, fip := range port.FixedIPs {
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -142,50 +137,37 @@ func (s *Service) InstanceCreate(openStackCluster *infrav1.OpenStackCluster, mac
...
-
+func createInstance(is *Service, object runtime.Object, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error) {
 	accessIPv4 := ""
-	networkList := i.Networks
-	portsList := []servers.Network{}
-	for _, network := range *networkList {
+	portList := []servers.Network{}
+
+	for _, network := range *i.Networks {
 		network := network
```
chrischdi, line 152: If I'm not mssing anything then we could also get rid of this line?

pkg/cloud/services/compute/instance.go
```diff
@@ -482,47 +381,83 @@ func getServerNetworks(networkClient *gophercloud.ServiceClient, networkParams [
...
+	if err != nil {
+		record.Warnf(object, "FailedCreateTrunk", "Failed to create trunk %s: %v", trunkName, err)
+		return nil, err
+	}
+
+	record.Eventf(object, "SuccessfulCreateTrunk", "Created trunk %s with id %s", trunk.Name, trunk.ID)
+	return trunk, nil
+}
+
+func replaceAllAttributestag(is *Service, object runtime.Object, trunkID string, tags []string) error {
```
chrischdi, line 451: Is this func is specific to trunks? If so I think we should add some context to the function name, otherwise we could rename `trunkID` to `resourceID` like it is named in the `ReplaceAll` func

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +530,148 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
...
+	return deleteInstance(s, object, instance.ID)
+}
+
+func deletePort(is *Service, object runtime.Object, portID string) error {
+	port, err := ports.Get(is.networkClient, portID).Extract()
+	if err != nil {
+		if capoerrors.IsNotFound(err) {
+			return nil
+		}
+	}
```
chrischdi, line 557: Shouldn't the error still get returned if it is not `IsNotFound`?
```suggestion
		return err
	}
```

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +530,148 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
...
+	if instance == nil || instance.ID == "" {
+		return nil
+	}
+
+	port, err := ports.Get(is.networkClient, portID).Extract()
+	if err != nil {
+		if capoerrors.IsNotFound(err) {
+			return nil
+		}
+	}
```
chrischdi, line 602: Same here: houldn't the error still get returned if it is not `IsNotFound`?
```suggestion
		return err
	}
```

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +530,148 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
...
+	return nil
+}
+
+func deleteTrunk(is *Service, object runtime.Object, portID string) error {
+	port, err := ports.Get(is.networkClient, portID).Extract()
+	if err != nil {
+		if capoerrors.IsNotFound(err) {
+			return nil
+		}
+	}
```
chrischdi, line 623: Same here: houldn't the error still get returned if it is not `IsNotFound`?
```suggestion
		return err
	}
```

chrischdi (commented):
Hey,

I did take a look on the PR, I hope this helps :-)

pkg/cloud/services/compute/instance.go
hidekazuna: Without this, I encountered the following error by `make lint`
```
pkg/cloud/services/compute/instance.go:157:65: G601: Implicit memory aliasing in for loop. (gosec)
                port, err := getOrCreatePort(is, object, clusterName, i.Name, &network, i.SecurityGroups)
```

pkg/cloud/services/compute/instance.go
hidekazuna: I want to express the context by function naming `replaceAllAttributestag` and argument name `trunkID`.

sbueringer:
@chrischdi please ping me when you're happy, then I'll also take a look :)

pkg/cloud/services/compute/instance.go
chrischdi: I don't know if this would be a good pattern. If there later is a code review and someone does a `replaceAllAttributestag(is, object, port.ID, i.Tags)` the context would be gone.

Also:
If we keep the name, it should be called `AttributesTags` (upper-case T, and s), just the package name is all lowercase and inside the gophercloud package it is also called `AttributesTags`

hidekazuna: @chrischdi So you want to rename to `replaceAllAttributesTagOfTrunk` or something?
I agree with upper-case T. This function name should start with small letter, since this function is not public.

chrischdi: I'd prefer something like that, however I would be also ok with `replaceAllAttributesTag` if this is still your favourite :relieved:

pkg/cloud/services/compute/instance.go
hidekazuna: Done

pkg/cloud/services/compute/instance.go
hidekazuna: Renamed `replaceAllAttributestag` to `replaceAllAttributesTags`. Sorry forgot to say ending with small s.

chrischdi:
/lgtm

@sbueringer ping

pkg/cloud/services/compute/instance.go
```diff
@@ -194,54 +175,31 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
...
+	if err != nil {
+		return nil, err
+	}
+	return instance, nil
+}
+
+func createServer(is *Service, object runtime.Object, portList []servers.Network, accessIPv4 string, i *infrav1.Instance) (*infrav1.Instance, error) {
+	imageID, err := getImageID(is, i.Image)
+	if err != nil {
+		return nil, fmt.Errorf("create new server err: %v", err)
```
sbueringer, line 200:
```suggestion
		return nil, fmt.Errorf("create new server: %v", err)
```
Can you drop those `err` strings everywhere, please?

pkg/cloud/services/compute/utils.go
```diff
@@ -0,0 +1,132 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package compute
+
+import (
```
sbueringer, line 19: I don't really see the benefit of moving this to a separate file if those funcs are all only used by instance.go

pkg/cloud/services/compute/instance.go
```diff
@@ -194,54 +175,31 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
...
-				return nil, fmt.Errorf("tagging trunk for server err: %v", err)
-			}
-		}
 	}
 
 	if i.Subnet != "" && accessIPv4 == "" {
-		if errd := deletePorts(is, portsList); errd != nil {
-			return nil, fmt.Errorf("no ports with fixed IPs found on Subnet %q: error cleaning up ports: %v", i.Subnet, errd)
+		if errd := deletePorts(is, object, portList); errd != nil {
+			return nil, errd
```
sbueringer, line 185:
```suggestion
		if err := deletePorts(is, object, portList); err != nil {
			return nil, err
```

pkg/cloud/services/compute/instance.go
```diff
@@ -194,54 +175,31 @@ func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infr
...
 		}
 		return nil, fmt.Errorf("no ports with fixed IPs found on Subnet %q", i.Subnet)
 	}
 
+	instance, err := createServer(is, object, portList, accessIPv4, i)
```
sbueringer, line 190: Does it really make sense to split this func? I don't see a good semantic difference between createInstance and createServer and createServer is only ever called by createInstance

pkg/cloud/services/compute/instance.go
```diff
@@ -142,50 +137,36 @@ func (s *Service) InstanceCreate(openStackCluster *infrav1.OpenStackCluster, mac
...
 }
 
-func createInstance(is *Service, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error) {
-	// Get image ID
-	imageID, err := getImageID(is, i.Image)
-	if err != nil {
-		return nil, fmt.Errorf("create new server err: %v", err)
-	}
-
+func createInstance(is *Service, object runtime.Object, clusterName string, i *infrav1.Instance) (*infrav1.Instance, error) {
```
sbueringer, line 147: Hm object is too generic imho as a variable name. Can we call it at least `eventObject` (in case that's the only context in which it's used)

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +529,164 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
 	}
 	// get and delete trunks
 	for _, port := range instanceInterfaces {
-		err := attachinterfaces.Delete(s.computeClient, instance.ID, port.PortID).ExtractErr()
-		if err != nil {
+		if err = deleteAttachinterface(s, object, instance.ID, port.PortID); err != nil {
```
sbueringer, line 532:
```suggestion
		if err = deleteAttachInterface(s, object, instance.ID, port.PortID); err != nil {
```

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +529,164 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
...
+	if port == nil {
+		return nil
+	}
+
+	err = attachinterfaces.Delete(is.computeClient, instanceID, portID).ExtractErr()
+	if err != nil {
+		if capoerrors.IsNotFound(err) {
+			return nil
+		}
+		record.Warnf(object, "FailedDeleteAttachinterface", "Failed to delete attachinterface: instance %s, port %s: %v", instance.ID, port.ID, err)
```
sbueringer, line 609:
```suggestion
		record.Warnf(object, "FailedDeleteAttachInterface", "Failed to delete attachinterface: instance %s, port %s: %v", instance.ID, port.ID, err)
```

pkg/cloud/services/compute/instance.go
```diff
@@ -596,61 +529,164 @@ func (s *Service) DeleteInstance(object runtime.Object, instanceName string) err
...
+		return err
+	}
+
+	record.Eventf(object, "SuccessfulDeleteAttachinterface", "Deleted attachinterface: instance %s, port %s", instance.ID, port.ID)
```
sbueringer, line 613:
```suggestion
	record.Eventf(object, "SuccessfulDeleteAttachInterface", "Deleted attachInterface: instance %s, port %s", instance.ID, port.ID)
```

pkg/cloud/services/compute/utils.go
hidekazuna: OK, let's go back to instance.go.

pkg/cloud/services/compute/instance.go
hidekazuna: Thanks question. That is What I considered. If `createInstance` is "business logic" which means call `getOrCreatePort`, and if trunk true, call `getOrCreateTrunk` then `createServer`, current is better. Otherwise, we should not split function, `createInstance` includes `createServer` as main logic. I got your suggestion and will not split function.

pkg/cloud/services/compute/instance.go
sbueringer: @hidekazuna Whad to you think about this one?

hidekazuna: @sbueringer Sorry I missed, updated.

sbueringer:
@hidekazuna Imho you can squash + merge now

hidekazuna:
> @hidekazuna Imho you can squash + merge now

@sbueringer Thanks, Let7s squash and merge.

hidekazuna:
> > @hidekazuna Imho you can squash + merge now
> 
> @sbueringer Thanks, Let7s squash and merge. again lgtm, please.

PR #867, opened 2021-05-13
Add more logs for LB creation

jichenjc:

helpful for me to debug some LB issues...

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -50,7 +50,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 		return err
 	}
 	if lb == nil {
-		s.logger.Info("Creating load balancer", "name", loadBalancerName)
+		s.logger.Info(fmt.Sprintf("Creating load balancer in subnet: %q", openStackCluster.Status.Network.Subnet.ID), "name", loadBalancerName)
 		lbCreateOpts := loadbalancers.CreateOpts{
 			Name:        loadBalancerName,
 			VipSubnetID: openStackCluster.Status.Network.Subnet.ID,
@@ -90,7 +90,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return err
 		}
 		if listener == nil {
-			s.logger.Info("Creating load balancer listener", "name", lbPortObjectsName)
+			s.logger.Info("Creating load balancer listener", "name", lbPortObjectsName, "lb-id", lb.ID)
 			listenerCreateOpts := listeners.CreateOpts{
 				Name:           lbPortObjectsName,
 				Protocol:       "TCP",
@@ -116,7 +116,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return err
 		}
 		if pool == nil {
-			s.logger.Info("Creating load balancer pool", "name", lbPortObjectsName)
+			s.logger.Info(fmt.Sprintf("Creating load balancer pool for listener %q", listener.ID), "name", lbPortObjectsName, "lb-id", lb.ID)
 			poolCreateOpts := pools.CreateOpts{
 				Name:       lbPortObjectsName,
 				Protocol:   "TCP",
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -50,7 +50,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 		return err
 	}
 	if lb == nil {
-		s.logger.Info("Creating load balancer", "name", loadBalancerName)
+		s.logger.Info("Creating load balancer", "name", loadBalancerName, "vipsubnetid", openStackCluster.Status.Network.Subnet.ID)
```
sbueringer, line 53: I'm not sure if we should put all this data in key/value pairs. I think this makes mostly sense for fields where you can filter for later on through tools like Kibana/Loki. So I would only add the loadbalancer id (with key "id") as key value pair and the other fields into the message itself (although I would add the lb id to all those log lines so it's easy to filter on logs for this specific lb later on)

jichenjc: ok, works for me , I only find it's really hard to find corresponding logs ..

jichenjc: this one is for VM creation .. so the param is helpful for me to know what's given as param

jichenjc: the others I only add the ID per suggestion , please let me know your opinoin

sbueringer: I'm not against adding this info. I just wouldn't add it as key value pair :) I make an example:

```suggestion
		s.logger.Info(fmt.Sprintf("Creating load balancer in subnet: %q", openStackCluster.Status.Network.Subnet.ID), "name", loadBalancerName)
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -90,7 +90,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return err
 		}
 		if listener == nil {
-			s.logger.Info("Creating load balancer listener", "name", lbPortObjectsName)
+			s.logger.Info("Creating load balancer listener", "name", lbPortObjectsName, "lbid", lb.ID)
```
sbueringer, line 93:
```suggestion
			s.logger.Info("Creating load balancer listener", "name", lbPortObjectsName, "lb-id", lb.ID))
```
I think `-` is the convention. Wanted to open an issue soon for that, just didn't get to it.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -116,7 +116,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return err
 		}
 		if pool == nil {
-			s.logger.Info("Creating load balancer pool", "name", lbPortObjectsName)
+			s.logger.Info("Creating load balancer pool", "name", lbPortObjectsName, "listenerID", listener.ID)
```
sbueringer, line 119:
```suggestion
			s.logger.Info(fmt.Sprintf("Creating load balancer pool to listener %q", listener.ID), "name", lbPortObjectsName, "lb-id", lb.ID)
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -138,7 +138,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return err
 		}
 		if monitor == nil {
-			s.logger.Info("Creating load balancer monitor", "name", lbPortObjectsName)
+			s.logger.Info("Creating load balancer monitor", "name", lbPortObjectsName, "poolID", pool.ID)
```
sbueringer, line 141:
```suggestion
			s.logger.Info(fmt.Sprintf("Creating load balancer monitor for pool %q", pool.ID), "name", lbPortObjectsName, "lb-id", lb.ID)
```

pkg/cloud/services/loadbalancer/loadbalancer.go
sbueringer: Please see my other suggestions below. Something like this would make it possible to filter on the lb id to get all logs realted to as specific lb. If logs related to a specific listener should be found that can be still done by matching on the message.

I just think we shouldn' introduce arbitrary "dimensions" to our logs. I think key/value pairs in logs can be seen simlar to labels in metrics.

pkg/cloud/services/loadbalancer/loadbalancer.go
sbueringer: Hm we should either drop the `lb-` prefix from the lb-id or should prefix the name accordingly. Can we try to find out what the best practice is?

sbueringer (approved):

PR #872, opened 2021-05-20
🏃 Validate OpenStackMachineTemplate spec.template.spec immutability

### LINKED ISSUE #858: Block updates on the Template resources OpenStackMachineTemplate
/kind feature

Issue is based on this Slack thread: https://app.slack.com/client/T09NY5SBT/C8TSNPY4T/thread/C8TSNPY4T-1618296322.489700
Rest is (mostly) copied from CAPZ

As documented in https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html

> These MachineTemplate resources should be immutable, unless the infrastructure provider documentation indicates otherwise for certain fields (see below for more details).
> Some infrastructure providers may, at their discretion, choose to support in-place modifications of certain infrastructure machine template fields. This may be useful if an infrastructure provider is able to make changes to running instances/machines, such as updating allocated memory or CPU capacity. In such cases, however, Cluster API will not trigger a rolling update.

Currently, updates to OpenStackMachineTemplate are allowed which is misleading as it results in no changes being applied to the running OpenStackMachines and no rolling update being triggered. We should block updates and instruct the user to create a new OpenStackMachineTemplate instead + point to the doc above in the error message.

How: add a validation webhook and reject all updates to OpenStackMachineTemplate.

xref:
* Azure issue: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1255
* Azure PR: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1275
* Slack thread: https://app.slack.com/client/T09NY5SBT/C8TSNPY4T/thread/C8TSNPY4T-1618296322.489700

johannesfrey:
Signed-off-by: Johannes Frey <johannes.frey@daimler.com>

This PR blocks changes to `OpenStackMachineTemplate.Spec.Template.Spec` and provides a message that a new resource should be created if the user does try to mutate it.

Credits where credit's due. This is mainly copy-pasted from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1275 :slightly_smiling_face:.

Fixes #858

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [X] adds unit tests

/hold

<sub>Johannes Frey [johannes.frey@daimler.com](mailto:johannes.frey@daimler.com) Daimler TSS GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>

### DIFF
--- a/api/v1alpha4/openstackmachinetemplate_webhook.go
+++ b/api/v1alpha4/openstackmachinetemplate_webhook.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1alpha4
 
 import (
-	"errors"
 	"reflect"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -26,6 +25,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
+
+// OpenStackMachineTemplateImmutableMsg ...
+const OpenStackMachineTemplateImmutableMsg = "OpenStackMachineTemplate spec.template.spec field is immutable. Please create a new resource instead. Ref doc: https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html"
 
 func (r *OpenStackMachineTemplate) SetupWebhookWithManager(mgr manager.Manager) error {
 	return builder.WebhookManagedBy(mgr).
@@ -51,9 +53,17 @@ func (r *OpenStackMachineTemplate) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *OpenStackMachineTemplate) ValidateUpdate(old runtime.Object) error {
+	var allErrs field.ErrorList
 	oldOpenStackMachineTemplate := old.(*OpenStackMachineTemplate)
-	if !reflect.DeepEqual(r.Spec, oldOpenStackMachineTemplate.Spec) {
-		return errors.New("openstackMachineTemplateSpec is immutable")
+
+	if !reflect.DeepEqual(r.Spec.Template.Spec, oldOpenStackMachineTemplate.Spec.Template.Spec) {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "template", "spec"), r, OpenStackMachineTemplateImmutableMsg),
+		)
+	}
+
+	if len(allErrs) != 0 {
+		return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 	}
 
 	return nil
--- a/api/v1alpha4/openstackmachinetemplate_webhook_test.go
+++ b/api/v1alpha4/openstackmachinetemplate_webhook_test.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


api/v1alpha4/openstackmachinetemplate_webhook.go
```diff
@@ -51,9 +53,17 @@ func (r *OpenStackMachineTemplate) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *OpenStackMachineTemplate) ValidateUpdate(old runtime.Object) error {
+	var allErrs field.ErrorList
 	oldOpenStackMachineTemplate := old.(*OpenStackMachineTemplate)
-	if !reflect.DeepEqual(r.Spec, oldOpenStackMachineTemplate.Spec) {
-		return errors.New("openstackMachineTemplateSpec is immutable")
```
jichenjc, line 56: um...  looks like we did this kind of check before here? @sbueringer from your original issue
seems we still allow such changes before this PR? or we return wrong error ?

sbueringer: I actually wasn't aware that we already had this in place :)

But I think the new version is nicer and probably in line with other providers @johannesfrey ?

jichenjc: Yes I agree this is better :)

the question is whether we had problem before (by using existing code)? if so then with new PR, the problem might not be fixed?

sbueringer: No it's completely fine imho. I'll test if it actually works. My issue was not triggered because I saw the problem in practice. I had a discussion the cluster-api Slack and then I thouht "well we also need this", and wasn't aware that we already had an implementation for it.

johannesfrey: The only change is that with this PR we don't compare the `OpenStackMachineTemplate.Spec` anymore but rather `OpenStackMachineTemplate.Spec.Template.Spec`. This is also done in [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1275/files#diff-5acce21862b7c3e7b1ca6b85ba34ae2a798a488972d16ea6275655beb1053fc0R58). Is this correct or should we rather block changes to `OpenStackMachineTemplate.Spec` like before?

sbueringer: @johannesfrey Good point. I'll take a closer look next week and try to find out

jichenjc: I think we are using deep copy anyway?

jichenjc: I gave a try here;

```
root@jitest19:~/abc# kubectl edit openstackmachinetemplate
openstackmachinetemplate.infrastructure.cluster.x-k8s.io/capi-openstack-control-plane skipped
error: openstackmachinetemplates.infrastructure.cluster.x-k8s.io "capi-openstack-md-0" could not be patched: admission webhook "validation.openstackmachinetemplate.infrastructure.x-k8s.io" denied the request: openstackMachineTemplateSpec is immutable
You can run `kubectl replace -f /tmp/kubectl-edit-fh9fo.yaml` to try this update again.
root@jitest19:~/abc#
```

sbueringer: @jichenjc what does deepcopy have to do with this PR? :)

jichenjc: @sbueringer we are using deep copy of openstackmachinetemplate.Spec before ? 
so actually previous code works fine as I showed above... so this PR's commit msg is not accurate
maybe an improvement on the wording but not provide immutability enhancement as we already have it..

sbueringer: I guess you mean deepequal?

jichenjc: my bad .. yes, I thought deepequal but typed deepcopy ..

PR #863, opened 2021-05-08
Add metrics of gophercloud (POST,UPDATE and DELETE) actions

jichenjc:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of #856 

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #876, opened 2021-05-21
✨ Add feature to create ports with custom options

### LINKED ISSUE #788: add the possibility to create custom ports

macaptain:
This commit adds a new field to the v1alpha4 API: OpenStackMachineTemplateSpec.ports.

The list of ports are added per instance. Each port may be customized with options. These ports are created in addition to any ports created in the `networks:` field.

If `networks:` is not specified, only the ports specified in `ports:` will be created and attached to the instance. If neither `networks:` nor `ports:` are specified, the instance will be connected to the default cluster network and subnet.

This feature is very much based on the work on @jsen-, and a lot of credit goes there for the implementation:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/778

Example spec to give you some idea. Creates exactly three ports, two ports on the same network on different subnets with different VNIC types and a further example showing how to add security groups.

```
metadata:
  name: basic-1-control-plane
  namespace: default
spec:
  template:
    spec:
      ports:
      - networkId: 2f796e80-65e4-4589-b92d-77a6c019fe35
        description: primary    
        vnicType: normal
        fixedIps:
        - subnetId: 74ba4ed6-239a-44f4-81ac-95cc65b540b3
          ipAddress: 192.168.1.35
      - networkId: 2f796e80-65e4-4589-b92d-77a6c019fe35
        description: direct    
        vnicType: direct
        fixedIps:
        - subnetId: b6e81653-2bd6-4334-ab93-a7c7ec8bc66f
          ipAddress: 10.10.10.10
      - networkId: e6810d0c-85c5-41ce-b1d4-c2795611b9a0
        description: secondary
        vnicType: direct
        securityGroups:
        - one
        - two
      cloudName: openstack
      cloudsSecret:
        name: basic-1-cloud-config
      flavor: 2C-4GB-100GB
      image: Ubuntu_20.04_node
```

I'd like to be able to have instances on different networks with different settings per port. In particular, customizable VNIC types and customizable security groups.

Fixes #788

This PR is about trying to get https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/778 merged to master, so firstly, huge credit + thanks to @jsen- for the implementation (plus the OpenShift fork was a very helpful reference so thanks to those working on that too). I've taken on board the comments in the original PR and:
* Rebased on master after #862 was merged
* Kept the API camel case instead of using gophercloud's structs
* Fixed the API generation warnings from v1alpha4 to v1alpha3

In addition, I've changed the behavior slightly with regard to defaults:
* Usually if `networks:` isn't specified, you get your port on the default cluster network. Now if neither `networks:` nor `ports:` is specified, you get this default.
* For a port, the field `NetworkID` used to be required. Now if unspecified, it defaults to the default cluster network.
* For a port, the field `NameSuffix` used to be mandatory. Now if unspecified, it defaults to the 0-based index of the port of the instance.

I've tried to keep the API consistent with this project, but also similar to the OpenShift fork, maybe @iamemilio could comment on that.

**TODOs**:

- [x] squashed commits
- [x] verify that the API change meets all my requirements
- [x] manual testing
- [ ] add e2e tests: TODO in follow-up PR
- [x] rebase after #871 is merged
- if necessary:
  - [ ] includes documentation: TODO in follow-up PR
  - [x] adds unit tests

/hold


pkg/cloud/services/compute/instance.go
```diff
@@ -376,40 +409,71 @@ func getServerNetworks(networkClient *gophercloud.ServiceClient, networkParams [
...
+		}
 
-	if len(portList) != 0 {
-		return &portList[0], nil
-	}
+		// inherit port security groups from the instance if not explicitly specified
+		securityGroups := net.PortOpts.SecurityGroups
+		if securityGroups == nil {
+			securityGroups = instanceSecurityGroups
+		}
```
hidekazuna, line 434: Can this logic move to before 126? it is easier to understand constructing security groups(from line 114-126) then construct nets(from line 128-164).

pkg/cloud/services/compute/instance.go
```diff
@@ -376,40 +409,71 @@ func getServerNetworks(networkClient *gophercloud.ServiceClient, networkParams [
...
+		record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+		return port, nil
+	} else if len(existingPorts) == 1 {
+		return &existingPorts[0], nil
 	}
 
-	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
-	return port, nil
+	return nil, fmt.Errorf("multiple ports found with name \"%s\"", portName)
```
hidekazuna, line 476: Changing `len(existingPorts)` order is easier to understand.

```go
if len(existingPorts) == 1 {
  return &existingPorts[0], nil
}

if len(existingPorts) > 1 {
  return nil, fmt.Errorf("multiple ports found with name \"%s\"", portName)
}

// create port
description := net.PortOpts.Description
...
```

pkg/cloud/services/compute/instance.go
macaptain: The security groups can be set on a port and can be different per port, so I don't think we can put this before 126, before we've looped through the nets and ports in the spec.

SecurityGroups could be added as a field to the Network struct and we could set this default in lines 128-164, instead of in `getOrCreatePort`. I wonder also if constructing nets should be pulled into a separate function, as I agree it takes some understanding. Let me make those changes and you can tell me what you think.

hidekazuna: Thanks, I understand we can not move this before 126. Please make a separate function as you think.

api/v1alpha4/types.go
```diff
@@ -116,6 +116,32 @@ type SubnetFilter struct {
...
+type PortOpts struct {
+	// ID of the OpenStack network on which to create the port. If unspecified, create the port on the default cluster network.
+	NetworkID string `json:"networkId,omitempty"`
+	// Used to make the name of the port unique. If unspecified, instead the 0-based index of the port in the list is used.
+	NameSuffix   string `json:"nameSuffix,omitempty"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"adminStateUp,omitempty"`
+	MACAddress   string `json:"macAddress,omitempty"`
+	// Specify pairs of subnet and/or IP address. These should be subnets of the network with the given NetworkID.
+	FixedIPs       []FixedIP `json:"fixedIps,omitempty"`
```
sbueringer, line 128:
```suggestion
	FixedIPs       []FixedIP `json:"fixedIPs,omitempty"`
```

sbueringer: I now JSON capitalization is a mess in CAPO. I think I would capitalize IP for now and probably fix all the `Id`s in a follow-up issue. (at least we're still in alpha so we can fix things like this :))

pkg/cloud/services/compute/instance.go
```diff
@@ -152,12 +174,13 @@ func createInstance(is *Service, eventObject runtime.Object, clusterName string,
 	accessIPv4 := ""
 	portList := []servers.Network{}
 
-	for _, network := range *i.Networks {
+	for idx, network := range *i.Networks {
```
sbueringer, line 177:
```suggestion
	for i, network := range *i.Networks {
```
nit: I think I would prefer i as that's usually used

pkg/cloud/services/compute/instance.go
```diff
@@ -251,6 +274,15 @@ func createInstance(is *Service, eventObject runtime.Object, clusterName string,
 	return instance, nil
 }
 
+// Different ports must have unique names.
```
sbueringer, line 277: This godoc comment has to start with `getPortName` per convention

sbueringer (commented):
I gave it a first high-level review, overall lgtm.

 @iamemilio if you have some time to spare to review this, that would be great :)

pkg/cloud/services/compute/instance.go
```diff
@@ -376,38 +409,73 @@ func getServerNetworks(networkClient *gophercloud.ServiceClient, networkParams [
...
 
-	portCreateOpts := ports.CreateOpts{
+	if len(existingPorts) > 1 {
+		return nil, fmt.Errorf("multiple ports found with name \"%s\"", portName)
+	}
+
+	// no port found, so create the port
+	description := net.PortOpts.Description
+	if description == "" {
+		description = fmt.Sprintf("Created by cluster-api-provider-openstack cluster %s", clusterName)
```
sbueringer, line 436: Should be adjusted once: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/870/files is merged (which should be soon)

pkg/cloud/services/compute/instance.go
```diff
@@ -131,7 +132,28 @@ func (s *Service) CreateInstance(openStackCluster *infrav1.OpenStackCluster, mac
 		if err != nil {
 			return nil, err
 		}
-	} else {
+	}
+	if len(openStackMachine.Spec.Ports) > 0 {
+		for _, port := range openStackMachine.Spec.Ports {
```
iamemilio, line 137: You can actually skip the `len(openStackMachine.Spec.Ports) > 0` check here, since your for loop will take that into consideration anyway. 

https://play.golang.org/p/Ry8T8bcDGmo

api/v1alpha4/types.go
```diff
@@ -116,6 +116,32 @@ type SubnetFilter struct {
...
+	TenantID       string    `json:"tenantId,omitempty"`
+	ProjectID      string    `json:"projectId,omitempty"`
+	SecurityGroups *[]string `json:"securityGroups,omitempty"`
+
```
iamemilio, line 132: Could we include allowedAddressPairs as well? 

Examples of Implementation:
https://github.com/openshift/cluster-api-provider-openstack/blob/3024c78c7026e804f8c4b5765c652110f6df4d7a/pkg/apis/openstackproviderconfig/v1alpha1/types.go#L211
https://github.com/openshift/cluster-api-provider-openstack/blob/3024c78c7026e804f8c4b5765c652110f6df4d7a/pkg/apis/openstackproviderconfig/v1alpha1/types.go#L226-L229
https://github.com/openshift/cluster-api-provider-openstack/blob/3024c78c7026e804f8c4b5765c652110f6df4d7a/pkg/cloud/openstack/clients/machineservice.go#L348-L366

pkg/cloud/services/compute/instance.go
macaptain: The only issue is that `i` is already taken for the instance variable name. I'll use `i` for the index and rename the instance variable to `instance`.

api/v1alpha4/types.go
macaptain: Yes, of course, I'll add `allowedAddressPairs` too.

pkg/cloud/services/compute/instance.go
sbueringer: Ah, I would prefer renaming the instance. `i` is a really short variable name for a variable with that large of a scope

sbueringer:
@macaptain I think we will get https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/863 and https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/870 merged soon, then you're a bit safer against rebase conflicts

pkg/cloud/services/compute/instance.go
macaptain: That sounds good. `instance` is already used in this scope for something else, but I'll rename the loop variable to `i` and come up with a sensible name for the instance variable currently called `i`.

macaptain:
> @macaptain I think we will get #863 and #870 merged soon, then you're a bit safer against rebase conflicts

Sure, no problem, I don't mind rebasing when it's needed.

sbueringer:
> > @macaptain I think we will get #863 and #870 merged soon, then you're a bit safer against rebase conflicts
> 
> Sure, no problem, I don't mind rebasing when it's needed.

Yup, just wanted to let you know in case you want to wait a day or so with the rebase so you only have to do it once. Not sure how problematic the conflicts will be.

hidekazuna:
@macaptain we see stacktrace in https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/876/pull-cluster-api-provider-openstack-e2e-test/1398202842900074496/artifacts/clusters/bootstrap/controllers/capo-controller-manager/capo-controller-manager-7598cb9d55-lbhgd/manager.log

We have to care about [CreateBastion method](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/bastion.go#L25-L68) as well.

macaptain:
> @macaptain we see stacktrace in https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/876/pull-cluster-api-provider-openstack-e2e-test/1398202842900074496/artifacts/clusters/bootstrap/controllers/capo-controller-manager/capo-controller-manager-7598cb9d55-lbhgd/manager.log
> 
> We have to care about [CreateBastion method](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/bastion.go#L25-L68) as well.
> 
> 

Thanks, I'll have a look at this today and try to fix up this failure.

api/v1alpha3/conversion.go
```diff
@@ -118,3 +117,15 @@ func Convert_v1alpha3_OpenStackClusterSpec_To_v1alpha4_OpenStackClusterSpec(in *
 func Convert_v1alpha3_OpenStackMachineSpec_To_v1alpha4_OpenStackMachineSpec(in *OpenStackMachineSpec, out *v1alpha4.OpenStackMachineSpec, s conversion.Scope) error {
 	return autoConvert_v1alpha3_OpenStackMachineSpec_To_v1alpha4_OpenStackMachineSpec(in, out, s)
 }
+
+// Convert_v1alpha3_OpenStackClusterSpec_To_v1alpha4_OpenStackClusterSpec has to be added by us for the new portOpts
+// parameter in v1alpha4. There is no intention to support this parameter in v1alpha3, so the field is just dropped.
+func Convert_v1alpha4_Network_To_v1alpha3_Network(in *v1alpha4.Network, out *Network, s conversion.Scope) error {
```
jan-est, line 123: I think this should be like it says in the comment `Convert_v1alpha3_OpenStackClusterSpec_To_v1alpha4_OpenStackClusterSpec`, I think we don't need specific conversion function for Network since it is part of the `OpenStackMachineSpec`.

I am pretty sure when changing between api versions you just can't drop any fields that are not supported in another. You need to store and restore these field during conversion.

pkg/cloud/services/compute/instance.go
macaptain: I've updated to use this new function to get the default description (if none is provided in the port options).

api/v1alpha3/conversion.go
macaptain: I see that we'll drop fields if we go v1alpha4 -> v1alpha3 -> v1alpha4. This will also apply to other fields added/removed in v1alpha4 that aren't part of this change (e.g. .UserDataSecret, useOctavia). Can we raise and fix robust conversion between versions as a separate issue outside the scope of this change?

Happy to accept any suggestions on improving or fixing these conversion functions that I have written, because I have largely guessed at what they are supposed to be in order to get the code generation passing without warnings or errors.

iamemilio:
lgtm :)

hidekazuna:
@macaptain Is this still WIP?

api/v1alpha4/types.go
```diff
@@ -116,6 +116,38 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type PortOpts struct {
```
jichenjc, line 119: as follow up, we may need consider add some samples in how to use this opts (not this PR)

macaptain:
> @macaptain Is this still WIP?

I've taken WIP off the PR. I've been testing this manually and I'm happy with the implementation. The feature meets my requirements.

The only question for me is would you prefer I use 'Filters' for NetworkId / SubnetId / SecurityGroups so that these fields can be chosen by details other than Id. If you want, I can look at implementing that before the API gets fixed in place. If not, let's look to merge this as it is.

Thanks very much!

api/v1alpha4/types.go
macaptain: I can certainly add additional documentation in a follow up PR.

api/v1alpha4/types.go
```diff
@@ -116,6 +116,38 @@ type SubnetFilter struct {
...
+
+	// The ID of the host where the port is allocated
+	HostID string `json:"hostId,omitempty"`
+
+	// The virtual network interface card (vNIC) type that is bound to the neutron port.
+	VNICType string `json:"vnicType,omitempty"`
+}
+
+type FixedIP struct {
+	SubnetID  string `json:"subnetId"`
```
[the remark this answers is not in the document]
iamemilio: That would be cool. That is probably a good idea for a lot of these fields that only take IDs as an input

hidekazuna:
> The only question for me is would you prefer I use 'Filters' for NetworkId / SubnetId / SecurityGroups so that these fields can be chosen by details other than Id. If you want, I can look at implementing that before the API gets fixed in place. If not, let's look to merge this as it is.

I do not need filters for these fields. Another PR would be proposed if someone wants 😃 

/approve

hidekazuna:
@sbueringer @jichenjc PTAL

jichenjc:
/lgtm

let's merge this and continue evolution

PR #886, opened 2021-05-31
:running: Install binaries required for running tests

### LINKED ISSUE #887: Add missing binaries for envtest
For the envtest to usable, the ```etcd, kubectl and kube-apiserver``` binaries are required. As we have it now, none of the make targets install the binaries.
Due to this issue, the current test suit is skipped and none of the tests could run, preventing creation of new controllers’ tests.

Therefore, adding the following is beneficial.

1. Add a make target to install the missing binaries
2. Change the ```test``` target taking these binaries into account
3. Add a ```test-cover``` target

/kind feature

Xenwar:

Install binaries required for running tests and enables skipped test

Binary installer taken https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/Makefile

Fixes #887

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,13 @@ $(ARTIFACTS):
 
 .PHONY: test
 test: ## Run tests
-	go test -v ./...
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; go test ./...
+
+.PHONY: test-cover
+test-cover: ## Run tests with code coverage and code generate  reports
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; mkdir -p /tmp/out; go test -v -coverprofile=/tmp/out/coverage.out ./... $(TEST_ARGS)
+	go tool cover -func=out/coverage.out -o out/coverage.txt
+	go tool cover -html=out/coverage.out -o out/coverage.html
 
 # Can be run manually, e.g. via:
 # export OPENSTACK_CLOUD_YAML_FILE="$(pwd)/clouds.yaml"
--- a/scripts/fetch_ext_bins.sh
+++ b/scripts/fetch_ext_bins.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Enable tracing in this script off by setting the TRACE variable in your
+# environment to any value:
+#
+# $ TRACE=1 test.sh
+TRACE=${TRACE:-""}
+if [[ -n "${TRACE}" ]]; then
+  set -x
+fi
+
+k8s_version=1.19.2
+goarch=amd64
+goos="unknown"
+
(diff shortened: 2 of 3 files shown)


jichenjc:
@Xenwar do you mind create an issue to describe the gaps/issues that this PR addressed? it will be easier to know from there.. thanks

Xenwar:
> @Xenwar do you mind create an issue to describe the gaps/issues that this PR addressed? it will be easier to know from there.. thanks

Added a new issue and liked it. #887

Makefile
```diff
@@ -106,7 +106,13 @@ $(ARTIFACTS):
 
 .PHONY: test
 test: ## Run tests
-	go test -v ./...
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./...
+
+.PHONY: test-cover
+test-cover: ## Run tests with code coverage and code generate  reports
```
jichenjc, line 112: thanks for submit this,

I run locally and it works well , one nit is do we want to rename this to something like capi_coverage to make it more specifi?

Makefile
```diff
@@ -106,7 +106,13 @@ $(ARTIFACTS):
 
 .PHONY: test
 test: ## Run tests
-	go test -v ./...
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./...
```
jichenjc, line 109: do we want to run test also include cover or just the test itself?

Makefile
sbueringer: I would really like to keep this consistent with the main repo. I don't see a need to prefix this which anything to be honest :)

Makefile
sbueringer: Just the test, please compare with the main repo. I wouldn't want to do it differently here if we don't have to

sbueringer, line 109: @Xenwar Can you maybe drop the -v in the test target and add a test-verbose (analog to the cluster-api repo)

Makefile
```diff
@@ -106,7 +106,13 @@ $(ARTIFACTS):
...
+
+.PHONY: test-cover
+test-cover: ## Run tests with code coverage and code generate  reports
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; mkdir -p /tmp/out; go test -v -coverprofile=/tmp/out/coverage.out ./... $(TEST_ARGS)
```
sbueringer, line 113: Can we call the test target from above here? (please do it the same way as in the main repo, e.g. this also means using out/coverage.out instead the tmp folder)

scripts/fetch_ext_bins.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
```
sbueringer, line 2: I'm not sure where this is from but I would prefer the version from the cluster-api repo. In this script here we delete a folder when SKIP_FETCH_TOOLS is set, which is a bit strange, as imho it shouldn't do anything.

Overall we try to stay consistent with the main cluster-api repo in CAPO to reduce maintenance effort on our side.

Makefile
Xenwar: I agree with not prefixing it.

Makefile
Xenwar: Thanks. Removed

Makefile
Xenwar: thanks. done

Makefile
Xenwar: for the cover, there is another target.

scripts/fetch_ext_bins.sh
Xenwar: The capi repo does not have the script. But, the ones in Amazon, which which I took this script and the one in GCP use the same approach. 
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/scripts/fetch_ext_bins.sh

Xenwar:
@jichenjc @sbueringer
Sorry for the late reply, could not get time due to higher priority PRs. 
I have made the relevant changes as suggested.

jichenjc:
#938 merged, hopefully the gate will be better now

jichenjc:

#938 merged, let's try again

PR #897, opened 2021-06-07
:sparkles: add predicates.ResourceIsNotExternallyManaged to cluster controller

### LINKED ISSUE #859: Handle "cluster.x-k8s.io/managed-by" annotation
/kind feature

CAPI v1alpha4 introduced a mechanism to manage the cluster resource externally. This is triggered by the `managed-by` annotation. For more context, please see: 
* https://github.com/kubernetes-sigs/cluster-api/pull/4135
* https://github.com/kubernetes-sigs/cluster-api/pull/4303

Notes:
* We should make sure we do all that's necessary. I'm not entirely sure if it's enough to just ignore the cluster when the annotation is set. Please read the proposal for details
* Prereq: we need a rebase onto current CAPI before

asalkeld:
This PR implements https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20210203-externally-managed-cluster-infrastructure.md for the OpenStack provider.

Fixes #859

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -430,6 +430,9 @@ func reconcileNetworkComponents(log logr.Logger, osProviderClient *gophercloud.P
 }
 
 func (r *OpenStackClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
+	clusterToInfraFn := util.ClusterToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("OpenStackCluster"))
+	log := ctrl.LoggerFrom(ctx)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options).
 		For(&infrav1.OpenStackCluster{},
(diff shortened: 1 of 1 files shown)


hidekazuna:
This is practically part of #859 , right?

asalkeld:
> This is practically part of #859 , right?

Yes, thanks @hidekazuna I didn't see that issue at all.

controllers/openstackcluster_controller.go
```diff
@@ -450,9 +453,27 @@ func (r *OpenStackClusterReconciler) SetupWithManager(ctx context.Context, mgr c
 		).
 		Watches(
 			&source.Kind{Type: &clusterv1.Cluster{}},
-			handler.EnqueueRequestsFromMapFunc(util.ClusterToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("OpenStackCluster"))),
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
```
asalkeld, line 456: This code should take into account this: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/859#issuecomment-839477889

controllers/openstackcluster_controller.go
```diff
@@ -450,9 +453,27 @@ func (r *OpenStackClusterReconciler) SetupWithManager(ctx context.Context, mgr c
...
+				c := &infrav1.OpenStackCluster{}
+				if err := r.Client.Get(ctx, requests[0].NamespacedName, c); err != nil {
+					log.V(4).Error(err, "Failed to get OpenStack cluster")
+					return nil
+				}
+
+				if annotations.IsExternallyManaged(c) {
+					log.V(4).Info("OpenStackCluster is externally managed, skipping mapping.")
+					return nil
+				}
```
chrischdi, line 471: Isn't the very same done already by setting the predicate in l 477? So this func may be obsolete?

asalkeld: Hi @chrischdi, see https://kubernetes.slack.com/archives/C8TSNPY4T/p1620749340371300

No, this is getting a stream of cluster object events and then mapping them to openstackClusters.
The below predicates are only working on the base object (in this case the cluster, which does not have the annotation - the openstackCluster has the annotation)

chrischdi: Ah good point :-) thank you fopr clarification.

controllers/openstackcluster_controller.go
```diff
@@ -450,9 +453,27 @@ func (r *OpenStackClusterReconciler) SetupWithManager(ctx context.Context, mgr c
 		).
 		Watches(
 			&source.Kind{Type: &clusterv1.Cluster{}},
-			handler.EnqueueRequestsFromMapFunc(util.ClusterToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("OpenStackCluster"))),
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+				requests := clusterToInfraFn(o)
+				if requests == nil {
```
sbueringer, line 458:
```suggestion
				if len(requests) > 0 {
```
Should catch both cases and then we're safe against nil pointers in l.463

asalkeld: https://github.com/kubernetes-sigs/cluster-api/blob/master/util/util.go#L195-L222
It either returns nil or an array of len == 1.. I could change it to 
```
if len(requests) < 1 {
   return nil
}
```

sbueringer: Ups of course, accidentally inverted the logic :)

PR #909, opened 2021-06-15
🐛update logic to support more instance state

### LINKED ISSUE #889: Support machine instance state "migrating"
/kind bug

An OpenStack machine was ["live migrated"](https://docs.openstack.org/nova/pike/admin/live-migration-usage.html), but the `openstackmachine` resource ended up being stuck in `MIGRATING` state:

```shell
$ kubectl get openstackmachine
NAME                                       CLUSTER          INSTANCESTATE   READY   PROVIDERID                                          MACHINE
prod-1-osmt-control-plane-g2-2-4-004vxjz   prod-1-cluster   MIGRATING       true    openstack:///2b3fb754-bf0d-4b25-a820-0720953e8429   prod-1-kubeadm-control-plane-t2cnm
[...]
```

```json
$ kubectl get openstackmachine prod-1-osmt-control-plane-g2-2-4-004vxjz -o json | jq -r .status
{
  "addresses": [
    {
      "address": "10.224.1.30",
      "type": "InternalIP"
    }
  ],
  "errorMessage": "OpenStack instance state \"MIGRATING\" is unexpected",
  "errorReason": "UpdateError",
  "instanceState": "MIGRATING",
  "ready": true
}
```

...but the migration in OpenStack seems to have completed successfully:

Skärmavbild 2021-06-01 kl  17 11 37

```
$ openstack server show prod-1-osmt-control-plane-g2-2-4-004vxjz -f json | jq .status
"ACTIVE"
```

Reconciles progressing, and not being blocked by invalid state on the instance.

It seems we only accept `ACTIVE` and `BUILDING` states, not `MIGRATING`:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/release-0.3/controllers/openstackmachine_controller.go#L403-L412

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.3
- Cluster-API version: 
- OpenStack version:
- Minikube/KIND version:

### LINKED ISSUE #1037: supported machine status list and corresponding actions
/kind bug

see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/889 and its related discussions
there are multiple status in nova VM and we need a way to support all of them instead checking status and supoort them one by one 
e.g we support MIGRATING then SHUTOFF then might be something else...

so tracking this issue from different perspective

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

### LINKED ISSUE #1038: Instance keep SHUTOFF in openstackmachine
/kind bug

after VM shutoff / start, the nova is reporting VM ACTIVE now , but the openstackmachine instancestate always SHUTOFF (I wait for 24 hours + )

```
# nova list
+--------------------------------------+-------------------------------------+--------+------------+-------------+-------------------------------------------------------------------------+
| ID                                   | Name                                | Status | Task State | Power State | Networks                                                                |
+--------------------------------------+-------------------------------------+--------+------------+-------------+-------------------------------------------------------------------------+
| dddf685b-1a34-46dd-92cf-9f30bd7cd6b6 | capi-quickstart-control-plane-lkwlb | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-quickstart=10.6.0.225, 172.24.4.142 |
| d19ac136-10c2-479c-810c-06fac0829a0b | capi-quickstart-md-0-rvkns          | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-quickstart=10.6.0.128               |
+--------------------------------------+-------------------------------------+--------+------------+-------------+-------------------------------------------------------------------------+

```

```
# kubectl get openstackmachine
NAME                                  CLUSTER           INSTANCESTATE   READY   PROVIDERID                                          MACHINE
capi-quickstart-control-plane-lkwlb   capi-quickstart   ACTIVE          true    openstack:///dddf685b-1a34-46dd-92cf-9f30bd7cd6b6   capi-quickstart-control-plane-c95bs
capi-quickstart-md-0-rvkns            capi-quickstart   SHUTOFF         true    openstack:///d19ac136-10c2-479c-810c-06fac0829a0b   capi-quickstart-md-0-7d869f6fb9-6kxwx

```

the openstackmachine state become ACTIVE after VM startup from stop status

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

jichenjc:

Fixes #889 
Fixes #1037 
Fixes #1038 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -227,9 +227,6 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
-
 	// InstanceStateActive is the string representing an instance in an active state.
 	InstanceStateActive = InstanceState("ACTIVE")
 
@@ -241,6 +238,9 @@ var (
 
 	// InstanceStateShutoff is the string representing an instance in a shutoff state.
 	InstanceStateShutoff = InstanceState("SHUTOFF")
+
+	// InstanceStateDeleted is the string representing an instance in a deleted state.
+	InstanceStateDeleted = InstanceState("DELETED")
 )
 
 // Bastion represents basic information about the bastion node.
--- a/api/v1alpha4/types.go
+++ b/api/v1alpha4/types.go
@@ -279,9 +279,6 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
-
 	// InstanceStateActive is the string representing an instance in an active state.
 	InstanceStateActive = InstanceState("ACTIVE")
 
@@ -293,6 +290,9 @@ var (
 
 	// InstanceStateShutoff is the string representing an instance in a shutoff state.
 	InstanceStateShutoff = InstanceState("SHUTOFF")
+
+	// InstanceStateDeleted is the string representing an instance in a deleted state.
+	InstanceStateDeleted = InstanceState("DELETED")
 )
 
 // Bastion represents basic information about the bastion node.
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -64,6 +64,7 @@ type OpenStackMachineReconciler struct {
 
 const (
 	waitForClusterInfrastructureReadyDuration = 15 * time.Second
+	waitForInstanceBecomeActiveToReconcile    = 60 * time.Second
 )
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachines,verbs=get;list;watch;create;update;patch;delete
@@ -343,11 +344,19 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 	case infrav1.InstanceStateActive:
 		logger.Info("Machine instance is ACTIVE", "instance-id", instanceStatus.ID())
 		openStackMachine.Status.Ready = true
-	case infrav1.InstanceStateBuilding:
-		logger.Info("Machine instance is BUILDING", "instance-id", instanceStatus.ID())
-	default:
+	case infrav1.InstanceStateError:
+		// Error is unexpected, thus we report error and never retry
 		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("OpenStack instance state %q is unexpected", instanceStatus.State()))
 		return ctrl.Result{}, nil
+	case infrav1.InstanceStateDeleted:
+		// we should avoid further actions for DELETED VM
+		logger.Info("Instance state is DELETED, no actions")
+		return ctrl.Result{}, nil
+	default:
+		// The other state is normal (for example, migrating, shutoff) but we don't want to proceed until it's ACTIVE
+		// due to potential conflict or unexpected actions
+		logger.Info("Waiting for instance to become ACTIVE", "instance-id", instanceStatus.ID(), "status", instanceStatus.State())
+		return ctrl.Result{RequeueAfter: waitForInstanceBecomeActiveToReconcile}, nil
 	}
 
 	if openStackCluster.Spec.ManagedAPIServerLoadBalancer {
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -293,6 +293,9 @@ var (
 
 	// InstanceStateShutoff is the string representing an instance in a shutoff state.
 	InstanceStateShutoff = InstanceState("SHUTOFF")
+
+	// InstanceStateDeleted is the string representing an instance in a deleted state.
+	InstanceStateDeleted = InstanceState("DELETED")
 )
 
 // Bastion represents basic information about the bastion node.


controllers/openstackmachine_controller.go
```diff
@@ -336,9 +337,14 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
-	default:
+	case infrav1.InstanceStateError:
+		// Error is unexpected, thus we report error and never retry
 		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("OpenStack instance state %q is unexpected", instance.State))
 		return ctrl.Result{}, nil
+	default:
+		// The other state is normal (for example, migrating) but we don't want to proceed until it's ACTIVE
+		// due to potential conflict or unexpected actions
+		return ctrl.Result{RequeueAfter: waitForInstanceBecomeActiveToReconcile}, nil
 	}
```
jichenjc, line 348: to be honst, follow this code, we only do floating ip association which is safe whenever the VM is in which state (the floating ip operation not on VM , but on port) .. so I think it's ok to accept those status (so that means no requeue)

but to be safe, seems only allow ACTIVE is also reasonable (eventually the VM will go ACTIVE ?)

hidekazuna: Ideally we need to care all statuses: https://docs.openstack.org/api-ref/compute/?expanded=show-server-details-detail,list-servers-detail#list-servers
I found there is no `BUILDING` status.

hidekazuna: In addtion to `ERROR`, `DELETED`, `UNKNOWN` statuses never become `ACTIVE` again.
I think other statuses could become `ACTIVE` by automatically(`BUILD`) or manually.

controllers/openstackmachine_controller.go
```diff
@@ -336,9 +337,14 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateBuilding:
```
hidekazuna, line 338: This should be deleted because there is not `BUILDING` status.

jichenjc: I don't even notice that, will check again, thanks for the info~

controllers/openstackmachine_controller.go
jichenjc: @hidekazuna BUILDING is correct, 
https://github.com/openstack/nova/blob/master/nova/compute/vm_states.py#L37

hidekazuna: Not correct. status is determined by `vm_state` and `task_state`.
function: https://github.com/openstack/nova/blob/master/nova/api/openstack/common.py#L115-L124
dictionary: https://github.com/openstack/nova/blob/master/nova/api/openstack/common.py#L45-L112

This means if vm_state is BUILDING, status is always BUILD.

jichenjc: you are right, the status is BUILD 
```

nova --debug show 1d36ba54-4dfa-432a-b773-dda237770b28

"status": "BUILD",
...
"OS-EXT-STS:vm_state": "building",
```

jichenjc:
@hidekazuna updated to BUILD now :)

controllers/openstackmachine_controller.go
```diff
@@ -334,11 +335,16 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 	case infrav1.InstanceStateActive:
 		logger.Info("Machine instance is ACTIVE", "instance-id", instance.ID)
 		openStackMachine.Status.Ready = true
-	case infrav1.InstanceStateBuilding:
-		logger.Info("Machine instance is BUILDING", "instance-id", instance.ID)
-	default:
+	case infrav1.InstanceStateBuild:
+		logger.Info("Machine instance is BUILD", "instance-id", instance.ID)
+	case infrav1.InstanceStateError:
```
hidekazuna, line 340: case infrav1.InstanceStateError, infrav1.InstanceStateDeleted, infrav1.InstanceStateUnknown:

Because DELETED, UNKNOWN statuses also never become ACTIVE again.

controllers/openstackmachine_controller.go
```diff
@@ -334,11 +335,16 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 	case infrav1.InstanceStateActive:
 		logger.Info("Machine instance is ACTIVE", "instance-id", instance.ID)
 		openStackMachine.Status.Ready = true
-	case infrav1.InstanceStateBuilding:
-		logger.Info("Machine instance is BUILDING", "instance-id", instance.ID)
-	default:
+	case infrav1.InstanceStateBuild:
```
hidekazuna, line 338: This should be deleted, because VM with BUILD status is on the way to ACTIVE,  thus should go into Requeue.

mdbooth: I agree, although the additional logging is nice. I wonder if we could add a generic log message to the default case below?

controllers/openstackmachine_controller.go
mdbooth: UNKNOWN can become ACTIVE. It is returned in certain circumstances if the control plane can't reach the compute, but doesn't mean that the VM has failed. IIRC the primary use cases are Edge and during cell maintainance on multi-cell deployments.

It might also be better not to handle DELETED here, because that will end reconciliation in failure. DELETED presumably means the user deleted the VM for some reason, almost certainly in the hope that it would fix a problem. If we continue to ignore DELETED here we'll automatically create a new machine when the deletion completes, which is most likely the intended behaviour.

jichenjc:
/remove-lifecycle stale

@hidekazuna @mdbooth maybe you can take another look? Thanks

jichenjc:
with latest code change, the status is valid now

```
    "status": {
        "addresses": [
            {
                "address": "10.6.0.49",
                "type": "InternalIP"
            }
        ],
        "instanceState": "SHUTOFF",
        "ready": true
    }
}

```

jichenjc:
@hidekazuna @mdbooth can you help on this? Thanks

mdbooth:
> @hidekazuna @mdbooth can you help on this? Thanks

I've been giving this some thought, especially as I'm just about to do something similar for volume creation in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1030.

I don't think we should even try to enumerate all possible instance states: it's a tight coupling of an api that's fairly loosely defined, and can be changed in the future by adding more states anyway. Instead we should explicitly handle states that we care about, and ignore all other states.

So honestly I would personally modify this logic to be something like:

```golang
        logger.Info("Machine state", "instance-id", instanceStatus.ID(), "state", instanceStatus.State())
	switch instanceStatus.State() {
	case infrav1.InstanceStateActive:
                // This is the only state in which we want to continue reconciling
		openStackMachine.Status.Ready = true
                // This is the only fatally unreconcilable state
		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("OpenStack instance is in error state", instanceStatus.State()))
		return ctrl.Result{}, nil
	default:
                // Unknown state. Wait 10 seconds to see if it goes away
                logger.Info("Waiting for instance to become ACTIVE")
                return ctrl.Result{RequeueAfter: 10}, nil
	}
```

jichenjc:
>>>should explicitly handle states that we care about, and ignore all other states.

@mdbooth infrav1.InstanceStateDeleted is a temp status and we handled it specially
the others are almost same to what you suggested now...


I added a log line per your comments above.. thanks

api/v1alpha3/types.go
```diff
@@ -227,8 +227,8 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
+	// InstanceStateBuild is the string representing an instance in a building state.
+	InstanceStateBuild = InstanceState("BUILD")
```
hidekazuna, line 231: We can delete this variable if this is not used.

api/v1alpha4/types.go
```diff
@@ -279,8 +279,8 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
+	// InstanceStateBuild is the string representing an instance in a building state.
+	InstanceStateBuild = InstanceState("BUILD")
```
hidekazuna, line 283: We can delete this variable if this is not used.

controllers/openstackmachine_controller.go
```diff
@@ -343,11 +344,18 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
 		openStackMachine.Status.Ready = true
-	case infrav1.InstanceStateBuilding:
-		logger.Info("Machine instance is BUILDING", "instance-id", instanceStatus.ID())
-	default:
+	case infrav1.InstanceStateError:
+		// Error is unexpected, thus we report error and never retry
 		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("OpenStack instance state %q is unexpected", instanceStatus.State()))
 		return ctrl.Result{}, nil
+	case infrav1.InstanceStateDeleted:
+		// we should avoid further actions for DELETED VM
```
hidekazuna, line 352: better to add log.

controllers/openstackmachine_controller.go
```diff
@@ -343,11 +344,18 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
+		return ctrl.Result{}, nil
+	default:
+		// The other state is normal (for example, migrating, shutoff) but we don't want to proceed until it's ACTIVE
+		// due to potential conflict or unexpected actions
+		logger.Info("Waiting for instance to become ACTIVE")
```
mdbooth, line 357: Might be best to log the current state here, too.

mdbooth (approved):
Agree with all the comments, but I'd still be happy with the change as-is.

Issue #913, opened 2021-06-18
Unable to disable port security at the network level


Xenwar:
It is not possible to disable port security at the network level. the following is logged. 
```
{
    "NeutronError": {
        "type": "PortSecurityAndIPRequiredForSecurityGroups",
        "message": "Port security must be enabled and port must have an IP address in order to use security groups.",
        "detail": ""
    }
}
```
also, 
```
(openstack) server show c04f824e-f0b4-47f6-9e75-2b92a0ec3cc3 -c fault -f value
{
   "code":500,
   "created":"2021-06-18T12:25:07Z",
   "message":"Exceeded maximum number of retries. Exceeded max scheduling attempts 5 for instance c04f824e-f0b4-47f6-9e75-2b92a0ec3cc3. Last exception: Network requires port_security_enabled and subnet associated in order to apply security groups."
}
```

Security groups and AllowedAddressPairs need to be added to a port conditionally. 

Example manifest
```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: OpenStackCluster
metadata:
  name: test-osc
  namespace: default
spec:
  disablePortSecurity: true
```
/kind bug

hidekazuna:
This is equal to #911 

/close

Xenwar:
#911 is created for the feature to add port level configuration. 
 This one is a bug report.

hidekazuna:
Aha, understand.

/close cancel

hidekazuna:
I confirmed disablePortSecurity works.

A piece of manifest:
```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: OpenStackCluster
metadata:
  name: nonha
  namespace: default
spec:
  cloudName: openstack
  cloudsSecret:
    name: nonha-cloud-config
    namespace: default
  disablePortSecurity: true
  #  managedSecurityGroups: true
  dnsNameservers:
  - 10.0.0.11
  nodeCidr: 10.6.0.0/24
```

Created network:
```sh
(python) hidekazu@controller:~$ openstack network show k8s-clusterapi-cluster-default-nonha | grep port_security
| port_security_enabled     | False                                                                                                                                                                            |
```

Created port:
```sh
(python) hidekazu@controller:~$ openstack port show 51cc6cc9-bd20-4018-8ae6-c1141f5b21ec | grep port_security
| port_security_enabled   | False                                                                                                                                                                            |
[...]

PR #914, opened 2021-06-21
[WIP] ✨Disable/enable PortSecurity at port level

### LINKED ISSUE #911: Add a feature to configure Port level PortSecurity
Add a feature to configure Port level PortSecurity such that

- Each port is configured independently. When disabled, both SecurityGroups and allowedAddressPairs values on that port should be removed. specified by `OpenStackMachineTemplate.spec..template.spec.ports[0].disablePortSecurity: false/true`

- When PortSecurity is disabled at the network level, then it is also disabled at each port level. specified by (OpenStackCluster.spec.disablePortSecurity: false)

/kind feature

Xenwar:
This PR adds the capability of enabling/disabling PortSecurity at the port level.

Example Manifest
```yaml
kind: OpenStackCluster
metadata:
  name: basic-8
  namespace: default
spec:
  disablePortSecurity: false
```

```yaml
metadata:
  name: basic-8-control-plane
  namespace: default
spec:
  template:
    spec:
      ports:
      - description: "primary with PortSecurity enabled, required for proper k8s cluster"
        disablePortSecurity: false
        allowedAddressPairs:
        - macAddress: "a4:b8:34:2b:ab:01"
          ipAddress: "10.10.10.10/24"
      - description: "secondary with PortSecurity disabled"
        disablePortSecurity: true
        allowedAddressPairs:
        - macAddress: "a4:b8:34:2b:ab:02"
          ipAddress: "20.20.20.20/24"
```
Example output:
`
```
(openstack) port show primary-port
+-----------------------+--------------------------------------------------------------------+
| Field                 | Value                                                              |
+-----------------------+--------------------------------------------------------------------+
| allowed_address_pairs | ip_address='10.10.10.10/24', mac_address='a4:b8:34:2b:ab:01'       |
| description           | primary with PortSecurity enabled, required for proper k8s cluster |
| port_security_enabled | True                                                               |
| security_group_ids    | 37e42488-1417-4e2c-980f-d2ef71ffbacc                               |
+-----------------------+--------------------------------------------------------------------+
(openstack) port show secondary-port
+-----------------------+--------------------------------------+
| Field                 | Value                                |
+-----------------------+--------------------------------------+
| allowed_address_pairs |                                      |
| description           | secondary with PortSecurity disabled |
| port_security_enabled | False                                |
| security_group_ids    |                                      |
+-----------------------+--------------------------------------+
```
`

Fixes #911


config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
```diff
@@ -1264,6 +1264,11 @@ spec:
                               type: array
                             description:
                               type: string
+                            disablePortSecurity:
+                              description: DisablePortSecurity disables the port security
```
hidekazuna, line 1268: Where this description come from? Description should be auto generated from a comment in api/v1alpha4/types.go.

pkg/cloud/services/compute/instance.go
```diff
@@ -505,6 +515,28 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+	if portOpts.DisablePortSecurity == true || networkDisabledPortSecurity == true {
```
hidekazuna, line 518: Can't we create port with/without port security not updating after creating port?

config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
Xenwar: You are right. I did not re-generate the yaml files after removing the comment. 
Will add the comment back.

pkg/cloud/services/compute/instance.go
Xenwar: no, we cannot. 
At the port level, PortSecurity is immutable. 
For the network level, there is a bug reporter (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/913)

hidekazuna: I think it is possible from [gophercloud docs](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/portsecurity/doc.go#L94-L120).

Xenwar: Indeed. The link you provided is more efficient as it does not need retrieving the port to update it, I will update my PR accordingly.

Xenwar: Thanks a lot

hidekazuna:
@Xenwar I think we need to discuss motivation and spec more. Please see #911 #913

Xenwar:
> @Xenwar I think we need to discuss motivation and spec more. Please see #911 #913

Thanks, I will try it again and update the issue accordingly.

Xenwar:
@hidekazuna 
Having difficult combining these two options. 
The  `ports.Create(s.networkClient,` takes only one of them. Option2 does not allow setting HostID, VNICType and Profile properties. 
Any idea if combining them is possible ?
```

	option1 := portsbinding.CreateOptsExt{
		CreateOptsBuilder: createOpts,
		HostID:            portOpts.HostID,
		VNICType:          portOpts.VNICType,
		Profile:           nil,

	}
	option2 := portsecurity.PortCreateOptsExt{
		CreateOptsBuilder:   createOpts,
		PortSecurityEnabled: &enablePortSecurity,
	}
```

hidekazuna:
> @hidekazuna
> Having difficulty combining these two options.
> The `ports.Create(s.networkClient,` takes only one of them. Option2 does not allow setting HostID, VNICType and Profile properties.
> Any idea if combining them is possible ?
> 
> ```
> 
> 	option1 := portsbinding.CreateOptsExt{
> 		CreateOptsBuilder: createOpts,
> 		HostID:            portOpts.HostID,
> 		VNICType:          portOpts.VNICType,
> 		Profile:           nil,
> 
> 	}
> 	option2 := portsecurity.PortCreateOptsExt{
> 		CreateOptsBuilder:   createOpts,
> 		PortSecurityEnabled: &enablePortSecurity,
> 	}
> ```

@Xenwar  Thanks investigating. I have no idea to combine them as of now.

Xenwar:
@hidekazuna Thanks.

iamemilio:
I just made a similar PR not realizing this was already out there. Lets combine efforts and share homework: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/921

Rather than inheriting the port security settings from the cluster spec, I think its better to just inherit them from the network (which openstack does automatically on create) since these ports can be created on any network in theory. This is why I chose to make it *bool.

iamemilio:
Another thing to watch out for is that because we typically set the security groups at the instance level, OpenStack will apply that security to all port interfaces of that instance. This will cause ugly bugs when setting the port security on those ports.

Xenwar:
> I just made a similar PR not realizing this was already out there. Lets combine efforts and share homework: #921
> 
> Rather than inheriting the port security settings from the cluster spec, I think its better to just inherit them from the network (which openstack does automatically on create) since these ports can be created on any network in theory. This is why I chose to make it *bool.

Thanks, we can combine efforts.
On inheriting portSecurity from the network, my reasoning was there should be a single source of truth for the setting. So, if one changes the setting at the OSCluster spec, then it is propagated to the the Network. 

Also, Network level setting is relevant when PortSecurity is disabled at the network level so the setting is disabled for each port. i.e. Ports do not inherit the setting from the network in all cases.

Xenwar:
> Another thing to watch out for is that because we typically set the security groups at the instance level, OpenStack will apply that security group to all port interfaces of that instance. This will cause ugly bugs when setting the port security on those ports.

We have a use case to enable/disable PortSecurity at the port(s) level. I am not quite sure yet at what point the bug you mentioned could show up, during Port creation or during an update as we have the option of specifying security groups at the port level.

We could probably open a discussion on slack on the this and the other comment.

iamemilio:
/assign @iamemilio 
going to assign myself to this so that I get notifications :)

hidekazuna:
@Xenwar This PR was replaced with #921 and we can close this PR, right?

Xenwar:
> @Xenwar This PR was replaced with #921 and we can close this PR, right?
Thanks for updating the issue. 

I was testing #921  this today and it does not seem to be ready for me.
We should probably unhold this one and proceed with it since it does provide the functionality. 
Probably, a follow up PR can handle the improvements and optimisation. 

@iamemilio can provide more info on this.

hidekazuna:
@Xenwar hold/unfold does not mean ready for review. If PR is not ready for review, rename title starting with WIP, please.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -327,6 +327,13 @@ A floating IP is created and associated to the bastion host automatically, but y
 
 If `managedSecurityGroups: true`, security group rule opening 22/tcp is added to security groups for bastion, controller, and worker nodes respectively. Otherwise, you have to add `securityGroups` to the `bastion` in `OpenStackCluster` spec and `OpenStackMachineTemplate` spec template respectively.
 
+If PortSecurity disable for all ports either by setting `OpenStackCluster.spec.disablePortSecurity: true` or for each port individually, `managedSecurityGroups: false` is the only valid option. The following is not a valid coonfiguration and is not supported.
```
hidekazuna, line 330: This is in "Accessing nodes through the bastion host via SSH" section, not security group section. How about moving to [Ports](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/book/src/clusteropenstack/configuration.md#ports)?

Xenwar: Thanks, moved

Xenwar:
This PR is no longer needed. 
Feature is merged by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/921

PR #915, opened 2021-06-23
📖 Upgrading a cluster without LBaaS is not supported

### LINKED ISSUE #912: Upgrading clusters without Octavia
/kind bug

I've tried upgrading k8s using the following command:

```
kubectl patch kubeadmcontrolplane control-plane --type=json \
  -p '[{"op": "replace", "path": "/spec/version", "value": "1.18.12"}]'
```

We didn't have Octavia installed, so what happened was that the floating IP moved to the new instance, which was trying to join the existing cluster using the floating ip that had already been moved.

I was expecting the k8s upgrade operation to complete successfully.

I'm not sure if this can be considered a bug or just an undocumented limitation. Is there a way in which we could get the cluster upgrade operation to work without having Octavia (Neutron LBaaS)? One possible approach would be to use the private IPs instead of the floating IP as part of the upgrade workflow.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.3.4
- Cluster-API version: v0.3.16
- OpenStack version: Victoria
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): 1.18
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

hidekazuna:

Upgrading a cluster without LBaaS is not supported.

Fixes #912 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ previous cluster managers such as [kops][kops] and
 - Native Kubernetes manifests and API
 - Choice of Linux distribution (as long as a current cloud-init is available)
 - Support for single and multi-node control plane clusters
-- Deploy clusters with and without LBaaS available
+- Deploy clusters with and without LBaaS available (only cluster with LBaaS can be upgraded)
 - Support for security groups
 - cloud-init based nodes bootstrapping
 


hidekazuna:
@jichenjc PTAL

jichenjc:
@hidekazuna can we open an issue to track this feature? thanks

/lgtm
/hold cancel

PR #921, opened 2021-06-29
✨ Apply Port Security to Ports

iamemilio:
:sparkles: Being able to set port security on each port specifically allows users
to use applications that dont work with OpenStack port security on
networks that have it enabled.


pkg/cloud/services/compute/instance.go
```diff
@@ -504,6 +505,25 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 		return nil, err
 	}
 
+	if portOpts.PortSecurity != nil {
```
jichenjc, line 508: seems CreateOpts has those 2 definition
```
SecurityGroups      *[]string     `json:"security_groups,omitempty"`
        AllowedAddressPairs []AddressPair `json:"allowed_address_pairs,omitempty"`
```

why we don't combine this with create function ?

pkg/cloud/services/compute/instance.go
```diff
@@ -504,6 +505,25 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
...
+			// Remove allowed address pairs and security groups from port if possible to avoid openstack errors
+			// when port security is disabled on a port.
+			portUpdateOpts.SecurityGroups = &[]string{}
+			portUpdateOpts.AllowedAddressPairs = &[]ports.AddressPair{}
+		}
+		updateOpts := portsecurity.PortUpdateOptsExt{
+			UpdateOptsBuilder:   portUpdateOpts,
+			PortSecurityEnabled: portOpts.PortSecurity,
+		}
+		port, err = ports.Update(s.networkClient, port.ID, updateOpts).Extract()
```
jichenjc, line 520: if we really need use `update` , we need add metrics like line 503 above

pkg/cloud/services/compute/instance.go
iamemilio: I will have to test that. My worry is that OpenStack would apply the security group to the node if its set on the instance regardless of the settings on the port.

iamemilio: If I am not mistaken, in my tests when security groups are applied to an instance they can overwrite the settings on the port. I can double check this though, it was a long time ago.

api/v1alpha4/types.go
```diff
@@ -136,6 +136,10 @@ type PortOpts struct {
 
 	// The virtual network interface card (vNIC) type that is bound to the neutron port.
 	VNICType string `json:"vnicType,omitempty"`
+
+	// DisablePortSecurity applies an update to a port that enables or disables the port security when set.
+	// It will use the network defaults when unset.
```
Xenwar, line 141:
```suggestion
	// DisablePortSecurity applies an update to a port that enables or disables the port security when set.
	// When not set, it takes the value of the corresponding field at network level.
```

api/v1alpha4/types.go
```diff
@@ -136,6 +136,10 @@ type PortOpts struct {
 
 	// The virtual network interface card (vNIC) type that is bound to the neutron port.
 	VNICType string `json:"vnicType,omitempty"`
+
+	// DisablePortSecurity applies an update to a port that enables or disables the port security when set.
+	// When not set, it takes the value of the corresponding field at the network level.
+	DisablePortSecurity *bool `json:"disablePortSecurity:omitempty"`
```
Xenwar, line 142:
```suggestion
	DisablePortSecurity bool `json:"disablePortSecurity:omitempty"`
```
If not specified, it should default to false.

Xenwar: Currently, the e2e test is failing for that reason. 
```
                Causes: [
                    {
                        Type: "FieldValueInvalid",
                        Message: "Invalid value: \"null\": spec.template.spec.ports.disablePortSecurity:omitempty in body must be of type boolean: \"null\"",
                        Field: "spec.template.spec.ports.disablePortSecurity:omitempty",
                    },
                ],
```

iamemilio: I think it being a pointer makes sense, and I see other fields that use this syntax, so its probably just a little bug somewhere. The reason I prefer a pointer here is because we can't know for sure if a user is going to be attaching a port from a network with port security enabled or disabled, so how can we know what to set the default value to? By having a pointer, the user can leave it nil,  and that allows us to infer that the user just wants this to not be modified by CAPO/use the network defaults.

macaptain, line 142:
```suggestion
	DisablePortSecurity *bool `json:"disablePortSecurity,omitempty"`
```

iamemilio: oooh lol that would do it

iamemilio:
looks like a nil pointer error, I think I know what is causing it

pkg/cloud/services/compute/instance.go
```diff
@@ -471,11 +472,13 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 		AllowedAddressPairs: []ports.AddressPair{},
 	}
 
-	for _, ap := range portOpts.AllowedAddressPairs {
-		createOpts.AllowedAddressPairs = append(createOpts.AllowedAddressPairs, ports.AddressPair{
-			IPAddress:  ap.IPAddress,
-			MACAddress: ap.MACAddress,
-		})
+	if portOpts.DisablePortSecurity == nil || !*portOpts.DisablePortSecurity {
```
Xenwar, line 475: I think this can be simplified by making the type bool, not a pointer.

pkg/cloud/services/compute/instance.go
```diff
@@ -505,6 +508,28 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+
+	if portOpts.DisablePortSecurity != nil {
+		mc2 := metrics.NewMetricPrometheusContext("port", "update")
+		portUpdateOpts := ports.UpdateOpts{}
+
+		if *portOpts.DisablePortSecurity {
```
Xenwar, line 516: The same here. 
if the type is just bool, it will be false by default. 
Basically, 513 and 514 are not needed unless the value is true and ca be moved to after 516

pkg/cloud/services/compute/instance_test.go
```diff
@@ -28,6 +28,7 @@ func Test_getPortName(t *testing.T) {
 		opts         *infrav1.PortOpts
 		netIndex     int
 	}
+	var True = true
```
Xenwar, line 31:
```suggestion
	var iTrue = true
```

pkg/cloud/services/compute/instance_test.go
```diff
@@ -48,6 +49,11 @@ func Test_getPortName(t *testing.T) {
 			args: args{"test-1-instance", &infrav1.PortOpts{}, 4},
 			want: "test-1-instance-4",
 		},
+		{
+			name: "with PortOpts name suffix",
+			args: args{"test-1-instance", &infrav1.PortOpts{NameSuffix: "foo2", NetworkID: "bar", DisablePortSecurity: &True}, 4},
```
Xenwar, line 54:
```suggestion
			args: args{"test-1-instance", &infrav1.PortOpts{NameSuffix: "foo2", NetworkID: "bar", DisablePortSecurity: &iTrue}, 4},
```

Xenwar (changes requested):

pkg/cloud/services/compute/instance_test.go
Xenwar: `pull-cluster-api-provider-openstack-test` is failing due to that.

iamemilio: Ack, yeah it doesn't like the style. I think doing ```var x = y``` is also bad style and should be ```x := y```

pkg/cloud/services/compute/instance_test.go
```diff
@@ -28,6 +28,7 @@ func Test_getPortName(t *testing.T) {
 		opts         *infrav1.PortOpts
 		netIndex     int
 	}
+	True := true
```
mdbooth, line 31: I'm sure this is idiomatic, but 🤦‍♂️

pkg/cloud/services/compute/instance.go
mdbooth: There's a critical difference here. There are genuinely 3 behaviours:

* DisablePortSecurity = true -> Disable port security
* DisablePortSecurity = false -> Do not disable port security
* DisablePortSecurity = nil -> Inherit port security setting from the network

It's detailed here: https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#id53

pkg/cloud/services/compute/instance.go
mdbooth: I don't think we need this Update. I can't see why we wouldn't just pass PortSecurityEnabled during the creation above. We're hitting neutron twice for no obvious reason.

mdbooth (changes requested):
I'd like to see the Update call removed.

In terms of API change, I believe this can be safely added to v1alpha4 without breaking the api contract.

pkg/cloud/services/compute/instance.go
chrischdi: The only reason would be to set an existing OpenStackMachine to enable or disable port security. But I don't know if this is even allowed (immutability?).

chrischdi (commented):
I'd also prefer, if possible to set the value on creation of the port in line 496/499 ?!

From https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/914 : would this need to get adapted to the `bastion.go` too?

pkg/cloud/services/compute/instance_test.go
iamemilio: This is unfortunately the only way I could create a pointer to a bool without upsetting the GO gods

pkg/cloud/services/compute/instance.go
iamemilio: Is it possible to do a **create** call in gophercloud with more than one extension applied?

api/v1alpha4/types.go
iamemilio: I am going to resolve this due to the failure being determined to be a flake. The conversation around the pointer is happening in a lower comment.

pkg/cloud/services/compute/instance.go
Xenwar: Thanks for clarifying it. One case though, What happens for the following case. 
NetworkLevel = disable && PortLevel=enable ?

Xenwar:
There are some issues I noticed while trying to add port level PortSecurity. It might help if we define some test 
cases. 

Terminology:
```
networkLevel: OpenStackCluster.OpenStackCluster.disablePortSecurity
portLevel: OpenStackMachineTemplate.spec.template.spec.ports[0].disablePortSecurity 
```

Case 1:
```
networkLevel = false/true && no portLevel info at all
Expected: portLevel = false/true (inherits network level)
Adding PortOpts here might help https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/2b1d1bc46c71c97dcdb42e872a8adf56091488fb/pkg/cloud/services/compute/instance.go#L177

```

case 2:
```
networkLevel = false/true && portLevel provided, but missing disablePortSecurity field
Expected: portLevel = false/true (inherits network level)
```

case 3:
```
networkLevel = false &&  portLevel = false
Expected: portLevel = false
```

case 4:
```
networkLevel = false &&  portLevel = true
Expected: portLevel = true
```
Running manual tests against these cases could help to see more corner cases.

iamemilio:
> There are some issues I noticed while trying to add port level PortSecurity. It might help if we define some test
> cases.
> 
> Terminology:
> 
> ```
> networkLevel: OpenStackCluster.OpenStackCluster.disablePortSecurity
> portLevel: OpenStackMachineTemplate.spec.template.spec.ports[0].disablePortSecurity 
> ```
> 
> Case 1:
> 
> ```
> networkLevel = false/true && no portLevel info at all
> Expected: portLevel = false/true (inherits network level)
> Adding PortOpts here might help https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/2b1d1bc46c71c97dcdb42e872a8adf56091488fb/pkg/cloud/services/compute/instance.go#L177
> ```
This use case confuses me a little. If there is not port level info, a port should not be created as far as I know.
> 
> case 2:
> 
> ```
> networkLevel = false/true && portLevel provided, but missing disablePortSecurity field
> Expected: portLevel = false/true (inherits network level)
> ```
> 
> case 3:
> 
> ```
> networkLevel = false &&  portLevel = false
> Expected: portLevel = false
> ```
> 
> case 4:
> 
> ```
> networkLevel = false &&  portLevel = true
> Expected: portLevel = true
> ```
> 
> Running manual tests against these cases could help to see more corner cases.

These sound reasonable. I think that these behaviors should be consistent since they are executed largely by openstack, not our code. What you are more likely to bump into are usability issues or complexities caused by security groups or allowed address pairs. Let us know what you find.

Xenwar:
Here is my test result for cases 2,3 and 4. To make is to reproduce, I have added some sample data. 
First, `disablePortSecurity: false `for network level
OpenStackMachineTemplate
```
spec:
  template:
    spec:
      ports:
      - description: "port-0: PortSecurity not provided"
      - description: "port-1: PortSecurity enabled"
        disablePortSecurity: false
        allowedAddressPairs:
        - macAddress: "a4:b8:34:2b:ab:01"
          ipAddress: "10.10.10.10/24"
      - description: "port-2: PortSecurity disabled"
        disablePortSecurity: true
        allowedAddressPairs:
        - macAddress: "a4:b8:34:2b:ab:02"
          ipAddress: "20.20.20.20/24"
```
OpenStackCluster
```
spec:
  disablePortSecurity: false
  managedSecurityGroups: true
```

Result:
- None of the ports are created
- Error is generated 

`E0717 09:53:18.179270      13 controller.go:302] controller-runtime/manager/controller/openstackmachine "msg"="Reconciler error" "error"="error creating Openstack instance: Bad request with: [POST https://kna1.citycloud.com:9696/v2.0/ports], error message: {\"NeutronError\": {\"type\": \"PortSecurityAndIPRequiredForSecurityGroups\", \"message\": \"Port security must be enabled and port must have an IP address in order to use security groups.\", \"detail\": \"\"}}" "name"="basic-1-control-plane-hpk8x" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" `

Next, `disablePortSecurity: true` at network level
OpenStackCluster
```
spec:
  disablePortSecurity: true
  managedSecurityGroups: false
```
Result:
- only one port (port-0: PortSecurity not provided ) is created successfully.
```
(openstack) port show cbe9645c-76da-4080-af17-f7a94286d224 -c description -c allowed_address_pairs -c port_security_enabled -c port_security_enabled -c security_group_ids
+-----------------------+-----------------------------------+
| Field                 | Value                             |
+-----------------------+-----------------------------------+
| allowed_address_pairs |                                   |
| description           | port-0: PortSecurity not provided |
| port_security_enabled | False                             |
| security_group_ids    |                                   |
+-----------------------+-----------------------------------+
```

- Error for the remaining ones. 

`
E0717 09:57:58.888067      13 controller.go:302] controller-runtime/manager/controller/openstackmachine "msg"="Reconciler error" "error"="error creating Openstack instance: Expected HTTP response code [] when accessing [POST https://kna1.citycloud.com:9696/v2.0/ports], but got 409 instead\n{\"NeutronError\": {\"type\": \"AddressPairAndPortSecurityRequired\", \"message\": \"Port Security must be enabled in order to have allowed address pairs on a port.\", \"detail\": \"\"}}" "name"="basic-1-control-plane-k49kd" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" 
`

Xenwar:
For case 1, the code that creates the port. is here https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9a849c865c5510eea394dc365b32eb79a25fc7f0/pkg/cloud/services/compute/instance.go#L171

To make is clear, here is the test result. 

OpenStackMachineTemplate
( ports: Not provides )

```
spec:
  template:
    spec:
      cloudName: openstack-1
```
  
OpenStackCluster

```
spec:
  disablePortSecurity: false
  managedSecurityGroups: true
```

Result:
- No port is created for the following reason.

`E0717 10:14:32.443914      13 controller.go:302] controller-runtime/manager/controller/openstackmachine "msg"="Reconciler error" "error"="error creating Openstack instance: Bad request with: [POST https://kna1.citycloud.com:9696/v2.0/ports], error message: {\"NeutronError\": {\"type\": \"PortSecurityAndIPRequiredForSecurityGroups\", \"message\": \"Port security must be enabled and port must have an IP address in order to use security groups.\", \"detail\": \"\"}}" "name"="basic-1-control-plane-25rx2" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" `

 and , when PortSecurity is disabled at network level
OpenStackCluster

```
spec:
  disablePortSecurity: true
  managedSecurityGroups: false
```
  
Result:
Port is created successfully.
```
(openstack) port show 80a9c217-dbc0-44e6-9d8b-47428dee2253
+-----------------------+-----------------------------------------------------------+
| Field                 | Value                                                     |
+-----------------------+-----------------------------------------------------------+
| allowed_address_pairs |                                                           |
| description           | Created by cluster-api-provider-openstack cluster basic-1 |
| port_security_enabled | False                                                     |
| security_group_ids    |                                                           |
+-----------------------+-----------------------------------------------------------+
```

pkg/cloud/services/compute/instance.go
jichenjc: opened a question https://github.com/gophercloud/gophercloud/issues/2195 ,hopefully we can get some comments

pkg/cloud/services/compute/instance.go
iamemilio: it should always follow what is set at the port level

pkg/cloud/services/compute/instance.go
```diff
@@ -471,14 +484,25 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
...
+		AllowedAddressPairs: addressPairs,
 	}
 
-	for _, ap := range portOpts.AllowedAddressPairs {
-		createOpts.AllowedAddressPairs = append(createOpts.AllowedAddressPairs, ports.AddressPair{
-			IPAddress:  ap.IPAddress,
-			MACAddress: ap.MACAddress,
-		})
+	var finalCreateOpts ports.CreateOptsBuilder
+	finalCreateOpts = createOpts
```
chrischdi, line 491: Can't we instead directly set the createOpts as ports.CreateOptsBuilder so we don't have to define the `finalCreateOpts`?

chrischdi (commented):
Change now looks straight forward to me :-) It get's added when set.

There's nothing we have to consider to get updated, am I right? Should we make the `disablePortSecurity` fields immutable via validation?

pkg/cloud/services/compute/instance.go
iamemilio: yeah, I can clean this up a little

iamemilio:
yep, we cut the update out, and nothing should need it. Is that necessary? Where would I add that?

chrischdi:
> yep, we cut the update out, and nothing should need it. Is that necessary? Where would I add that?

This could be done in the `*_webhook.go` files where the `ValidateUpdate` funcs are. Example from `AWS`: 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1alpha4/awscluster_webhook.go#L74

pkg/cloud/services/compute/instance_test.go
```diff
@@ -48,6 +49,11 @@ func Test_getPortName(t *testing.T) {
 			args: args{"test-1-instance", &infrav1.PortOpts{}, 4},
 			want: "test-1-instance-4",
 		},
+		{
+			name: "with PortOpts name suffix",
+			args: args{"test-1-instance", &infrav1.PortOpts{NameSuffix: "foo2", NetworkID: "bar", DisablePortSecurity: &truePtr}, 4},
```
tobiasgiese, line 54:
```suggestion
			args: args{"test-1-instance", &infrav1.PortOpts{NameSuffix: "foo2", NetworkID: "bar", DisablePortSecurity: pointer.Bool(true)}, 4},
```

Just a hint, not really a finding: you could use `k8s.io/utils/pointer`

iamemilio: ah, that is convenient. Its unusually difficult to declare a bool pointer in go.

chrischdi (commented):
/lgtm

@jichenjc what do you think about blocking changes for DisablePortSecurity?

@Xenwar does it fulfill the use-case(s) you thought about?

chrischdi:
/retitle ✨ Apply Port Security to Ports

Xenwar:
@chrischdi Thanks for asking. 
I have re-tested the uses cases one by one, and all of them were successful in creating a functioning cluster.

iamemilio:
thanks @Xenwar for testing

jichenjc:
/approve

great~,  can anyone help another quick look? Thanks

PR #924, opened 2021-06-30
🐛 Fix failure to create server with specified tags

### LINKED ISSUE #917: Nova Microversion support
/kind bug

I don't know if this is reproducable in 'normal' CAPO.

Calling computeService.CreateInstance() with a machine that specifies tags fails with an OpenStack error indicating that tags are not supported. This is because tags support was added in microversion 2.52, which we weren't specifying.

I pushed something to https://github.com/shiftstack/cluster-api-provider-openstack/tree/upstream-working to get me working, but this isn't suitable for merging in its current state.

What I'm going for in this patch is something which bumps the microversion to the minimum required by the machine specified. Ideally it would live in some library code, perhaps even in GopherCloud. However, this specific code is mutating global state (s.computeClient) based on ephemeral state (instance), which seems like an anti-pattern.

We most likely need to maintain a minimum microversion in any case. However, we do need to retain the ability to vary microversion so we can add support for new OpenStack features without breaking users on older versions.

I don't want to over-design this feature yet, though. Best to wait until it gets messy and refactor. Something like what I've written but which only affects the current execution of CreateInstance would likely be sufficient.

Calling it from here:
https://github.com/shiftstack/machine-api-provider-openstack

**Environment:**

Calling CAPO as a library from OpenShift machine-api-provider-openstack

mdbooth:
Trying to create a machine with specified tags would fail with:
  Additional properties are not allowed ('tags' was unexpected)

We fix this by raising the minimum supported microversion to 2.53. Note
that this means we no longer support versions of OpenStack prior to
Pike.

Fixes #917

Note that this change has the potential to break code anywhere in CAPO by changing the behaviour of the Nova API. From a quick review of the microversion history I think we should be OK (the change to Flavor looked most significant, but it doesn't affect us yet). CI is especially important here.

I don't think a unit test would achieve anything here. We could assert that the microversion has been set, but that kind of unit test is usually just noise. It is the E2E tests which are vital here. We are presumably missing an E2E test which sets a server tag?

I'll push a second patch which updates the docs.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [X] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -333,7 +333,13 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
 		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to get or create floating IP for bastion: %v", err))
 		return errors.Errorf("failed to get or create floating IP for bastion: %v", err)
 	}
-	err = computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP)
+	port, err := computeService.GetManagementPort(instance)
+	if err != nil {
+		err = errors.Errorf("getting management port for bastion: %v", err)
+		handleUpdateOSCError(openStackCluster, err)
+		return err
+	}
+	err = networkingService.AssociateFloatingIP(openStackCluster, fp, port.ID)
 	if err != nil {
 		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to associate floating IP with bastion: %v", err))
 		return errors.Errorf("failed to associate floating IP with bastion: %v", err)
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -27,7 +27,6 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/common/extensions"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
-	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/schedulerhints"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
@@ -610,16 +609,31 @@ func (s *Service) getImageID(imageName string) (string, error) {
 	}
 }
 
-func (s *Service) AssociateFloatingIP(instanceID, floatingIP string) error {
-	opts := floatingips.AssociateOpts{
-		FloatingIP: floatingIP,
+// GetManagementPort returns the port which is used for management and external
+// traffic. Cluster floating IPs must be associated with this port.
+func (s *Service) GetManagementPort(instance *infrav1.Instance) (*ports.Port, error) {
+	mc := metrics.NewMetricPrometheusContext("port", "list")
+	portOpts := ports.ListOpts{
+		DeviceID: instance.ID,
+		FixedIPs: []ports.FixedIPOpts{
+			{
+				IPAddress: instance.IP,
+			},
+		},
+		Limit: 1,
 	}
-	mc := metrics.NewMetricPrometheusContext("floating_ip", "update")
-	err := floatingips.AssociateInstance(s.computeClient, instanceID, opts).ExtractErr()
+	allPages, err := ports.List(s.networkClient, portOpts).AllPages()
 	if mc.ObserveRequest(err) != nil {
-		return err
+		return nil, fmt.Errorf("lookup management port for server %s: %w", instance.ID, err)
 	}
-	return nil
+	allPorts, err := ports.ExtractPorts(allPages)
+	if err != nil {
+		return nil, err
+	}
+	if len(allPorts) < 1 {
+		return nil, fmt.Errorf("did not find management port for server %s", instance.ID)
+	}
+	return &allPorts[0], nil
 }
 
 func (s *Service) DeleteInstance(eventObject runtime.Object, instanceName string) error {
--- a/pkg/cloud/services/compute/service.go
+++ b/pkg/cloud/services/compute/service.go
@@ -39,6 +39,17 @@ type Service struct {
 	logger            logr.Logger
 }
 
+/*
+ NovaMinimumMicroversion is the minimum Nova microversion supported by CAPO
+ 2.53 corresponds to OpenStack Pike
+
+ For the canonical description of Nova microversions, see
+ https://docs.openstack.org/nova/latest/reference/api-microversion-history.html
+
+ CAPO uses server tags, which were added in microversion 2.52.
+*/
+const NovaMinimumMicroversion = "2.53"
+
 // NewService returns an instance of the compute service.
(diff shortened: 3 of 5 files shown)


mdbooth:
> 
> 
> [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and at(@) or hashtag(#) mentions are not allowed in commit messages.
> 
> **The list of commits with invalid commit messages**:
> 
>     * [aa5d63d](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commits/aa5d63d6ee83b0cfaa3845a03d56909679724771) Add a minimum Nova microversion
> 
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

So you can't reference a GH issue in a commit message? That's a horrendous anti-pattern.

pkg/cloud/services/compute/service.go
```diff
@@ -39,6 +39,17 @@ type Service struct {
...
+/*
+ The minimum Nova microversion supported by CAPO
+ 2.53 corresponds to OpenStack Pike
+
+ For the canonical description of Nova microversions, see
+ https://docs.openstack.org/nova/latest/reference/api-microversion-history.html
+
+ CAPO uses server tags, which were added in microversion 2.52.
+*/
+const NovaMinimumMicroversion = "2.53"
```
jichenjc, line 51: I think microversion is >= ? so 2.52 is good here?

mdbooth: I rounded it up to a released version, and Pike is 2.53.

sbueringer:
> > [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and at(@) or hashtag(#) mentions are not allowed in commit messages.
> > **The list of commits with invalid commit messages**:
> > ```
> > * [aa5d63d](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commits/aa5d63d6ee83b0cfaa3845a03d56909679724771) Add a minimum Nova microversion
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).
> 
> So you can't reference a GH issue in a commit message? That's a horrendous anti-pattern.

They are blocking mentioning of users in commit messages because otherwise the users are getting notified in forks etc. I'm not entirely sure if it's the same with the issues. It probably is.

mdbooth:
This seems to be failing on the same thing every time. I'm not going to retest it any more until I've actually tried to debug it locally. Unless you know differently I'm going to assume this could be a real bug.

jichenjc:
#927 is the bug I opened though I don't have too much time to debug until now.. need keep an eye

jichenjc:
#938 merged and supposed to fix flaky gate (we actually gave a few tries and all work) not sure your PR's failure is same issue and @mdbooth can you help to rebase locally and let's give another try? Thanks

jichenjc:

not see it's rebased , let's give another try

mdbooth:
I found the problem: we were failing to associate a floating IP because [the Nova associate floating IP proxy API has been deprecated](https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail,add-associate-floating-ip-addfloatingip-action-deprecated-detail#create-server) and returns 404 since microversion 2.44.

pkg/cloud/services/compute/instance.go
```diff
@@ -610,16 +609,29 @@ func (s *Service) getImageID(imageName string) (string, error) {
 	}
 }
 
-func (s *Service) AssociateFloatingIP(instanceID, floatingIP string) error {
-	opts := floatingips.AssociateOpts{
-		FloatingIP: floatingIP,
+func (s *Service) GetControlPlanePort(instance *infrav1.Instance) (*ports.Port, error) {
```
hidekazuna, line 612: GetPort is better because this is used for getting port of bastion instance also.

pkg/cloud/services/compute/instance.go
```diff
@@ -610,16 +609,29 @@ func (s *Service) getImageID(imageName string) (string, error) {
...
+			},
+		},
+		Limit: 1,
 	}
-	mc := metrics.NewMetricPrometheusContext("floating_ip", "update")
-	err := floatingips.AssociateInstance(s.computeClient, instanceID, opts).ExtractErr()
+	allPages, err := ports.List(s.networkClient, portOpts).AllPages()
 	if mc.ObserveRequest(err) != nil {
-		return err
+		return nil, fmt.Errorf("lookup control plane port for server %s: %w", instance.ID, err)
```
hidekazuna, line 625: lookup port for server %s: %w is better because this is used for port of bastion also.

controllers/openstackcluster_controller.go
```diff
@@ -333,7 +333,13 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
 		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to get or create floating IP for bastion: %v", err))
 		return errors.Errorf("failed to get or create floating IP for bastion: %v", err)
 	}
-	err = computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP)
+	port, err := computeService.GetControlPlanePort(instance)
+	if err != nil {
+		err = errors.Errorf("getting control plane port for bastion: %v", err)
```
hidekazuna, line 338: This port is not control plane port, so should be "getting port for bastion: %v".

jichenjc:
/lgtm

thanks for find the root cause
except the minor issues pointed out by hidekazuna 
it's good to go, so leave this for final unhold

pkg/cloud/services/compute/instance.go
mdbooth: I'm at risk of bike-shedding myself here :)

I want to differentiate this port from other ports because it's 'special'. It's the port that will have an IP on NodeCIDR, and it's the only port which we can guarantee is on the correct subnet to associate a floating ip with. It's the port whose first fixed IP will populate instance.IP.

What kind of special? I called it the 'primary' port first, but I didn't like that and changed it to the 'control plane' port. That's an overloaded term, though, and potentially confusing when applied to the bastion host as you point out. I guess it could be the 'management' port?

It's not just any port, though: it's this specific port. It's the <insert name here> port 😂

I don't want to get too caught up on this, though, because I think I'm going to have to come back to this stuff soon anyway. I need to address the issue of `GetIPFromInstance` returning only the first internal port without sorting, so I'm fully expecting to rework this anyway. For example, ideally I think we would determine both this port and instance.ID from cluster.Subnet.

I'm going to go with 'management' port for now, and I'll add an approriate comment block explaining what differentiates it from other ports. If that doesn't work I'll go to GetPort and come back to the issue of what to call it while tackling the wider issue. I'm guessing I should probably also open an issue to discuss the other thing first?

mdbooth:
I forgot I already opened https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/926 about the 'primary' IP being non-deterministic. I should flesh that out.

PR #930, opened 2021-07-09
reassociate Floating IP if first associate fails

### LINKED ISSUE #927: gate: high failure ratio because of Workload cluster (without lb) creation
/kind bug

several PRs suffer same issue and retry 1-2 times still have same issue
```
[1] 
[1] [91m[1m• Failure [1891.637 seconds][0m
[1] e2e tests
[1] [90m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/test/e2e/suites/e2e/e2e_test.go:49[0m
[1]   Workload cluster (without lb)
[1]   [90m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/test/e2e/suites/e2e/e2e_test.go:133[0m
[1]     [91m[1mIt should be creatable and deletable [It][0m
[1]     [90m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/test/e2e/suites/e2e/e2e_test.go:175[0m
[1] 
[1]     [91mTimed out after 1800.001s.
[1]     Expected
[1]         <bool>: false
[1]     to be true[0m
[1] 
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/924/pull-cluster-api-provider-openstack-e2e-test/1410422201777131520/build-log.txt

need check what's happened lead to so high failure ratio

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

jichenjc:

Fixes #927 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
@@ -53,6 +54,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/loadbalancer"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/provider"
+	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
 )
 
 // OpenStackMachineReconciler reconciles a OpenStackMachine object.
@@ -353,7 +355,31 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be got or created: %v", err))
 			return ctrl.Result{}, nil
 		}
-		err = computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP)
+
+		// There is some possibility that floating IP is created but not able to be find right after that
+		// so give a retry here, see #927 for further information.
+		backoff := wait.Backoff{
+			Steps:    10,
+			Duration: time.Second,
+			Factor:   1.25,
+			Jitter:   0.1,
+		}
+		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+			if err := computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP); err != nil {
+				if capoerrors.IsNotFound(err) {
+					// not found, timing issue
+					logger.Info("Floating IP association failed, will retry.", "instance-id", instance.ID, "floating-ip", fp.FloatingIP)
+					return false, nil
+				}
+				// real error occurs
+				return false, err
+			}
+			return true, nil
+		})
+		if err == wait.ErrWaitTimeout {
+			err = fmt.Errorf("floating IP association timeout %s", fp.FloatingIP)
+		}
+
 		if err != nil {
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be associated: %v", err))
 			return ctrl.Result{}, nil


jichenjc:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/927#issuecomment-876994877

is the primary reason for this PR

jichenjc:
@hidekazuna @chrischdi can you help to review this ? THanks

controllers/openstackmachine_controller.go
```diff
@@ -353,10 +353,17 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be got or created: %v", err))
 			return ctrl.Result{}, nil
 		}
+
 		err = computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP)
 		if err != nil {
-			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be associated: %v", err))
-			return ctrl.Result{}, nil
+			if apierrors.IsNotFound(err) {
+				time.Sleep(2 * time.Second)
```
tobiasgiese, line 360: What Do you think about something like this? (untested)

```go
		backoff := wait.Backoff{
			Steps:    10,
			Duration: time.Second,
			Factor:   1.25,
			Jitter:   0.1,
		}
		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
			if err := computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP); err != nil {
				return false, nil
			}
			return true, nil
		})
		if err != nil {
			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be associated: %v", err))
			return ctrl.Result{}, nil
		}
```

chrischdi:
> @hidekazuna @chrischdi can you help to review this ? THanks

Sure, i'll take a look at it tomorrow.

controllers/openstackmachine_controller.go
jichenjc: yes, this seems better approach, thanks~

controllers/openstackmachine_controller.go
```diff
@@ -353,7 +354,23 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
+		backoff := wait.Backoff{
+			Steps:    10,
+			Duration: time.Second,
+			Factor:   1.25,
+			Jitter:   0.1,
+		}
+		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+			if err := computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP); err != nil {
+				logger.Info("Floating IP association failed, will retry.", "instance-id", instance.ID, "floating-ip", fp.FloatingIP)
+				return false, err
```
chrischdi, line 369: returning `false, err` will lead to the backoff to stop instead of doing the next retry.
So we have to log only and do `return false, nil`. The inner error could be set to a error variable defined outside the retryFunc so it could get added as infromation in line 375 if wanted.

controllers/openstackmachine_controller.go
```diff
@@ -353,7 +354,23 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
+			}
+			return true, nil
+		})
+
 		if err != nil {
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be associated: %v", err))
```
chrischdi, line 375: If we always have `return false, nil` inside the `wait.ExponentialBackoff`, the `err` here will always be `timed out waiting for the condition`, as addition we may want the inner error here too?

chrischdi (commented):
backoff seems ok for me, however we will have to change the `return false` part to get retries :-)

Best would be if we have some kind of condition to check prior to associating the floating IP, but this may get hard to debug.

I'm thinking of e.g. the instance must be in `ACTIVE` or the Port of the instance must be in `ACTIVE`, otherwise the associate may fail and we have to do retries. However this may be hard to find out and debug. 
All in all the retry may do it too.

Just for my understanding: the reconcilation is not able to fix the issue too because `handleUpdateMachineError` will set an error state which results in not reconciling to this point again right?

jichenjc:
> 
> 
> backoff seems ok for me, however we will have to change the `return false` part to get retries :-)
> 
> Best would be if we have some kind of condition to check prior to associating the floating IP, but this may get hard to debug.
> 
> I'm thinking of e.g. the instance must be in `ACTIVE` or the Port of the instance must be in `ACTIVE`, otherwise the associate may fail and we have to do retries. However this may be hard to find out and debug.
> All in all the retry may do it too.


I have no clue why the gate suddenly have such issue
maybe some thing changed in the env ... I think retry is easier way as wait for ACTIVE actually is another kind of retry as well but makes code logic more complicated

> 
> Just for my understanding: the reconcilation is not able to fix the issue too because `handleUpdateMachineError` will set an error state which results in not reconciling to this point again right?

yes, check following log you will find the reconcile is skipped at such situation
though I am not sure it's the best approach so I think that's something we need consider later
for now, maybe we can fix the gate first

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/924/pull-cluster-api-provider-openstack-e2e-test/1412692551504236544/artifacts/clusters/bootstrap/controllers/capo-controller-manager/capo-controller-manager-6f4cdd5947-42qv4/manager.log

I0707 09:26:57.481439       1 openstackmachine_controller.go:258] controller-runtime/manager/controller/openstackmachine "msg"="Error state detected, skipping reconciliation"

controllers/openstackmachine_controller.go
jichenjc: ok, will give another thought and reply back here

chrischdi:
> > backoff seems ok for me, however we will have to change the `return false` part to get retries :-)
> > Best would be if we have some kind of condition to check prior to associating the floating IP, but this may get hard to debug.
> > I'm thinking of e.g. the instance must be in `ACTIVE` or the Port of the instance must be in `ACTIVE`, otherwise the associate may fail and we have to do retries. However this may be hard to find out and debug.
> > All in all the retry may do it too.
> 
> I have no clue why the gate suddenly have such issue
> maybe some thing changed in the env ... I think retry is easier way as wait for ACTIVE actually is another kind of retry as well but makes code logic more complicated
> 
> > Just for my understanding: the reconcilation is not able to fix the issue too because `handleUpdateMachineError` will set an error state which results in not reconciling to this point again right?
> 
> yes, check following log you will find the reconcile is skipped at such situation
> though I am not sure it's the best approach so I think that's something we need consider later
> for now, maybe we can fix the gate first
> 
> https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/924/pull-cluster-api-provider-openstack-e2e-test/1412692551504236544/artifacts/clusters/bootstrap/controllers/capo-controller-manager/capo-controller-manager-6f4cdd5947-42qv4/manager.log
> 
> I0707 09:26:57.481439 1 openstackmachine_controller.go:258] controller-runtime/manager/controller/openstackmachine "msg"="Error state detected, skipping reconciliation"

Yes, a long-term alternative would to not set the machine to an error state if floating IP associate fails, to have a unlimited retry (maybe as addition to the backoff :-) ). I think if we get an error which could get retried later we should either not set the machine to error state or be able to retry depending on which kind of error is set on the machine.

controllers/openstackmachine_controller.go
```diff
@@ -353,7 +354,32 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
+		// so give a retry here, see #927 for further information.
+		backoff := wait.Backoff{
+			Steps:    10,
+			Duration: time.Second,
+			Factor:   1.25,
+			Jitter:   0.1,
+		}
+		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+			if err := computeService.AssociateFloatingIP(instance.ID, fp.FloatingIP); err != nil {
+				if apierrors.IsNotFound(err) {
```
chrischdi, line 368: I think you wanted to use `capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"` 's `IsNotFound?

Also maybe it's worth to `return false, nil` on `capoerrors.IsRetryable`? :-)

jichenjc:
>>>Yes, a long-term alternative would to not set the machine to an error state if floating IP associate fails, to have a unlimited retry (maybe as addition to the backoff :-) ). I think if we get an error which could get retried later we should either not set the machine to error state or be able to retry depending on which kind of error is set on the machine.

agree, that need some idempotent consideration and not sure whether it has any side effect ,so for now, let's do as smaller change as we can to move forward :)

controllers/openstackmachine_controller.go
jichenjc: I am not sure IsRetryable is suitable here 
for example, conflict or badrequest we should not retry anymore

chrischdi (commented):
/lgtm

tobiasgiese:
/lgtm
even if i'm not allowed to ;)

chrischdi:
/approve
still have to figure this out :-)

PR #934, opened 2021-07-12
✨ Allow Trunk configuration at a Port level

Xenwar:
Allow Trunk configuration at a Port level

This PR enables and disables port level trunk setting as follows.

- If user provides trunk field (true or false) at Port level, then that value is used for the port.
- If user does not provide trunk field at Port level, then the port inherits the trunk field of the instance. 

Example spec:
```
metadata:
  name: basic-1-control-plane
  namespace: default
spec:
  template:
    spec:
  template:
    spec:
      trunk: true
      ports:
      - description: "Port 1, no trunk specified"
      - description: "Port 2, trunk disabled"
        trunk: false
      - description: "Port 3, trunk enabled"
        trunk: true
      cloudName: openstack-1
      flavor: 2C-4GB-100GB
      identityRef:
        kind: Secret
        name: basic-1-cloud-config
      image: Ubuntu_20.04_node
```
Output:

```
(openstack) network trunk list  --sort-column Name
+--------------------------------------+-------------------------------+--------------------------------------+-----------------------------------------------------------+
| ID                                   | Name                          | Parent Port                          | Description                                               |
+--------------------------------------+-------------------------------+--------------------------------------+-----------------------------------------------------------+
| 03bbab0b-96d9-4226-bded-08ed8b3449ce | basic-1-control-plane-th8l2-0 | 6b6445f6-13af-4669-a301-cc79a79bf8ef | Created by cluster-api-provider-openstack cluster basic-1 |
| c09f979e-333f-4651-a4ff-93d7a72afc2d | basic-1-control-plane-th8l2-2 | fcd0a7ff-138c-4125-9f2a-42a2c32b13e5 | Created by cluster-api-provider-openstack cluster basic-1 |
+--------------------------------------+-------------------------------+--------------------------------------+-----------------------------------------------------------+
```

```
(openstack) port list -c Name -c Description -c ID --sort-column Name
+-------------------------------+----------------------------+--------------------------------------+
| Name                          | Description                | ID                                   |
+-------------------------------+----------------------------+--------------------------------------+
|                               |                            | 1dc33e63-3e4e-492a-86c4-70b304b58986 |
|                               |                            | 35830ee6-d984-4a1f-8e1a-703e024caca8 |
|                               |                            | e650d8ec-4786-4790-a6c9-befe7db57fae |
| basic-1-control-plane-th8l2-0 | Port 1, no trunk specified | 6b6445f6-13af-4669-a301-cc79a79bf8ef |
| basic-1-control-plane-th8l2-1 | Port 2, trunk disabled     | a944ec80-e147-41e8-aa6b-bbef1a47fe37 |
| basic-1-control-plane-th8l2-2 | Port 3, trunk enabled      | fcd0a7ff-138c-4125-9f2a-42a2c32b13e5 |
+-------------------------------+----------------------------+--------------------------------------+
(openstack) 
```

### DIFF
--- a/api/v1alpha4/types.go
+++ b/api/v1alpha4/types.go
@@ -130,6 +130,8 @@ type PortOpts struct {
 	ProjectID           string        `json:"projectId,omitempty"`
 	SecurityGroups      *[]string     `json:"securityGroups,omitempty"`
 	AllowedAddressPairs []AddressPair `json:"allowedAddressPairs,omitempty"`
+	// Enables and disables trunk at port level. If not provided, openStackMachine.Spec.Trunk is inherited.
+	Trunk *bool `json:"trunk,omitempty"`
 
 	// The ID of the host where the port is allocated
 	HostID string `json:"hostId,omitempty"`
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -145,11 +145,16 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster, 
 		}
 	}
 	for i, port := range openStackMachine.Spec.Ports {
+		pOpts := &openStackMachine.Spec.Ports[i]
+		// No Trunk field specified for the port, inherit openStackMachine.Spec.Trunk.
+		if pOpts.Trunk == nil {
+			pOpts.Trunk = &openStackMachine.Spec.Trunk
+		}
 		if port.NetworkID != "" {
 			nets = append(nets, infrav1.Network{
 				ID:       port.NetworkID,
 				Subnet:   &infrav1.Subnet{},
-				PortOpts: &openStackMachine.Spec.Ports[i],
+				PortOpts: pOpts,
 			})
 		} else {
 			nets = append(nets, infrav1.Network{
@@ -157,7 +162,7 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster, 
 				Subnet: &infrav1.Subnet{
 					ID: openStackCluster.Status.Network.Subnet.ID,
 				},
-				PortOpts: &openStackMachine.Spec.Ports[i],
+				PortOpts: pOpts,
 			})
 		}
 	}
@@ -167,6 +172,9 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster, 
 			ID: openStackCluster.Status.Network.ID,
 			Subnet: &infrav1.Subnet{
 				ID: openStackCluster.Status.Network.Subnet.ID,
+			},
+			PortOpts: &infrav1.PortOpts{
+				Trunk: &openStackMachine.Spec.Trunk,
 			},
 		}}
 	}
@@ -181,22 +189,14 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 		if network.ID == "" {
 			return nil, fmt.Errorf("no network was found or provided. Please check your machine configuration and try again")
 		}
-
+		iTags := []string{}
+		if len(instanceSpec.Tags) > 0 {
+			iTags = instanceSpec.Tags
+		}
 		portName := getPortName(instanceSpec.Name, network.PortOpts, i)
-		port, err := s.getOrCreatePort(eventObject, clusterName, portName, network, instanceSpec.SecurityGroups)
+		port, err := s.getOrCreatePort(eventObject, clusterName, portName, network, &instanceSpec.SecurityGroups, iTags)
 		if err != nil {
 			return nil, err
-		}
-
-		if instanceSpec.Trunk {
-			trunk, err := s.getOrCreateTrunk(eventObject, clusterName, instanceSpec.Name, port.ID)
-			if err != nil {
-				return nil, err
-			}
-
-			if err = s.replaceAllAttributesTags(eventObject, trunk.ID, instanceSpec.Tags); err != nil {
-				return nil, err
-			}
 		}
 
 		for _, fip := range port.FixedIPs {
@@ -420,7 +420,7 @@ func (s *Service) getServerNetworks(networkParams []infrav1.NetworkParam) ([]inf
 	return nets, nil
 }
 
-func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string, portName string, net infrav1.Network, instanceSecurityGroups []string) (*ports.Port, error) {
+func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string, portName string, net infrav1.Network, instanceSecurityGroups *[]string, tags []string) (*ports.Port, error) {
 	mc := metrics.NewMetricPrometheusContext("port", "list")
 	allPages, err := ports.List(s.networkClient, ports.ListOpts{
 		Name:      portName,
@@ -467,7 +467,7 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 
 		// inherit port security groups from the instance if not explicitly specified
 		if securityGroups == nil {
-			securityGroups = &instanceSecurityGroups
+			securityGroups = instanceSecurityGroups
 		}
 	}
 
@@ -523,6 +523,18 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+	if portOpts.Trunk != nil && *portOpts.Trunk {
+		trunk, err := s.getOrCreateTrunk(eventObject, clusterName, port.Name, port.ID)
+		if err != nil {
+			record.Warnf(eventObject, "FailedCreateTrunk", "Failed to create trunk for port %s: %v", portName, err)
+			return nil, err
+		}
+		if err = s.replaceAllAttributesTags(eventObject, trunk.ID, tags); err != nil {
+			record.Warnf(eventObject, "FailedReplaceTags", "Failed to replace trunk tags %s: %v", portName, err)
+			return nil, err
+		}
+	}
+
 	return port, nil
 }
 
@@ -548,11 +560,11 @@ func (s *Service) getOrCreateTrunk(eventObject runtime.Object, clusterName, trun
 		PortID: portID,
 	}).AllPages()
 	if mc.ObserveRequest(err) != nil {
-		return nil, fmt.Errorf("searching for existing trunk for server: %v", err)
+		return nil, fmt.Errorf("searching for existing trunk for port: %v", err)
 	}
 	trunkList, err := trunks.ExtractTrunks(allPages)
 	if err != nil {
-		return nil, fmt.Errorf("searching for existing trunk for server: %v", err)
+		return nil, fmt.Errorf("searching for existing trunk for port: %v", err)
 	}
 
 	if len(trunkList) != 0 {
--- a/api/v1alpha4/zz_generated.deepcopy.go
+++ b/api/v1alpha4/zz_generated.deepcopy.go
@@ -812,6 +812,11 @@ func (in *PortOpts) DeepCopyInto(out *PortOpts) {
 		*out = make([]AddressPair, len(*in))
 		copy(*out, *in)
 	}
+	if in.Trunk != nil {
+		in, out := &in.Trunk, &out.Trunk
+		*out = new(bool)
+		**out = **in
+	}
 	if in.Profile != nil {
 		in, out := &in.Profile, &out.Profile
 		*out = make(map[string]string, len(*in))
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
@@ -1322,6 +1322,10 @@ spec:
                               type: array
                             tenantId:
                               type: string
+                            trunk:
+                              description: Enables and disables trunk at port level.
+                                If not provided, openStackMachine.Spec.Trunk is inherited.
+                              type: boolean
                             vnicType:
                               description: The virtual network interface card (vNIC)
                                 type that is bound to the neutron port.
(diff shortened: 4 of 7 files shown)


dulek:
This is awesome, I'm just interested if the logic here is fine. I'd expect that `trunk` setting at instance level would just change the default behavior for all ports but you could still be able to overwrite it by specifying it explicitly on the port level. If I understand the proposal correctly, currently the logic only allows that when instance level `trunk` is `true`.

Xenwar:
> This is awesome, I'm just interested if the logic here is fine. I'd expect that `trunk` setting at instance level would just change the default behavior for all ports but you could still be able to overwrite it by specifying it explicitly on the port level. If I understand the proposal correctly, currently the logic only allows that when instance level `trunk` is `true`.

True. At the moment, instance level overrides all ports if it is set to false. 
If you have a use to make the port-level as a source of truth, we can make changes as well.

Xenwar:
We can remove the conditional here https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/934/files#diff-5e7dbfd52305a21a11fb94fab201062ae2c7b851db93c2a8443a1d82d7aa7936R506
Then, ports would dictate trunk setting, regardless instance level value.

dulek:
Thank you, this seems more logical to me. The main reason here is that I'm looking at this from Kuryr perspective which requires you to have trunk set only on the main port of the instance. This means that I can now set instance's `trunk` to false and only set `trunk` to true for that single port that requires it. The previous version of this code would force me to set `trunk` to false for each port that does not require it.

pkg/cloud/services/compute/instance.go
```diff
@@ -508,10 +504,22 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+
+	if portOpts.Trunk {
```
mdbooth, line 508: This is a problem.

Previously if we specified trunk on the instance and a number of ports without any trunk setting, those ports would inherit the trunk setting of the instance. Here we're *always* using the trunk setting of the port, which now defaults to False rather than inheriting the trunk setting of the instance. This will cause a change of behaviour.

I think we want the following behaviour:
1. instance=true, port=undefined, trunk=true
2. instance=true, port=true, trunk=true
3. instance=true, port=false, trunk=false
4. instance=false, port=undefined, trunk=false
5. instance=false, port=true, trunk=true
6. instance=false, port=false, trunk=false

If we define port=undefined to be false (the behaviour here), that breaks case 1.
If we define port=undefined to be true, that breaks case 4.

IOW think we need to make portOpts.Trunk a *bool instead of a bool.

mdbooth (changes requested):

mdbooth:
Also, where are we on testing this kind of functionality?

pkg/cloud/services/compute/instance.go
```diff
@@ -188,23 +191,16 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 			return nil, fmt.Errorf("no network was found or provided. Please check your machine configuration and try again")
 		}
 
+		iTags := []string{}
+		if len(instance.Tags) > 0 {
+			iTags = instance.Tags
+		}
```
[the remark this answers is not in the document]
Xenwar: Tried to pass the instance.Tags as is. However, it is not always passed to the `createInstance`. line 195 is a guard against that.

pkg/cloud/services/compute/instance.go
```diff
@@ -508,10 +504,22 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreatePort", "Created port %s with id %s", port.Name, port.ID)
+
+	if portOpts.Trunk {
+		trunkDescription := names.GetDescription(clusterName)
+		trunk, err := s.getOrCreateTrunk(eventObject, trunkDescription, port.Name, port.ID)
```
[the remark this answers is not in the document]
Xenwar: Yes, this PR is creating a new trunk for every port marked as trunked. 
However, it is configurable as is shown in the description of this PR.

pkg/cloud/services/compute/instance.go
Xenwar: We are not inheriting instance level setting. As discussed with @dulek above, each port makes its own decision, regardless of what the instance says.
for (1) and (4), the check was on if a port is specified via `Portopts`, if not it is set to the instance level https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1673261708882c975a746bd0c2bdd5b62bca673a/pkg/cloud/services/compute/instance.go#L178

Is it so that there are other ways of specifying multiple ports ?

Xenwar:
> Also, where are we on testing this kind of functionality?

I was waiting for the merge of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/935 before adding tests, but add a test for it.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
mdbooth: I don't understand. What breaks?

pkg/cloud/services/compute/instance.go
mdbooth: However, I think we previously *did* inherit the instance setting. Let's take a specific example:

```
      trunk: true
      ports:
      - description: "primary 1"
        foo: bar
      - description: "primary 2"
        foo: baz
```

Without this change, both of these ports will be trunked. This change will cause both of the above ports to now be trunked because `trunk` is not specified, and being a `bool` it will default to `false`. IOW this change causes a change in previously defined behaviour. I think that would be a regression.

I think the fix is simple: if we make `trunk` on the port a `*bool` instead of a `bool` we can define a third behaviour for the undefined case which is compatible with the previous behaviour.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
mdbooth: Please ignore this because it's out of scope for your change and consistent with the previous behaviour. However, now that I've seen it I don't understand why trunks are used this way. That could just be because I don't properly understand trunks, though.

mdbooth (changes requested):
I still think this change would introduce a regression without changing `portopts.trunk` from `bool` to `*bool`.

pkg/cloud/services/compute/instance.go
Xenwar: Thanks. Now, I understood the consequence of not making it a pointer. 
Making changes to address it. Will also update the PR description accordingly.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
Xenwar: Will try to get more specific data on this.

pkg/cloud/services/compute/instance.go
Xenwar: Done.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
Xenwar: Got some logs. 
Cause: When no tag is provided `OpenStackMachineTemplate.spec.template.spec.template.spec.tags`, the following error occurs. 

logs:
`E0716 20:27:52.384885     846 openstackmachine_controller.go:389] controller-runtime/manager/controller/openstackmachine "msg"="OpenStack instance cannot be created: error creating Openstack instance: Missing input for argument [Tags]" "error"="UpdateError" "cluster"="basic-1" "machine"="basic-1-control-plane-wsrg7" "name"="basic-1-control-plane-4p2jc" "namespace"="default" "openStackCluster"="basic-1" "openStackMachine"="basic-1-control-plane-4p2jc" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" `

The problem does not occur for all ports when user specified multiple ports. It only occurs to some of them. Which is the reason why I added that check.

Xenwar:
@mdbooth  added an e2e test

api/v1alpha4/types.go
```diff
@@ -130,6 +130,8 @@ type PortOpts struct {
 	ProjectID           string        `json:"projectId,omitempty"`
 	SecurityGroups      *[]string     `json:"securityGroups,omitempty"`
 	AllowedAddressPairs []AddressPair `json:"allowedAddressPairs,omitempty"`
+	// Enables and disables trunk at port level. Inhertis from the corresponding network level field, if not provided.
```
adduarte, line 133: just a nit. is this comment correct(needed?)  wouldn't "network level field" be openStackMachine.Spec.network.Trunk? 
But the port inherits openStackMachine.Spec.Trunk (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/934/files#diff-5e7dbfd52305a21a11fb94fab201062ae2c7b851db93c2a8443a1d82d7aa7936R146)

Perhaps drop the comment to avoid confusion..

pkg/cloud/services/compute/instance.go
```diff
@@ -151,19 +152,24 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster,
 		}
 	}
 	for i, port := range openStackMachine.Spec.Ports {
+		pOpts := &openStackMachine.Spec.Ports[i]
+		// No Trunk field specified for the port, inherit network level
```
adduarte, line 156: "inherits  openStackMachine.Spec.Trunk"

api/v1alpha4/types.go
Xenwar: True, will correct it. We need the comment is it is needed, visible when running `kubectl explain`
I am also working on PortSecurity, confused the two.

pkg/cloud/services/compute/instance.go
Xenwar: done

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
dulek: @mdbooth: It's a Neutron trunk, which might be a bit confusing. In Neutron a *port* can have a trunk created and then you add other ports as *subports* in that trunk. In Kuryr we use that to dynamically add subports which serve as pods interfaces. What you're proposing is doable, but I'm not sure if I see value in subports added statically. Anyway if we'd want to implement that, I'd rather add a `subports` field in `PortOpts` and get subports defined there.

pkg/cloud/services/compute/instance.go
dulek: Thanks @mdbooth, I missed this detail.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
adduarte: @dulek is the name of the trunk singinificant to kuryr? Does kuryr look up trunks per port (parent port.name for example), or by instance

Xenwar:
Once trunk is supported, a test for trunk with the following content can be added https://gist.github.com/Xenwar/d2a9a120531aef09f7670b57d43e0857

Xenwar:
The trunk feature is tested on separate openstack environment with trunk support.
I was unable to add the E2E test due to the CAPO test infra does not support trunk yet, report https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/961

Once it supports trunk, the following test can be added, PortSecurity is also tested
https://gist.github.com/Xenwar/d2a9a120531aef09f7670b57d43e0857

pkg/cloud/services/compute/instance.go
```diff
@@ -152,19 +152,24 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster,
 		}
 	}
 	for i, port := range openStackMachine.Spec.Ports {
+		pOpts := &openStackMachine.Spec.Ports[i]
+		// No Trunk field specified for the port, inherits openStackMachine.Spec.Trunk.
+		if pOpts.Trunk == nil {
+			pOpts.Trunk = &openStackMachine.Spec.Trunk
+		}
```
tobiasgiese, line 159: You have to create the variable `pOpts` to not override `port.Trunk`, correct?
Or can we do sth like this? And pass `PortOpts: &port` in l.164 and l.172

```suggestion
		// No Trunk field specified for the port, inherits openStackMachine.Spec.Trunk.
		if port.Trunk == nil {
			port.Trunk = &openStackMachine.Spec.Trunk
		}
```

pkg/cloud/services/compute/instance.go
```diff
@@ -527,21 +527,34 @@ func (s *Service) getOrCreatePort(eventObject runtime.Object, clusterName string
...
+			record.Warnf(eventObject, "FailedReplaceTags", "Failed to replace trunk tags %s: %v", portName, err)
+			return nil, err
+		}
+	}
+
 	return port, nil
 }
 
-func (s *Service) getOrCreateTrunk(eventObject runtime.Object, clusterName, trunkName, portID string) (*trunks.Trunk, error) {
+func (s *Service) getOrCreateTrunk(eventObject runtime.Object, description, trunkName, portID string) (*trunks.Trunk, error) {
```
tobiasgiese, line 546: I would keep consistent to the other methods and keep the `clusterName` in the arguments. We're doing this several times like this

tobiasgiese (commented):
two nits

jichenjc:
> 
> 
> The trunk feature is tested on separate openstack environment with trunk support.
> I was unable to add the E2E test due to the CAPO test infra does not support trunk yet, report #961
> 
> Once it supports trunk, the following test can be added, PortSecurity is also tested
> https://gist.github.com/Xenwar/d2a9a120531aef09f7670b57d43e0857

thanks for the info, let's merge this after review comments done,  I agree we can track the issue and fix later on

pkg/cloud/services/compute/instance.go
Xenwar: Thanks. done.

pkg/cloud/services/compute/instance.go
Xenwar: pOpts is created to avoid code duplication. 
The check needs to be done for each Port over the loop.

Xenwar:
I have addressed on the of the comments. @tobiasgiese would you please check again to see if I have addresses the issues. 
Thanks.

pkg/cloud/services/compute/instance.go
tobiasgiese: But you don't have to set `pOpts := &openStackMachine.Spec.Ports[i]` if `port` is already `openStackMachine.Spec.Ports[i]` because of the for loop

Xenwar:
@jichenjc
The last comments are addressed. Do we merge it now ?
Thanks.

jichenjc:
/approve 

thanks a lot for your help~

PR #944, opened 2021-07-19
📖 Adjust README.md to fit v0.4.0 and bump k8s to v1.21.3

### LINKED ISSUE #943: v1alpha4: Update Readme

With the upcoming CAPO release v0.4.0 we have to update the [README.md](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/README.md).
At least the compatibility matrix has to be adapted to the new version.

As the readme is part of the release (i.e., in the release archive), it should be done before we're releasing.

part of #745

/kind feature

tobiasgiese:

Fixes #943

Bumped Kubernetes version in the e2e tests to v1.21.2.
Further, adapted the tables format according to [the CAPI supported version in the CAPI books](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/reference/versions.md) 

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/docs/book/src/clusteropenstack/configuration.md
+++ b/docs/book/src/clusteropenstack/configuration.md
@@ -42,7 +42,7 @@ Note: You can use [the template file](../templates/cluster-template.yaml) by man
 # Using 'external-cloud-provider' flavor
 clusterctl config cluster capi-quickstart \
   --flavor external-cloud-provider \
-  --kubernetes-version v1.20.6 \
+  --kubernetes-version v1.21.3 \
   --control-plane-machine-count=3 \
   --worker-machine-count=1 \
   > capi-quickstart.yaml
@@ -50,7 +50,7 @@ clusterctl config cluster capi-quickstart \
 # Using 'without-lb' flavor
 clusterctl config cluster capi-quickstart \
   --flavor without-lb \
-  --kubernetes-version v1.20.6 \
+  --kubernetes-version v1.21.3 \
   --control-plane-machine-count=1 \
   --worker-machine-count=1 \
   > capi-quickstart.yaml
--- a/README.md
+++ b/README.md
@@ -37,27 +37,35 @@ previous cluster managers such as [kops][kops] and
 
 This provider's versions are compatible with the following versions of Cluster API:
 
-| | Cluster API v1alpha1 (v0.1) | Cluster API v1alpha2 (v0.2) | Cluster API v1alpha3 (v0.3) |
-|-------------------------------------------------|---|---|---|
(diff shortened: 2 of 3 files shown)


chrischdi (commented):
/ok-to-test

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -42,15 +42,15 @@ Note: You can use [the template file](../templates/cluster-template.yaml) by man
 # Using 'external-cloud-provider' flavor
 clusterctl config cluster capi-quickstart \
   --flavor external-cloud-provider \
-  --kubernetes-version v1.20.6 \
+  --kubernetes-version v1.21.2 \
```
chrischdi, line 45: maybe we could consider using `v1.21.3`? (capi is still using v1.21.2 I think?)

tobiasgiese: Correct, CAPI is using v1.21.2, that's why I decided to also use this version.
But okay for me to use the latest stable version

sbueringer: I upgraded to 1.21.2 in CAPI because that was the latest release at that point. But in CAPI we also try to align the version we use for testing to our documentation. Not sure which version is used for testing in CAPO :)

sbueringer: But the main CAPO quickstart doc is in the book anyway. So I guess it makes sense to align to the same version, but it doesn't really matter.

tobiasgiese: @sbueringer the version for testing is also changed in this PR to v1.21.2 according to CAPI :)
I would suggest we not test a newer version as CAPI to avoid edge cases.
WDYT?

sbueringer: Ups, overlooked that file.

It's fine to just use v1.21.2. It's pretty safe that v1.21.3 will work too. Imho it doesn't really matter which one you use

tobiasgiese: Okay, if it really doesn't matter if we use a newer (patch) version than CAPI, then I'll change it to the current stable version

chrischdi: :+1: for directly going to v1.21.3 . Hopefully we don't hit https://github.com/kubernetes/kubernetes/issues/102676 :-) but let's see (would also be the case for v1.21.2)

chrischdi (commented):
/lgtm

PR #883, opened 2021-05-29
🏃 Refactor: do not use networkclient in compute package

hidekazuna:
**What this PR does / why we need it**:

To makes clear the responsibility of networking package,  this PR removes networkClient usage from compute package.

- Moved port related logic in instance.go to newly created port.go in networking package.
- `getServerNetworks` method renamed `GetNetworks` and moved to  network.go
- `getSecurityGroups` method moved to securitygroups.go

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #935, opened 2021-07-12
✨ Add mock client and 2 tests for networking package

macaptain:
Refactors the Service in the networking package to use a client with CRUD functions for manipulating OpenStack resources. This interface is combined with mockgen code generation to make the package unit testable.

Two tests are added to illustrate how writing tests for the package will work. The list / filter networks and subnets functions are also slightly simplified.

**What this PR does / why we need it**: 

A refactor of the code to allow unit tests to be written for code that uses gophercloud to interface with OpenStack.

**Which issue(s) this PR fixes**:
Goes some way to fixing #843.

**Special notes for your reviewer**: 

If you like this change, my plan is to move some of the 'ports' logic to the networking package and write unit tests for it, and then apply the same changes, adding mock clients for the Service in the compute package.

I tried a few things out for making the code unit testable, including using gomega's ghttp server for mocking (too much JSON), or trying to write smaller interfaces (ended up being a lot of boiler plate and manual writing of mocks), and ended up settling on this. It's a similar pattern to what's used in cluster-api-provider-aws (see instance_test.go and the generated mocks: https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/pkg/cloud/services/ec2), as well as using the gomega library for matching (as recommended in the Cluster API book section on Testing.)

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

Issue #843, opened 2021-04-09
Improve unit test coverage


sbueringer:
/kind feature

We now have e2e / integration / conformance tests but basically zero unit tests. I think we should change that.
I think we need at least unit tests on 2 layers:
* CAPO controllers
* OpenStack / gophercloud utils

I would start with an initial implementation to establish patterns how we do unit tests for those layers

Some considerations:
* controller-runtime: fakeclient vs envtest
* how to mock gophercloud
* how to unit test gophercloud

Notes:
* I expect that we have to refactor some of our code to reduce the complexity and make it unit-testable (we can probably adjust the gocyclo linter to  about 10-20 (only if appropriate))

After the initial implementation I would create a list of additional unit tests for follow-up PRs, which I would then flag with help wanted / good first issue.

sbueringer:
/unassign

Didn't have time for this. So let's see if somebody wants to pick it up until the next minor release

sbueringer:
Related to this slack conversation: https://kubernetes.slack.com/archives/CFKJB65G9/p1623237546041600

@macaptain  just fyi, this is the corresponding issue :)

Are you currently on it?

macaptain:
@sbueringer, yes I was hoping to have a go this week.

/assign

sbueringer:
@macaptain Perfect, great!

tobiasgiese:
@macaptain are you still on this?

macaptain:
@tobiasgiese, I completed an initial implementation of mocking the OpenStack API to support unit tests in CAPO (#935).

Per the issue description:

> After the initial implementation I would create a list of additional unit tests for follow-up PRs, which I would then flag with help wanted / good first issue.

I expect we could do this now. @mdbooth had a really promising draft #1030 mocking out the interfaces in the compute module which would also help!

/unassign

PR #950, opened 2021-07-22
🏃 Refactor all network client calls into networking package

macaptain:
This refactors all networking client behavior into the networking package. The network client has been removed from the compute Service struct, instead, we use the networking Service. Ports and trunking logic have been moved out of instance.go in the compute package into port.go and trunk.go in the networking package.

The process for refactoring is simple enough:

1. Delete network client from compute package
2. Move any functions using the network client to the networking package
3. Refactor the functions to call the networking client interface instead of gophercloud directly

This will allow us to write unit tests for ports and trunk functionality using the mocks in the networking package. It also considerably tidies up instance.go, since now all networking client calls are made in the networking package only.

Continuing work on #843, finishes #883 after #935 was merged.

~~This is going to cause some rebasing to be required on open PRs around trunks and port security features, namely #934, #914, #921, but will at least allow these features to have unit tests written for them. I think the rebase should be straightforward, the `getOrCreatePort` function in instance.go has simply moved to `GetOrCreatePort` in the networking package. Alternatively, I don't mind waiting and rebasing this PR if those features are ready to go in.~~

PR is ready to go!

- [x] squashed commits

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -22,19 +22,12 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/gophercloud/gophercloud/openstack/common/extensions"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/schedulerhints"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
-	netext "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/portsbinding"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/portsecurity"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
@@ -47,20 +40,12 @@ import (
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
 	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
-	"sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/names"
 )
 
 const (
 	timeoutInstanceCreate       = 5
 	retryIntervalInstanceStatus = 10 * time.Second
-
-	timeoutTrunkDelete       = 3 * time.Minute
-	retryIntervalTrunkDelete = 5 * time.Second
-
-	timeoutPortDelete       = 3 * time.Minute
-	retryIntervalPortDelete = 5 * time.Second
-
-	timeoutInstanceDelete = 5 * time.Minute
+	timeoutInstanceDelete       = 5 * time.Minute
 )
 
 func (s *Service) CreateInstance(openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, clusterName string, userData string) (instance *InstanceStatus, err error) {
@@ -87,7 +72,7 @@ func (s *Service) CreateInstance(openStackCluster *infrav1.OpenStackCluster, mac
 	}
 
 	if openStackMachine.Spec.Trunk {
-		trunkSupport, err := s.getTrunkSupport()
+		trunkSupport, err := s.networkingService.GetTrunkSupport()
 		if err != nil {
 			return nil, fmt.Errorf("there was an issue verifying whether trunk support is available, please disable it: %v", err)
 		}
@@ -111,7 +96,7 @@ func (s *Service) CreateInstance(openStackCluster *infrav1.OpenStackCluster, mac
 	instanceSpec.Tags = machineTags
 
 	// Get security groups
-	securityGroups, err := s.getSecurityGroups(openStackMachine.Spec.SecurityGroups)
+	securityGroups, err := s.networkingService.GetSecurityGroups(openStackMachine.Spec.SecurityGroups)
 	if err != nil {
 		return nil, err
 	}
@@ -194,7 +179,7 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 			iTags = instanceSpec.Tags
 		}
 		portName := getPortName(instanceSpec.Name, network.PortOpts, i)
-		port, err := s.getOrCreatePort(eventObject, clusterName, portName, network, &instanceSpec.SecurityGroups, iTags)
+		port, err := s.networkingService.GetOrCreatePort(eventObject, clusterName, portName, network, &instanceSpec.SecurityGroups, iTags)
 		if err != nil {
 			return nil, err
 		}
@@ -211,7 +196,7 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 	}
 
 	if instanceSpec.Subnet != "" && accessIPv4 == "" {
-		if err := s.deletePorts(eventObject, portList); err != nil {
+		if err := s.networkingService.DeletePorts(eventObject, portList); err != nil {
 			return nil, err
 		}
 		return nil, fmt.Errorf("no ports with fixed IPs found on Subnet %q", instanceSpec.Subnet)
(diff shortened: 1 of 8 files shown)


macaptain:
Hi @hidekazuna, thanks for reviewing and merging #935. I saw you closed #883. Hopefully this PR does the same thing after #935.

/assign @hidekazuna

jichenjc:
I am not sure how big trouble this can cause to other ongoing PR
maybe we need let those functional PRs merge then do the refactory all-in-one.. 
ok to any suggestions instead

macaptain:
> I am not sure how big trouble this can cause to other ongoing PR
> maybe we need let those functional PRs merge then do the refactory all-in-one..
> ok to any suggestions instead

I agree with this, there is a lot of discussion on those other PRs and I don't want to complicate them any further. I am happy to hold the refactor until they are merged.

hidekazuna:
> I agree with this, there is a lot of discussion on those other PRs and I don't want to complicate them any further. I am happy to hold the refactor until they are merged.

@macaptain Sorry late. Thank you for taking my PR into account. This PR looks fine for me as of now. It is OK to wait for other PRs not to bother them.

macaptain:
Now I'm just waiting for #934 to be merged and then I'll rebase and this PR will be ready for review again.

hidekazuna:
@macaptain Now that #934 has been merged, rebase please, if you do not mind.

pkg/cloud/services/networking/port.go
```diff
@@ -0,0 +1,251 @@
...
+			return nil, nil
+		}
+		return nil, fmt.Errorf("get port %q detail failed: %v", portID, err)
+	}
+	return port, nil
+}
+
+// GetPortFromInstanceIP returns at most one port attached to the instance with given ID
+// and with the IP address provided.
+func (s *Service) GetPortFromInstanceIP(instanceID string, ip string) ([]ports.Port, error) {
```
macaptain, line 57: This is the only function I had to create (to avoid a circular dependency with `GetManagementPort`). All other functions in this file are moved from instance.go.

macaptain:
> @macaptain Now that #934 has been merged, rebase please, if you do not mind.

@hidekazuna - I've rebased and this is ready for review again. Although the diff is quite large, the steps to make this change were not so complicated (they are in the PR description). I think the side-by-side diff will show what has moved. Thanks!

hidekazuna:
> 
> 
> > @macaptain Now that #934 has been merged, rebase please, if you do not mind.
> 
> @hidekazuna - I've rebased and this is ready for review again. Although the diff is quite large, the steps to make this change were not so complicated (they are in the PR description). I think the side-by-side diff will show what has moved. Thanks!

@macaptain Thank s rebasing and advising to review. I will review in a few days.

hidekazuna:
@jichenjc PTAL

jichenjc:
I see some issue reported https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/993
which likely to be a regression of #935 and this PR seems to be a follow up of #935 

not sure whether this PR will make things worse or has conflict to the proposed fix of #993 
need more time to read the code and maybe @macaptain @mdbooth can help us in understand more

before that, I Think how about let's hold this?

hidekazuna:
> before that, I Think how about let's hold this?

@jichenjc I agree to hold this PR. revisit this PR after #993 merged.

mdbooth:
I haven't yet reviewed properly, but in general I'm very much in favour of continuing the refactor started by #935. The regression it introduced was caused by 2 very minor additions (checks for a zero length result) to a couple of networking functions which weren't strictly part of the refactor. However, it enables a pattern of unit testing which we can use to catch exactly this kind of bug in the future. I would like to see the same refactor in the compute package.

I suspect this change is going to conflict with *everything*, so I'd personally be in favour of merging it asap. I will try to review today.

pkg/cloud/services/networking/client.go
```diff
@@ -163,21 +171,55 @@ func (c networkClient) ListPort(opts ports.ListOptsBuilder) ([]ports.Port, error
...
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+	return port, nil
 }
 
 func (c networkClient) DeletePort(id string) error {
-	return ports.Delete(c.serviceClient, id).ExtractErr()
+	mc := metrics.NewMetricPrometheusContext("port", "delete")
+	return mc.ObserveRequest(ports.Delete(c.serviceClient, id).ExtractErr())
```
mdbooth, line 184: This would be a regression because we're now logging an error in prometheus on delete of an object which has already been deleted where we previously were not. I think we need to make this, and all Delete* methods in this interface, call ObserveRequestsIgnoreNotFound.

We have the same issue in all Delete* methods.

pkg/cloud/services/networking/trunk.go
```diff
@@ -0,0 +1,128 @@
...
+	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/names"
+)
+
+const (
+	timeoutTrunkDelete       = 3 * time.Minute
+	retryIntervalTrunkDelete = 5 * time.Second
+)
+
+func (s *Service) GetTrunkSupport() (bool, error) {
```
mdbooth, line 39: Aside: I wonder how expensive this call is. We call it on every create and delete, but it doesn't seem like it's ever going to change. And if it did change, would it break inter-op with servers created before the change? What if it changed between create and delete? 🤔

Not for this PR!

pkg/cloud/services/networking/client.go
```diff
@@ -163,21 +171,55 @@ func (c networkClient) ListPort(opts ports.ListOptsBuilder) ([]ports.Port, error
...
+	}
+	return trunk, nil
+}
+
+func (c networkClient) DeleteTrunk(id string) error {
+	mc := metrics.NewMetricPrometheusContext("trunk", "delete")
+	return mc.ObserveRequest(trunks.Delete(c.serviceClient, id).ExtractErr())
+}
+
+func (c networkClient) ListTrunk(opts trunks.ListOptsBuilder) ([]trunks.Trunk, error) {
```
mdbooth, line 214: Bikeshed: This should be plural: ListTrunks (along with all the other List* methods).

Leave it for now unless you want to change all the others in a separate commit, and it's obviously not super important! It does make me wince, though 😉

pkg/cloud/services/networking/trunk.go
```diff
@@ -0,0 +1,128 @@
...
+		record.Warnf(eventObject, "FailedCreateTrunk", "Failed to create trunk %s: %v", trunkName, err)
+		return nil, err
+	}
+
+	record.Eventf(eventObject, "SuccessfulCreateTrunk", "Created trunk %s with id %s", trunk.Name, trunk.ID)
+	return trunk, nil
+}
+
+func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, trunkID string, tags []string) error {
+	mc := metrics.NewMetricPrometheusContext("trunk", "update")
```
mdbooth, line 83: This is inconsistent. We should move this prometheus logging into the client, consistent with all the other logging.

That is going to inevitably change the logging from "trunk", "update" to something like "attributetags", "update" 🤔 I personally think that's ok, certainly less serious than the difference between logging errors and not logging errors on delete. Interested in other opinions, though.

pkg/cloud/services/networking/port.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		record.Warnf(eventObject, "FailedDeletePort", "Failed to delete port %s with id %s: %v", port.Name, port.ID, err)
+		return err
+	}
+
+	record.Eventf(eventObject, "SuccessfulDeletePort", "Deleted port %s with id %s", port.Name, port.ID)
+	return nil
+}
+
+func (s *Service) DeletePorts(eventObject runtime.Object, nets []servers.Network) error {
```
mdbooth, line 228: nit: I personally wouldn't have moved this method. Its function is to iterate over servers.Network, which is a compute thing. Honestly I'd inline it in createInstance.

pkg/cloud/services/networking/port.go
```diff
@@ -0,0 +1,251 @@
...
+		if err := s.DeletePort(eventObject, n.Port); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Service) GarbageCollectErrorInstancesPort(eventObject runtime.Object, instanceName string) error {
+	portList, err := s.client.ListPort(ports.ListOpts{
+		Name: instanceName,
```
mdbooth, line 239: Not for this PR, but I'm pretty sure this is broken. This is listing ports which have the same name as the instance, but we don't create ports with the same name as the instance. I wonder if we should be deleting these by DeviceID instead 🤔

*NOT FOR THIS PR!*

pkg/cloud/services/compute/service.go
```diff
@@ -33,7 +33,6 @@ type Service struct {
 	projectID         string
 	computeClient     *gophercloud.ServiceClient
 	identityClient    *gophercloud.ServiceClient
-	networkClient     *gophercloud.ServiceClient
```
mdbooth, line 36: 🥳

pkg/cloud/services/compute/instance.go
```diff
@@ -211,7 +196,7 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 	}
 
 	if instanceSpec.Subnet != "" && accessIPv4 == "" {
-		if err := s.deletePorts(eventObject, portList); err != nil {
+		if err := s.networkingService.DeletePorts(eventObject, portList); err != nil {
```
mdbooth, line 199: nit: As mentioned elsewhere, I would personally have left this method here.

mdbooth (changes requested):
This mostly looks good to me. Just some minor issues which should be addressed before merging. Nits and bikesheds are up to you 😉

I would personally be in favour of merging this quickly despite the merge conflicts it will inevitably cause.

jichenjc:
/approve
/hold cancel

ok, let's merge this given agreement and clearance of regression issue
and for all the nits, I guess we have chance to submit follow up PRs

macaptain (commented):
Thanks for the thoughtful review @mdbooth ! I'll be addressing your points in follow-up PRs. I've raised #997 for making the metrics usage consistent.

@jichenjc - thanks for merging. This refactor means it's now possible to write unit tests for some of the network and ports logic, which will make this logic easier to maintain in future.

macaptain:
@jichenjc oops, I managed to hide all my comments in response to @mdbooth :laughing:. I don't have permission to unhide them, so if you can unhide for me, that would be great.

jichenjc:
@macaptain I don't know how to unhide them :( 
@hidekazuna do you know ?

PR #960, opened 2021-08-04
✨ Cleanup and refactor InstanceExists and DeleteInstance

mdbooth:
This PR primarily renames InstanceExists to GetInstanceByName to better represent its purpose.

In renaming this method it became clear that we were using the OpenStack API very inefficiently in DeleteInstance, so we clean that up, too. These changes fall into 2 categories:

* Don't re-fetch an object which we already have
* Handle 404 errors during DELETE operations

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -667,7 +667,7 @@ func (s *Service) DeleteInstance(eventObject runtime.Object, instance *infrav1.I
 	}
 	// get and delete trunks
 	for _, port := range instanceInterfaces {
-		if err = s.deleteAttachInterface(eventObject, instance.ID, port.PortID); err != nil {
+		if err = s.deleteAttachInterface(eventObject, instance, port.PortID); err != nil {
 			return err
 		}
 
@@ -697,7 +697,10 @@ func (s *Service) deletePort(eventObject runtime.Object, portID string) error {
 	err = util.PollImmediate(retryIntervalPortDelete, timeoutPortDelete, func() (bool, error) {
 		mc := metrics.NewMetricPrometheusContext("port", "delete")
 		err := ports.Delete(s.networkClient, port.ID).ExtractErr()
-		if mc.ObserveRequest(err) != nil {
+		if mc.ObserveRequestIgnoreNotFound(err) != nil {
+			if capoerrors.IsNotFound(err) {
+				record.Eventf(eventObject, "SuccessfulDeletePort", "Port %s with id %d did not exist", port.Name, port.ID)
+			}
 			if capoerrors.IsRetryable(err) {
 				return false, nil
 			}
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -811,18 +777,14 @@ func (s *Service) getPort(portID string) (port *ports.Port, err error) {
...
-		return nil
-	}
+func (s *Service) deleteInstance(eventObject runtime.Object, instance *infrav1.Instance) error {
 	mc := metrics.NewMetricPrometheusContext("server", "delete")
-	err = servers.Delete(s.computeClient, instance.ID).ExtractErr()
-	if mc.ObserveRequest(err) != nil {
+	err := servers.Delete(s.computeClient, instance.ID).ExtractErr()
+	if mc.ObserveRequestIgnoreNotFound(err) != nil {
+		if capoerrors.IsNotFound(err) {
+			record.Eventf(eventObject, "SuccessfulDeleteServer", "Server %s with id %s does not exist", instance.Name, instance.ID)
```
jichenjc, line 785: not sure whether we want to add event or a log is fine ... not a bit deal

mandre: I believe it doesn't matter that much as long as we're consistent with the rest of the project, and it seems we currently not logging anything or adding event when the resource we want to delete doesn't exist.

@mdbooth can we perhaps remove the successful deletion events from your PR in case of 404 and add them globally as a new PR?

mandre (commented):
Nice cleanup, yay for reducing API calls and handling 404s.

pkg/cloud/services/compute/instance.go
mdbooth: Also replying to @mandre :)

My thinking here was that a failure to delete due to 404 indicates a race, because in all cases we're deleting an object which we've recently fetched. As we shouldn't be racing with ourselves this should be an unusual event. I think it's sufficiently different to regular the 'try to fetch instance, instance doesn't exist' case that it merits different handling. It certainly merits at least a log message.

I picked SuccessfulDeleteServer to be consistent with the other case where we called Delete and the result was that the server was gone, but it's really a toss-up.

mandre: Not objecting to creating a SuccessfulDeleteServer event in case the resource was deleted prior to the `servers.Delete()` call. My point is about consistency: if we're going to log 404 during deletion with a SuccessfulDelete event, we should be consistent and make sure we also do the same for `deleteAttachInterface` and `deleteTrunk`.

In deleteAttachInterface:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/5fecf27c37822518d5ec43e727345af122f24170/pkg/cloud/services/compute/instance.go#L715-L717

In deleteTrunk:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/5fecf27c37822518d5ec43e727345af122f24170/pkg/cloud/services/compute/instance.go#L746-L748

mdbooth: Done

jichenjc:
I think it's good enough now , someone can help lgtm then let's get this in

tobiasgiese:
@mdbooth you should squash your commits before merging
/lgtm

Issue #926, opened 2021-06-30
Instance.IP is set randomly when a server has multiple networks


mdbooth:
/kind bug

This is from code inspection while working on MAPO: https://github.com/shiftstack/machine-api-provider-openstack
There is no reproducer, I could have misread the code, and/or this may not be relevant to CAPO.

The problem is where we're iterating over a server's Addresses in here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/services/compute/instance.go#L916-L930
```go
		for _, network := range networkList {
			var netInterface networkInterface
			b, _ := json.Marshal(network)
			err = json.Unmarshal(b, &netInterface)
			if err != nil {
				return nil, fmt.Errorf("extract IP from instance err: %v", err)
			}
			if netInterface.Version == 4.0 {
				if netInterface.Type == "floating" {
					addrMap["floating"] = netInterface.Address
				} else {
					addrMap["internal"] = netInterface.Address
				}
			}
		}
```

If the server has multiple networks we will return the last one returned by the API which is not a floating IP as the machine's control plane IP.

The problem here is that as this interface is currently used it doesn't have enough context to determine which network of multiple networks is the control plane network. When we call GetInstance() or InstanceExists() we're passing in an OpenStack identifier. However, to determine which is the control plane IP we additionally need to know which is the control plane network. AFAICT the canonical reference for this would be OpenStackMachine.Spec.Subnet.

I'm currently thinking of a refactor. I think serverToInstance() needs this additional context. I also note that GetIPFromInstance is called by an E2E test, and duplicates a bunch of code. I wonder if it might be cleaner to have it use an exported serverToInstance instead. I haven't got a firm plan in mind, yet.

jichenjc:
looks like the code you pointed out seems you are right, the last one will be selected 
`addrMap["internal"] = netInterface.Address`

 but I think floating ip in current implementation is mandatory? if so if we have 2 networks then
it means likely the internal IP will be selected by some chance? then it might be a severe bug..

mdbooth:
`instance.IP` is currently used in 2 places:

* https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/22f77e62fb271232dc34365ce1b9791e905392b0/controllers/openstackmachine_controller.go#L392-L400
[...]

PR #971, opened 2021-08-16
:sparkles: Refactor: Don't use infrav1.Instance internally

mdbooth:
This is predominantly a non-functional change. It contains some minor
changes to log messages, and omits a check for instance name being empty
prior to delete in OpenStackMachine controller because this check is no
longer relevant when we delete by ID. Apart from that it contains no
(deliberate) functional changes.

Currently we use infrav1.Instance as our internal representation of an
OpenStack instance. There are several problems with this:

* infrav1.Instance is part of the API, so is hard to change for internal
  use
* infrav1.Instance is used to represent both the 'spec' and 'status' of
  an instance
* When used as a spec for the Bastion host, it allows fields to be set
  which will be ignored
* When used as the status of an instance fetched from OpenStack not all
  fields are populated, leading to potential accidental usage errors

This change creates several new types representing different types of
instance data purely for internal use. This has the following
advantages:

* As they are not API types they can be updated easily
* When used in a function signature they make it clear what data is
  being used
* The type system will prevent accidental use of uninitialised data

The new types are:
* InstanceSpec
* InstanceIdentifier
* InstanceStatus
* InstanceNetworkStatus

They are defined and described in `instance_types.go`.

The primary driver for this work is to eventually fix the documented
errors in InstanceNetworkStatus in IP() and FloatingIP().

This is a refactor in support of a future change to fix https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/926. I would like to merge these in separate PRs to separate potential errors introduced by a refactor from the functional changes required to fix the issue.

I suggest the easiest way to read this change is:

* Read the commit message
* Read instance_types.go
* Everything else is pretty much a mechanical consequence

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -196,24 +196,30 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
 		return err
 	}
 
-	instance, err := computeService.GetInstanceByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
+	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
 	if err != nil {
 		return err
 	}
 
-	if instance != nil && instance.FloatingIP != "" {
-		if err = networkingService.DisassociateFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
-			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to disassociate floating IP: %v", err))
-			return errors.Errorf("failed to disassociate floating IP: %v", err)
+	if instanceStatus != nil {
+		instanceNS, err := instanceStatus.NetworkStatus()
+		if err != nil {
+			return err
 		}
-		if err = networkingService.DeleteFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
-			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete floating IP: %v", err))
-			return errors.Errorf("failed to delete floating IP: %v", err)
-		}
-	}
 
-	if instance != nil && instance.Name != "" {
-		if err = computeService.DeleteInstance(openStackCluster, instance); err != nil {
+		floatingIP := instanceNS.FloatingIP()
+		if floatingIP != "" {
+			if err = networkingService.DisassociateFloatingIP(openStackCluster, floatingIP); err != nil {
+				handleUpdateOSCError(openStackCluster, errors.Errorf("failed to disassociate floating IP: %v", err))
+				return errors.Errorf("failed to disassociate floating IP: %v", err)
+			}
+			if err = networkingService.DeleteFloatingIP(openStackCluster, floatingIP); err != nil {
+				handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete floating IP: %v", err))
+				return errors.Errorf("failed to delete floating IP: %v", err)
+			}
+		}
+
+		if err = computeService.DeleteInstance(openStackCluster, instanceStatus.InstanceIdentifier()); err != nil {
 			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete bastion: %v", err))
 			return errors.Errorf("failed to delete bastion: %v", err)
 		}
@@ -308,16 +314,20 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
 		return err
 	}
 
-	instance, err := computeService.GetInstanceByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
+	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
 	if err != nil {
 		return err
 	}
-	if instance != nil {
-		openStackCluster.Status.Bastion = instance
+	if instanceStatus != nil {
+		bastion, err := instanceStatus.APIInstance()
+		if err != nil {
+			return err
+		}
+		openStackCluster.Status.Bastion = bastion
 		return nil
 	}
 
-	instance, err = computeService.CreateBastion(openStackCluster, cluster.Name)
+	instanceStatus, err = computeService.CreateBastion(openStackCluster, cluster.Name)
 	if err != nil {
 		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to reconcile bastion: %v", err))
 		return errors.Errorf("failed to reconcile bastion: %v", err)
@@ -333,7 +343,7 @@ func reconcileBastion(log logr.Logger, osProviderClient *gophercloud.ProviderCli
 		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to get or create floating IP for bastion: %v", err))
 		return errors.Errorf("failed to get or create floating IP for bastion: %v", err)
 	}
-	port, err := computeService.GetManagementPort(instance)
+	port, err := computeService.GetManagementPort(instanceStatus)
 	if err != nil {
 		err = errors.Errorf("getting management port for bastion: %v", err)
 		handleUpdateOSCError(openStackCluster, err)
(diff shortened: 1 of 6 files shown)


controllers/openstackcluster_controller.go
```diff
@@ -196,24 +196,34 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
 		return err
 	}
 
-	instance, err := computeService.GetInstanceByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
+	instanceStatus, err := computeService.GetInstanceByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
```
jichenjc, line 199: not sure we need rename to GetInstanceStatusByName?

pkg/cloud/services/compute/instance_types.go
```diff
@@ -0,0 +1,171 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
jichenjc, line 2: nit : 2021

pkg/cloud/services/compute/instance_types.go
```diff
@@ -0,0 +1,171 @@
...
+func (is *InstanceStatus) State() infrav1.InstanceState {
+	return infrav1.InstanceState(is.server.Status)
+}
+
+func (is *InstanceStatus) SSHKeyName() string {
+	return is.server.KeyName
+}
+
+// APIInstance returns an infrav1.Instance object for use by the API.
+func (is *InstanceStatus) APIInstance() (*infrav1.Instance, error) {
```
jichenjc, line 95: looks like we have more param for Metadata, ConfigDrive etc for Infrav1.Instance
will it be inited here or somewhere else?

controllers/openstackcluster_controller.go
mdbooth: I'm inclined to leave it for now because it's already getting a bit long, but it's easy to change if we decide it's confusing. I think it should be ok, though, because now everything returns InstanceStatus.

pkg/cloud/services/compute/instance_types.go
mdbooth: That's the intention, although I don't currently think we'll add any paramaters to APIInstance() specifically. We will almost definitely add parameters to NetworkStatus(), though.

mdbooth:
I've rebased on to master and fixed the copyright nit. No other changes.

jichenjc:
/approve

ok, I think we can do additional update over time, so far it looks great

controllers/openstackcluster_controller.go
tobiasgiese: +1 for rename it to `GetInstanceStatusByName`

controllers/openstackcluster_controller.go
```diff
@@ -196,24 +196,34 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
...
 	}
 
-	if instance != nil && instance.FloatingIP != "" {
-		if err = networkingService.DisassociateFloatingIP(openStackCluster, instance.FloatingIP); err != nil {
-			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to disassociate floating IP: %v", err))
-			return errors.Errorf("failed to disassociate floating IP: %v", err)
+	instanceNS, err := instanceStatus.NetworkStatus()
+	if err != nil {
+		return err
+	}
```
tobiasgiese, line 207: You should move this block inside the `instanceStatus != nil` condition, as it's only used there

controllers/openstackcluster_controller.go
```diff
@@ -196,24 +196,34 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
...
+
+	if instanceStatus != nil {
+		floatingIP := instanceNS.FloatingIP()
+		if err != nil {
```
tobiasgiese, line 211: You don't define `err` here, you can remove the err handling as no error will be returned

controllers/openstackmachine_controller.go
```diff
@@ -230,15 +230,20 @@ func (r *OpenStackMachineReconciler) reconcileDelete(ctx context.Context, logger
...
-		}
+	if err = computeService.DeleteInstance(openStackMachine, instance.InstanceIdentifier()); err != nil {
+		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("error deleting OpenStack instance %s with ID %s: %v", instance.Name(), instance.ID(), err))
+		return ctrl.Result{}, nil
+	}
+
+	ns, err := instance.NetworkStatus()
+	if err != nil {
+		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("error getting network status for OpenStack instance %s with ID %s: %v", instance.Name(), instance.ID(), err))
+		return ctrl.Result{}, nil
```
tobiasgiese, line 241: Do you need the current status here? If yes, you should call `GetInstanceByName` (or `GetInstanceStatusByName`) again.

tobiasgiese: Also `instance` should be called `instanceStatus` :)

tobiasgiese (commented):
I have a few findings. But all in all it looks good :)

controllers/openstackcluster_controller.go
mdbooth: Whoops. FloatingIP() returned an error in an earlier version and I missed this!

controllers/openstackmachine_controller.go
mdbooth: I don't think we want to fetch again here. Were you thinking of some specific reason we might need to? We'd potentially race with the prior delete, so it might even be a bug to re-fetch.

mdbooth (commented):
Thanks for a thorough review, @tobiasgiese! I'm not in a hurry to land this as I still haven't written the fix for the actual multi-network bug, so I'd prefer to address these before merging.

With 2 people thinking I should rename the GetInstance functions I'll do that, too.

controllers/openstackmachine_controller.go
tobiasgiese: Why we are deleting the instance and after that checking for the instance status? Shouldn't we first check the status (i.e., to get the fips) and if that fails, we don't want to delete the instance. Otherwise we could have orphaned fips, right?

mdbooth: We're creating a InstanceNetworkStatus here so we can get the FloatingIP. This doesn't go back to OpenStack; it's just doing some extra work on the previously returned server object. The only reason it can fail is json parsing. All we need to know is which FloatingIP to delete.

The previous code also had this failure mode, but it was in the return of DeleteInstance()->GetInstance()->serverToInstance()->GetIPFromInstance(). Moving that call up to the 'top level' in the controller is the primary purpose of this refactor, because it means we can pass additional arguments to NetworkStatus() so we can have more context. I don't think it currently matters for FloatingIP(), for for IP() it means we can pass OpenStackCluster as an argument here and use the additional context to determine the correct 'primary' IP from those returned by OpenStack.

mdbooth: Incidentally, for robustness we should probably reverse the order of these deletions. The reason is that if deletion of the server succeeds but the deletion of the floating ip fails, the next reconcile will short-cut because there's no server and we'll leak the floating ip.

This patch is just a refactor, though, so I don't want to fix that here.

tobiasgiese: That's exactly the point I was worried about. We should first delete the floating IPs and the instance right after the successful deletion.
But okay for me to fix that after this refactor PR.

Further, we could maybe improve the error message for the failed json parsing. `error getting network status for OpenStack instance` is a bit misleading, tbh.

tobiasgiese: And starting errors with `error` is a bit redundant. But we have to fix that in the complete repo anyway

controllers/openstackcluster_controller.go
mdbooth: This would also have been a bug!

tobiasgiese: nil pointer incoming :)

tobiasgiese:
/lgtm cancel

waiting for the findings to be updated

mdbooth:
That change is intended to address all review comments. Please check!

tobiasgiese:
/lgtm
👍🏻

PR #973, opened 2021-08-18
:sparkles: Allow clusters without a floating IP for the API server

### LINKED ISSUE #875: no external network but a vlan network  in openstack environment
/kind feature

Hi. We have some customers whose openstack environment does not have an external network. They use vlan network. The management cluster and tenant network are connected through vlan. Thus there is no floating ip. How does OpenStackCluster handle such a situation? Because the external network seems to be necessary in OpenstackCluster Reconcile.

mkjpryor:

This PR allows clusters to be created without a floating IP attached to the API server. As a result, it is also able to modify the reconciliation to be proceed-able even when there are no external networks available.

This is accomplished by adding a new backwards-compatible flag to the API resource - `OpenStackCluster.spec.disableAPIServerFloatingIP` - that determines whether or not to attempt to attach a floating IP to the API server.

Note that in a lot of cases having a floating IP means that the API server is exposed to the internet, which is not always desirable if there is an alternative access route.

This patch is particularly useful for the following cases:

  1. Control plane nodes are to go directly onto a VLAN that is also available outside of the project.
  2. The management cluster lives in the same project as the workload cluster.

Fixes #875 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
@mkjpryor this is great 

can you help fix the CLA and test failure? thanks

api/v1alpha4/openstackcluster_types.go
```diff
@@ -58,14 +58,29 @@ type OpenStackClusterSpec struct {
...
+	// If not using a managed load balancer, cluster configuration will fail without additional
+	// configuration to manage the VIP on the control plane machines, which falls outside of
+	// the scope of this controller.
+	// +optional
+	DisableAPIServerFloatingIP bool `json:"disableAPIServerFloatingIP"`
+
+	// APIServerFloatingIP is the floatingIP which will be associated with the API server.
+	// The floatingIP will be created if it does not already exist.
+	// If not specified, a new floatingIP is allocated.
+	// This field is not used if DisableAPIServerFloatingIP is set.
```
jichenjc, line 83: DisableAPIServerFloatingIP is false ?

api/v1alpha4/openstackcluster_types.go
```diff
@@ -58,14 +58,29 @@ type OpenStackClusterSpec struct {
...
+	// management and workload clusters or when the management cluster is inside the
+	// project network.
+	// This option requires that the API server use a VIP on the cluster network so that the
+	// underlying machines can change without changing ControlPlaneEndpoint.Host.
+	// When using a managed load balancer, this VIP will be managed automatically.
+	// If not using a managed load balancer, cluster configuration will fail without additional
+	// configuration to manage the VIP on the control plane machines, which falls outside of
+	// the scope of this controller.
+	// +optional
+	DisableAPIServerFloatingIP bool `json:"disableAPIServerFloatingIP"`
```
jichenjc, line 78: sorry, is the no floating ip means only fixed ip, no floating ip in tenant network or it must be using provider network?

api/v1alpha4/openstackcluster_types.go
mkjpryor: That should say

> This field is not used if DisableAPIServerFloatingIP is true.

It defaults to false, meaning that a floating IP will be used by default, which is the existing behaviour.

api/v1alpha4/openstackcluster_types.go
mkjpryor: It literally does what it says on the tin and disables the floating IP for the API server. It makes no other assumptions about the networking.

However the cluster will only create successfully when the management cluster and workload cluster are on the same network, which could be a tenant-specific VXLAN or a shared provider VLAN.

That means that it is possible to create a workload cluster whose API server is not accessible, in which case the cluster creation will fail. For example, if you set this flag when using the automatically created cluster network, then the API server load balancer will only exist on the newly created network and will not be accessible by the management cluster.

mkjpryor: What we can't support without a fair amount of extra work, but would be nice to support in the future, is the case where you have two networks - a workload-cluster-specific VXLAN and a VLAN provider "control network".

In this case, the management cluster lives on the "control network", and the worker nodes for the workload cluster live on the cluster-specific VXLAN. The control-plane nodes for the workload cluster are connected to both, with the management cluster accessing the API server via the "control network" (without a floating IP), and the worker nodes accessing the API server via the cluster-specific network.

In order to do this, we would need to make a load balancer for the API *on each network*, and also be able to tell the corresponding `kubeadmcontrolplane` to include IP SANs for each IP in the TLS certificate for the API server. Ideally this wouldn't require any configuration from the user as the IPs for the load balancers won't be known until they are deployed.

jichenjc: ok , I admin I am not fully understand the whole topo here but I think it's anyway a future item
maybe create another issue with `topo picture` will be easy to understand the desired way

for now, with following statement I think we are good enough for this PR.. thanks 

>>>will only create successfully when the management cluster and workload cluster are on the same network

JohnGarbutt:
Looks like a similar approach is taken here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/991

@mkjpryor does the above patch to everything your use case needs?

mkjpryor:
I signed it

mkjpryor:
@jichenjc 

I have signed the CNCF CLA via my organisation - any idea why this hasn't propagated to here yet?

controllers/openstackcluster_controller.go
```diff
@@ -438,44 +438,69 @@ func reconcileNetworkComponents(log logr.Logger, osProviderClient *gophercloud.P
...
+		} else if !openStackCluster.Spec.DisableAPIServerFloatingIP {
+			// If floating IPs are not disabled, get one to use as the VIP for the control plane
+			fp, err := networkingService.GetOrCreateFloatingIP(openStackCluster, clusterName, openStackCluster.Spec.APIServerFloatingIP)
+			if err != nil {
+				handleUpdateOSCError(openStackCluster, errors.Errorf("Floating IP cannot be got or created: %v", err))
+				return errors.Errorf("Floating IP cannot be got or created: %v", err)
+			}
+			host = fp.FloatingIP
+		} else {
+			// This case is not managed for now (i.e. no load balancer + no floating IP)
```
petrutlucian94, line 491: Can we skip this error? That's exactly what we're doing: using kube-vip and setting allowedAddressPairs. This is all handled at the template level: https://paste.opendev.org/raw/809295/

mkjpryor: @petrutlucian94

In the case you are describing, the test if !openStackCluster.Spec.ControlPlaneEndpoint.IsValid() will fail and you will not enter this block. If you get to the point where that error is returned, it actually is an error that will prevent the cluster from deploying.

In the case where you are using kube-vip, you will actually need the ability to specify the fixed IP ahead of time, which this PR doesn't support but could easily be added.

petrutlucian94: Oh, right. This is under the ``!openStackCluster.Spec.ControlPlaneEndpoint.IsValid()`` branch, all good then. Sorry, got distracted by the comment. May be worth updating the comment though.

mkjpryor: @petrutlucian94

Having said that, as discussed before I would prefer to support an `APIServerFixedIP` option rather than requiring `ControlPlaneEndpoint.Host` to contain an IP address. In the case where there is no load balancer, not specifying it would be an error, so this check does need an additional clause.

petrutlucian94: Yep, sounds good to me.

mkjpryor:
@petrutlucian94 

Latest commit adds support for using either `apiServerFixedIP` or `controlPlaneEndpoint.host` to specify a fixed IP for the API server, with the former taking precedence if both are given.

Can you confirm that this works for your use case with kube-vip? Note that my flag is the inverse to yours, i.e. `disableAPIServerFloatingIP` with a default of false. I thought this makes it more obvious that this is an active opt-out of default functionality (which is to assign a floating IP).

petrutlucian94:
> @petrutlucian94
> 
> Latest commit adds support for using either `apiServerFixedIP` or `controlPlaneEndpoint.host` to specify a fixed IP for the API server. In the case where both are given, `apiServerFixedIP` will be used for the fixed IP and `controlPlaneEndpoint.host` will be used as the control plane address, enabling the case we discussed before where `controlPlaneEndpoint.host` is a DNS name pointing to `apiServerFixedIP`.
> 
> Can you confirm that this works for your use case with kube-vip? Note that my flag is the inverse to yours, i.e. `disableAPIServerFloatingIP` with a default of false. I thought this makes it more obvious that this is an active opt-out of default functionality (which is to assign a floating IP).

That's great, thanks! I'll give it a try as soon as possible.

mkjpryor:
@petrutlucian94 Did you have a chance to test with kube-vip yet? I can confirm that these changes still work for my use case with Octavia and dynamically allocated fixed IPs.

@jichenjc What else do we need to get this merged?

Thanks both!

petrutlucian94:
> @petrutlucian94 Did you have a chance to test with kube-vip yet? I can confirm that these changes still work for my use case with Octavia and dynamically allocated fixed IPs.

Not yet, had a very busy schedule today, meeting over meeting :). I'll test your patch first thing tomorrow.

petrutlucian94:
@mkjpryor I've tested the PR and it works as expected for our use case as well (pinned fixed IPs and kube-vip), thanks! I've closed my PR (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/991) since this one is covering more scenarios.

Just one minor nit: I'm wondering if those templates should be updated as well: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/master/templates

mkjpryor:
> @mkjpryor I've tested the PR and it works as expected for our use case as well (pinned fixed IPs and kube-vip), thanks! I've closed my PR (#991) since this one is covering more scenarios.
> 
> Just one minor nit: I'm wondering if those templates should be updated as well: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/master/templates

@petrutlucian94 That is great - thanks for testing!

I have no strong opinion whether the templates are updated or not. I don't use the templates (I am deploying CAPI clusters using Helm), so I don't have much motivation to do it...

mkjpryor:
@jichenjc 

Now that we have confirmed this patch fulfils the desired use cases and it passes the checks, what else do we need to do to get it merged?

jichenjc:
/approve

@hidekazuna do you want to take a look at this (as you closed other PR which is fixed by this PR as well? )

hidekazuna:
@mkjpryor Current code seems include not only `OpenStackCluster.spec.disableAPIServerFloatingIP` but also `OpenStackCluster.spec.apiServerFixedIP`. Could you update PR comment or [documentation](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/docs/book/src/clusteropenstack/configuration.md#optional-configuration)? I think we have to update External network and floating IP or create new section at all.

mkjpryor:
@hidekazuna 

Thanks for looking at this. I will add some text to the docs.

controllers/openstackcluster_controller.go
```diff
@@ -438,44 +438,75 @@ func reconcileNetworkComponents(log logr.Logger, osProviderClient *gophercloud.P
...
 			return errors.Errorf("failed to reconcile load balancer: %v", err)
 		}
 	}
 
+	if !openStackCluster.Spec.ControlPlaneEndpoint.IsValid() {
+		var host string
+		// If there is a load balancer use the floating IP for it if set, falling back to the internal IP
+		switch {
+		case openStackCluster.Spec.ManagedAPIServerLoadBalancer:
+			if openStackCluster.Status.Network.APIServerLoadBalancer.IP != "" {
```
hidekazuna, line 477: If we need load balancer in this case only, we can move L460-470 in this case.

pkg/cloud/services/networking/network.go
```diff
@@ -72,7 +72,10 @@ func (s *Service) ReconcileExternalNetwork(openStackCluster *infrav1.OpenStackCl
 
 	switch len(networkList) {
 	case 0:
-		return fmt.Errorf("external network not found")
+		// Not finding an external network is fine
```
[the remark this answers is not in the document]
mkjpryor: @hidekazuna

As part of this patch, I made a change to the external network reconciliation so that an installation can proceed even if an external network is not present.

I would prefer not to skip the external network reconciliation completely, as there is an edge case which it would be nice to support:

  * A shared "external" network is available that connects to the public internet
  * OpenStack provider manages an internal for the cluster and a router connecting it to the "external" network
  * Control-plane nodes also belong to a shared "control" network, which the management cluster is on

In this case, the router and external network are important for `LoadBalancer` services to work, but there is no need to expose the control-plane nodes to the internet as the management cluster can access them over the "control" network.

I've not tried to configure this with the existing options, and it may not be possible, but it would definitely be nice to support it in the future and skipping the external network reconciliation would mean it definitely won't work.

controllers/openstackcluster_controller.go
mkjpryor: @hidekazuna 

We can't move L460-470 here, because this code only runs if `!openStackCluster.Spec.ControlPlaneEndpoint.IsValid()`.

Consider the case where the user sets `managedAPIServerLoadBalancer: true`, `controlPlaneEndpoint.host: <dns name pointing to floating IP>` and `apiServerFloatingIP: <IP>`. If we moved the code from L460-470 to here, the load-balancer would not be created, even though it would be needed.

pkg/cloud/services/networking/network.go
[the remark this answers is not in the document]
mkjpryor: Also, you might still want to have a bastion with a floating IP. This PR only prevents a floating IP being assigned to the API server.

seanschneeweiss:
fixes #983

api/v1alpha4/openstackcluster_types.go
```diff
@@ -58,16 +58,41 @@ type OpenStackClusterSpec struct {
...
 	APIServerFloatingIP string `json:"apiServerFloatingIP,omitempty"`
 
+	// APIServerFixedIP is the fixed IP which will be associated with the API server.
+	// In the case where the API server has a floating IP but not a managed load balancer,
+	// this field is not used.
+	// If a managed load balancer is used and this field is not specified, a fixed IP will
+	// be dynamically allocated for the load balancer.
+	// If a managed load balancer is not used AND the API server floating IP is disabled,
+	// this field MUST be specified and should correspond to a pre-allocated port that
+	// holds the fixed IP to be used as a VIP.
```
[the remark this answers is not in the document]
mkjpryor: @seanschneeweiss 

It has `omitempty`, so if not given gets set to the empty string. Would you prefer an `+optional` annotation? From the perspective of the required YAML when creating a CRD, it makes not much difference.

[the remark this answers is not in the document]
seanschneeweiss: Ah okay, then it is fine like it is now. Thank you.

mkjpryor:
@hidekazuna 

I had to rebase, but hopefully you are still happy to merge this?

hidekazuna:
> @hidekazuna
> 
> I had to rebase, but hopefully you are still happy to merge this?

Sure, Let's merge.

/lgtm

PR #975, opened 2021-08-20
Add gc for error instance's port

### LINKED ISSUE #976: A port of error instance blocks deleting managed security group
/kind bug

When failed to create instance with error state in OpenStack, currently DeleteInstance logic didn't delete port of the error instance.
Because, error instance's port detached from the instance by OpenStack.

If a port of error instance is still living until deleting managed security group, deleting a security group fails because the port is not deleted and is still associated with the security group.
As a result, OpenStackCluster deletion reconcile loop stuck.

OpenStackMachine controller delete a port with error instance

**Environment:**

- Cluster-API version: v0.4.1
- OpenStack version: ocata
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): v1.20.1

Bo0km4n:
When failed to create instance with error state in OpenStack, currently `DeleteInstance` logic didn't delete port of the error instance.
Because, error instance's port detached from the instance by OpenStack.

If a port of error instance is still living until deleting managed security group, deleting a security group fails because the port is not deleted and is still associated with the security group.
As a result, OpenStackCluster deletion reconcile loop stuck.

To resolve above problem, I added logic of deleting port of error instance.

Fixes #976

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -653,9 +653,30 @@ func (s *Service) GetManagementPort(instanceStatus *InstanceStatus) (*ports.Port
 	return &allPorts[0], nil
 }
 
-func (s *Service) DeleteInstance(eventObject runtime.Object, instance *InstanceIdentifier) error {
+func (s *Service) garbageCollectErrorInstancesPort(eventObject runtime.Object, instance *InstanceStatus) error {
+	pager, err := ports.List(s.networkClient, ports.ListOpts{
+		Name: instance.Name(),
+	}).AllPages()
+	if err != nil {
+		return err
+	}
+	portList, err := ports.ExtractPorts(pager)
+	if err != nil {
+		return err
+	}
+	for _, p := range portList {
+		if err := s.deletePort(eventObject, p.ID); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *Service) DeleteInstance(eventObject runtime.Object, instance *InstanceStatus) error {
 	mc := metrics.NewMetricPrometheusContext("server_os_interface", "list")
-	allInterfaces, err := attachinterfaces.List(s.computeClient, instance.ID).AllPages()
+	instanceIdentifier := instance.InstanceIdentifier()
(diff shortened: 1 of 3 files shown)


jichenjc:
yes, this is a valid fix , I suggest you create an issue so that we can track the fix ,appreciate the report and fix

Bo0km4n:
@jichenjc Thanks, I created an issue about this problem #976

jichenjc:
code logic is reasonable to me

pkg/cloud/services/compute/instance.go
```diff
@@ -653,9 +653,28 @@ func (s *Service) GetManagementPort(instanceStatus *InstanceStatus) (*ports.Port
...
+	}
+	portList, err := ports.ExtractPorts(pager)
+	if err != nil {
+		return err
+	}
+	if len(portList) < 1 {
+		return nil
+	}
+
+	return s.deletePort(eventObject, portList[0].ID)
```
tobiasgiese, line 671: Even if we only support a single port per instance I would suggest to loop over the `portList` and delete all ports. It does not hurt and it feels more stable.
WDYT?
cc @jichenjc

tobiasgiese: Also we don't have to check for `if len(portList) < 1` in this case

Bo0km4n: Thanks @tobiasgiese, I agree it and modify code to loop through portList and delete each port.

Bo0km4n: I fixed it. What about it ? @tobiasgiese

tobiasgiese: Perfect, thanks :)

Please squash your commits. Then I will give my lgtm

Bo0km4n: Ok, I squashed.

tobiasgiese:
/lgtm
Thanks 👍🏻

PR #984, opened 2021-09-03
🐛Fix Multi network support

### LINKED ISSUE #926: Instance.IP is set randomly when a server has multiple networks

mdbooth:

Fixes multi-network support by making 'primary' IP selection deterministic.

Currently contains only an E2E test which reproduces the bug and therefore fails. Note that the current behaviour is non-deterministic, so it is possible the test may occasionally pass. It should usually fail, though.

Fixes #926

/hold

### DIFF
--- a/test/e2e/data/infrastructure-openstack/cluster-template-multi-network.yaml
+++ b/test/e2e/data/infrastructure-openstack/cluster-template-multi-network.yaml
@@ -1,0 +1,223 @@
+---
+apiVersion: cluster.x-k8s.io/v1alpha4
+kind: Cluster
+metadata:
+  name: ${CLUSTER_NAME}
+  labels:
+    cni: "${CLUSTER_NAME}-crs-0"
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks: ["192.168.0.0/16"] # CIDR block used by Calico.
+    serviceDomain: "cluster.local"
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    kind: OpenStackCluster
+    name: ${CLUSTER_NAME}
+  controlPlaneRef:
+    kind: KubeadmControlPlane
+    apiVersion: controlplane.cluster.x-k8s.io/v1alpha4
+    name: ${CLUSTER_NAME}-control-plane
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+kind: OpenStackCluster
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
+  cloudName: ${OPENSTACK_CLOUD}
+  identityRef:
+    name: ${CLUSTER_NAME}-cloud-config
+    kind: Secret
+  managedAPIServerLoadBalancer: true
+  managedSecurityGroups: true
+  nodeCidr: 10.6.0.0/24
+  dnsNameservers:
+  - ${OPENSTACK_DNS_NAMESERVERS}
+  bastion:
+    enabled: true
+    instance:
+      flavor: ${OPENSTACK_BASTION_MACHINE_FLAVOR}
+      image: ${OPENSTACK_BASTION_IMAGE_NAME}
+      sshKeyName: ${OPENSTACK_SSH_KEY_NAME}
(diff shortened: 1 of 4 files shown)


mdbooth:
Something doesn't look right here: the multi-network test appears to fail (successfully), but the suite reports that `capo-e2e: e2e tests MachineDeployment misconfigurations Should fail to create MachineDeployment with invalid subnet or invalid availability zone` failed instead. Need to investigate.

jichenjc:
I took a quick look at test result, seems like a familiar pattern, I knew our test is tested in parellel so not sure it's affected by newly added test case..

let's give another try

mdbooth:
I looked at the logs, and it's kinda interesting. We seem to get a different IP every time we reconcile an OpenStackMachine, so you see the logs full of deleting an LB member because its IP address changed and then failing to add a new one. However, eventually it picks the right one and succeeds. Each member will eventually come up, but it won't stay up. This is enough to make the create/delete cluster test pass, though.

I don't want to over-complicate this test just to exercise a weird non-deterministic bug which we're about to fix. It's still a reasonable test IMHO, so I'll probably leave it. I might look at a unit test instead.

jichenjc:

ok, let's give another try, and need check the logs you mentioned so that we can open an issue to track

mdbooth:
It's ok, now that I think of it the behaviour is obvious and expected given the bug that we're already tracking in #926. This test is really just intended to reproduce that bug. It does reproduce it, but it still doesn't cause the test to fail.

You can find examples here: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/984/pull-cluster-api-provider-openstack-e2e-test/1435117248539791360/artifacts/clusters/bootstrap/controllers/capo-controller-manager/capo-controller-manager-6c94d569b7-gjwp5/manager.log

e.g.

```
I0907 06:36:37.880119       1 loadbalancer.go:279] controller-runtime/manager/controller/openstackmachine "msg"="Deleting load balancer member (because the IP of the machine changed)" "cluster"="cluster-e2e-rsuqi3" "machine"="cluster-e2e-rsuqi3-control-plane-hvjf6" "namespace"="e2e-rsuqi3" "openStackCluster"="cluster-e2e-rsuqi3" "openStackMachine"="cluster-e2e-rsuqi3-control-plane-6vj5q" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" "name"="k8s-clusterapi-cluster-e2e-rsuqi3-cluster-e2e-rsuqi3-kubeapi-6443-cluster-e2e-rsuqi3-control-plane-6vj5q"
```

mdbooth:
Closed in favour of #1004

PR #985, opened 2021-09-06
:bug: Add alpha3 Resources to Scheme

shaardie:
Upgrading from alpha3 to alpha4 is only possible, if the schemes are added.
Otherwise the controller is not able to fetch those resources.

This PR adds the old alpha3 schemes to the controller. We need it, because otherwise it is impossible to use those resources and the conversion does not work properly.

This is pretty similar to https://github.com/kubernetes-sigs/cluster-api/pull/4417

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
+	infrav1old "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha3"
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha4"
 	"sigs.k8s.io/cluster-api-provider-openstack/controllers"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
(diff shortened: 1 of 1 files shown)


shaardie:
Signed the CLA

chrischdi:
@shaardie : do you see any need to add the `clusterv1old`? I don't think they are necessary for the conversion webhook of `infrav1` resources.

shaardie:
> @shaardie : do you see any need to add the `clusterv1old`? I don't think they are necessary for the conversion webhook of `infrav1` resources.

I do not know. I am not that familar with the code base and honestly did not try to add only  `infrav1old`. But I can adapt the PR, if you want it removed.

chrischdi:
@shaardie : as you had been first: if you adapt the change we could merge your PR imho. If you don't care that much about getting this PR merged we could also merge #988 so you would not need to adjust it.

shaardie:
I updated the PR respectively.

shaardie:
The failed test does not look like it has someting to do with this PR.

tobiasgiese:
@shaardie you can retrigger the test manually :)

chrischdi:
@jichenjc or @hidekazuna ptal :-)

chrischdi:
By the way: that the conversion is not working is also visible in the kube-apiserver logs :-)

jichenjc:
/approve

>>>that the conversion is not working is also visible in the kube-apiserver logs :-)

ok, we need go through our logs to see any obvious error :)

PR #986, opened 2021-09-06
:bug: Fix Conversion from alpha3 to alpha4

shaardie:

This patch sets the `Kind` in `IdentityRef` to `Secret` during the conversion
from the `CloudSecret` to the `OpenStackIdentityReference`. Without a `Kind`
set, the admission hook rejects the converted resource, because it is invalid, which makes it
impossible to upgrade.


jichenjc:
/approve

thanks for the fix~

PR #990, opened 2021-09-08
[WIP] ✨load balancers: delete orphaned load balancers from same network

### LINKED ISSUE #842: Cleanup loadbalancers created by ccm

seanschneeweiss:

A cluster deletion also deletes the CCM. Potentially, load balancers managed by CCM remain in the cluster's network. This
blocks network deletion. This change deletes all remaining load balancers running in the network defined by `OpenStackCluster.Status.Network.ID`.

Fixes #842.

This PR also adds some error wrapping, but only for funcs that are related to this PR. If feasible, I could adapt other `errors.Errorf` to `errors.Wrap` in a separate PR.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold
<sub>Sean Schneeweiss <sean.schneeweiss@daimler.com>, Daimler TSS GmbH, [Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)</sub>

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -17,7 +17,6 @@ limitations under the License.
 package loadbalancer
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
@@ -25,6 +24,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 	"sigs.k8s.io/cluster-api/util"
@@ -318,13 +318,18 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 	return nil
 }
 
+// DeleteLoadBalancer deletes a load balancer identified by its name and all child objects (cascade).
 func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
 	loadBalancerName := getLoadBalancerName(clusterName)
 	lb, err := s.checkIfLbExists(loadBalancerName)
 	if err != nil {
 		return err
 	}
+	return s.deleteLoadBalancer(openStackCluster, lb)
+}
 
+// deleteLoadBalancer deletes the load balancer and all child objects (cascade).
+func (s *Service) deleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, lb *loadbalancers.LoadBalancer) error {
 	if lb == nil {
 		return nil
 	}
@@ -332,15 +337,15 @@ func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster,
 	if lb.VipPortID != "" {
 		fip, err := s.networkingService.GetFloatingIPByPortID(lb.VipPortID)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "get floating IP by port ID %s", lb.VipPortID)
 		}
 
 		if fip != nil && fip.FloatingIP != "" {
 			if err = s.networkingService.DisassociateFloatingIP(openStackCluster, fip.FloatingIP); err != nil {
-				return err
+				return errors.Wrapf(err, "dissassociate floating IP %s", fip.FloatingIP)
 			}
 			if err = s.networkingService.DeleteFloatingIP(openStackCluster, fip.FloatingIP); err != nil {
-				return err
+				return errors.Wrapf(err, "delete floating IP %s", fip.FloatingIP)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -318,48 +318,80 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
...
 }
 
+// DeleteLoadBalancersByNetworkID deletes all remaining load balancers in the network
+// identified by openStackCluster.Status.Network.ID.
+func (s *Service) DeleteLoadBalancersByNetworkID(openStackCluster *infrav1.OpenStackCluster) error {
+	if openStackCluster.Status.Network == nil || openStackCluster.Status.Network.ID == "" {
+		return errors.Errorf("expected status.network.id for OpenStackCluster %s/%s", openStackCluster.Namespace, openStackCluster.Name)
+	}
+	lbList, err := loadbalancers.List(s.loadbalancerClient, loadbalancers.ListOpts{
+		VipNetworkID: openStackCluster.Status.Network.ID,
```
seanschneeweiss, line 375: Please let me know if we should get the network id similar to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/221d55b6112e92b90f2f946acaeaffc37baecfd7/pkg/cloud/services/networking/network.go#L148-L149 or do it like I proposed.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -318,48 +318,80 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
...
+		return errors.Wrapf(err, "delete loadbalancer by ID %s", lb.ID)
 	}
 
 	record.Eventf(openStackCluster, "SuccessfulDeleteLoadBalancer", "Deleted load balancer %s with id %s", lb.Name, lb.ID)
 	return nil
 }
 
+// DeleteLoadBalancersByNetworkID deletes all remaining load balancers in the network
+// identified by openStackCluster.Status.Network.ID.
+func (s *Service) DeleteLoadBalancersByNetworkID(openStackCluster *infrav1.OpenStackCluster) error {
```
jichenjc, line 370: one question here about the logic
I think multiple cluster share one network is a valid use case? e.g I give 10.0.0.16/24 and I want my clusters let's say 2 clusters in this network range , then we will delete all LB in case we delete one cluster? 
should we add some annotation or tag to mark this LB belong to this cluster and only handle update to it?

or we already have `project` so only visible to the LB that belongs to this cluster?

seanschneeweiss: The deletion of the load balancer is only executed if `openStackCluster.Spec.NodeCIDR` is set. It looks like the network is anyway deleted some code lines below, so deleting the attached load balancers should be fine.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ed56be16af090c926a610ab03a70d6ba96962573/controllers/openstackcluster_controller.go#L158-L169

Maybe I'm missing something, but having 2 clusters in the same network would require at least one cluster that has an empty `openStackCluster.Spec.NodeCIDR` but some values for `openStackCluster.Spec.Network` and `openStackCluster.Spec.Subnet`.

jichenjc: >>>the network is anyway deleted some code lines below, so deleting the attached load balancers should be fine.

>>>Maybe I'm missing something, but having 2 clusters in the same network would require at least one cluster that has an empty openStackCluster.Spec.NodeCIDR but some values for openStackCluster.Spec.Network and openStackCluster.Spec.Subnet.

not sure whether there are use case like this , e.g 
2 clusters ,first create with NodeCIDR then 2nd use the Network/Subnet etc

so maybe we can document some restriction(not sure we already have? )
e.g if you want to create multiple clusters, the network/subnet lifecycle is not included in CAPO?
and the network/subnet that create through NodeCIDR automatically should not be used for 
further cluster provisioning (otherwise as you mentioned, it might lead to some issue?)

mkjpryor:
Does this have to be load-balancers on the same network? Can we not just delete load-balancers that match the naming convention with the cluster's name substituted?

By default, load-balancers are made with the name `kube_service_<clustername>_<namespace>_<servicename>`.

An alternative approach, which would not be implemented in this provider but probably in the core provider, would be to delete all services of type `LoadBalancer` prior to deleting the cluster. This would also remove the OpenStack load-balancers, but would also work for all the public clouds as well.

A similar approach could be taken for Cinder volumes created using the Cinder CSI (which are also orphaned currently when a cluster is deleted). The generic approach here would be to delete all PVCs before deleting the cluster, which would also work with all other storage providers. Of course, in this case an opt-out (or explicit opt-in) is more important to avoid potential data loss.

mdbooth:
I haven't yet reviewed this properly, but this sounds like something we could cover with an E2E test? For example, in the existing CCM create/delete test, could we provision something in the cluster which creates an LB, I'm guessing a Service? Then we could test that we correctly clean it up.

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: On the contrary, I think we should update DeleteNetwork to delete by ID.

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: I would expect this to be a use case for multiple networks. i.e. If you want private networking between multiple clusters I would still expect each cluster to have its own control plane network, but share one or more additional dataplane networks.

Have I misunderstood the use case here?

seanschneeweiss:
/remove-lifecycle stale

Will rebase and continue to work on this after summer holidays.

mnaser:
@seanschneeweiss do you think you might have sometime to help get this landed?

mnaser:
@seanschneeweiss fyi, openstack magnum deals with this similarly, so this might help you with the logic:

https://github.com/openstack/magnum/blob/0ee8abeed0ab90baee98a92cab7c684313bab906/magnum/common/octavia.py#L75

bl0m1:
I think this would be a good addition to CAPO. Generally I agree with what has been said in previous comments. since CAPO already deletes networks that has been created I se no problem deleting load balancers on the cluster network.

Just to make sure that everyone using CAPO understands this feature a few lines added to the documentation might be useful.

seanschneeweiss:
I'm closing this PR as the parent issue was closed due to inactivity and there might be other resources that should be deleted as well. Not only load balancers.
This unmaintained repository might help as reference: https://github.com/giantswarm/cluster-api-cleaner-openstack

Issue #983, opened 2021-09-02
make floating IPs optional


petrutlucian94:
/kind feature

**Describe the solution you'd like**

When not using Octavia, CAPO expects the control plane endpoint address to be a floating IP and tries to associate it with the instance port: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1f7a85455539f70d52fbc8f4f31241b71ed66e11/controllers/openstackmachine_controller.go#L363-L380

This is not always the case, some Openstack clouds do not use floating IPs at all. We're assigning the control plane address using kube-vip at the guest level, thus avoiding the Octavia dependency.

We'd like to have a config option such as ``useControlPlaneFIP`` to allow skipping the floating IP assignment.

**Anything else you would like to add:**

It's a trivial change, I can go ahead and open a PR. I wanted to check that you're ok with it first.

jichenjc:
@petrutlucian94 thanks for the issue report, I think it's reasonable 
I myself don't know much about kube-vip so will do more homework
the question I am having is if we are using provider network (no tenant network at all) , with the new option, it should be also ok to deploy even without kube-vip ?

blankburian:
Even when using Octavia, it would be beneficial to have such an option to not use FIPs and (in our case) set a fixed VIP instead. We run Octavia on Neutron with OVN and are unable to connect to the FIP from the nodes, as the return packets have the wrong sender address. As a workaround, we use static VIPs instead of FIPs.

petrutlucian94:
> the question I am having is if we are using provider network (no tenant network at all) , with the new option, it should be also ok to deploy even without kube-vip ?

Yep, kube-vip is not required. However, some form of load balancing would be desired in order to allow scaling or upgrading the cluster, in which case Octavia might be used as an alternative.

petrutlucian94:
> Even when using Octavia, it would be beneficial to have such an option to not use FIPs and (in our case) set a fixed VIP instead. We run Octavia on Neutron with OVN and are unable to connect to the FIP from the nodes, as the return packets have the wrong sender address. As a workaround, we use static VIPs instead of FIPs.

[...]

PR #991, opened 2021-09-08
:sparkles: Make floating IPs optional

petrutlucian94:
Make floating IPs optional

At the moment, CAPO expects the control plane address to be a floating
IP. We're going to make it configurable through the "useControlPlaneFIP"
option.

In order to remain backwards compatible, it will default to "true".
If set to "false", we'll just use the control plane IP address without
attempting to create/attach a floating IP.

closes: #983

### DIFF
--- a/api/v1alpha4/openstackcluster_types.go
+++ b/api/v1alpha4/openstackcluster_types.go
@@ -68,6 +68,10 @@ type OpenStackClusterSpec struct {
 	// already exists.
 	APIServerFloatingIP string `json:"apiServerFloatingIP,omitempty"`
 
+	// UseControlPlaneFIP specifies whether to use a floating IP for the
+	// control plane address. Default: true.
+	UseControlPlaneFIP *bool `json:"useControlPlaneFIP,omitempty"`
+
 	// APIServerPort is the port on which the listener on the APIServer
 	// will be created
 	APIServerPort int `json:"apiServerPort,omitempty"`
--- a/api/v1alpha4/zz_generated.deepcopy.go
+++ b/api/v1alpha4/zz_generated.deepcopy.go
@@ -326,6 +326,11 @@ func (in *OpenStackClusterSpec) DeepCopyInto(out *OpenStackClusterSpec) {
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
+	if in.UseControlPlaneFIP != nil {
+		in, out := &in.UseControlPlaneFIP, &out.UseControlPlaneFIP
+		*out = new(bool)
+		**out = **in
+	}
 	if in.APIServerLoadBalancerAdditionalPorts != nil {
 		in, out := &in.APIServerLoadBalancerAdditionalPorts, &out.APIServerLoadBalancerAdditionalPorts
 		*out = make([]int, len(*in))
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
@@ -1655,6 +1655,10 @@ spec:
                 items:
                   type: string
                 type: array
+              useControlPlaneFIP:
+                description: 'UseControlPlaneFIP specifies whether to use a floating
+                  IP for the control plane address. Default: true.'
+                type: boolean
             type: object
           status:
             description: OpenStackClusterStatus defines the observed state of OpenStackCluster.
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclustertemplates.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclustertemplates.yaml
@@ -647,6 +647,10 @@ spec:
                         items:
                           type: string
                         type: array
+                      useControlPlaneFIP:
+                        description: 'UseControlPlaneFIP specifies whether to use
+                          a floating IP for the control plane address. Default: true.'
+                        type: boolean
                     type: object
                 required:
                 - spec
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -354,13 +354,15 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 		return ctrl.Result{}, nil
 	}
 
+	// For backwards compatibility reasons, UseControlPlaneFIP defaults to true.
+	useFIP := openStackCluster.Spec.UseControlPlaneFIP == nil || *openStackCluster.Spec.UseControlPlaneFIP
 	if openStackCluster.Spec.ManagedAPIServerLoadBalancer {
 		err = r.reconcileLoadBalancerMember(logger, osProviderClient, clientOpts, openStackCluster, machine, openStackMachine, instanceNS, clusterName)
 		if err != nil {
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("LoadBalancerMember cannot be reconciled: %v", err))
 			return ctrl.Result{}, nil
 		}
-	} else if util.IsControlPlaneMachine(machine) {
+	} else if util.IsControlPlaneMachine(machine) && useFIP {
 		fp, err := networkingService.GetOrCreateFloatingIP(openStackCluster, clusterName, openStackCluster.Spec.ControlPlaneEndpoint.Host)
 		if err != nil {
 			handleUpdateMachineError(logger, openStackMachine, errors.Errorf("Floating IP cannot be got or created: %v", err))
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -44,21 +44,35 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	loadBalancerName := getLoadBalancerName(clusterName)
 	s.logger.Info("Reconciling load balancer", "name", loadBalancerName)
 
-	lb, err := s.getOrCreateLoadBalancer(openStackCluster, loadBalancerName, openStackCluster.Status.Network.Subnet.ID, clusterName)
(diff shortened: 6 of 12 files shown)


jichenjc:
thanks for the PR
as this will add non-floating ip case, I think it's worthy to either add a new e2e or change one of existing e2e test cases to make sure we are good at such case? that is, no LB, no floating ip but we are still able to create a cluster?

petrutlucian94:
> thanks for the PR
> as this will add non-floating ip case, I think it's worthy to either add a new e2e or change one of existing e2e test cases to make sure we are good at such case? that is, no LB, no floating ip but we are still able to create a cluster?

Makes sense, I'll add some tests as soon as possible.

jichenjc:
>>>Makes sense, I'll add some tests as soon as possible.

another thoughts is current devstack create private, share and public network by default
```
+--------------------------------------+---------+----------------------------------------------------------+
| id                                   | name    | subnets                                                  |
+--------------------------------------+---------+----------------------------------------------------------+
| 366ad8b0-c6e9-44e9-96e7-493bdc02af99 | public  | 850e0a75-48ce-44a2-bedc-a721e19a7489                     |
|                                      |         | fd0fca0b-c46b-416f-ba59-f3156244193b                     |
| 57f6fc00-cf33-4d82-9e40-ee43e0255789 | shared  | a85c5742-4e7d-4164-bf4e-94f90869a93a 192.168.233.0/24    |
| c0d5e6e7-8b2f-44a0-8c08-2025134a80fa | private | 3287c4ef-e328-40fc-8c40-1a5f46f27fab fd94:d415:97fb::/64 |
|                                      |         | 60b05817-359b-4628-b710-b025ceed3a3c 10.0.0.0/26         |
+--------------------------------------+---------+----------------------------------------------------------+
```

so not sure we are able to access the VM we created by using those network (I didn't try devstack provider network before) so if we can't and no other way to do for now, we can make the test case as TODO ..

petrutlucian94:
> so not sure we are able to access the VM we created by using those network (I didn't try devstack provider network before) so if we can't and no other way to do for now, we can make the test case as TODO ..

I agree that this test might rely too much on environment specifics and that we should probably go with a TODO for now.

To access the private networks, I'm usually running the tests within an Openstack VM or manually attach a private network port to the host. As for the public (provider) network, some environments may not allow using it directly as the hypervisors may not have access to the external network, being accessible only to the Neutron gateway. Plus, in most cases DHCP is disabled on the provider network.

I'll send a minor update though, fixing the job failures.

petrutlucian94:
Ok, so the job is passing now but I've realized that it might be worth adding a parameter to the default cluster template file. I'll provide a follow up.

mkjpryor:
This change is very similar to the one I proposed in #973. 

My understanding of the code in this PR is that, in the case where you don’t want a floating IP, you have to specify the fixed IP for the load balancer using `ControlPlaneEndpoint.Host`, i.e. it cannot use a dynamically allocated IP. Is that the case? My patch is able to use a dynamically allocated fixed IP for the load balancer. 

The code in my patch has the opposite problem in that it cannot use a specific fixed IP, but this could be easily fixed by adding an extra `APIServerFixedIP` field that corresponds to `APIServerFloatingIP`. 

What do you think?

petrutlucian94:
> This change is very similar to the one I proposed in #973.
> 
> My understanding of the code in this PR is that, in the case where you don’t want a floating IP, you have to specify the fixed IP for the load balancer using `ControlPlaneEndpoint.Host`, i.e. it cannot use a dynamically allocated IP. Is that the case? My patch is able to use a dynamically allocated fixed IP for the load balancer even in the case where this is no floating IP.
> 
> The code in my patch has the opposite problem in that it cannot use a specific fixed IP, but this could be easily fixed by adding an extra `APIServerFixedIP` field that corresponds to `APIServerFloatingIP`.
> 
> What do you think is the best way to progress this functionality?

Indeed, the changes are very similar. However, this one allows assigning the ``ControlPlaneEndpoint.Host`` address to the load balancer as a static IP. If unset, it will simply dynamically allocate a fixed IP. I'm not sure if we actually need the ``APIServerFixedIP`` option.

Since this PR already covers both cases and is ready to merge, I'd prefer this one.

mkjpryor:
@petrutlucian94 

Have you actually tested that (i.e. the dynamic allocation of a fixed IP only for the LB)? I would have thought that this part of the code - https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/controllers/openstackcluster_controller.go#L441 - would mean a **floating** IP is always allocated unless you specifically pass a fixed IP using `ControlPlaneEndpoint.Host`, even if your new flag is set.  Perhaps I am missing something though?

My patch is able to use a dynamically allocated **fixed**, not floating, IP for the load balancer in the case where the flag is set. 

I also dislike the approach of relying on an IP being in `ControlPlaneEndpoint.Host` without an option to override, as with the floating IP case. What if you want to use a DNS name as the `ControlPlaneEnpoint.Host`? This is supported in the floating IP case by also specifying APIServerFloatingIP`. 

My PR requires a rebase to resolve a very small conflict, and then it too would be ready to resolve. I’m not necessarily advocating my code over yours - I just want to be sure we are covering all the cases.

petrutlucian94:
> @petrutlucian94
> 
> Have you actually tested that (i.e. the dynamic allocation of a fixed IP only for the LB)? I would have thought that this part of the code - https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/controllers/openstackcluster_controller.go#L441 - would mean a **floating** IP is always allocated unless you specifically pass a fixed IP using `ControlPlaneEndpoint.Host`, even if your new flag is set. It won’t be assigned to the load balancer, but would be used as the control plane host. Perhaps I am missing something though?
> 

Oh, you're right. I didn't get to test the dynamic allocation of a fixed IP, I was just looking at the part that's creating the loadbalancer.

> My patch is able to use a dynamically allocated **fixed**, not floating, IP for the load balancer in the case where the flag is set.
> 
> I also dislike the approach of relying on an IP being in `ControlPlaneEndpoint.Host` without an option to override, as with the floating IP case. What if you want to use a DNS name as the `ControlPlaneEnpoint.Host`? This is supported in the floating IP case by also specifying APIServerFloatingIP`.

Good point.

> My PR requires a rebase to resolve a very small conflict, and then it too would be ready to merge. I’m not necessarily advocating my code over yours - I just want to be sure we are covering all the cases.

At this point, I'm ok either way. Let's get one of the PRs in and then have a follow up covering the other scenario (pinned or dynamic fixed IPs). Btw, I took another look at your PR and it's breaking our use case (no Octavia lb and no FIP), we're using kube-vip + allowedAddressPairs, handled at the template level: https://paste.opendev.org/raw/809295/. Still, with minimal changes your PR will also work for us.

mkjpryor:
@petrutlucian94 

If I add the ability to specify `APIServerFixedIP` to my PR, and then remove the error that you commented on, I think that would then work for your use case right?

petrutlucian94:
> @petrutlucian94
> 
> If I add the ability to specify `APIServerFixedIP` to my PR, and then remove the error that you commented on, I think that would then work for your use case right?

Sounds good, thanks! Please let me know if you make those changes, I can help testing it.

mkjpryor:
@petrutlucian94 

Just wanted to check - when you use kube-vip without an Octavia load balancer, are you pre-creating a port to hold the fixed IP that you use for the VIP? I don't see how it works otherwise. I think that is OK as a requirement for this case.

petrutlucian94:
> @petrutlucian94
> 
> Just wanted to check - when you use kube-vip without an Octavia load balancer, are you pre-creating a port to hold the fixed IP that you use for the VIP? I don't see how it works otherwise. I think that is OK as a requirement for this case.

Indeed, I'm creating a placeholder port, otherwise the address will get reused. I'm not using CAPO for that though.

I guess it may be worth adding a CAPO CRD for neutron ports at some point.

mkjpryor:
> Indeed, I'm creating a placeholder port, otherwise the address will get reused. I'm not using CAPO for that though.
> 
> I guess it may be worth adding a CAPO CRD for neutron ports at some point.

I was thinking that CAPO could manage a port for the VIP and make sure allowed address pairs are set for the control plane hosts. I don't think you would need a new CRD for this, just an extra field on `OpenStackCluster.Status` to store the VIP port info.

Coordinating the management of kube-vip on the control plane hosts is more difficult though. What you have implemented is specific to the `KubeadmControlPlane`, and any (as far I am aware) there is no mechanism for the infrastructure provider to inject extra manifests into the control plane.

However I would suggest we do this as a future enhancement and get the PR in as it stands now for the time being.

mkjpryor:
@petrutlucian94 

Once you have tested the changes to my PR, are you happy to close this one as WONTMERGE?

petrutlucian94:
> > Indeed, I'm creating a placeholder port, otherwise the address will get reused. I'm not using CAPO for that though.
> > I guess it may be worth adding a CAPO CRD for neutron ports at some point.
> 
> I was thinking that CAPO could manage a port for the VIP and make sure allowed address pairs are set for the control plane hosts. I don't think you would need a new CRD for this, just an extra field on `OpenStackCluster.Status` to store the VIP port info.
> 
> Coordinating the management of kube-vip on the control plane hosts is more difficult though. What you have implemented is specific to the `KubeadmControlPlane`, and (as far I am aware) there is no mechanism for the infrastructure provider to inject extra manifests into the control plane.
> 
> However I would suggest we do this as a future enhancement and get the PR in as it stands now for the time being.

Definitely, that's outside the scope for now. What's important for us is to be able to use kube-vip, even if that requires a bit of extra work that CAPO is currently unaware of (creating that placeholder port or injecting the kube-vip static pod manifest using the KubeadmConfig/KubeadmControlPlane spec).

> Once you have tested the changes to my PR, are you happy to close this one as WONTMERGE?

For sure, thanks again for making the changes to cover our use case as well!

petrutlucian94:
Closed in favor of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/973, which covers more scenarios.

petrutlucian94:
Closed in favor of #973, which covers more scenarios.

PR #992, opened 2021-09-10
✨Add InstanceStatus.AvailabilityZone()

mdbooth:
This is currently only required by MAPO, where it's used to label the
node with its AZ. It doesn't fetch any additional data from OpenStack,
just exposes it in Go.

CAPO doesn't need it, yet. MAPO does, though, it's 'free', and CAPO will almost certainly require it eventually.

Fixes #

**TODOs**:

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -880,19 +880,18 @@ func (s *Service) GetInstanceStatus(resourceID string) (instance *InstanceStatus
 	if resourceID == "" {
 		return nil, fmt.Errorf("resourceId should be specified to get detail")
 	}
+
 	mc := metrics.NewMetricPrometheusContext("server", "get")
-	server, err := servers.Get(s.computeClient, resourceID).Extract()
+	var server ServerExt
+	err = servers.Get(s.computeClient, resourceID).ExtractInto(&server)
 	if mc.ObserveRequestIgnoreNotFound(err) != nil {
 		if capoerrors.IsNotFound(err) {
 			return nil, nil
 		}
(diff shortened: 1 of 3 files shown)


jichenjc:
@mdbooth out of curiousity, what is MAPO ?
and I think you want CAPO to expose the InstanceStatus.AvailabilityZone()
so that MAPO can use it?

hidekazuna:
Failure domains is introduced in CAPI v1alpha3: https://cluster-api.sigs.k8s.io/developer/providers/v1alpha2-to-v1alpha3.html#optional-support-failure-domains
[CAPZ maps a failure domain to an availability zone](https://capz.sigs.k8s.io/topics/failure-domains.html) and we do as well.

jichenjc:
>>>CAPZ maps a failure domain to an availability zone and we do as well.

yes , we have such mapping for now but I don't know what is MAPO so not sure the use case or we already have something that can fit for the requirement..

mdbooth:
MAPO is what we're renaming our (very!) old fork of CAPO to. It's still using an API based on v1alpha1. Code is currently here: https://github.com/shiftstack/machine-api-provider-openstack. In order to share code with current CAPO we've decided that the best course is to maintain our own API layer, but share all the library code. Our current work is around closing feature gaps.

CAPO can currently explicitly specify a failure domain, which is mapped to an AZ, on create. This works when you know in advance which AZ you want to use. However, there's another case: you need to ensure resources are in the same AZ (e.g. servers and volumes), but you don't care which one. In this case you don't specify an AZ on create and let the scheduler pick. The AZ which was picked is returned with the created server.

This patch simply allows us to read the AZ field which is already being returned.

PR #994, opened 2021-09-13
🐛 Don't make unnecessary REST API calls in getServerNetworks

### LINKED ISSUE #993: Regression: Specifying subnet filter without network fails when it previously did not
/kind bug

In OpenShift we add default networks to machines by subnet tag, e.g.:

```yaml
      networks:
      - filter: {}
        subnets:
        - filter:
            name: cluster-dsal-59p8x-nodes
            tags: openshiftClusterID=cluster-dsal-59p8x
```

This now returns an error. It previously succeeded.

We should find the requested subnet in any network, which was the previous behaviour.

This was regressed by #935 in an interesting way. It turns out that the previous behaviour in this case was to list networks with an empty filter, which returns all networks, then execute a separate subnet query per-network and add all matches subnets. This worked, but was obviously also incredibly inefficient.

#935 was principally about adding a mock interface, but incidentally also modified GetSubnetsByFilter and GetNetworksByFilter to return an error if there were no results. This breaks us, because we're querying every network individually, but only one of them contains the subnet we're looking for. The first network we query which doesn't contain the target subnet will return an error.

I'm not 100% convinced this is a desirable change, but it did highlight a very inefficient use of the OpenStack REST API. We can address this just by making a more efficient API call.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): ac9c503a52d3c2311389fc549455e5abbdaa6a70

mdbooth:

If we specify a network with a subnet filter but no network filter, we now execute a single subnet query rather than executing one for every network. We also no longer execute queries at all when the user has specified a uuid explicitly.

Fixes #993

This PR is intentionally structured in 2 commits:

The first adds unit tests for getServerNetworks (ironically using the excellent work in #935, which also happened to be the commit which introduced the regression). The second fixes getServerNetworks and makes the necessary changes to the unit tests.

The purpose of doing this is to highlight functional changes introduced by the second commit. These can be clearly seen in the unit test changes:

* We make fewer gophercloud calls without changing the returned data
* The target case no longer returns an error

Apart from the target case, we can see that the output of the test cases are unaffected.

### DIFF
--- a/api/v1alpha4/types.go
+++ b/api/v1alpha4/types.go
@@ -55,7 +55,9 @@ type SecurityGroupFilter struct {
 }
 
 type NetworkParam struct {
-	// The UUID of the network. Required if you omit the port attribute.
+	// Optional UUID of the network.
+	// If specified this will not be validated prior to server creation.
+	// Required if `Subnets` specifies a subnet by UUID.
 	UUID string `json:"uuid,omitempty"`
 	// A fixed IPv4 address for the NIC.
 	FixedIP string `json:"fixedIP,omitempty"`
@@ -85,10 +87,12 @@ type Filter struct {
 }
 
 type SubnetParam struct {
-	// The UUID of the network. Required if you omit the port attribute.
+	// Optional UUID of the subnet.
+	// If specified this will not be validated prior to server creation.
+	// If specified, the enclosing `NetworkParam` must also be specified by UUID.
 	UUID string `json:"uuid,omitempty"`
 
-	// Filters for optional network query
+	// Filters for optional subnet query
 	Filter SubnetFilter `json:"filter,omitempty"`
 }
 
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -313,38 +313,71 @@ func applyServerGroupID(opts servers.CreateOptsBuilder, serverGroupID string) se
 
 func (s *Service) getServerNetworks(networkParams []infrav1.NetworkParam) ([]infrav1.Network, error) {
 	var nets []infrav1.Network
+
+	addSubnet := func(netID, subnetID string) {
+		nets = append(nets, infrav1.Network{
+			ID: netID,
+			Subnet: &infrav1.Subnet{
+				ID: subnetID,
+			},
+		})
+	}
+
+	addSubnets := func(networkParam infrav1.NetworkParam, netID string) error {
+		if len(networkParam.Subnets) == 0 && netID != "" {
+			addSubnet(netID, "")
+			return nil
+		}
+
+		for _, subnet := range networkParam.Subnets {
+			// Don't lookup subnet if it was specified explicitly by UUID
+			if subnet.UUID != "" {
+				// If subnet was supplied by UUID then network
+				// must also have been supplied by UUID.
+				if netID == "" {
+					return fmt.Errorf("validation error adding network for subnet %s: "+
+						"network uuid must be specified when subnet uuid is specified", subnet.UUID)
+				}
+
+				addSubnet(netID, subnet.UUID)
+			} else {
+				subnetOpts := subnets.ListOpts(subnet.Filter)
+				if netID != "" {
+					subnetOpts.NetworkID = netID
+				}
+				subnetsByFilter, err := s.networkingService.GetSubnetsByFilter(&subnetOpts)
+				if err != nil {
+					return err
+				}
+				for _, subnetByFilter := range subnetsByFilter {
+					addSubnet(subnetByFilter.NetworkID, subnetByFilter.ID)
+				}
+			}
+		}
+
+		return nil
+	}
+
 	for _, networkParam := range networkParams {
+		// Don't lookup network if we specified one explicitly by UUID
+		// Don't lookup network if we didn't specify a network filter
+		// If there is no explicit network UUID and no network filter,
+		// we will look for subnets matching any given subnet filters in
+		// all networks.
+		if networkParam.UUID != "" || networkParam.Filter == (infrav1.Filter{}) {
+			if err := addSubnets(networkParam, networkParam.UUID); err != nil {
+				return nil, err
+			}
+			continue
+		}
 		opts := networks.ListOpts(networkParam.Filter)
-		opts.ID = networkParam.UUID
 		ids, err := s.networkingService.GetNetworkIDsByFilter(&opts)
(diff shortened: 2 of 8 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -382,38 +382,64 @@ func (s *Service) getSecurityGroups(securityGroupParams []infrav1.SecurityGroupP
...
+		if len(networkParam.Subnets) == 0 && netID != "" {
+			addSubnet(netID, "")
+			return nil
+		}
+
+		for _, subnet := range networkParam.Subnets {
+			// Don't lookup subnet if it was specified explicitly by UUID
+			if subnet.UUID != "" {
+				addSubnet(netID, subnet.UUID)
+				continue
```
iamemilio, line 405: nit pick, IMO continue is just a more confusing way to write if {} else {}, but its a very minor nit pick

mdbooth: Yeah. Sometimes it's worth it if the following block is longer, but I agree an else block would have been clearer here.

I'll change this if I have to update the patch for any other reason.

iamemilio:
overall, LGTM. Really appreciate the test to cover this code.

pkg/cloud/services/compute/instance.go
```diff
@@ -382,38 +382,64 @@ func (s *Service) getSecurityGroups(securityGroupParams []infrav1.SecurityGroupP
...
+	addSubnets := func(networkParam infrav1.NetworkParam, netID string) error {
+		if len(networkParam.Subnets) == 0 && netID != "" {
+			addSubnet(netID, "")
+			return nil
+		}
+
+		for _, subnet := range networkParam.Subnets {
+			// Don't lookup subnet if it was specified explicitly by UUID
+			if subnet.UUID != "" {
+				addSubnet(netID, subnet.UUID)
```
tobiasgiese, line 404: I think you should validate if `netID` is also not an empty string.

tobiasgiese: Or don't we need the network ID here, as the subnet ID is sufficient?

mdbooth: You are right. There are 2 ways we can go here:

* Fetch the subnet and therefore the network
* Return a validation error

I'm inclined to return the error as this is most efficiently fixed by the caller.

/hold

mdbooth: Thinking out loud:

Q: How is this different to specifying a subnet filter without a network?
A: A UUID implies the caller already has the details of a specific resource. We are already trusting the caller if they specify a UUID (we use it without further validation): they will get a server create error instead if it was incorrect.

IOW if the caller specifies a network or subnet by UUID, we are explicitly delegating the validation of those UUIDs to the caller. If the caller specifies a filter we execute it for them, and let them know if something was wrong (i.e. it was empty).

Is that a good enough argument? If so, should we codify it in the docs?

tobiasgiese: I think your assumption is correct. We should also document this preservation, I agree.
We don't have to make any code changes then. But it might be helpful to add a small comment. WDYT?

mdbooth: Thanks. I will:

* update to return a validation here. If it's not sufficiently obvious from the validation error I'll also add a comment.
* add a unit test to cover this error.
* try to find some documentation to update.

On the last point, I want to make the following explicit in the docs:

* The controller will not validate an explicit network or subnet UUID.
* An invalid network or subnet UUID will result in a server create error.
* When specifying subnet UUID, network UUID must also be specified.

mdbooth:
/hold cancel
@tobiasgiese Is this better?

api/v1alpha4/types.go
```diff
@@ -85,10 +87,12 @@ type Filter struct {
...
 
 type SubnetParam struct {
-	// The UUID of the network. Required if you omit the port attribute.
+	// Optional UUID of the subnet.
+	// If specified this will not be validated prior to server creation.
+	// If specified, the enclosing `NetworkParam` must also be specified by UUID.
 	UUID string `json:"uuid,omitempty"`
 
-	// Filters for optional network query
+	// Filters for optional subnet query
```
macaptain, line 95: So you can search for a subnet by ID (your use case), you just have to do it in the SubnetFilter. Makes sense.

macaptain (approved):
The test cases are exhaustive so I can see what's going on and the implementation looks good to me. It's also useful from the tests to see that when specifying UUIDs for networks explicitly, then no OpenStack calls are made. (I expect if UUID is incorrect, then we fail at getOrCreatePort time.)

Something I was reminded of: when a network has multiple subnets and we choose that network with a filter that matches multiple subnets, we end up with a port on every subnet of that network, which is a little bit surprising to me, but it's the way this has always worked. It's good to see the tests cover that too.

pkg/cloud/services/compute/instance_test.go
```diff
@@ -64,3 +71,298 @@ func Test_getPortName(t *testing.T) {
...
+			s := &Service{
+				networkingService: networkingService,
+			}
+
+			got, err := s.getServerNetworks(tt.networkParams)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Service.getServerNetworks() error = %+v, wantErr %+v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
macaptain, line 363: Minor point: I've noticed the tests don't use Gomega for pattern matching.

The Cluster API book suggests we should use Gomega for testing (https://cluster-api.sigs.k8s.io/developer/testing.html#gomega): 

> In Cluster API all the test MUST use Gomega assertions.

You get a bit nicer output on errors if you use Gomega. An example of how to use Gomega for matching test assertions is here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6aa74945653aefffc76528d96fff41fe19c66e9b/pkg/cloud/services/networking/floatingip_test.go#L83-L84

pkg/cloud/services/compute/instance.go
mdbooth: I updated this.

pkg/cloud/services/compute/instance_test.go
mdbooth: As it's trivial and MUST is in caps I updated it.

macaptain:
lgtm

PR #998, opened 2021-09-15
:sparkles: Implement allowAllInClusterTraffic flag

### LINKED ISSUE #995: Ability to specify security group rules for CNIs other than Calico
/kind feature

It would be great to be able to specify the ingress rules for the managed security groups for CNIs other than Calico, defaulting to the Calico rules if not otherwise specified. This is better than pre-creating security groups for each CNI as it still enables the user to easily have separate security groups for each cluster that are tied to the cluster lifecycle.

Assuming that managed security groups are requested, the allow-all egress rules would always be added. Then if custom ingress rules are specified, they would be used for the control plane and worker groups. If no custom rules are specified, the Calico groups used now will be used.

An alternative, maybe even better, solution would be to have a set of supported CNIs that we maintain rulesets for. If a user wants to use an unsupported CNI, they can use unmanaged security groups to do this or submit a patch to add the rules.

mkjpryor:

This PR implements an additional flag that allows the creation of managed security groups that permit all in-cluster traffic. Note that all traffic from outside the cluster (except API server and NodePort services) is still blocked.

This makes it much easier to deploy a CNI other than Calico while still using managed security groups.

In the future, we could look at maintaining rules for multiple CNIs which can be selected from, or allowing users to specify custom rules for the managed security groups. However I think this represents a good catch-all.

Fixes #995 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha4/openstackcluster_types.go
+++ b/api/v1alpha4/openstackcluster_types.go
@@ -75,13 +75,20 @@ type OpenStackClusterSpec struct {
 	// APIServerLoadBalancerAdditionalPorts adds additional ports to the APIServerLoadBalancer
 	APIServerLoadBalancerAdditionalPorts []int `json:"apiServerLoadBalancerAdditionalPorts,omitempty"`
 
-	// ManagedSecurityGroups defines that kubernetes manages the OpenStack security groups
-	// for now, that means that we'll create security group allows traffic to/from
-	// machines belonging to that group based on Calico CNI plugin default network
-	// requirements: BGP and IP-in-IP for master node(s) and worker node(s) respectively.
-	// In the future, we could make this more flexible.
+	// ManagedSecurityGroups determines whether OpenStack security groups for the cluster
+	// will be managed by the OpenStack provider or whether pre-existing security groups will
+	// be specified as part of the configuration.
+	// By default, the managed security groups have rules that allow the Kubelet, etcd, the
+	// Kubernetes API server and the Calico CNI plugin to function correctly.
 	// +optional
 	ManagedSecurityGroups bool `json:"managedSecurityGroups"`
+
+	// AllowAllInClusterTraffic is only used when managed security groups are in use.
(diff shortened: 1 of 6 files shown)


jichenjc:
/lgtm

thanks for the PR ~
I think it's a reasonable change, might need follow up on doc/template etc

hidekazuna:
/approve

I want to hear other reviewer's opinion so hold for a while.

tobiasgiese:
I like this feature 👍🏻 
And like @jichenjc already mentioned, a brief follow up in the docs would be helpful.

Further, you could maybe implement an e2e test to verify the connectivity.

/lgtm

mkjpryor:
@tobiasgiese 

Do you want the tests/docs for this PR? The docs should be super-quick. I actually think the current E2E tests would pass if we set this flag to `true` in the test manifests, because as far as I can see there is no test that checks the Calico connectivity either (other than that the cluster can start). I suppose the test to add would be set the flag and use a different CNI.

tobiasgiese:
> @tobiasgiese
> 
> Do you want the tests/docs for this PR? The docs should be super-quick. I actually think the current E2E tests would pass if we set this flag to `true` in the test manifests, because as far as I can see there is no test that checks specifically for Calico connectivity either (other than that the cluster can start). I suppose the test to add would be set the flag and use a different CNI.

You can add the docs to this PR, sure.
Regarding the e2e tests: I think it's enough to just enable `AllowAllInClusterTraffic` in one of the e2e templates to verify that the CNI is working. I thought about some connectivity tests, but I think that's not necessary (or a bit overengineered).

seanschneeweiss:
Great addition to make the use of CAPO and initial setup easier.

/approve

mkjpryor:
Added some docs and set the flag in one of the E2E test manifests which still pass. 

Let me know ASAP if there is anything else this needs to be merged.

Issue #961, opened 2021-08-05
Enable trunk on CAPO CI devstack


Xenwar:
We are adding features to enable and disable trunk at a port level, https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/934 and https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/929. 

However, the CAPO CI devstack setup does not support trunk and the PR could not be tested properly due to that.

Therefore, change on the `local.conf` or other configuration files to enable trunk could useful for this PR and future trunk related PRs.

Once trunk is supported, a test for trunk with the following content can be added https://gist.github.com/Xenwar/d2a9a120531aef09f7670b57d43e0857

/kind feature

PR #999, opened 2021-09-20
Enable trunking in CI devstack

mdbooth:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #961

/hold
/cc @Xenwar

PR #1000, opened 2021-09-20
[WIP] Test env trunk setup for e2e test

Xenwar:
Tests if trunk setup for e2e is correct.
Once that is verified, tests specific to trunk feature are added.

Tests: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/999
Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/961

### DIFF
--- a/hack/ci/devstack-default-cloud-init.yaml.tpl
+++ b/hack/ci/devstack-default-cloud-init.yaml.tpl
@@ -84,7 +84,7 @@ write_files:
     ENABLED_SERVICES+=,g-api,g-reg
 
     # Octavia-Neutron
-    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3
+    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3,neutron-trunk
     ENABLED_SERVICES+=,neutron-metadata-agent,neutron-qos
     # Octavia
     ENABLED_SERVICES+=,octavia,o-api,o-cw,o-hm,o-hk,o-da
--- a/test/e2e/data/infrastructure-openstack/cluster-template-without-lb.yaml
+++ b/test/e2e/data/infrastructure-openstack/cluster-template-without-lb.yaml
@@ -108,6 +108,7 @@ metadata:
 spec:
   template:
     spec:
+      trunk: true
       flavor: ${OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR}
       image: ${OPENSTACK_IMAGE_NAME}
       sshKeyName: ${OPENSTACK_SSH_KEY_NAME}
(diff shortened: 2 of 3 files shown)


mdbooth:
> 
> 
> Tests if trunk setup for e2e is correct.
> Once that is verified, tests specific to trunk feature are added.
> 
> Tests: #999

As #999 isn't merged you'll need to rebase this on top of it to test.

> Fixes: #961

I think #999 strictly fixes that.

*However*, if you rebase on top of #999 then my commit will be included in this PR and I can just drop #999. That would make most sense, tbh.

mdbooth:
I'm not going to mark this ok-to-test right now because until it's rebased it will definitely fail.

BUT, you should also submit the form to become a member of this community, then you won't need it.

Xenwar:
@mdbooth Thanks for the comment.
Apologies, not quite sure I assumed it was merged :)

jichenjc:
somehow I was told the env we have doesn't support trunk at all (due to hardware ?? )but I don't know the detail
did I remeber wrong thing or we fixed that problem?

Xenwar:
> somehow I was told the env we have doesn't support trunk at all (due to hardware ?? )but I don't know the detail
> did I remeber wrong thing or we fixed that problem?

That seems to be the case. Tried two different versions and both fail due to timeout. I have also, in the past, tried to enable trunk using a different configuration options and got the same result. 
I think you are right in that trunk may not be supported at the hardware level.
```
[2]     [91mTimed out after 1800.000s.
[2]     Expected
[2]         <bool>: false
[2]     to be true[0m
```

mdbooth:
> 
> 
> > somehow I was told the env we have doesn't support trunk at all (due to hardware ?? )but I don't know the detail
> > did I remeber wrong thing or we fixed that problem?
> 
> That seems to be the case. Tried two different versions and both fail due to timeout. I have also, in the past, tried to enable trunk using a different configuration options and got the same result.
> I think you are right in that trunk may not be supported at the hardware level.
> 
> ```
> [2]     �[91mTimed out after 1800.000s.
> [2]     Expected
> [2]         <bool>: false
> [2]     to be true�[0m
> ```

If I get the chance I'll look into this further, but I wouldn't *expect* hardware to make any difference to networking in a single-node devstack. I would expect this to work. Are you sure the test didn't fail for some other reason?

Xenwar:
@mdbooth , I tried two versions of the test and both failed. afaik it is trunk related.  
To be sure, I can comment out the trunk part and test. If it passes, then we can be sure it is the trunk.

Xenwar:
The test has passed with trunk field commented out: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1000/pull-cluster-api-provider-openstack-e2e-test/1440750894147702784

```
+                       // tTrue := true.
                        disablePSTrue := true
                                {
+                                       Description: "primary",
+                                       // Trunk:               &tTrue,.

```
next, will enable both

mdbooth:
In the morning I'll try this change in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1009, see if it works with OVN.

Xenwar:
> In the morning I'll try this change in #1009, see if it works with OVN.

Thanks. Will add comment on it on how to make test the trunk setup.

PR #1003, opened 2021-09-21
🏃 Standardize metrics in networking package

### LINKED ISSUE #997: Prometheus metrics when making API calls are inconsistent
/kind bug

Prometheus metrics for OpenStack API calls are made throughout the codebase as `mc.ObserveRequest(err)` calls. They're inconsistent in places:

* Sometimes we make a call without recording a metric
* Sometimes deletion events which return 404 (not found) are recorded as errors and sometimes not
* ReplaceAttributes call is recorded as 'trunk update', but it should probably be recorded as its own type

Ideally all Prometheus metric counters for OpenStack API calls are consistent.

After refactoring the networking logic into a separate client (#935), metrics recordings for networks can be tidied up from around the codebase and all put in client.go. If we apply a similar refactor to the compute package, we should be able to move all the `metrics` package usage into client implementation files only and see that all the metrics gathering is consistent.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 754371d1ebb96202ba73474f3b9f7e5a4d3eb7d9
- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

macaptain:
This ensures that all calls made by the client in the networking package have a Prometheus metric associated with them.

Delete and Get calls are standardized to not record an error in the metrics if the resource in OpenStack is not found. (However, the not found error will still be returned by the function as nil, err.)

Finally, the replace all attributes tags call is no longer recorded as a trunk update event.

**What this PR does / why we need it**: based off fixing a comment in @mdbooth 's review of #950.

Partially fix #997 (for networking related calls, but compute related calls are still TODO.)

**TODOs**:
- [x] squashed commits

/hold

### DIFF
--- a/pkg/cloud/services/networking/client.go
+++ b/pkg/cloud/services/networking/client.go
@@ -108,7 +108,12 @@ func (c networkClient) RemoveRouterInterface(id string, opts routers.RemoveInter
 }
 
 func (c networkClient) ReplaceAllAttributesTags(resourceType string, resourceID string, opts attributestags.ReplaceAllOptsBuilder) ([]string, error) {
-	return attributestags.ReplaceAll(c.serviceClient, resourceType, resourceID, opts).Extract()
+	mc := metrics.NewMetricPrometheusContext("attributes_tags", "replace_all")
+	tags, err := attributestags.ReplaceAll(c.serviceClient, resourceType, resourceID, opts).Extract()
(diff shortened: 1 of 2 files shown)


mdbooth (approved):
Another advantage of this pattern is that it makes this kind of consistency much easier.

PR #1004, opened 2021-09-22
✨Return all NodeAddresses in OpenStackMachine.Status.Addresses

### LINKED ISSUE #926: Instance.IP is set randomly when a server has multiple networks

### LINKED ISSUE #1001: Report all machine addresses
/kind feature

CAPO should report *all* addresses of a machine in OpenstackMachine.Status.Addresses, not just a single internal IP and a single external IP.

This would make the behaviour of CAPO consistent with the only 2 other providers I checked, AWS and GCP, both of which report all addresses.

*AWS:*
reconcileNormal:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/151fb40cbaa2d87d846f4a0f620c2dbf8b440283/controllers/awsmachine_controller.go#L600

SDKToInstance:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/151fb40cbaa2d87d846f4a0f620c2dbf8b440283/pkg/cloud/services/ec2/instances.go#L780

getInstanceAddresses:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/151fb40cbaa2d87d846f4a0f620c2dbf8b440283/pkg/cloud/services/ec2/instances.go#L791-L818

*GCP*
Reconcile:
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/582ef8e5548ee0357689dd9987bf70ac3839695d/cloud/services/compute/instances/reconcile.go#L45-L61

Another motivation for doing this is removing a use of the problematic instanceStatus.NetworkStatus().IP() and FloatingIP() methods.

mdbooth:
This commit makes 2 closely related changes:
* Add NetworkStatus.Addresses(), and return this in OpenStackMachine
* Update IP() and FloatingIP() to return individual addresses by
  network name

There are also a couple of incidental changes:
* Don't bother with DisassociateFloatingIP before DeleteFloatingIP for
  bastion host
* Delete floating IPs before server during machine delete: avoids
  orphaned floating IPs if floating IP delete fails.

Fixes #1001, fixes #926


mdbooth:
Multiple addresses: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1004/pull-cluster-api-provider-openstack-e2e-test/1440688542236807168/artifacts/clusters/bootstrap/resources/e2e-0vpq66/Machine/cluster-e2e-0vpq66-control-plane-28crg.yaml

controllers/openstackcluster_controller.go
```diff
@@ -206,16 +207,14 @@ func deleteBastion(log logr.Logger, osProviderClient *gophercloud.ProviderClient
...
-		floatingIP := instanceNS.FloatingIP()
-		if floatingIP != "" {
-			if err = networkingService.DisassociateFloatingIP(openStackCluster, floatingIP); err != nil {
-				handleUpdateOSCError(openStackCluster, errors.Errorf("failed to disassociate floating IP: %v", err))
-				return errors.Errorf("failed to disassociate floating IP: %v", err)
-			}
-			if err = networkingService.DeleteFloatingIP(openStackCluster, floatingIP); err != nil {
-				handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete floating IP: %v", err))
-				return errors.Errorf("failed to delete floating IP: %v", err)
+		for _, address := range addresses {
```
jichenjc, line 212: just curious, what's the reason behind the removal of disasociate floating ip for bastion only?

controllers/openstackmachine_controller.go
```diff
@@ -224,28 +224,27 @@ func (r *OpenStackMachineReconciler) reconcileDelete(ctx context.Context, logger
 
 	if instanceStatus == nil {
 		logger.Info("Skipped deleting machine that is already deleted")
```
jichenjc, line 226: maybe add instance ID here?

pkg/cloud/services/compute/instance_types.go
```diff
@@ -132,51 +135,100 @@ func (is *InstanceStatus) InstanceIdentifier() *InstanceIdentifier {
...
-			if vif.Version == 4.0 && vif.Type != "floating" {
-				return vif.Address
+			var addressType corev1.NodeAddressType
+			switch address.Type {
+			case "floating":
+				addressType = corev1.NodeExternalIP
+			case "fixed":
+				addressType = corev1.NodeInternalIP
+			default:
+				// Ignore unknown address types
```
jichenjc, line 174: maybe add some log as reference? though should not reach here..

pkg/cloud/services/compute/instance_types.go
```diff
@@ -132,51 +135,100 @@ func (is *InstanceStatus) InstanceIdentifier() *InstanceIdentifier {
...
 }
+
+// IP returns the first listed ip of an instance for the given network name.
+func (ns *InstanceNetworkStatus) IP(networkName string) string {
+	return ns.firstAddressByNetworkAndType(networkName, corev1.NodeInternalIP)
+}
+
+// FloatingIP returns the first listed floating ip of an instance for the given
+// network name.
+func (ns *InstanceNetworkStatus) FloatingIP(networkName string) string {
```
jichenjc, line 232: I guess I lost some background
usually ,multiple network is reasonable, so we have multiple NICs from various network
but multiple NIC from one network is not so familiar to me, anyway, seems no harm if we 
return the first ip

jichenjc:
/lgtm

overall looks great, just very some minor questions.. thanks~

controllers/openstackcluster_controller.go
mdbooth: Consistency. We've never done it for controllers: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/cf5e5cb826db47cf79ccd3d8bdd2a779f3c584c0/controllers/openstackmachine_controller.go#L246-L251

Also it's redundant, because the floating IP will be disassociated automatically on deletion.

jichenjc: ok, thanks for the confirmation
I don't know the disassociate also happen on deletion.. thanks:)

pkg/cloud/services/compute/instance_types.go
mdbooth: I thought about this, and I'm not sure.

The most likely reason that we would get an unknown address type here is that OpenStack introduced a new address type which we don't use, yet. This would be a new feature, and logging something here because of a new feature we don't support yet would be very noisy. I guess we could log this and IPv6 addresses at debug level just in case?

Interestingly, I was speaking to a Nova developer about `addresses` yesterday, and his impression is that `addresses` is more likely to go away entirely rather than gain new features because it's viewed as a proxy API. The replacement would be to directly query neutron for ports associated with the device id. The only reason I didn't make that change here is that as long as Nova is caching NetworkInfo, getting it from Nova requires 1 less API call.

TL;DR I might add this at debug.

pkg/cloud/services/compute/instance_types.go
mdbooth: We're actually only using this method in 1 place: APIInstance() above. And the only thing that uses APIInstance() is the bastion host. I guess we're not likely to put a second floating IP on the bastion host, so this is really just something to make the behaviour defined if we ever did.

pkg/cloud/services/compute/instance_types.go
mdbooth: I've added this.

controllers/openstackmachine_controller.go
mdbooth: I added this, but then removed it again because after looking at how we create the log object I realised it's redundant. We don't have instance ID here because the machine has been deleted. I had added machine name, but that's already included by the logger: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/cf5e5cb826db47cf79ccd3d8bdd2a779f3c584c0/controllers/openstackmachine_controller.go#L88

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -192,6 +198,59 @@ var _ = Describe("e2e tests", func() {
...
+				Namespace:         namespace.Name,
+				MachineDeployment: *md[0],
+			})
+			controlPlaneMachines := framework.GetControlPlaneMachinesByCluster(ctx, framework.GetControlPlaneMachinesByClusterInput{
+				Lister:      e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
+				ClusterName: clusterName,
+				Namespace:   namespace.Name,
+			})
+			Expect(len(workerMachines)).To(Equal(1))
+			Expect(len(controlPlaneMachines)).To(Equal(3))
```
macaptain, line 250: I see that this creates a four node cluster with the machines having ports on two different networks. Did you also want to test your functions? I think you should be able to see e.g. `controlPlaneMachines[0].Status.Addresses`, if the controller is working, this should show multiple addresses, since they get assigned in reconciliation: 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/fcbf20746b7dfe3b7776c1d7420c341e1225d9cb/controllers/openstackmachine_controller.go#L334

macaptain (approved):
Looks great! Unit tests and e2e tests, we are spoiled.

pkg/cloud/services/compute/instance_types.go
```diff
@@ -132,51 +137,101 @@ func (is *InstanceStatus) InstanceIdentifier() *InstanceIdentifier {
...
+func (ns *InstanceNetworkStatus) Addresses() []corev1.NodeAddress {
+	// We want the returned order of addresses to be deterministic to make
+	// it easy to detect changes and avoid unnecessary updates. Iteration
+	// over maps is non-deterministic, so we explicitly iterate over the
+	// address map in lexical order of network names. This order is
+	// arbitrary.
+	networks := make([]string, 0, len(ns.addresses))
+	for network := range ns.addresses {
+		networks = append(networks, network)
+	}
```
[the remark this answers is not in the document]
mdbooth: I read your comment, looked at my code, and agreed with you.

I was trying to work out why I wrote it this way when I realised the code is actually correct, because the range iterates over only the keys of the map, not the elements.

I think that's a fairly clear indication this code needs another comment 😂 I'll add one.

test/e2e/suites/e2e/e2e_test.go
mdbooth: This was a lazy late addition, tbh. I'd previously written it as a standalone create/delete test which I expected to fail. However, it turns out that it eventually comes up because although IP selection was non-deterministic and failed when it picked the wrong one, the reconciler would just retry and eventually it picked the right one 🤷‍♂️

You are right: this test would be massively improved by testing the resulting addresses.

pkg/cloud/services/compute/instance_types.go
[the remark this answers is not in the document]
mdbooth: Added a comment.

test/e2e/suites/e2e/e2e_test.go
mdbooth: I've expanded the e2e test to explicitly check ports and addresses.

Incidentally, @EmilienM you may want to do something like this in #1008? I'm iterating over openstack ports per machine, which I think it what we were trying to do in that test.

macaptain (approved):

chrischdi (commented):
/lgtm

PR #1006, opened 2021-09-23
✨Add webhook of openstackcluster, no update allowed for now

### LINKED ISSUE #1005: Webhook for openstackcluster
/kind feature

we have webhook for openstackmachines now
we need add webhook for openstackcluster and in turn, don't allow openstackcluster's spec to be
updated ..

jichenjc:

Fixes #1005 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha4/openstackcluster_webhook.go
+++ b/api/v1alpha4/openstackcluster_webhook.go
@@ -17,6 +17,10 @@ limitations under the License.
 package v1alpha4
 
 import (
+	"reflect"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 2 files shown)


jichenjc:
@hidekazuna take a quick look?

seanschneeweiss (commented):
/lgtm

api/v1alpha4/openstackcluster_webhook.go
```diff
@@ -64,10 +68,34 @@ func (r *OpenStackCluster) ValidateCreate() error {
...
 	if r.Spec.IdentityRef != nil && r.Spec.IdentityRef.Kind != defaultIdentityRefKind {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	newOpenStackClusterSpec := newOpenStackCluster["spec"].(map[string]interface{})
+	oldOpenStackClusterSpec := oldOpenStackCluster["spec"].(map[string]interface{})
+
+	// allow changes to ControlPlaneEndpoint
+	delete(oldOpenStackClusterSpec, "controlPlaneEndpoint")
+	delete(newOpenStackClusterSpec, "controlPlaneEndpoint")
```
mdbooth, line 93: Looking at the cluster and machine controllers, we MUST be able to set controlPlaneEndpoint if it is not currently set, but I don't think we support changing it if it is already set.

This change would be an improvement over the current situation so I wouldn't say hold the change for this, but it might be good to include this nuance for completeness if you can.

config/webhook/manifests.yaml
```diff
@@ -25,7 +25,7 @@ webhooks:
     - CREATE
     - UPDATE
     resources:
-    - openstackcluster
+    - openstackclusters
```
mdbooth, line 28: I assume this change is for consistency because all other resources are referenced by their plural?

jichenjc: we must set it , otherwise the webhook doesn't work at all..

I guess the reason is the resource we created in CRD is plural but not 100% sure..

api/v1alpha4/openstackcluster_webhook.go
jichenjc: yeah.. the reason is we will patch the spec in our code in controller
without those 2 lines , the deployment of cluster will fail..

mdbooth: Right. I was thinking something like:

```golang
  // Only allow controlPlaneEndpoint to be updated if it has not been set previously
  if testThatAllFieldsAreEmptySomehow(oldOpenStackClusterSpec["controlPlaneEndpoint"]) {
    delete(oldOpenStackClusterSpec, "controlPlaneEndpoint")
    delete(newOpenStackClusterSpec, "controlPlaneEndpoint")
  }
```

config/webhook/manifests.yaml
mdbooth: That's some strange voodoo: I thought these were equivalent. It's more consistent like this anyway, so 🤷‍♂️ That can remain a mystery for another time.

api/v1alpha4/openstackcluster_webhook.go
jichenjc: ok, will take this a TODO (with other webhook like openstackmachine into one PR)

thanks~

PR #1008, opened 2021-09-23
✨ getOrCreatePort: add support to configure QoS profile ID on ports

### LINKED ISSUE #1007: Adding support to create Neutron ports with QoS policy ID
/kind feature

Adding support to create the ports with the QoS policy ID, see API doc:
https://docs.openstack.org/api-ref/network/v2/?expanded=show-port-details-detail#id54

EmilienM:
**What this PR does / why we need it:**

Adding support to configure Neutron ports with a specific QoS profile
ID:
https://docs.openstack.org/api-ref/network/v2/?expanded=show-port-details-detail#id54

This feature is already supported by Gophercloud:
https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/qos/policies/requests.go

Fixes #1007

### DIFF
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/portsbinding"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/portsecurity"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/qos/policies"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/cluster-api/util"
@@ -139,6 +140,14 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 		SecurityGroups:      securityGroups,
 		AllowedAddressPairs: addressPairs,
 		FixedIPs:            fixedIPs,
+	}
+
+	if portOpts.QoSPolicyID != "" {
+		policyID := portOpts.QoSPolicyID
+		createOpts = policies.PortCreateOptsExt{
+			CreateOptsBuilder: createOpts,
+			QoSPolicyID:       policyID,
+		}
 	}
 
 	if portOpts.DisablePortSecurity != nil {
--- a/test/e2e/suites/e2e/e2e_test.go
+++ b/test/e2e/suites/e2e/e2e_test.go
@@ -201,6 +201,7 @@ var _ = Describe("e2e tests", func() {
 
 	Describe("Workload cluster (multiple attached networks)", func() {
 		var extraNet1, extraNet2 *networks.Network
+		var policyID string
 
 		BeforeEach(func() {
 			var err error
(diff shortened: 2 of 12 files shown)


pkg/cloud/services/networking/port.go
```diff
@@ -141,6 +142,14 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 		FixedIPs:            fixedIPs,
 	}
 
+	if portOpts.QoSPolicyID != nil {
+		policyID := portOpts.QoSPolicyID
+		createOpts = policies.PortCreateOptsExt{
```
EmilienM, line 147: Note that the policy has to exist before the port gets created. But since we have to provide an UUID to it, we expect the policy to exist. In the case of the policy can't be found (e.g. owned by another tenant), Neutron should return an API error where the policy can't be found.

jichenjc: we have error check at line 170 below, so should be ok to avoid the check

jichenjc:
/approve cancel

oops, the failure is valid .. need some update
pkg/cloud/services/networking/port.go:145:26: invalid operation: portOpts.QoSPolicyID != nil (mismatched types string and nil)

pkg/cloud/services/networking/port.go
mdbooth: This is also consistent with what I've done in #994: if the user explicitly gives us a resource UUID we trust it without further validation. If it's wrong, a subsequent operation will fail and they can see the error about that.

mdbooth (commented):
@EmilienM Can you think of a way we could reasonably test this? Perhaps creating a QoS policy in an E2E test and creating a port to use it?

mdbooth:
@EmilienM You need to run `make generate`

EmilienM:
> @EmilienM Can you think of a way we could reasonably test this? Perhaps creating a QoS policy in an E2E test and creating a port to use it?

yes, this will imply the qos extension enabled in Neutron. I can look at that at some point, certainly.

jichenjc:
> 
> 
> > @EmilienM Can you think of a way we could reasonably test this? Perhaps creating a QoS policy in an E2E test and creating a port to use it?
> 
> yes, this will imply the qos extension enabled in Neutron. I can look at that at some point, certainly.

we can customize the devstack , it should be not that hard , some changes 
already done here
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1009/files
you can refer to

mdbooth:
> 
> 
> > > @EmilienM Can you think of a way we could reasonably test this? Perhaps creating a QoS policy in an E2E test and creating a port to use it?
> > 
> > 
> > yes, this will imply the qos extension enabled in Neutron. I can look at that at some point, certainly.
> 
> we can customize the devstack , it should be not that hard , some changes
> already done here
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1009/files
> you can refer to

Looks like QoS is already in there.

EmilienM:
I'm looking at e2e testing today.

EmilienM:
TODO: I need to finish the e2e test to confirm that we actually get the port with the right QoS Policy ID. Will do asap.

mdbooth:
@EmilienM re the e2e test, could it live with the other custom ports test instead? I was hoping that test might grow into a mega test of everything.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -229,6 +230,12 @@ var _ = Describe("e2e tests", func() {
 
 			os.Setenv("CLUSTER_EXTRA_NET_1", extraNet1.ID)
 			os.Setenv("CLUSTER_EXTRA_NET_2", extraNet2.ID)
+
+			policyName := "demo-policy"
+			policyID, err = shared.GetOpenStackQosPolicyID(e2eCtx, policyName)
+			Expect(err).To(BeNil())
+			Expect(policyID).NotTo(BeNil(), "Neutron QoS policy does not exist: %s", policyName)
+			os.Setenv("CLUSTER_QOS_POLICY_ID", policyID)
```
EmilienM, line 238: I'm missing the bits later in this test to confirm that the port has been created with a specific Policy ID.
@mdbooth I might use your help on this one

hidekazuna:
@EmilienM v1beta1 API is merged(#1047). update to change v1beta1 API, please.

PR #1009, opened 2021-09-23
Enable trunking in CI devstack

mdbooth:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
/approve

agree use ovn and up to date configurations
so let's wait for test result with trunk enabled

chrischdi:
Just a question: so in the future all tests will run with the trunk enabled devstack right?
There are no downsides to expect regarding a OpenStack using not the ovn setup?

/lgtm

Xenwar:
In order to avoid having another PR for testing the setup, modify the manifests to enable trunk. That way, you can get an immediate feedback. example shown below. 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1000/files#diff-6e511039e9f238d265053cbd9ab382a54b997df9b4a9790bb2ffbf0f0b29404eR111

Once trunk setup is verified, you can change back the manifests.

mdbooth:
> Just a question: so in the future all tests will run with the trunk enabled devstack right?
> There are no downsides to expect regarding a OpenStack using not the ovn setup?
> 
> /lgtm 

In theory, no. In practise we have stumbled across different bugs in different neutron backends in the past.

mdbooth:
Still seeing lack of trunk support in the logs:

```
E0924 10:37:56.260404       1 controller.go:304] controller-runtime/manager/controller/openstackmachine "msg"="Reconciler error" "error"="error creating Openstack instance: there is no trunk support. Please disable it" "name"="cluster-e2e-0kv26f-control-plane-g2x4g" "namespace"="e2e-0kv26f" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" 
```

This isn't super urgent for me but I would very much like to see it working, so I'm going to leave this open to hack on it. I plan to:
* Test this devstack config manually to see if there's anything obviously wrong with it.
* Investigate pulling openstack service config and logs from devstack as artifacts.

There was mention of this possibly being a hardware issue, but unless I'm missing something important the network with the trunk on it is virtual, so hardware shouldn't be relevant. Am I missing something important?

If anybody else wants to take this on instead, feel free! Please let me know here or on slack so we don't duplicate effort.

mdbooth:
I think I've worked out the fundamental issue here: the job is running with a pre-installed devstack image and doesn't actually use the local.conf we were updating. I'm testing a change which will hopefully change that. We can then decide whether we want to move to OVN or try the much simpler change which just adds the neutron-trunk service.

Xenwar:
@mdbooth Thanks for taking the time. 
Would it be possible to override the local.conf file by supplying the same file via cloud-init ?

mdbooth:
> 
> 
> @mdbooth Thanks for taking the time.
> Would it be possible to override the local.conf file by supplying the same file via cloud-init ?

That's what my current attempts are (hopefully) doing 🤞

mdbooth:
It worked 😁

```
I0924 17:25:42.806163       1 recorder.go:104] controller-runtime/manager/events "msg"="Normal"  "message"="Created trunk cluster-e2e-f7cs8y-control-plane-9w47w-0 with id 9e0b176d-2c06-46b5-9324-7e1d3a55d24a" "object"={"kind":"OpenStackMachine","namespace":"e2e-f7cs8y","name":"cluster-e2e-f7cs8y-control-plane-9w47w","uid":"ef6a9922-0392-420a-a181-bcf600a299db","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha4","resourceVersion":"3564"} "reason"="SuccessfulCreateTrunk"
```

mdbooth:
@sbueringer Interested in your opinion on the approach to customising devstack.

jichenjc:
>>>the job is running with a pre-installed devstack image and doesn't actually use the local.conf we were updating. 

this is new to me , but I am glad we figure it out :)

chrischdi:
I think using a pre-installed devstack makes the testing way faster / was the reason to do so. 

The preinstalled image gets fetched here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/hack/ci/devstack-on-gce-project-install.sh#L134

jichenjc:
>>>I think using a pre-installed devstack makes the testing way faster / was the reason to do so.

agree we should use pre-install image to speed up test
then can we update the devstack image per request above?

mdbooth:
> 
> 
> I think using a pre-installed devstack makes the testing way faster / was the reason to do so.
> 
> The preinstalled image gets fetched here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/hack/ci/devstack-on-gce-project-install.sh#L134

This patch still uses the preinstalled image. The preinstalled image still runs stack.sh to actually install devstack: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/48c7bb90fc36a9b605ed079b47a6407621399ea2/hack/ci/devstack-preinstalled-cloud-init.yaml.tpl#L37-L38

The problem we were hitting is that the preinstalled image also contains a hardcoded local.conf. All this change does is allow us to change local.conf. It shouldn't impact time to install devstack.

mdbooth:
The most recent failure was due to an OpenStack bug: http://lists.openstack.org/pipermail/openstack-discuss/2021-September/025101.html

chrischdi:
> > > > I think using a pre-installed devstack makes the testing way faster / was the reason to do so.
> 
> agree we should use pre-install image to speed up test then can we update the devstack image per request above?

Fine for me :+1:

I think on the long-term / on a PR both local.conf files should be in sync. So during the PR the new one get's used and after merging we can finally update the image and also have it active there :-)

jichenjc:
/lgtm

leave to other folks to take another look~

mdbooth:
https://review.opendev.org/c/openstack/devstack/+/811399 has merged in devstack, so retesting without the workaround.

PR #1012, opened 2021-09-27
✨Allow update for some param only first time update

jichenjc:

Only allow the field's update at first time, as that will be updated by CAPI internally,
after that, customer's update will be rejected..

follow up of #1006 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
@mdbooth this is follow up of #1006

mdbooth (approved):
Nice!

Do we have a good way to test these, yet? The E2E tests, I guess, but that seems like a big hammer.

mdbooth:
I was thinking of [envtest](https://book.kubebuilder.io/reference/envtest.html).

To get there I think we need to merge #886, then implement interfaces for the remainder of our services (as we currently have for networking only) to allow us to cleanly mock out OpenStack calls. This will allow us to write tests like https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller_test.go.

I'm not proposing we do all that before merging your patch, btw! It's going to be a lot of work.

PR #596, opened 2020-07-24
Remove Tag creation for ports and add description for it

jichenjc:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #543

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

Issue #543, opened 2020-04-06
disableServerTags: true is not working


AllaeddineEL:
/kind bug

**What steps did you take and what happened:**
I'm trying to provision Kubernetes Cluster on Top of our OpenStack Provider, which it doesn't support port tags.
I've disabled the server tags `disableServerTags: true` in the `OpenStackCluster` config and I have no tags in the `OpenStackMachineTemplate`

**What did you expect to happen:**
Create OpenStack cluster without tags  

**Anything else you would like to add:**
```
$ kubectl describe openstackmachine devops-control-plane-trf55
Name:         devops-control-plane-trf55
Namespace:    default
Labels:       cluster.x-k8s.io/cluster-name=devops
              cluster.x-k8s.io/control-plane=
              kubeadm.controlplane.cluster.x-k8s.io/hash=724735395
Annotations:  <none>
API Version:  infrastructure.cluster.x-k8s.io/v1alpha3
Kind:         OpenStackMachine
Metadata:
  Creation Timestamp:  2020-04-06T14:29:27Z
  Finalizers:
    openstackmachine.infrastructure.cluster.x-k8s.io
  Generation:  1
  Owner References:
    API Version:           controlplane.cluster.x-k8s.io/v1alpha3
    Kind:                  KubeadmControlPlane
    Name:                  devops-control-plane
    UID:                   d22ddafb-17f8-4c79-ad6d-5b285a48a335
    API Version:           cluster.x-k8s.io/v1alpha3
    Block Owner Deletion:  true
    Controller:            true
    Kind:                  Machine
    Name:                  devops-control-plane-n88ww
    UID:                   db838279-5626-4e23-8126-7c33b6640e02
  Resource Version:        2889
  Self Link:               /apis/infrastructure.cluster.x-k8s.io/v1alpha3/namespaces/default/openstackmachines/devops-control-plane-trf55
  UID:                     d62c98e5-c9ff-40f0-a91e-5052f3299a8d
Spec:
  Cloud Name:  otc_doa
  Clouds Secret:
    Name:       devops-cloud-config
    Namespace:  default
  Flavor:       s2.large.1
  Image:        Standard_CentOS_7_prev_packer_test
  Security Groups:
    Name:        devops-k8s-master
    Name:        devops-k8s-master
    Name:        devops-local-ssh
    Name:        devops-k8s-ssh
    Name:        devops-k8s
  Ssh Key Name:  kubernetes-devops
Status:
[...]

PR #1016, opened 2021-09-29
Port Tagging Regression

iamemilio:
:bug: Tagging of port resources was mistakenly removed at some point, but not
for trunking. To make things consistent once again, this enables tagging of ports
created during instanceCreate(). :bug:


pkg/cloud/services/networking/port.go
```diff
@@ -161,6 +161,10 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", portName, err)
 		return nil, err
 	}
+	if err = s.replaceAllAttributesTags(eventObject, port.ID, tags); err != nil {
```
EmilienM, line 164: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1016/files#diff-787f1b361511084b9cd8e6918dfb552589bc1698aa1aa13c9cff9fafde0377aaR131-R142

How are we creating ports with tags if we don't add it here? ^

EmilienM: Also `replaceAllAttributesTags` was done for Trunks, not Ports (see the implementation).

iamemilio: Thanks for the review, I extended the replaceAllAttributesTags function to support ports and tags.

jichenjc:
I think we might consider add this into e2e 
so we can check resources like servers/ports has such ports thus we won't
have this regression?

mdbooth:
Incidentally, there is a neutron extension (proposed by Michal Dulko) called tag_ports_during_bulk_creation which, despite its name, also allows port tagging on regular creation. IIUC 'base' neutron does not support tagging ports on creation.

However:
* the extension only landed in Ussuri
* additionally, neutron extensions must be enabled explicitly by the operator in config

This means that, in practise, we're rarely going to encounter a cloud where we can use this feature. Until we pass a point where we can make this feature part of our minimum supported configuration, which feels many years away, I don't see any benefit in maintaining 2 ways of doing this. We should ensure that 2-step tagging is as robust as possible.

mdbooth:
```
pkg/cloud/services/networking/service.go:98: File is not `gofumpt`-ed (gofumpt)
```

Pretty sure that's import ordering.

chrischdi:
Anyone did take a look at #543 which was the reason for removing the tags again in #596 ?

It is now fine to add it back?

chrischdi:
Maybe we have to hide adding tags behind a feature gate / allow users to disable the tagging via a feature gate?! WDYT?

mdbooth:
> 
> 
> Maybe we have to hide adding tags behind a feature gate / allow users to disable the tagging via a feature gate?! WDYT?

I'd be more inclined to add tagging support to our minimum requirements. At this point there are no supported versions of OpenStack which don't have it.

pkg/cloud/services/networking/service.go
```diff
@@ -78,3 +83,23 @@ func NewTestService(projectID string, client NetworkClient, logger logr.Logger)
...
 }
+
+// replaceAllAttributesTags replaces all tags on a neworking resource.
+// the value of resourceType must match one of the allowed constants: trunkResource or portResource.
+func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceType string, resourceID string, tags []string) error {
+	if resourceType != trunkResource && resourceType != portResource {
+		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Invalid resourceType argument in function call")
+		return fmt.Errorf("invalid argument: resourceType, %s, does not match allowed arguments: %s or %s", resourceType, trunkResource, portResource)
+	}
+
```
chrischdi, line 94: is it even worth checking it?

OpenStack would return an error on wrong types and we have it hardcoded.
Also if we add it, wouldn't it make sense to implement it in the `networkClient.ReplaceAllAttributesTags`?

But also ok for me as is.

chrischdi (commented):
/approve

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -483,6 +487,7 @@ func makeOpenStackMachineTemplateWithPortOptions(namespace, clusterName, name st
 						Name: fmt.Sprintf("%s-cloud-config", clusterName),
 					},
 					Ports: *portOpts,
+					Tags:  machineTags,
```
mdbooth, line 490: Oh, we really can't tag ports individually? I think we're going to have to fix that 🤔

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -173,6 +176,7 @@ var _ = Describe("e2e tests", func() {
 
 			port := plist[0]
 			Expect(port.Description).To(Equal("primary"))
+			Expect(port.Tags).To(ContainElement(testTag))
```
mdbooth, line 179: Hmm, this test is broken 🤔 It's dumping all ports with a description of "primary" with no other constraint. This could easily pass due to a concurrently running test creating a different port with this name.

However... Good News! We can now tag ports 😁

Please could you ensure that the test tag is unique to the test, then fetch ports by tag and ensure you get the one with description "primary"? Suggest using clusterName as a tag would work well.

mdbooth (changes requested):

pkg/cloud/services/networking/port.go
```diff
@@ -161,6 +161,10 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", portName, err)
 		return nil, err
 	}
+	if err = s.replaceAllAttributesTags(eventObject, portResource, port.ID, tags); err != nil {
```
macaptain, line 164: If `len(tags) == 0`, then would this OpenStack call be necessary? We may be able to be save a call per port creation in the cases where tags are not specified on the instance.

iamemilio:
> Invalid resourceType would be a programming error. I personally think it would be safer to panic() here, which ensures it's less likely to be missed in testing.
>
> Not worth a respin just for that, though.

Panic or return the error and exit? Do we want any errors to be causing a panic? It would stop the service from running.

pkg/cloud/services/networking/service.go
```diff
@@ -78,3 +83,23 @@ func NewTestService(projectID string, client NetworkClient, logger logr.Logger)
 		logger:    logger,
 	}
 }
+
+// replaceAllAttributesTags replaces all tags on a neworking resource.
+// the value of resourceType must match one of the allowed constants: trunkResource or portResource.
+func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceType string, resourceID string, tags []string) error {
+	if resourceType != trunkResource && resourceType != portResource {
+		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Invalid resourceType argument in function call")
```
mdbooth, line 91: nit: Warnf is redundant. The only reason we're here is because somebody's hacking on it and they messed up.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -153,26 +154,30 @@ var _ = Describe("e2e tests", func() {
...
-				InfraMachineTemplate:    makeOpenStackMachineTemplateWithPortOptions(namespace.Name, clusterName, md3Name, customPortOptions),
+				InfraMachineTemplate:    makeOpenStackMachineTemplateWithPortOptions(namespace.Name, clusterName, md3Name, customPortOptions, machineTags),
 			})
 
 			shared.Byf("Waiting for custom port to be created")
 			var plist []ports.Port
 			var err error
 			Eventually(func() int {
-				plist, err = shared.DumpOpenStackPorts(e2eCtx, ports.ListOpts{Description: "primary"})
+				plist, err = shared.DumpOpenStackPorts(e2eCtx, ports.ListOpts{Description: "primary", Tags: testTag})
```
mdbooth, line 173: Nice. Don't need Description here any more, but it sounds like there's plenty more to do to this test so we can finesse it later.

mdbooth (approved):
/lgtm

Let's get this merged!

mdbooth:
> 
> 
> /test pull-cluster-api-provider-openstack-e2e-test

I investigated this for a while but couldn't find an obvious reason for the failure with the available logs. As it passed with a re-run, it's possible it was just an ephemeral failure. Perhaps a repo was down 🤷‍♂️ We'll worry about it more if we keep seeing it.

PR #1020, opened 2021-10-06
🏃 Add unit tests for getOrCreatePort func

macaptain:
The `getOrCreatePort` func is quite complicated, taking many arguments and options and with conditional calls to OpenStack. This change adds unit tests for the function, which should make it clearer what OpenStack API calls are made under which conditions, and shows off all the supported features.

**What this PR does / why we need it**: Improves unit test coverage in the networking package, helps avoid future regressions in the ports logic

Helps with #843

**TODOs**:
- [x] squashed commits
- [x] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/networking/port_test.go
+++ b/pkg/cloud/services/networking/port_test.go
@@ -1,0 +1,371 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/networking/port_test.go
```diff
@@ -0,0 +1,373 @@
...
+			&ports.Port{Name: "foo-port-1", ID: portID1},
+			false,
+		},
+	}
+
+	eventObject := &infrav1.OpenStackMachine{}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.name == "creates port with tags passed to function" {
+				t.Skip("TODO: Bug! Tags aren't applied to the port. See #1016. Then ensure that the port tags are applied.")
```
macaptain, line 344: Note that I've skipped a test! This depends on #1016 being merged and then hopefully we will avoid this regression again in future.

PR #1030, opened 2021-10-27
✨Cinder AZ and volume type support

mdbooth:
This PR adds the ability to specify availability zone and volume type for a machine's root volume. In doing this it makes some API cleanups and adds 2 new API fields.

N.B. RootVolume is a child of both OpenStackMachineSpec and Instance, which is a child of OpenStackClusterStatus (Bastion), so changes affect those 2 objects.

Firstly we cleanup some useless fields which were copied directly from the Nova API. RootVolume currently contains:

```golang
type RootVolume struct {
        SourceType string `json:"sourceType,omitempty"`
        SourceUUID string `json:"sourceUUID,omitempty"`
        DeviceType string `json:"deviceType,omitempty"`
        Size       int    `json:"diskSize,omitempty"`
}
```

* `SourceType` is the source of the data used to initialise the root volume. In our case this is always glance, and therefore any value other than `image` will not work.
* `SourceUUID` is the image ID. We already have this specified on the parent object (either OpenStackMachineSpec or Instance), so it is redundant.
* `DeviceType` is the type of hardware device the hypervisor will present the volume to the server as. In our case we always want this to be `disk`.

We remove all 3 of these fields as they are redundant.

Secondly we add 2 new fields. RootVolume becomes:

```golang
type RootVolume struct {
        Size             int    `json:"diskSize,omitempty"`
        VolumeType       string `json:"volumeType,omitempty"`
}
```

By default we will create a root volume in the same availability zone as the server. This can be overridden by specifying RootVolume.AvailabilityZone. This may even be necessary, for example in the (not recommended, but still supported and actually found in practise) event that cinder and nova do not have availability zones with the same names.

By default the root volume will be created using the cloud's default volume type, but this can be overridden by specifying RootVolume.VolumeType.

In addition to unit testing, we update the multi-AZ E2E test to boot all servers from volume with a mix of configurations. The control plane servers are spread across 2 AZs and specify only size, so we expect them to have root volumes in the same AZ as the server, and use the default volume type. The workers are all scheduled to the alternate AZ, so to exercise the cinder AZ functionality we specify that all their root volumes should be in the primary AZ, which is not the default behaviour. Additionally we create a new volume type and specify that type for all the workers.

**TODOs**:

- [X] Wait for #1063 to merge

/hold until #1063 merges


pkg/cloud/services/compute/instance.go
```diff
@@ -226,24 +229,26 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 		AccessIPv4:       accessIPv4,
 	}
 
-	serverCreateOpts = applyRootVolume(serverCreateOpts, instanceSpec.RootVolume)
+	volume, err := s.getOrCreateRootVolume(eventObject, instanceSpec)
+	if err != nil {
+		return nil, fmt.Errorf("error creating root volume: %w", err)
```
jichenjc, line 234: nit: should this be error get or create ?

pkg/cloud/services/compute/instance.go
```diff
@@ -277,24 +282,98 @@ func getPortName(instanceName string, opts *infrav1.PortOpts, netIndex int) stri
...
+	return &volumeList[0], nil
+}
+
+func (s *Service) getOrCreateRootVolume(eventObject runtime.Object, instanceSpec *InstanceSpec) (*volumes.Volume, error) {
+	rootVolume := instanceSpec.RootVolume
+	if hasRootVolume(rootVolume) {
+		return nil, nil
+	}
+
+	// TODO: Remove SourceType and DeviceType from the API because we don't support customising them
```
jichenjc, line 318: todo: I think we can put those validation into webhook later on?
as if we check (guess we have another option to delegate this check to openstack itself?) we can always do at API level?

mdbooth: These are taken directly from the nova `block_device_mapping_v2` field to create: https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server. This is a problem as this change stops using this API to create volumes. This is because the Nova API is a proxy API, and doesn't expose the full functionality of Cinder. By policy Nova will not generally add new functionality to proxy APIs. They prefer to remove them in favour of the client making explicit calls, which is what we're doing here.

The possible values of `SourceType` are:
* `blank`: a blank disk
* `image`: initialise from a glance image
* `snapshot`: initialise from a volume snapshot
* (`volume`: attach a specific volume, which isn't relevant here)

`blank` doesn't work for a root volume, so the only values which make any sense are `image` and `snapshot`. As far as I'm aware we only support glance images, so any value other than `image` would be a mistake. We might as well remove `SourceType` entirely.

`DeviceType` relates to how the hypervisor exposes the disk to the guest VM. Potential values include:
* `disk`
* `cdrom`
* `floppy`

Unless somebody makes a K8s distro which fits in 1.44MB or which boots from a Live CD, any value other than `disk` is a mistake. We might as well remove it entirely.

jichenjc: yes, I knew those proxy and param pass :) thanks for point that out ~

I am only thinking about reject those at API layer as actually those are bad API request
so instead of fail later, we fail early ~

but another thought comes out ,we have many many places might fit into such check
put all those into API layer might make the validation too big ...

mdbooth: You mean adding them to the validating web hook in this PR?

Given that we're about to cut v1alpha5 and hopefully soon v1beta1 I was hoping to slip in the breaking API change of just removing them entirely. I'm still not entirely sure how that works with branching, though, and of course I also need to make the case for breaking the API.

However, would I still want to add it for validating <v1alpha5? 🤔 And if we decide against removing the API entirely in this PR, the validating webhook also sounds like the better option.

As for making the validation too big, it's a very cheap test to execute. I haven't really looked at the validation in detail yet, so I don't know how much code complexity it would add.

iamemilio: I would be in favor of removing the API fields that serve no function other than as a trap for users. Its not like there is even a use case or a reason they should exist, so why support and validate them in the first place?

pkg/cloud/services/compute/instance.go
```diff
@@ -206,10 +204,15 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 		})
 	}
 
-	if instanceSpec.Subnet != "" && accessIPv4 == "" {
-		if err := s.deletePorts(eventObject, portList); err != nil {
-			return nil, err
+	// Ensure we delete the ports we created on any error prior to instance create
+	deletePorts := portList
+	defer func() {
```
iamemilio, line 209: I think this is cool syntax, but its also way more confusing then just having the function return an error and checking it when it returns.

iamemilio:
overall, looks good. I'll give it a once over again when the test passes.

pkg/cloud/services/compute/instance.go
mdbooth: That's what the previous code did, and it missed several cases already 😉 And I'm adding another one.

This is more robust than requiring adding the same check at every potential exit point of the function. I will eventually add unit tests for this behaviour, btw.

iamemilio: ok, I can live with it I suppose.

mdbooth: I have now added unit tests!

mdbooth:
> /hold the test failure suggests a volume wasn't deleted, which obviously shouldn't happen

We only expect 2 volumes with the custom volume type to be created. I can see in the devstack controller logs that we only create 2 volumes with this type, and that both of these volumes are deleted in cinder when the server is deleted in nova. This includes deletion of the backing LV, so they are most definitely gone. They are deleted in cinder more than a minute before the failed call to delete volume type. This all seems fine, so it's not clear to me yet why the delete volume type failed.

mdbooth:
> > /hold the test failure suggests a volume wasn't deleted, which obviously shouldn't happen
> 
> We only expect 2 volumes with the custom volume type to be created. I can see in the devstack controller logs that we only create 2 volumes with this type, and that both of these volumes are deleted in cinder when the server is deleted in nova. This includes deletion of the backing LV, so they are most definitely gone. They are deleted in cinder more than a minute before the failed call to delete volume type. This all seems fine, so it's not clear to me yet why the delete volume type failed.

I have reproduced this error manually in my local devstack:

* Create a volume type (as admin)
* Create a volume from image (as tenant)
* Attach the volume to a server with delete_on_termination (as tenant)
* Delete the server (as tenant)
* Verify that the volume no longer exists
* Delete the volume type (as admin) -> fails

The issue seems to be that in creating the volume from image, cinder creates a volume  called `image-<volume id>` with the same type as the target volume, which it then snapshots to create the requested volume. The 'image cache' volume is not deleted when the volume itself is deleted, and appears to be what is preventing the deletion of the volume type.

I'll ask some cinder folks what the best thing to do is, but possibilities include:

* Manually delete 'image cache' volumes before deleting the volume type.
* Just don't delete the volume type.
* If cinder consider this a bug (they might not) and it's fixed in a subsequent release, upgrade to that release.

Of these I'm inclined to favour just not deleting the volume type: we blow the devstack away straight after anyway. I'll update this PR to do that and add a comment.

jichenjc:
/hold 

not a blocker for anything :)
as this is a bigger PR now hope to get more people take a look before one lgtm to merge

pkg/cloud/services/compute/instance.go
```diff
@@ -303,24 +335,95 @@ func getPortName(instanceName string, opts *infrav1.PortOpts, netIndex int) stri
...
+	if err != nil {
+		return nil, err
+	}
+	if volume != nil {
+		if volume.Size != size {
+			return nil, fmt.Errorf("exected to find volume %s with size %d; found size %d", name, size, volume.Size)
 		}
+
+		s.logger.Info("using existing root volume %s", name)
+		return volume, nil
```
hidekazuna, line 384: This returns a volume selected by name and size only. I wonder if there are two volumes in different availability zones or volume types, I am not sure that could happen though.

pkg/cloud/services/compute/instance.go
```diff
@@ -258,7 +259,40 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
...
+	if err != nil {
+		return nil, fmt.Errorf("error in get or create root volume: %w", err)
+	}
+
+	instanceCreateTimeout := getTimeout("CLUSTER_API_OPENSTACK_INSTANCE_CREATE_TIMEOUT", timeoutInstanceCreate)
+	instanceCreateTimeout *= time.Minute
+
+	// Wait for volume to become available
+	if volume != nil {
+		err = util.PollImmediate(retryIntervalInstanceStatus, instanceCreateTimeout, func() (bool, error) {
```
hidekazuna, line 272: `instanceCreateTimeout` is used but actually this waits for volume to become available so that it is better to introduce new variable. We can follow up this issue after this PR merged because time out settings are not consistently defined: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/33a98520ea942a1e011f249d36d8541c84bbb558/pkg/cloud/services/compute/instance.go#L41-L43

mdbooth: Honestly I'd probably just rename the variable to be something like openStackRetryInterval. In practise nobody is ever going to tune these individually, and there would (hopefully!) be very minimal benefit if they did.

However, my grand plan (see my other comment reply!) is to push this polling out to the controller. Blocking in the reconcile loop is potentially problematic anyway.

pkg/cloud/services/compute/instance.go
mdbooth: Firstly, I agree. It's hopefully unlikely in practise, but I'd really like to eliminate everywhere that we rely on resource name because it introduces the potential for a naming conflict, and it's less efficient for the API.

[As mentioned elsewhere](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1095#pullrequestreview-848928494) I'd also like to clean up how this works to simplify the resource creation/deletion logic.

My current idea for how to do this is to track explicitly-created cloud resources in Status. This would look something like:

```
    function ReconcileRootVolume(osMachine) (reconciled bool) {
        if osMachine.Status.RootVolume != "" {
            return volumeIsAvailable(rootVolumeUUID)
        }
        
        osMachine.Status.RootVolume = createRootVolume()
        return false
    }
```

This also makes cleanup more robust because you know you have to delete a root volume [iff](https://en.wikipedia.org/wiki/If_and_only_if) RootVolume is set.

The same would also be true of ports. It would also mean we could create ports while waiting for the root volume to become available by moving polling entirely out of the create function, and simply asking the controller to call reconcile again in 10 seconds.

I was *extremely* tempted to do this in this patch, but I didn't because this patch is already big enough, and already makes a significant API change in the Spec. I very much want to come back to it, though.

In the meantime I don't think it's possible to avoid doing this match on name because that implies we'd have to store a uuid somewhere, and we don't yet have anywhere to put it.

pkg/cloud/services/compute/instance.go
hidekazuna: I agree creating instance logic is too big and hard to review now. We really need refactoring.

pkg/cloud/services/compute/instance.go
hidekazuna: Thanks for not clean up creation/deletion logic at the same time for reviewers 😃 I agree with your idea using uuid in Status to identify resources.

iamemilio:
code looks good. Lots of test coverage. Thanks @mdbooth 
/lgtm

mdbooth:
@jichenjc Are you comfortable removing the hold on this?

PR #1032, opened 2021-10-27
✨E2E error logging improvements

mdbooth:

This makes 2 improvements to error logging in E2E tests:

* Specific improvements in gomega error reporting in the multi-network test
* Stop using --stream for ginkgo output


mdbooth:
> /test pull-cluster-api-provider-openstack-e2e-test

Looks like this was required due to a flake in devstack 🤔 Lets keep an eye on how common this is. If this happens too often, this would be a more compelling reason to bring back preinstalled than the time saved.

tobiasgiese:
Because of

> Stop using --stream for ginkgo output

Should we also stop using `-stream` for the ginkgo conformance tests?
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c12f58b03ffc918ad1122a5412b1eaf7c8955560/Makefile#L145

Because the target `test-e2e` will also trigger the target `test-e2e-image-prerequisites`

PR #1040, opened 2021-11-04
🏃Remove most uses of submakes in Makefile

mdbooth:
This commit series systematically removes most uses of submakes in the top level Makefile. It was fairly easy to write and I personally think it results in a more maintainable Makefile.

**HOWEVER**, I'm not aware of any bugs it addresses (yet), there's obviously the possibility that it has introduced bugs, and we don't have great test coverage of the Makefile, so... is it worth it?

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,21 +357,13 @@ IMAGE_PATCH_DIR := $(ARTIFACTS)/image-patch
 $(IMAGE_PATCH_DIR): $(ARTIFACTS)
 	mkdir -p $@
 
-.PHONY: $(RELEASE_DIR)/$(MANIFEST_FILE).yaml
-$(RELEASE_DIR)/$(MANIFEST_FILE).yaml:
-	$(MAKE) compiled-manifest \
-		PROVIDER=$(MANIFEST_FILE) \
-		OLD_IMG=$(CONTROLLER_ORIGINAL_IMG) \
-		MANIFEST_IMG=$(CONTROLLER_IMG) \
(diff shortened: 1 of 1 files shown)


jichenjc:
does this conflict to #1039 ?

tobiasgiese:
> does this conflict to #1039 ?

This is just a WIP PR and should not be merged, see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1039#issuecomment-961402570

@mdbooth you should add the WIP label in such cases

/title WIP Remove .EXPORT_ALL_VARIABLES

iamemilio:
/lgtm

This is more readable too IMO

jichenjc:
I think it's a great PR
the question I am having is as commit msg said, how to test those ..
if we are sure current test is good enough and overtime we can add more test to cover those newly added rules, I can live with that, just need some agreement .. thanks

mdbooth:
Seems like there's some agreement to merge this.

/hold cancel

hidekazuna:
I look back again this PR due to time has passed. I take a look at CAPA and found our current Makefile is like CAPA. I am not familiar with Makefile so that I want to continue to refer to Makefile in CAPA as of now.

/lgtm cancel

mdbooth:
> I look back again this PR due to time has passed. I take a look at CAPA and found our current Makefile is like CAPA. I am not familiar with Makefile so that I want to continue to refer to Makefile in CAPA as of now.
> 
> /lgtm cancel

That's entirely reasonable, and I also didn't feel confident pushing it due to the lack of test coverage. I'll drop this.

PR #1048, opened 2021-11-15
🐛Fixes for running E2E tests locally against devstack on OpenStack

mdbooth:

If you want to test CI changes against a local OpenStack.

Fixes #

**TODOs**:

- [X] squashed commits
- if necessary:
  - [X] includes documentation


docs/book/src/development/development.md
```diff
@@ -138,6 +138,17 @@ You must configure systemd and iptables as described in https://kind.sigs.k8s.io
 
 You must install the `podman-docker` package to emulate the docker cli tool. However, this is not sufficient on its own as described below.
 
+### Configure slirp4netns to use a different CIDR
+
+By default we configure the controller and worker on 10.0.2.0/24. Rootless podman uses slirp4netns for networking, and unfortunately by default it uses the same CIDR internally. This prevents the management cluster from being able to route to the CI devstack.
```
hidekazuna, line 143: If we change the default CIDR, do we no longer need to configure it?

mdbooth: Yes, we could certainly do that, too. It would just require updating these 3 values: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/b6695f2051feeebc674924cb71faf5a56cdb2ede/hack/ci/create_devstack.sh#L42-L44.

Incidentally, I did wonder if 10.0.2.0/24 was some kind of convention. If we changed it to something else, would it merit a comment explaining why it isn't 10.0.2.0/24?

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -117,12 +117,46 @@
...
+  permissions: 0755
+  content: |
+    #!/bin/bash
+
+    set -o -x
+
+    # Add environment variables for auth/endpoints
+    echo 'source /opt/stack/devstack/openrc admin admin' >> /opt/stack/.bashrc
+
+    source /opt/stack/devstack/openrc admin admin
```
jichenjc, line 130: a separate question

do we have any places that we use `demo` which is non-admin user and tenant to do some test? 
I knew it's kind of overkill but just want to see whether we already do it or potentially can do it

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -117,12 +117,46 @@
...
+
+    # Upload the images so we don't have to upload them from Prow
+    /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2021-03-27/ubuntu-2004-kube-v1.18.15.qcow2
```
jichenjc, line 133: later on maybe we can make those images as variable so easy to change

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -117,12 +117,46 @@
...
+    for host in $(openstack compute service list --service nova-compute -f value -c Host)
+    do
+        openstack aggregate add host "$aggregateid" "$host"
+    done
+
+    # the flavors are created in a way that we can execute at least 2 e2e tests in parallel (overall we have 32 vCPUs)
+    openstack flavor delete m1.tiny
+    openstack flavor create --ram 512 --disk 1 --vcpus 1 --public --id 1 m1.tiny --property hw_rng:allowed='True'
+    openstack flavor delete m1.small
+    openstack flavor create --ram 4192 --disk 10 --vcpus 2 --public --id 2 m1.small --property hw_rng:allowed='True'
```
jichenjc, line 147: should be it 8192 or 4096?

docs/book/src/development/development.md
```diff
@@ -138,6 +138,17 @@ You must configure systemd and iptables as described in https://kind.sigs.k8s.io
...
+
+To fix this, add the following to `~/.config/containers/containers.conf`:
```
jichenjc, line 145: just curious and education purpose :) , 
this is for containerd or cri-o or docker? 
can you point the issue about why we are suffering this to give more background on rootless podman 
etc so anyone hit the issue will be easier to understand the reason behind that?

docs/book/src/development/development.md
hidekazuna: I am not sure why CIDR is 10.0.2.0/24. Let's update to 10.0.3.0/24 and see how it goes.

mdbooth:
Rebased because I'm using this actively in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1030. No changes.

docs/book/src/development/development.md
mdbooth: The issue is specific to a management kind cluster running in rootless podman, which we use when running the E2E tests. The kind cluster uses 10.0.2.0/24 internally, so it can't route to the controller and worker.

As suggested by @hidekazuna I changed the IPs of the controller and worker instead so this workaround is no longer necessary.

PR #1049, opened 2021-11-16
🐛Pull Calico from quay.io instead of docker.io

mdbooth:
docker.io is rate limited without authentication, quay.io is not. This
is causing failures in local CI testing because provisioned machines
can't pull the calico image and therefore never become Ready.

We likely avoid this issue in CI due to the implementation of rate
limiting in docker.io, but it would not be impossible to hit it there,
either.

Calico publishes images to both repositories, so this just works.

### DIFF
--- a/test/e2e/data/cni/calico.yaml
+++ b/test/e2e/data/cni/calico.yaml
@@ -3739,7 +3739,7 @@ spec:
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
         - name: upgrade-ipam
-          image: docker.io/calico/cni:v3.20.1
+          image: quay.io/calico/cni:v3.20.1
           command: ["/opt/cni/bin/calico-ipam", "-upgrade"]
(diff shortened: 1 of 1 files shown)


tobiasgiese:
/lgtm
👍🏻

mdbooth:
@sbueringer don't suppose you could bless this?

sbueringer:
> @sbueringer don't suppose you could bless this?

It looks fine for me, but someone else should :)

jichenjc:
I thought quay.io need some kind of auth @mdbooth 
maybe I remember wrong thing as the test passed
please confirm this then use unhold to merge :) thanks

/approve
/hold

mdbooth:
> I thought quay.io need some kind of auth @mdbooth maybe I remember wrong thing as the test passed please confirm this then use unhold to merge :) thanks

Can confirm that quay.io doesn't require any credentials to pull a public image, and it doesn't have rate limiting.

/hold cancel

PR #1053, opened 2021-11-19
:sparkles:  Add tags on security groups created by capo

### LINKED ISSUE #1044: Add tags to security groups when ManagedSecurityGroups: true
/kind feature

Security groups created when `ManagedSecurityGroups: true` should be tagged when Tags are defined on `OpenStackCluster` similarly to how network/subnet/router etc are being tagged

- i'd like to work on this one
- Question: Should the tags be propagated to `openStackCluster.Status` similarly to how it's done for network/subnet/router? (it's not done the floatingIp of the loadbalancer but i think that's a bit of a different use case)

nikParasyr:
When ManagedSecurityGroups: true, security groups created by CAPO
are tagged with the tags on OpenStackCluster.spec.tags. Update
e2e tests to include tests for various resources that are tagged.

Fixes #1044 

1. Tags on SG are only added on creation, they are not propagated to cluster.status nor reconciled. Please let me know if you think any of the 2 needs to be added.
1. I didn't find any unit tests for the securitygroups but i updated the e2e tests. I used an existing test scenario as creating a new one seemed like an overkill. Also added tests for other resources that are also tagged

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -19,6 +19,7 @@ package networking
 import (
 	"fmt"
 
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 
@@ -546,6 +547,16 @@ func (s *Service) createSecurityGroupIfNotExists(openStackCluster *infrav1.OpenS
 			record.Warnf(openStackCluster, "FailedCreateSecurityGroup", "Failed to create security group %s: %v", groupName, err)
 			return err
 		}
+
+		if len(openStackCluster.Spec.Tags) > 0 {
+			_, err = s.client.ReplaceAllAttributesTags("security-groups", group.ID, attributestags.ReplaceAllOpts{
+				Tags: openStackCluster.Spec.Tags,
+			})
+			if err != nil {
+				return err
+			}
+		}
+
 		record.Eventf(openStackCluster, "SuccessfulCreateSecurityGroup", "Created security group %s with id %s", groupName, group.ID)
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/networking/securitygroups.go
```diff
@@ -546,6 +548,17 @@ func (s *Service) createSecurityGroupIfNotExists(openStackCluster *infrav1.OpenS
 			record.Warnf(openStackCluster, "FailedCreateSecurityGroup", "Failed to create security group %s: %v", groupName, err)
 			return err
 		}
+
+		if len(openStackCluster.Spec.Tags) > 0 {
+			mc := metrics.NewMetricPrometheusContext("security_group", "update")
```
hidekazuna, line 553: we use [ReplaceAllAttributesTags](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/client.go#L110) function in networking package so that we do not need to care metrics every time.

nikParasyr: should have seen that. Updated it. Thanks

nikParasyr:
@hidekazuna This is my first PR under the k8s umbrella so sorry for the noob question, I read the docs for the pull request process but i'm not sure what more is required. There is no `lgtm` label and im not sure whether i should ping more approvers for this one. Thank you

hidekazuna:
@nikParasyr Both member and reviewer can also add `lgtm` label. Please check out [Community membership](https://github.com/kubernetes/community/blob/master/community-membership.md#community-membership)

test/e2e/shared/openstack.go
```diff
@@ -135,6 +137,132 @@ func dumpOpenStackImages(providerClient *gophercloud.ProviderClient, clientOpts
...
+	providerClient, clientOpts, err := getProviderClient(e2eCtx)
+	if err != nil {
+		_, _ = fmt.Fprintf(GinkgoWriter, "error creating provider client: %s\n", err)
+		return nil, nil
+	}
+
+	computeClient, err := openstack.NewComputeV2(providerClient, gophercloud.EndpointOpts{Region: clientOpts.RegionName})
+	if err != nil {
+		return nil, fmt.Errorf("error creating compute client: %v", err)
+	}
```
tobiasgiese, line 150: I think it's better to move the creation of the clients into a helper function.
Further, the usage of the gomage `Expect` would be better instead of the usual `err != nil` condition in the e2e ginkgo tests.

Issue #1023, opened 2021-10-07
Consolidate Networks and Ports in OpenStackMachineSpec


mdbooth:
/kind feature

**Describe the solution you'd like**
Currently in MachineSpec we have 2 different ways to add a vif to a VM: [Networks](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/007a8278293efdded741831de584fcc4a1d6867e/api/v1alpha4/openstackmachine_types.go#L55) and [Ports](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/007a8278293efdded741831de584fcc4a1d6867e/api/v1alpha4/openstackmachine_types.go#L59) which are defined as lists of [NetworkParam](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/007a8278293efdded741831de584fcc4a1d6867e/api/v1alpha4/types.go#L57-L68) and [PortOpts](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/007a8278293efdded741831de584fcc4a1d6867e/api/v1alpha4/types.go#L123-L154) respectively.

The difference can be summarised as:

Networks:
* Networks and Subnets can be specified explicitly (which is also efficient since  #994) or by filter.
* Cannot specify port options

Ports:
* Networks and Subnets must be specified explicitly
* Possible specify a port with no subnet
* Can specify all port options

Apart from being confusing for the user, the difference is also a source of potential maintenance bugs and different ports can accidentally end up with slightly different semantics.

I think we could easily create a consolidated API which combines the flexible network selection of Networks with the ability to specify all port options of Ports.

iamemilio:
> The other (very) minor consideration is whether network and subnet should be plural in your proposal, since the filter can match multiple networks and I expect the intention is that the behavior would be forEach like the existing networks:. Then it would be strange that network singular specifies the creation of multiple ports. Perhaps the clearest would be if the filter part were explicit, kind of like how networks is now

I think the plural would be appropriate

macaptain:
> This can be handled like this:
> 
> ```yaml
> ports:
>   - network:
>       id: uuid
>     fixedIPs:
>       subnet:
>         networkID: uuid
> ```

The difference between that snippet and the one using the `networks:` API is that the `networks:` API will create a port per subnet. But your example will create a single port with multiple fixed IPs on different subnets.

iamemilio:
## Summary
[...]

PR #1052, opened 2021-11-19
🏃 Fix conversion-gen in Makefile

hidekazuna:
**What this PR does / why we need it**:

This PR fixes conversion-gen in Makefile.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

When proposing #1047 ,  `api/v1alpha3/zz_generated.conversion.go` and `api/v1alpha4/zz_generated.conversion.go` are generated manually. Do not worry for the PR 😄 

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

/hold

PR #970, opened 2021-08-12
update from v1alpha3 to v1alpha4 in Makefile

jichenjc:
**What this PR does / why we need it**:

in Makefile we are still using v1alpha3 (review Makefile found this) 
seems we need update to v1alph4 anyway though I didn't find changes after run the generate-go command

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #1059, opened 2021-11-22
:sparkles: Port rework

iamemilio:
This is the first half of the fix to #1023. These changes make the port API comperable to the network and subnet API, setting them up for deprecation in an upcoming set of changes. 

This is a re-upload of the former pull request because I destroyed the git history :scream_cat:


hidekazuna:
@iamemilio I think `zz_generated.conversion.go` are not properly generated since update from v1alpha3 to v1alpha4 in Makefile(#970) merged. I proposed a PR #1052. Please check `zz_generated.conversion.go` are generated correctly.

iamemilio:
That would explain why the conversion was having so many issues. I had to manually correct them. Hopefully I can undo that.

mdbooth:
I split this branch out into separate commits and fixed up the conversion unit tests: https://github.com/shiftstack/cluster-api-provider-openstack/tree/port-rework-split. Up to you how to use it.

go.mod
```diff
@@ -13,7 +13,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
+	golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871
```
hidekazuna, line 16: This is duplicate of #1072 , right?

pkg/cloud/services/networking/port.go
```diff
@@ -187,6 +191,32 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 	return port, nil
 }
 
+func (s *Service) getSubnetIDForFixedIP(subnet *infrav1.SubnetFilter, networkID string) (string, error) {
+	if subnet == nil {
```
hidekazuna, line 195: Is this normal case? if this is not normal, need to return error.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -330,7 +338,7 @@ spec:
   - cluster-tag
 ```
 
-To tag resources specific to a machine, add a value to the tags field in the `OpenStackMachineTemplate` spec like this:
+To tag resources specific to a machine, add a value to the tags field in `controlplane.yaml` and `machinedeployment.yaml` like this:
```
hidekazuna, line 341: What are `controlplane.yaml` and `machinedeployment.yaml`?

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -243,12 +243,20 @@ metadata:
   namespace: <cluster-name>
```
hidekazuna, line 243: L182: `please note that it is possible to get multiple networks as a result.` But network filter must only return one network in my understanding.
YAML example should change in Network Filters, Multiple Networks, and Subnet Filters.

go.mod
mdbooth: It is literally #1072! I rebased on top of it to make it easier to see compilation errors while refactoring.

pkg/cloud/services/networking/port.go
mdbooth: It's valid to specify a FixedIP with IPAddress only (no subnet). From the docs https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#ports

```
If you specify only an IP address, OpenStack Networking tries to allocate the IP address if the address is a valid IP for any of the subnets on the specified network.
```

mdbooth (changes requested):
Thanks for the review @hidekazuna!

@iamemilio would you like me to address the documentation issues, or do you have time to do it?

docs/book/src/clusteropenstack/configuration.md
iamemilio: the source of confusion you are having is that there is a difference between the networking api and the ports api right now. We intend to remove networks all together and pivot all the way to just using ports. That is going to be handled in a separate issue, which is why I chose to leave this alone for now

iamemilio: Network and subnet queries for ports should only return one value. But for networks and subnets, we have decided to implement this weird poweruser api.

iamemilio: for more on what is to come, see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1023

pkg/cloud/services/networking/port.go
hidekazuna: Thanks, I studied.

docs/book/src/clusteropenstack/configuration.md
hidekazuna: Yes, I see #1023 and asked. I understand I missed this is half of fixing #1023, thanks.

docs/book/src/clusteropenstack/configuration.md
hidekazuna: I confirmed this was fixed.

mdbooth:
/hold we need to rebase this to remove the bottom commit

@hidekazuna has concerns in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1072, and if we don't rebase we'd be pushing the commit anyway here. That fix was useful (to me, anyway) during development, but removing it now shouldn't have any effect.

mdbooth:
I've:
* Rebased to remove the bottom commit
* Squashed the documentation fix

mdbooth:
/hold cancel

Looking for an LGTM from somebody. Probably not appropriate for me to do it.

tobiasgiese:
> Looking for an LGTM from somebody.

I'll take a look

pkg/cloud/services/compute/instance.go
```diff
@@ -147,9 +145,22 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster,
...
+			if netID == "" {
+				netIDs, err := s.networkingService.GetNetworkIDsByFilter(port.Network.ToListOpt())
+				if err != nil {
+					return nil, err
+				}
+				if len(netIDs) > 1 {
+					return nil, fmt.Errorf("network filter for port %s returns more than one result", port.NameSuffix)
+				} else if len(netIDs) == 0 {
+					return nil, fmt.Errorf("network filter for port %s returns no networks", port.NameSuffix)
+				}
```
tobiasgiese, line 159: nit: I think we can reduce the `else if` here, just for the sake of readability.

api/v1alpha3/conversion.go
```diff
@@ -201,3 +201,58 @@ func Convert_Slice_v1alpha3_Network_To_Slice_v1beta1_Network(in *[]Network, out
...
 }
+
+func Convert_v1alpha3_SubnetFilter_To_v1beta1_SubnetFilter(in *SubnetFilter, out *v1beta1.SubnetFilter, s conversion.Scope) error {
+	out.Name = in.Name
+	out.Description = in.Description
+	if in.ProjectID != "" {
+		out.ProjectID = in.ProjectID
+	} else {
+		out.ProjectID = in.TenantID
+	}
```
tobiasgiese, line 212: nit: we could write this without the else

```suggestion
	out.ProjectID = in.TenantID
	if in.ProjectID != "" {
		out.ProjectID = in.ProjectID
	}
```

api/v1alpha4/conversion.go
```diff
@@ -107,3 +109,123 @@ func (r *OpenStackMachineTemplateList) ConvertFrom(srcRaw ctrlconversion.Hub) er
...
+	out.NotTags = in.NotTags
+	out.NotTagsAny = in.NotTagsAny
+	return nil
+}
+
+func Convert_v1alpha4_PortOpts_To_v1beta1_PortOpts(in *PortOpts, out *v1beta1.PortOpts, s conversion.Scope) error {
+	err := autoConvert_v1alpha4_PortOpts_To_v1beta1_PortOpts(in, out, s)
+	if err != nil {
+		return err
+	}
```
tobiasgiese, line 172: This could be written in short syntax.
We should probably think about adding the [ifshort golangci linter](https://github.com/esimonov/ifshort)

tobiasgiese (commented):
Just a few nits, don't take it to serious :)
rest /lgtm

api/v1alpha4/conversion.go
mdbooth: Do you know how to do that?

api/v1alpha3/conversion.go
mdbooth: This is a toss-up. I probably wouldn't bother updating this one.

api/v1alpha4/conversion.go
tobiasgiese: fun fact: [it's already enabled](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/.golangci.yml#L28). Are we not calling the golangci-linter with our presubmit prow jobs?

mdbooth:
@iamemilio @tobiasgiese raised 3 nits. 2 are in code that I wrote, 1 is in your code.

I like the idea of adding the ifshort linter, but as always when adding a new linter we'd have to add a bunch of cleanups, so I'm happy to leave this until then. The other one I'm personally ok with, so it's up to you whether to address the other one now, in a follow up, or not at all... Just unhold if you want to merge.

tobiasgiese:
> @iamemilio @tobiasgiese raised 3 nits. 2 are in code that I wrote, 1 is in your code.
> 
> I like the idea of adding the ifshort linter, but as always when adding a new linter we'd have to add a bunch of cleanups, so I'm happy to leave this until then. The other one I'm personally ok with, so it's up to you whether to address the other one now, in a follow up, or not at all... Just unhold if you want to merge.

It's totally fine to merge it as it is. They're just mini nits I came across.
I'll check if the golangci-linter is working as expected (because the `ifshort` linter is already enabled). But this is totally disjoint of  this PR!

iamemilio:
I would prefer to merge what we have now, then address the nits that the linter picks up when we have it running in an automated way

PR #1061, opened 2021-11-24
✨Add unit tests for compute service

mdbooth:
This PR adds unit tests for CreateInstance and getImageID in the compute service.

It is split out from #1030, which was getting too large.

It also highlights a bug with port cleanup which will be addressed by a separate PR.

/hold


pkg/cloud/services/networking/service.go
```diff
@@ -57,15 +57,9 @@ func NewService(providerClient *gophercloud.ProviderClient, clientOpts *clientco
...
-		projectID, err = provider.GetProjectID(providerClient, clientOpts.AuthInfo.ProjectName)
-		if err != nil {
-			return nil, fmt.Errorf("error retrieveing project id: %v", err)
-		}
-	}
-	if projectID == "" {
-		return nil, fmt.Errorf("failed to get project id")
+	projectID, err := provider.GetProjectID(providerClient, clientOpts)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieveing project id: %v", err)
```
macaptain, line 62: tiny typo that was already in the code, not a blocker
```suggestion
		return nil, fmt.Errorf("error retrieving project id: %v", err)
```

pkg/cloud/services/compute/service.go
```diff
@@ -29,11 +29,8 @@ import (
 )
 
 type Service struct {
-	provider          *gophercloud.ProviderClient
 	projectID         string
-	computeClient     *gophercloud.ServiceClient
-	identityClient    *gophercloud.ServiceClient
-	imagesClient      *gophercloud.ServiceClient
+	computeService    Client
```
macaptain, line 33: I guess it's a philosophical difference at what point a Client becomes a Service. :smile: I think this is fine, the networking service is definitely a wrapper around the networking client. If the bare compute client serves its purpose, it makes sense to have `computeService` and `networkingService` named consistently in this struct.

pkg/cloud/services/provider/provider.go
```diff
@@ -170,7 +170,16 @@ type projects struct {
...
+func GetProjectID(client *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts) (string, error) {
+	if clientOpts.AuthInfo.ProjectID != "" {
+		return clientOpts.AuthInfo.ProjectID, nil
+	}
+
+	projectName := clientOpts.AuthInfo.ProjectName
+	if projectName == "" {
+		return "", fmt.Errorf("failed to get project id")
+	}
+
```
macaptain, line 182: Good call moving this block of code into this function.

pkg/cloud/services/compute/instance.go
```diff
@@ -31,21 +30,22 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
-	"github.com/gophercloud/utils/openstack/compute/v2/flavors"
 	"k8s.io/apimachinery/pkg/runtime"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
-	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
```
macaptain, line 40: :partying_face:

pkg/cloud/services/compute/instance.go
```diff
@@ -31,21 +30,22 @@ import (
...
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
 	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
 )
 
+// Polling interval between retries during instance creation
+// N.B. This is overridden in unit tests.
+var retryIntervalInstanceStatus = 10 * time.Second
```
macaptain, line 44: Rather than using a global var, did you consider keeping this as a `const`, adding `retryInterval` as an argument to the function, and using the value of the const when the function is called? Then this value can be maintained as constant in the codebase but you can choose what to set it as in the tests.

pkg/cloud/services/compute/instance_test.go
```diff
@@ -366,3 +376,477 @@ func TestService_getServerNetworks(t *testing.T) {
...
+	return &infrav1.OpenStackMachine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      openStackMachineName,
+			Namespace: namespace,
+		},
+		Spec: infrav1.OpenStackMachineSpec{
+			// ProviderID is set by the controller
+			// InstanceID is set by the controller
+			// FloatingIP is only used by the cluster controller for the Bastion
+			// TODO: Test Networks, Ports, Subnet, and Trunk separately
```
macaptain, line 519: Mostly we have the networkingService covered by tests in the networking package now, I guess this is more to add some kind of integration test for both packages.

macaptain (approved):
Amazing work on the tests. I can see that they're all running and passing in the CI log. Good to see that they caught a couple of bugs around port deletion as well.

pkg/cloud/services/compute/service.go
mdbooth: Yeah, it's a client. This is easy to fix so I'll update.

pkg/cloud/services/provider/provider.go
mdbooth: I think we're going to have to fix this in order to support application credentials, btw, which is another reason to centralise it.

cc @pierreprinetti

pkg/cloud/services/compute/instance_test.go
mdbooth: Honestly I don't want to get too philosophical on what's a unit test and what's an integration test. I've been burned before by the maintenance burden and general uselessness of puritanically-scoped unit tests, so I've tried to keep it simple here.

* The scope of the test is the exported method (CreateInstance/DeleteInstance).
* The test boundary is what we can reasonably mock, which is currently the gophercloud calls.

If we were able to mock network service that might also be a practical boundary, but to be honest I quite the emergent benefit of revealing all the external API calls we're making, and they haven't felt *too* onerous yet.

pkg/cloud/services/compute/instance.go
mdbooth: I've added https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1061/commits/75ba1a99d206156d0f0889edc14a301f1b8027d8 to see if I like this better than the global. Have a look and tell me what you think. I'm personally undecided on which version is nicer, so I'm very much interested in a second opinion!

I'll squash this commit if we decide it's better, so:

/hold

pkg/cloud/services/compute/instance_test.go
macaptain: Yes, I do like that the tests show up all the API calls to OpenStack.

pkg/cloud/services/compute/instance.go
macaptain: I could be a bit paranoid about global variables, but I like passing the retry interval into the function just a little better. It has the drawback of adding a function argument which isn't absolutely essential, but I like that the dependency on the value is made clear in the function signature and that we maintain the original value as a const. What do you think? Maybe one day if more arguments are added to the function, we can tidy up by passing a config struct.

mdbooth: Ok, I'll squash the new commit into the existing one.

Thanks for the reviews!

mdbooth:
I think these are good to go.

/hold cancel

jichenjc:
/approve

works for me, thanks for the PR~

Issue #1060, opened 2021-11-23
Add Unit Tests to Cover generated go code


iamemilio:
/kind feature

cluster-api-provider-aws has unit tests to verify that the generated conversions function. We have introduced breakage in our go code during our upgrade to v1beta1 that could have been avoided by having tests like this in place.

See:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1alpha4/conversion_test.go
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1alpha3/conversion_test.go

hidekazuna:
Nice feature.

jichenjc:
as the PR merged, should we close this issue? @iamemilio

jichenjc:
@mdbooth I saw you submit the PR for this 
should we consider this as `done` ?

mdbooth:
I think so!

PR #1066, opened 2021-11-26
🐛Fix conversion of IdentityRef converting between v1alpha4 and v1beta1

### LINKED ISSUE #1065: IdentityRef is not preserved converting between v1alpha4 and v1beta1
/kind bug

Discovered by unit testing. IdentityRef annotated with `+k8s:conversion-gen=false`, possibly because it was new in v1alpha4 and we wanted to ignore it when converting from v1alpha3. However, we don't want to ignore it when converting to v1beta1, but it is still ignored.

mdbooth:
The most interesting parts of this PR are the new unit tests. The tests are little more than a 'do the fields line up' test, but they did highlight that we were missing conversion of IdentityRef.

I will be adding to these tests in #1030

Fixes #1065 #1060

### DIFF
--- a/api/v1alpha3/conversion_test.go
+++ b/api/v1alpha3/conversion_test.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha3
+
+import (
+	"testing"
+
+	fuzz "github.com/google/gofuzz"
+	"github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/api/apitesting/fuzzer"
+	runtime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


api/v1alpha3/conversion_test.go
```diff
@@ -0,0 +1,259 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
hidekazuna, line 2: 2020 should be 2021.

api/v1alpha4/conversion_test.go
```diff
@@ -0,0 +1,331 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
hidekazuna, line 2: same here.

hidekazuna (commented):
Overall lgtm, some nits.

tobiasgiese:
/lgtm
good catch, thanks for the unit tests 👍🏻

tobiasgiese:
@mdbooth what about #1060, shouldn't this fit here too?

mdbooth:
> @mdbooth what about #1060, shouldn't that also fit here?

Looks like it. Also I didn't realise we already had infrastructure for this! Those AWS tests are a whole lot more readable. I'll check out what they do.

tobiasgiese:
+1 for the *new* conversion unit tests, this is really much more readable 👌🏻

mdbooth:
Apart from more readable they're also significantly more thorough! They uncovered some missing pieces of CloudSecret<->IdentityRef conversion which I've hopefully rectified here.

iamemilio:
/lgtm
I love the fuzz tests

iamemilio:
/lgtm
I love the fuzz tests

api/v1alpha4/conversion_test.go
mdbooth: Ditto.

jichenjc:
/approve

thanks for the test cases :)

PR #1069, opened 2021-12-03
🏃Fix generated clouds.yaml for local e2e tests

mdbooth:
When creating multiple capo-e2e devstacks on the same shared cloud you will need to give them different names. We were also using this name when writing the local clouds.yaml. However, that requires also modifying the local e2e_conf.yaml. As there is no naming conflict between developers' local clouds.yaml copies these can safely be the same everywhere, which is easier to work with.


raildo (approved):
LGTM

Issue #1024, opened 2021-10-11
cluster-api v1beta1 adoption ?


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

I heard cluster-api v1beta1 is going to be out soon ... should we start to consider the adoption 
and corresponding v1beta1 version of CAPO as well?

update:
https://kubernetes.io/blog/2021/10/08/capi-clusterclass-and-managed-topologies/ might be something
related as well, need further anaylsis
**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

pfisterer:
I guess that would be very good. I failed to try cluster-api the first time today because the released version v1.0.0 fails with an error when using cluster-api-provider-openstack. Just for your reference, this is the corresponding (closed) issue: https://github.com/kubernetes-sigs/cluster-api/issues/5425

mkjpryor:
The v1beta1 versions of the Cluster API core resources won’t adopt v1alpha4 resources from infrastructure providers. So this needs sorting one way or another, either by this provider moving to v1beta1 or by Cluster API core resources not expecting infrastructure resources to match their version.

jichenjc:
from this https://github.com/kubernetes-sigs/cluster-api/issues/5425#issuecomment-943150790
seems yes, we need v1.0.x implementation for cluster-api now

>>>The current situation means we can’t upgrade the core Cluster API components to benefit from changes there.

I don't fully understand this ... are you suggesting that even upgrade to v1beta1, CAPO still can't work with cluster-api 1.0.x ?

mkjpryor:
@sbueringer I didn’t expect you personally to answer it! It definitely needs to be a community discussion.

What I am saying is that the discussion needs to happen ASAP. I don’t think we want to be lagging too far behind and, in my experience so far, the provider is pretty reliable. If we just need to freeze the API and say this is v1beta1 then we do need to have a discussion about anything else that we want in the API to be able to do that. 

We already have customers now saying “Cluster API is production ready - can we use it?” and we still have to say “well actually the OpenStack provider is still in alpha”, which they are less happy with.

mdbooth:
[...]

PR #1070, opened 2021-12-04
✨ Allow webhook changes to OpenStackCluster.Spec.Bastion

### LINKED ISSUE #1054: Bastion host is immutable on 0.5.0 while it wasnt on 0.4.0
/kind bug

On a `OpenStackCluster` I ussualy config the bastion specs and set `enable` to `false`. I toggle the enable flag during troubleshooting or checking various components on the nodes. And i disable it again. This is rather useful to not waste capacity when its not required and also not have a backdoor to the nodes when its not required either. 

- upgrade to CAPO `0.5.0` and rest of the components to `1.0.1`
- toggle the bastion enable flag to `true`
- error as it is immutable

- create the bastion host as this was possible on 0.4.0

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `0.5.0`
- Cluster-API version: `1.0.1`
- OpenStack version: `Stein`
- Kubernetes version (use `kubectl version`): many, shouldnt affect it
- OS (e.g. from `/etc/os-release`): ubuntu20.04, shouldnt affect it either

tobiasgiese:

This PR allows `OpenStackCluster.Spec.Bastion` to be updated (i.e., to be mutable).

Fixes #1054

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha4/openstackcluster_webhook.go
+++ b/api/v1alpha4/openstackcluster_webhook.go
@@ -17,12 +17,13 @@ limitations under the License.
 package v1alpha4
 
 import (
+	"fmt"
 	"reflect"
 
-	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -65,40 +66,24 @@ func (r *OpenStackCluster) ValidateCreate() error {
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
-func (r *OpenStackCluster) ValidateUpdate(old runtime.Object) error {
+func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 	var allErrs field.ErrorList
-
-	newOpenStackCluster, err := runtime.DefaultUnstructuredConverter.ToUnstructured(r)
-	if err != nil {
-		return apierrors.NewInvalid(GroupVersion.WithKind("OpenStackCluster").GroupKind(), r.Name, field.ErrorList{
-			field.InternalError(nil, errors.Wrap(err, "failed to convert new OpenStackCluster to unstructured object")),
-		})
-	}
-	oldOpenStackCluster, err := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
-	if err != nil {
-		return apierrors.NewInvalid(GroupVersion.WithKind("OpenStackCluster").GroupKind(), r.Name, field.ErrorList{
-			field.InternalError(nil, errors.Wrap(err, "failed to convert old OpenStackCluster to unstructured object")),
-		})
+	old, ok := oldRaw.(*OpenStackCluster)
+	if !ok {
+		return apierrors.NewBadRequest(fmt.Sprintf("expected an OpenStackCluster but got a %T", oldRaw))
 	}
 
 	if r.Spec.IdentityRef != nil && r.Spec.IdentityRef.Kind != defaultIdentityRefKind {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
-	newOpenStackClusterSpec := newOpenStackCluster["spec"].(map[string]interface{})
-	oldOpenStackClusterSpec := oldOpenStackCluster["spec"].(map[string]interface{})
-
-	// get controlPlaneEndpoint, something like {"host":"", "port":""}
-	cpe := oldOpenStackClusterSpec["controlPlaneEndpoint"].(map[string]interface{})
-
-	// allow change only for the first time
-	host, ok := cpe["host"].(string)
-	if ok && len(host) == 0 {
-		delete(oldOpenStackClusterSpec, "controlPlaneEndpoint")
-		delete(newOpenStackClusterSpec, "controlPlaneEndpoint")
+	// Allow change only for the first time.
+	if old.Spec.ControlPlaneEndpoint.Host == "" {
+		old.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{}
+		r.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{}
 	}
 
-	if !reflect.DeepEqual(oldOpenStackClusterSpec, newOpenStackClusterSpec) {
+	if !reflect.DeepEqual(old.Spec, r.Spec) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
 	}
 
--- a/api/v1alpha4/openstackcluster_webhook.go
+++ b/api/v1alpha4/openstackcluster_webhook.go
@@ -105,6 +105,18 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		r.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{}
 	}
 
+	// Allow changes to the bastion spec only if no bastion host is deployed (i.e. Spec.Bastion.Enabled=false).
+	if old.Status.Bastion == nil {
+		old.Spec.Bastion = &Bastion{}
+		r.Spec.Bastion = &Bastion{}
+	}
+
+	// Allow toggling the bastion enabled flag.
+	if old.Spec.Bastion != nil && r.Spec.Bastion != nil {
+		old.Spec.Bastion.Enabled = true
+		r.Spec.Bastion.Enabled = true
+	}
+
 	if !reflect.DeepEqual(old.Spec, r.Spec) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
 	}
--- a/api/v1alpha4/openstackcluster_webhook.go
+++ b/api/v1alpha4/openstackcluster_webhook.go
@@ -74,7 +74,29 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 	}
 
 	if r.Spec.IdentityRef != nil && r.Spec.IdentityRef.Kind != defaultIdentityRefKind {
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "identityRef", "kind"),
+				r.Spec.IdentityRef, "must be a Secret"),
+		)
+	}
+
+	// Allow changes to Spec.IdentityRef.Name.
+	if old.Spec.IdentityRef != nil && r.Spec.IdentityRef != nil {
+		old.Spec.IdentityRef.Name = ""
+		r.Spec.IdentityRef.Name = ""
+	}
+
+	// Allow changes to Spec.IdentityRef if it was unset.
+	if old.Spec.IdentityRef == nil && r.Spec.IdentityRef != nil {
+		old.Spec.IdentityRef = &OpenStackIdentityReference{}
+		r.Spec.IdentityRef = &OpenStackIdentityReference{}
+	}
+
+	if old.Spec.IdentityRef != nil && r.Spec.IdentityRef == nil {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "identityRef"),
+				r.Spec.IdentityRef, "field cannot be set to nil"),
+		)
 	}
 
 	// Allow change only for the first time.
--- a/api/v1alpha4/openstackcluster_webhook_test.go
+++ b/api/v1alpha4/openstackcluster_webhook_test.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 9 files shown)


api/v1alpha4/openstackcluster_webhook.go
```diff
@@ -17,12 +17,13 @@ limitations under the License.
...
 
 import (
+	"fmt"
 	"reflect"
 
-	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
```
seanschneeweiss, line 26: Not sure, but should v1alpha4 of CAPO use CAPI v1alpha4?
```suggestion
	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
```

tobiasgiese: For CAPO v0.5.0 we‘re using CAPI v1beta1 (see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1024#issuecomment-968406388)

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -0,0 +1,86 @@
...
+				},
+			},
+			newTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+					IdentityRef: &OpenStackIdentityReference{
+						Kind: "Secret",
+					},
+				},
+			},
```
mdbooth, line 71: These are the same. Did you intend to change IdentityRef.Name?

api/v1beta1/openstackcluster_webhook_test.go
```diff
@@ -0,0 +1,86 @@
...
+				},
+			},
+			newTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+					IdentityRef: &OpenStackIdentityReference{
+						Kind: "Secret",
+					},
+				},
+			},
```
mdbooth, line 71: As for v1alpha4

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -71,6 +71,58 @@ func TestOpenStackCluster_ValidateUpdate(t *testing.T) {
...
+					Bastion: &Bastion{
+						Instance: OpenStackMachineSpec{
+							CloudName: "foobarbaz",
+							Image:     "foobarbaz",
+							Flavor:    "medium",
+						},
+						Enabled: true,
+					},
+				},
+			},
```
mdbooth, line 101: This is an interesting one. I note that in this unit test we're changing both the bastion spec and the enabled flag. This is fine: we will create a new Bastion with the new spec.

However, note we also permit changing the bastion spec without flipping enabled, but this won't work. It won't fail, but the bastion will not be updated until the enabled flag is flipped to false and the bastion is deleted.

What do we want to do about this? Options:

1. Ignore it
2. Document it
3. Don't allow changing bastion spec if status.Bastion != nil
4. Make it work

For now I feel like 2 + 3 would be user friendly and relatively simple.

mdbooth (changes requested):
Thanks for the unit tests! Couple of discussion points.

api/v1alpha4/openstackcluster_webhook_test.go
tobiasgiese: You're totally right. This is a leftover from a previous test, but I will add the `IdentityRef.Name` 👍🏻

api/v1alpha4/openstackcluster_webhook_test.go
tobiasgiese: I totally agree that we should document this behavior.
Regarding 3: do you think it's worth the effort to update our `reconcileBastion` to delete the bastion if sth has changed? Or is it enough to only allow changes if the enabled flag is false? Both are fine for me, I just think from the customer's perspective.
WDYT?

mdbooth: I feel like automatically deleting and recreating the bastion if something changed would be the most user friendly, but I should strongly caveat that with: I don't feel like I understand all the potential use cases here. Have we firmly set an expectation that we can blow the bastion away? Nobody's running anything on the bastion which is going to suffer data loss? I'd hope not.

I also think it would be fine to do 3 and do 4 later. I don't think they're incompatible: if the user writes orchestration to delete and recreate the bastion if they need to change it, that will continue to work even if we remove the requirement to do it. We're also being nice to them in the meantime by pointing out that it doesn't work.

tobiasgiese: Okay, I'll change my PR to 2 + 3 👍🏻
Thanks!

tobiasgiese: I updated my PR, can you PTAL @mdbooth? 🙂 
Documentation will be pushed during the day.

api/v1alpha4/openstackcluster_webhook.go
```diff
@@ -77,6 +77,18 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	// Allow changes to Spec.IdentityRef.Name.
+	if old.Spec.IdentityRef != nil && r.Spec.IdentityRef != nil {
+		old.Spec.IdentityRef.Name = ""
+		r.Spec.IdentityRef.Name = ""
+	}
```
mdbooth, line 84: I *think* this is ok. Changing IdentityRef.Name is definitely a potential foot gun, but:
* I can't think of any reason to strictly forbid it
* There may be edge cases when it's useful to support allowing changing it

I'd appreciate a second opinion from anybody familiar with the original design decisions!

chrischdi: I was not part of the original design decisions but I also think it should be ok to allow changes.

E.g. maybe somebody wants to keep the old credential information and creating a new secret for the new credential.

tobiasgiese: CAPA is only disallowing the removal of a IdentityRef

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/awscluster_webhook.go#L136-L142

sbueringer: IIRC the identityref is only used during reconcile to get the OpenStack creds and then access the OpenStack API with them. (aka it is not used to store them somewhere in the cluster).

If I'm correct, I see no reason to deny changes to the identityRef

api/v1alpha4/openstackcluster_webhook_test.go
tobiasgiese: Added the documentation as well. Thx

api/v1alpha4/openstackcluster_webhook.go
mdbooth: That's also my understanding.

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -0,0 +1,172 @@
...
+	g := NewWithT(t)
+
+	tests := []struct {
+		name        string
+		oldTemplate *OpenStackCluster
+		newTemplate *OpenStackCluster
+		wantErr     bool
+	}{
+		{
+			name: "OpenStackCluster.Spec.IdentityRef with immutable spec",
```
mdbooth, line 35: nit: The tests names confused me, and perhaps are still confusing me a little. To me, mutability/immutability is a fundamental property of a thing. 'Immutable' to me means: "no changes at all, ever", whereas this is more like: "this specific change is not allowed".

For this specific test I might have written "Changing OpenStackCluster.Spec.IdentityRef.Kind is not allowed".

This is a nit! Please don't respin for this.

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -0,0 +1,172 @@
...
+func TestOpenStackCluster_ValidateCreate(t *testing.T) {
+	g := NewWithT(t)
+
+	tests := []struct {
+		name     string
+		template *OpenStackCluster
+		wantErr  bool
+	}{
+		{
+			name: "OpenStackCluster.Spec.IdentityRef with correct spec on create",
```
mdbooth, line 188: I like this name better 😀

api/v1alpha4/openstackcluster_webhook.go
```diff
@@ -105,6 +105,12 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		r.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{}
 	}
 
+	// Allow changes to the bastion spec only if no bastion host is deployed (i.e. Spec.Bastion.Enabled=false).
+	if old.Status.Bastion == nil {
+		r.Spec.Bastion = &Bastion{}
+		old.Spec.Bastion = &Bastion{}
+	}
```
mdbooth: Won't this prevent setting Enabled=false once the bastion is deployed? i.e. we can never undeploy it in order to make changes?

mdbooth (changes requested):
Side quest looks complete, but I think you've got a new problem in the main story line 😂

api/v1alpha4/openstackcluster_webhook_test.go
tobiasgiese: Updated the naming. WDYT?

api/v1alpha4/openstackcluster_webhook.go
tobiasgiese: Good catch 🙂 fixed it

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -111,6 +111,58 @@ func TestOpenStackCluster_ValidateUpdate(t *testing.T) {
...
+			oldTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+					Bastion: &Bastion{
+						Instance: OpenStackMachineSpec{
+							CloudName: "foobar",
+							Image:     "foobar",
+							Flavor:    "minimal",
+						},
+						Enabled: false,
```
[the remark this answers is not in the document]
tobiasgiese: fixed

api/v1alpha4/openstackcluster_webhook.go
tobiasgiese: I must have lost sight of the flag somehow 😅

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -0,0 +1,172 @@
...
+	g := NewWithT(t)
+
+	tests := []struct {
+		name        string
+		oldTemplate *OpenStackCluster
+		newTemplate *OpenStackCluster
+		wantErr     bool
+	}{
+		{
+			name: "OpenStackCluster.Spec.IdentityRef.Kind must always be Secret",
```
mdbooth, line 35: Thank you! 🙏

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -111,6 +111,124 @@ func TestOpenStackCluster_ValidateUpdate(t *testing.T) {
...
+							CloudName: "foobarbaz",
+							Image:     "foobarbaz",
+							Flavor:    "medium",
+						},
+						Enabled: true,
+					},
+				},
+			},
+			wantErr: true,
+		},
```
mdbooth, line 231: I think you're missing an interesting test here: flipping Enabled when the Bastion exists in status is allowed.

mdbooth (approved):
The clean unit tests make this so much easier to review, thanks!

I think there's an extra unit test you could add, but either way this
/lgtm

api/v1alpha4/openstackcluster_webhook_test.go
tobiasgiese: Good point. I added also another test `Disabling and changing the OpenStackCluster.Spec.Bastion while it's running is allowed`. This should be fine, as the bastion host will be destroyed anyway.

api/v1alpha4/openstackcluster_webhook_test.go
```diff
@@ -111,6 +111,164 @@ func TestOpenStackCluster_ValidateUpdate(t *testing.T) {
...
+		{
+			name: "Disabling the OpenStackCluster.Spec.Bastion while it's running is allowed",
```
mdbooth, line 233: 👍

mdbooth (approved):
/lgtm

Thanks for the extra test. As discussed in Slack, I was worried that allowing changes to the spec before the bastion had actually been removed could potentially introduce a race which somebody might hit if, e.g. they scripted it. This approach feels simple and correct.

mdbooth:
@tobiasgiese Tests are all passing. Do you want to remove the hold?

tobiasgiese:
> @tobiasgiese Tests are all passing. Do you want to remove the hold?

I‘m still waiting for the approve of a CAPO maintainer. Prow won‘t merge until it‘s approved.

tobiasgiese:
ptal @hidekazuna @jichenjc 🙂 thanks

mdbooth:
> > @tobiasgiese Tests are all passing. Do you want to remove the hold?
> 
> I‘m still waiting for the approve of a CAPO maintainer. Prow won‘t merge until it‘s approved.

That's ok. If you remove the /hold it will merge immediately after it's approved.

tobiasgiese:
> That's ok. If you remove the /hold it will merge immediately after it's approved.

I know that 🙂 we're using Prow at work.
I like to know when my PR is going to be merged. Also, I don't like to handover the responsibility (also as it's sometimes a approve-blocker for someone).

seanschneeweiss:
/LGTM
Agree very much that the tests make it easy to review 👍

hidekazuna:
@tobiasgiese You can remove /hold if you want 😃

tobiasgiese:
thx @hidekazuna :)

/unhold

PR #1072, opened 2021-12-06
🐛Bump golang.org/x/crypto and golang.org/x/net

mdbooth:
This fixes errors reported by gopls in the e2e tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
+	golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871
+	golang.org/x/net v0.0.0-20211123203042-d83791d6bcd9 // indirect
 	gopkg.in/ini.v1 v1.63.2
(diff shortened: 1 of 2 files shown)


hidekazuna:
@mdbooth Could you show us the error in detail or the pointer to the error?

mdbooth:
> @mdbooth Could you show us the error in detail or the pointer to the error?

Yes. VSCode, which is simply passing on gopls's error (IIUC) reports:

```
error while importing sigs.k8s.io/cluster-api/test/framework/clusterctl: missing go.sum entry for module providing package golang.org/x/crypto/openpgp (imported by github.com/google/go-github/v33/github); to add:
	go get github.com/google/go-github/v33/github@v33.0.0compiler
```

It reports this everywhere clusterctl is imported in the tests. The impact is that I can't see compilation errors in any of the tests in my IDE while refactoring.

Honestly I don't pretend to understand this message, but it's extremely convenient to make it go away.

hidekazuna:
Thanks understand. I do not understand the message too, but both CAPA and CAPZ use `golang.org/x/crypto v0.0.0-20210921155107-089bfa567519` so if this version can resolve the issue, I want to use the version.

jichenjc:
agree to open an issue to track what's really wrong :)

and whatever works should be good for us :)
so maybe we can be consistent to other provider as well

PR #1087, opened 2021-12-14
🐛 Add a failing unit test for leaking trunks

mdbooth:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance_test.go
+++ b/pkg/cloud/services/compute/instance_test.go
@@ -24,14 +24,20 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/golang/mock/gomock"
+	common "github.com/gophercloud/gophercloud/openstack/common/extensions"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/compute/instance_test.go
```diff
@@ -728,13 +788,64 @@ func TestService_CreateInstance(t *testing.T) {
...
+				networkRecorder.ReplaceAllAttributesTags("trunks", trunkUUID, attributestags.ReplaceAllOpts{Tags: []string{"test-tag"}}).Return(nil, nil)
+
+				// Looking up the second port fails
+				networkRecorder.ListPort(ports.ListOpts{
+					Name:      "test-openstack-machine-1",
+					NetworkID: networkUUID,
+				}).Return(nil, fmt.Errorf("test error"))
+
+				// We should cleanup the first port and its trunk
+				networkRecorder.DeleteTrunk(trunkUUID).Return(nil)
```
jichenjc, line 837: controller.go:137: missing call(s) to *mock_networking.MockNetworkClient.DeleteTrunk(is equal to 2cf74a9f-3e89-4546-9779-20f2503c4ae8 (string)) /home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute/instance_test.go:837

ok, here's the error and it's expected and it's caused by line 834 above

I agree this is reasonable that we should handle ,however, we have various logic after port creation
so , will all those places need to be handles or as you mentioned, put those stuffs into `status` field
so that we can avoid lost info and delete them in a whole bunch? I actually prefer to this way...thoughts?

jichenjc: ok, saw 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/compute/instance.go#L203

and seems if we add some logic there, it should be good enough

jichenjc: @mdbooth I assume if we do some error handling here 

e.g add delete trunk might help us 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/compute/instance.go#L213

see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1095 for some discussion points..

mdbooth:
This has been superceded.

PR #1094, opened 2021-12-16
Fail fast without create port at all

jichenjc:

fail fast, inspired by recent discussion in port leak 
for those kind of error handling, we can fail fast without go to create port then delete port stage
if we know it's definitely will fail...

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -194,6 +194,20 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 	accessIPv4 := ""
 	portList := []servers.Network{}
 
+	if instanceSpec.Subnet != "" && accessIPv4 == "" {
+		return nil, fmt.Errorf("no ports with fixed IPs found on Subnet %q", instanceSpec.Subnet)
+	}
+
+	imageID, err := s.getImageID(instanceSpec.Image)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -193,6 +193,20 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
...
 	accessIPv4 := ""
 	portList := []servers.Network{}
+	
+	if instanceSpec.Subnet != "" && accessIPv4 == "" {
+		return nil, fmt.Errorf("no ports with fixed IPs found on Subnet %q", instanceSpec.Subnet)
+	}
+
+	imageID, err := s.getImageID(instanceSpec.Image)
+	if err != nil {
+		return nil, fmt.Errorf("create new server: %v", err)
```
chrischdi, line 203: I think that's the wrong message 🙂

chrischdi (commented):
Not introduced by your PR but could be easily fixed.

Also ci looks like it wants some gofmt.

pkg/cloud/services/compute/instance.go
jichenjc: updated :)

PR #1101, opened 2021-12-27
Allow image uuid to be used in instanceSpec

### LINKED ISSUE #1100: Allow image uuid to be used in instanceSpec
/kind feature

we only allow image Name to be input as InstanceSpec part, input uuid lead to 
issue, so we need allow image UUID to be used as well

jichenjc:

Fixes #1100 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


mdbooth:
Looks good. What would be the best scope to unit test it?

jichenjc:
@mdbooth updated with a new test , thanks for reminder

chrischdi:
I think nobody does but this would break users which use image names that look like an id?! (Something for the changelog?!)

Also we check if the image exists when image name is used. Should we do that too if it is an UUID?

Nevertheless I think this change is fine :-)

/lgtm

jichenjc:
>>>Also we check if the image exists when image name is used. Should we do that too if it is an UUID?

um... you made a good point ,will double check this

jichenjc:
@chrischdi updated :)

chrischdi:
/lgtm

👍

mdbooth:
> > > > Also we check if the image exists when image name is used. Should we do that too if it is an UUID?
> 
> um... you made a good point ,will double check this

I doubt anybody does this because I'm pretty sure it would also break heuristics in the cli tools? @stephenfin would know for sure.

mdbooth:
I may be over-thinking this, but some thoughts:

We're still validating this in the event that we think it's a UUID by listing images with the given ID. In other places where the user passes us a UUID we're trusting it. If it's wrong the user will get an OpenStack error later, and it's nicer to the API(*). Would it be consistent to do the same here? Having said that, this is different from the other cases because this is using a heuristic to determine the type of the input, where as the other cases are unambiguous.

On that theme, is it worth adding a separate field for ImageID which is unambiguously an image id?

Or if we decide to go with the combined field and we validate it we could fall back to lookup by name if lookup by ID doesn't match?

I'm not necessarily opposed to the approach in this patch, btw. In practise I doubt we'll ever hit a real world issue doing a heuristic check on something that looks like a UUID. The neatnik in me would probably prefer a different field, and failing that to drop the validation anyway, but whatever we pick I'm not personally set on anything. This is the kind of API change which is hard to unroll later, though, so it's worth taking a moment to consider the options first.

(*) Incidentally, we have an internal OpenStack with *serious* glance API performance issues. This query likely takes a minute 😔

chrischdi:
Introducing a separate field sounds reasonable and more straight forward instead of having a hidden feature :-)

stephenfin:
Personally, I'd opt for a separate field. Tools like OSC allow you to specify a UUID or name because they're designed for use by humans who deal with names by default. Having to remember to use the UUID or pass a flag each time would be a pain. When you're writing config files or the likes this and you have an easy way to disambiguate, then I'm not sure why you wouldn't. I've included the logic used by OSC below in case it's helpful though.

Try to retrieve by UUID.
On success, return image.
On failure, try to list and filter by name.
If 0 matches, error.
If > 1 matches, error.
If exactly 1 match, return image.

The only variations in this depend on whether the user is admin and whether the API supports `all_tenants` or `all_projects` parameters. In the former case, a `GET` by UUID will obviously work for more than the current project if the user is admin. In the latter case, this parameter can/must be passed to allow admins to retrieve things by name if they exist in another project (unlike retrieving by UUID). We don't set this by default since it would result in more errors since there's a non-zero chance of a name being used in more than one project

jichenjc:
thanks all for the great comments :) 

will re-consider this and add more thoughts then re-submit the PR :) thanks~

jichenjc:
@mdbooth @chrischdi @stephenfin as we are already in beta version
I believe add a field named ImageUUID should be fine ?(but I will keep the `Image` variable as it seems ImageName is more reasonable) ？

mdbooth (commented):
Interested to hear alternative opinions on validating image uuid, btw. My thoughts on validating UUIDs in general are:

We fetch image id for name because the API requires us to, so generating an earlier error message for the user is really just a side-effect, not the primary purpose of it. They will still get an error message telling them the image ID is wrong, they'll just get it later during server create. There are also many other things we don't validate about the user input which will throw errors during server create, e.g.:

* The image is larger than the flavor's disk
* Insufficient quota
* Network/Subnet UUID
* If FailureDomain exists

My personal feeling is that as long as we're making the error generated by OpenStack easy for the user to find we're best not duplicating validation that OpenStack is already doing.

mdbooth:
Thanks!

/lgtm

pkg/cloud/services/compute/instance.go
```diff
@@ -410,30 +411,40 @@ func (s *Service) getServerNetworks(networkParams []infrav1.NetworkParam) ([]inf
...
 }
 
+// Helper function for getting image ID from name or ID.
+func (s *Service) getImageID(imageUUID, imageName string) (string, error) {
+	if imageUUID != "" {
+		// we return imageUUID without check
+		return imageUUID, nil
+	} else if imageName != "" {
+		return s.getImageIDFromName(imageName)
+	}
```
seanschneeweiss, line 443: We can omit the `else if` here.
```suggestion
	} 
	if imageName != "" {
		return s.getImageIDFromName(imageName)
	}
```

seanschneeweiss:
I really like this change, great idea to put it into a separate field.
/lgtm

jichenjc:
/hold cancel 

thanks all for the review and help~

Issue #1106, opened 2022-01-11
Mandatory Tags field for trunk setup


Xenwar:
When enabling trunk at port or instance level without tags field, provisioning a CP node is not possible for the following reason.

`E0104 10:21:04.379852       9 controller.go:317] controller/openstackmachine "msg"="Reconciler error" "error"="error creating Openstack instance: Missing input for argument [Tags]" "name"="basic-1-control-plane-jv45x" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" `

However, if tags field is introduced, tests pass and deployment continues.

```
      ports:
      - description: "primary with PortSecurity enabled, required for proper k8s cluster"
        disablePortSecurity: false
        trunk: true
        tags:
        - "tag1"
        allowedAddressPairs:
        - macAddress: "a4:b8:34:2b:ab:01"
          ipAddress: "10.10.10.10/24"
```

/kind bug

mdbooth:
Guessing this is because ReplaceAllAttributesTags doesn't like being called without any tags?

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/381e7c732c6b73512c72fa5fa6fdffcb8f22d102/pkg/cloud/services/networking/service.go#L83-L99

Xenwar:
@mdbooth I agree.
Trying to debug further locally. Perhaps, adding some default tag or returning without updating/replacing in the above code could solve the problem.

mdbooth:
> @mdbooth I agree. Trying to debug further locally. Perhaps, adding some default tag or returning without updating/replacing in the above code could solve the problem.

Thinking the latter makes most sense.

PR #1107, opened 2022-01-11
🐛  Avoid replacing tags when no tags are provided.

Xenwar:
When trunk is enabled at Port or Instance level and no tags are provided at either level, deployment fails. 
This PRs fixes this issue by avoiding performing the ReplaceAllAttributesTags when no tags are provided.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1106

### DIFF
--- a/pkg/cloud/services/networking/service.go
+++ b/pkg/cloud/services/networking/service.go
@@ -81,6 +81,10 @@ func NewTestService(projectID string, client NetworkClient, logger logr.Logger) 
 // replaceAllAttributesTags replaces all tags on a neworking resource.
 // the value of resourceType must match one of the allowed constants: trunkResource or portResource.
 func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceType string, resourceID string, tags []string) error {
+	if len(tags) == 0 {
+		s.logger.Info("IgnoreReplaceAllAttributesTags", "no tags provided: , %s", resourceID)
+		return nil
+	}
 	if resourceType != trunkResource && resourceType != portResource {
 		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Invalid resourceType argument in function call")
 		panic(fmt.Errorf("invalid argument: resourceType, %s, does not match allowed arguments: %s or %s", resourceType, trunkResource, portResource))


pkg/cloud/services/networking/service.go
```diff
@@ -92,6 +92,10 @@ func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceT
 		panic(fmt.Errorf("invalid argument: resourceType, %s, does not match allowed arguments: %s or %s", resourceType, trunkResource, portResource))
 	}
 
+	if len(tags) == 0 {
+		record.Warnf(eventObject, "IgnoreReplaceAllAttributesTags", "no tags provided: , %s", resourceID)
```
jichenjc, line 96: looks good and I guess  a `info` is reasonable ... as it's normal no tags input 

or we can check whether tags is empty before call this function , both works

Xenwar: Moved it to the stop. 
on the `info`, I was thinking the same. but, no such function for `record`
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/record/recorder.go

pkg/cloud/services/networking/service.go
```diff
@@ -81,6 +81,11 @@ func NewTestService(projectID string, client NetworkClient, logger logr.Logger)
 // replaceAllAttributesTags replaces all tags on a neworking resource.
 // the value of resourceType must match one of the allowed constants: trunkResource or portResource.
 func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceType string, resourceID string, tags []string) error {
+	if len(tags) == 0 {
+		s.logger.Info("IgnoreReplaceAllAttributesTags", "no tags provided: , %s", resourceID)
```
mdbooth, line 85: nit: I suspect this log isn't quite right, but I don't think it's important enough to hold up the fix.

/lgtm

Xenwar: What should change it to ?
I can make the changes on another (ongoing PR) which also related to trunk.

mdbooth: I don't think we're using this consistently at the moment which is why I only called this a nit, but I don't think the `%s` will be expanded in the second argument as you intend because this is a key/value pair. IIUC (and I'm not 100% on this without testing), it should be something like:

```golang
s.logger.Info("no tags provided to ReplaceAllAttributesTags")
```

Looking at how we're initialising the logger, I don't think we even need to log the resource ID because the machine name is already included:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/2a51dde1157f259efce1569c7901ceef27954c7b/controllers/openstackmachine_controller.go#L89

and

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/2a51dde1157f259efce1569c7901ceef27954c7b/controllers/openstackcluster_controller.go#L91

IIUC this log message is just going to look ugly with an un-expanded string in it. I'm still happy to accept a follow-up PR as I doubt it would confuse anybody in practise and it fixes a real bug.

/lgtm

mdbooth (approved):
IIUC this log message is just going to look ugly with an un-expanded string in it. I'm still happy to accept a follow-up PR as I doubt it would confuse anybody in practise and it fixes a real bug.

/lgtm

Issue #1111, opened 2022-01-13
Remove duplicate tags before making API call


Xenwar:
When updating/replacing tags, duplicate values are not allowed. The way we are combining instance level and port level trunk entries does not take this into account. see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/96fae1e439d98ed884be6a42a8b81dbc687ea9f1/pkg/cloud/services/networking/port.go#L169-L171

To ask to the user to put unique values is not reasonable as the manifest may continue large dataset. It might be easier to make changes on the code that collects the tags. 

The error generated due the duplicate tags is shown below:

`E0112 14:22:04.091531       1 openstackmachine_controller.go:411] controller/openstackmachine "msg"="OpenStack instance cannot be created: error creating Openstack instance: Bad request with: [PUT http://10.0.3.15:9696/v2.0/ports/4f35aad9-4e70-4713-be6b-7643e59b1813/tags], error message: {\"NeutronError\": {\"type\": \"InvalidInput\", \"message\": \"Invalid input for operation: Duplicate items in the list: '6ts5m2'.\", \"detail\": \"\"}}" "error"="UpdateError" "cluster"="cluster-e2e-g1tjuw" "machine"="cluster-e2e-g1tjuw-md-3-64df4c9649-sxwvn" "name"="cluster-e2e-g1tjuw-md-3-s8qcz" "namespace"="e2e-g1tjuw" "openStackCluster"="cluster-e2e-g1tjuw" "openStackMachine"="cluster-e2e-g1tjuw-md-3-s8qcz" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" 
`

/kind bug

PR #1112, opened 2022-01-13
Remove duplicate tags before making API calls

Xenwar:
Remove duplicate tags before making API calls.

Tags that come from instance and port levels could contain duplicate values, causing failure when making api call.
This PR removes duplicate values before making the call. The change is done on the `replaceAllAttributesTags` function in order to avoid making changes in multiple places where tags are collected.

fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1111

### DIFF
--- a/pkg/cloud/services/networking/service.go
+++ b/pkg/cloud/services/networking/service.go
@@ -89,15 +89,25 @@ func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceT
 		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Invalid resourceType argument in function call")
 		panic(fmt.Errorf("invalid argument: resourceType, %s, does not match allowed arguments: %s or %s", resourceType, trunkResource, portResource))
 	}
+	// remove duplicate values from tags
+	tagsMap := make(map[string]string)
+	for _, t := range tags {
+		tagsMap[t] = t
+	}
+
+	uniqueTags := []string{}
+	for k := range tagsMap {
+		uniqueTags = append(uniqueTags, k)
+	}
 
 	_, err := s.client.ReplaceAllAttributesTags(resourceType, resourceID, attributestags.ReplaceAllOpts{
-		Tags: tags,
+		Tags: uniqueTags,
 	})
 	if err != nil {
 		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Failed to replace all attributestags, %s: %v", resourceID, err)
 		return err
 	}
 
-	record.Eventf(eventObject, "SuccessfulReplaceAllAttributeTags", "Replaced all attributestags for %s with tags %s", resourceID, tags)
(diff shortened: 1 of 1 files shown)


Xenwar:
Note:  [WIP] is added to prevent merge while the manifests contain test data

iamemilio:
seems reasonable
/lgtm

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -32,8 +32,11 @@
     # Placement service needed for Nova
     ENABLED_SERVICES+=,placement-api,placement-client
     # Glance
-    ENABLED_SERVICES+=,g-api
+    ENABLED_SERVICES+=,g-api,g-reg
 
+    # Octavia-Neutron
+    Q_AGENT=linuxbridge
+    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3,neutron-trunk
```
jichenjc, line 39: what's this for? seems not related?

Xenwar:
Test passed. 
Changes done for testing purpose are removed.
This PR can be reviewed with just the required changes to fix the duplicate tags problem.

pkg/cloud/services/networking/service.go
```diff
@@ -89,15 +89,25 @@ func (s *Service) replaceAllAttributesTags(eventObject runtime.Object, resourceT
 		record.Warnf(eventObject, "FailedReplaceAllAttributesTags", "Invalid resourceType argument in function call")
 		panic(fmt.Errorf("invalid argument: resourceType, %s, does not match allowed arguments: %s or %s", resourceType, trunkResource, portResource))
 	}
+	// remove duplicate values from tags
```
jichenjc, line 92: as follow up ,we can create some ut (to be honest we don't have too much so not a mandatory thing for this PR )
to test this works ..

Xenwar: This one is more of a user's mistake that gives duplicate tags, not a use case. 
So, we may not need a test case for it. 
However, the previous test was run to validate the code: see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1112#issuecomment-1012206109

jichenjc: you are right :)
thanks for the great info :~

Xenwar:
For future reference, here is an entry from the capo log for the above input. 
`I0114 06:43:46.892107       1 recorder.go:103] events "msg"="Normal"  "message"="Replaced all attributestags for 9da95b85-fffb-41f7-9c8c-633f3060dc54 with tags [tag1 tag2 tag3]" "object"={"kind":"OpenStackMachine","namespace":"e2e-hphugc","name":"cluster-e2e-hphugc-control-plane-mvwx4","uid":"96a075e6-747b-4ef9-bf74-07489fb28b03","apiVersion":"infrastructure.cluster.x-k8s.io/v1beta1","resourceVersion":"3794"} "reason"="SuccessfulReplaceAllAttributeTags"
`

PR #1115, opened 2022-01-14
🐛 Prevent creation of floating IPs when reconcile load balancer in non ACTIVE state

### LINKED ISSUE #1114: Load balancer in state different from ACTIVE could cause the creation of many floating ip's
/kind bug

I would like to prevent the creation of unused floating ip's during load balancer reconcilation to avoid waste of resources. This happens, when a load balancer exists but it's state differs from 'ACTIVE'.

Actual behavior

- during reconcilation the controller waits for the load balancer to get ACTIVE
in case of timeout the load balancers the state of the load balancer is ignored and the error is hidden by a general timeout error ([loadbalancer.go#waitForLoadBalancerActive](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d407435ee95672f3480a7912ee22f66895b64daa/pkg/cloud/services/loadbalancer/loadbalancer.go#L528))

- if an existing load balancer is reconciled and this loadbalancer is in a state other than' 'ACTIVE', following network recources will be created, even if this loadbalancer is in state ERROR
    - [getOrCreateLoadbalancer](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d407435ee95672f3480a7912ee22f66895b64daa/pkg/cloud/services/loadbalancer/loadbalancer.go#L55) with existing load balancer in state 'ERROR'
    - [checkIfLbExists](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d407435ee95672f3480a7912ee22f66895b64daa/pkg/cloud/services/loadbalancer/loadbalancer.go#L440) finds the load balancer with that state 'ERROR'
    - which has no IPAddress, so a new one will be requested until the quota exeeds
([from here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d407435ee95672f3480a7912ee22f66895b64daa/pkg/cloud/services/loadbalancer/loadbalancer.go#L61))

Expected behavior

- an error during load balancer creation should prevent further tries to reconcile, 
any other state than ACTIVE should lead to an error

<sub>Michael Grote <michael.grote@daimler.com>, Daimler TSS GmbH, [Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)</sub>

mgrote:
ReconcileLoadBalancer check for appropiate load balancer state implemented to prevent creation of floating ip's for load balancer with states other than ACTIVE.
Documentation link to openstack load balancer states updated.

Reconcilation of load balancers in other states than ACTIVE could cause the creation of unused floating iIPs until the quotas exceeds.
This PR provides a check of the appropriate load balancer state before creating other resources.

Fixes #1114

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -40,6 +40,8 @@ const (
 	kubeapiLBSuffix string = "kubeapi"
 )
 
+const loadBalancerProvisioningStatusActive = "ACTIVE"
+
 func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string, apiServerPort int) error {
 	loadBalancerName := getLoadBalancerName(clusterName)
 	s.logger.Info("Reconciling load balancer", "name", loadBalancerName)
@@ -55,6 +57,9 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	lb, err := s.getOrCreateLoadBalancer(openStackCluster, loadBalancerName, openStackCluster.Status.Network.Subnet.ID, clusterName, fixedIPAddress)
 	if err != nil {
 		return err
+	}
+	if lb.ProvisioningStatus != loadBalancerProvisioningStatusActive {
+		return fmt.Errorf("load balancer %q is not in expected state %s, current state is %s", lb.ID, loadBalancerProvisioningStatusActive, lb.ProvisioningStatus)
 	}
 
 	var lbFloatingIP string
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -56,6 +56,9 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	if err != nil {
 		return err
 	}
+	if infrav1.InstanceState(lb.ProvisioningStatus) != infrav1.InstanceStateActive {
```
jichenjc, line 59: I thought we need check LB status instead of instance itself? 
not sure which one we want to check and use it ?

mgrote: Thank you for your comment. When I tried to reuse abstracted states I overshot the target.
Thanks for bringing clarity.

mgrote:
I signed it

mgrote:
I signed it

mgrote:
I signed it

jichenjc:
/approve

not sure we need keep creating the LB when first try failed.. seems reasonable to check ACTIVE before any other actions

seanschneeweiss:
/lgtm congratulations on your first PR for CAPO :smiley: 
I only `lgtm` as Michael is from the same team.

Issue #1116, opened 2022-01-18
Handle transient OpenStack API errors for OpenStackMachines


seanschneeweiss:
/kind feature

OpenStackMachine reconciliation stops (for ever) on transient and non transient OpenStack API errors
* reconciler does not differentiate between
  * transient API errors, e.g., InternalServerError, connectivity problems
  * non transient errors, e.g., OpenStack server in ERROR state
* this leads to a stopped reconciliation, even though a subsequent reconciliation would be most likely to succeed

**Describe the solution you'd like**
I would like CAPO to handle transient OpenStack API errors for OpenStackMachine deployments.

* Do not skip reconciliation for transient errors from the OpenStack API
* Do not set [FailureMessage/FailureReason](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/api/v1beta1/openstackmachine_types.go#L113-L132) on openstackmachine.status for temporary errors.
* [rethink when to return `ctrl.Result{}`, error, etc.](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/controllers/openstackmachine_controller.go#L305-L315)

**Anything else you would like to add:**

* [lines where the reconciliation is skipped](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/controllers/openstackmachine_controller.go#L260-L263)
* [pay attention to FailureMessage hints about transient errors](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/api/v1beta1/openstackmachine_types.go#L113-L132)
* have a look at https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/832 as conditions could help to not always use the "UpdateMachineError"
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/controllers/openstackmachine_controller.go#L403

<sub>Sean Schneeweiss <sean.schneeweiss@daimler.com>, Daimler TSS GmbH, [Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)</sub>

jichenjc:
[...]

PR #1118, opened 2022-01-21
✨Volume creation and deletion by reconciliation

mdbooth:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
weird the test not start ..

/retest

pkg/cloud/services/compute/instance.go
```diff
@@ -44,6 +44,39 @@ const (
 	timeoutInstanceDelete       = 5 * time.Minute
 )
 
+type OpenStackError struct {
```
jichenjc, line 47: if this is general, I think we should put into util or somewhere else?

pkg/cloud/services/compute/instance.go
```diff
@@ -336,6 +369,141 @@ func getPortName(instanceName string, opts *infrav1.PortOpts, netIndex int) stri
...
+			}
+			return false, fatalError(err)
+		}
+
+		switch createdVolume.Status {
+		case "available":
+			record.Eventf(eventObject, "VolumeAvailable", "Root volume; id=%s", volumeID)
+			rootVolumeStatus.Ready = true
+			return true, nil
+		case "error":
```
jichenjc, line 408: this will leave an error volume somewhere
do we delete it somewhere before we create a new one?

pkg/cloud/services/compute/instance.go
```diff
@@ -336,6 +369,141 @@ func getPortName(instanceName string, opts *infrav1.PortOpts, netIndex int) stri
 	return fmt.Sprintf("%s-%d", instanceName, netIndex)
 }
 
+func (s *Service) ReconcileRootVolume(eventObject runtime.Object, machine clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine) (bool, *OpenStackError) {
```
jichenjc, line 372: where's the function called? didn't find it..

mdbooth:
The tests didn't run because it's a WIP. I'm surprised it even compiles!

This is just a talking point at this stage in the context of the discussion about putting more explicit retries in code. I would prefer to move in the opposite direction and push all retries up to the controller level. My hope is that by reconciling separate resource creation separately we release ourselves from the 'compound complexity' of reconciling them all at the same time. A classic indicator of that would be how much simpler the unit tests would become!

I hope to finish fleshing this out to the point that it runs some time soon, but unfortunately I doubt it's going to be this week or next.

pkg/cloud/services/compute/instance.go
mdbooth: Yes. I only threw it in here for convenience. @seanschneeweiss would something like this address the fatal vs retriable errors issue (if used consistently).

pkg/cloud/services/compute/instance.go
mdbooth: We won't create a new one. Notice that it returns fatalError(). This means that the controller should put the Machine in a failed state, and the only thing left to do with it is delete the Machine. This is the same as the instance going into an error state.

Because we recorded the volume creation in the status object, when the controller deletes the machine we have a record that we need to cleanup this volume. A 404 here would similarly raise a fatal error because we know to created the volume (we have its uuid), but it no longer exists.

Note that going into the error state is different to, e.g. a 500 returned by cinder because rabbit was unhealthy. In that case we'll return a transientError(), so the controller knows to log it and retry after a back-off.

pkg/cloud/services/compute/instance.go
seanschneeweiss: This would most certainly be a solution to the problem, but as indicated in my [comment](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1116#issuecomment-1022047685) I hoped for an easier solution where we just reduce the error cases that call `handleUpdateMachineError`.

seanschneeweiss:
CAPZ seems to have a similar approach like you suggested:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e75d79fb7ddd20e0dd6d7bbe28cfb33bfc834253/azure/errors.go#L74-L128

jichenjc:
>>>My hope is that by reconciling separate resource creation separately we release ourselves from the 'compound complexity' of reconciling them all at the same time

agree with the approach , this should make the code clearer and logic clearer :) thanks~

mdbooth:
I'll create a new PR when I'm working on this again.

PR #1119, opened 2022-01-24
✨[WIP] Mgrote/service interface load balancer

mgrote:

For better testability it is useful to abstract a service interface for loadbalancers. That interface could be easily mocked e.g. with gomock for propper testing.
The same was already done for pkg/cloud/services/networking in #935.

Fixes # tbd

**TODOs**:

- [X] squashed commits
- [X] adds unit tests

  

/hold


pkg/cloud/services/loadbalancer/client.go
```diff
@@ -0,0 +1,231 @@
...
+	return listeners.ExtractListeners(allPages)
+}
+
+func (l lbClient) GetListener(id string) (*listeners.Listener, error) {
+	mc := metrics.NewMetricPrometheusContext("loadbalancer_listener", "get")
+	lb, err := listeners.Get(l.serviceClient, id).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+	return lb, nil
```
seanschneeweiss, line 119:
```suggestion
	listener, err := listeners.Get(l.serviceClient, id).Extract()
	if mc.ObserveRequest(err) != nil {
		return nil, err
	}
	return listener, nil
```

pkg/cloud/services/loadbalancer/client.go
```diff
@@ -0,0 +1,231 @@
...
+	listener, err := listeners.Create(l.serviceClient, opts).Extract()
+	if mc.ObserveRequest(err) != nil {
+		return nil, err
+	}
+	return listener, nil
+}
+
+func (l lbClient) ListListeners(opts listeners.ListOptsBuilder) ([]listeners.Listener, error) {
+	mc := metrics.NewMetricPrometheusContext("loadbalancer_listener", "list")
+	allPages, err := listeners.List(l.serviceClient, opts).AllPages()
```
seanschneeweiss, line 106: There are some occurences of `allPages` as variable name. I'd suggest to rename those to align more with the items that are in the list, e.g.
```suggestion
	listenerList, err := listeners.List(l.serviceClient, opts).AllPages()
```

pkg/cloud/services/loadbalancer/client.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/providers"
+
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
+	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
```
seanschneeweiss, line 30: Could you add the import alias to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/.golangci.yml#L64-L66

pkg/cloud/services/loadbalancer/service.go
```diff
@@ -49,8 +50,21 @@ func NewService(client *gophercloud.ProviderClient, clientOpts *clientconfig.Cli
...
-		logger:             logger,
+		loadbalancerClient: lbClient{
+			serviceClient: loadbalancerClient,
+		},
+		networkingService: networkingService,
+		logger:            logger,
 	}, nil
 }
+
+// NewLoadBalancerTestService returns a Service with no initialisation. It should only be used by tests.
```
seanschneeweiss, line 61:
```suggestion
// NewLoadBalancerTestService returns a Service with no initialization. It should only be used by tests.
```

seanschneeweiss (commented):
This is great. Helps a lot with testing and makes the load balancer code more readable by moving out the metrics context calls.

pkg/cloud/services/loadbalancer/client.go
seanschneeweiss: Sorry, the return value is not a list but rather pages. And this seems to be used like this throughout the repository. So it's perfectly fine like this.

seanschneeweiss (commented):
/lgtm

Thank you for updating the `boilerplate.generatego.txt`.

The previous test failed as the [image used by Prow uses Go version 1.17.6](https://github.com/kubernetes/test-infra/blob/d32b96b01d85940ce2aaf3adc9cc110988673c69/images/kubekins-e2e/variants.yaml#L25).

After @mgrote updated to Go version 1.17, the generated files were accepted by the test.

chrischdi (commented):
/approve

Great job, also for handling the boilerplate header :-)

mdbooth:
This looks great!
/approve

PR #1125, opened 2022-01-26
✨  Ensure trunk deletion

### LINKED ISSUE #1124: Wait for port detach to succeed when deleting trunk
Based what neutron driver is used, trunk deletion might require a detachment of port. While our local devstack setup does not require that, the CAPO e2e test infra requires it.  Basically, this can happen with drivers that demand that a trunk must be created or deleted on an unbound port.

Currently, the code tries to delete trunk as follows.

1. detach port
2. delete trunk
3. delete port

While the sequence above correct and should work even in the CAPO e2d infra, there is a possibility that `(2) delete trunk `is execute first. This is caused due to OpenStack API call to detach the port responding immediately with success status, despite the operation is not performed yet. This has resulted flaky failures. 

Error log:

`I0126 13:11:53.204487       1 recorder.go:103] events "msg"="Warning"  "message"="Failed to delete trunk cluster-e2e-6ufzl7-control-plane-9hpqc-0 with id 52190a49-381c-4a99-ae3e-02c34b4910c7: Expected HTTP response code [] when accessing [DELETE http://10.0.3.15:9696/v2.0/trunks/52190a49-381c-4a99-ae3e-02c34b4910c7], but got 409 instead\n{\"NeutronError\": {\"type\": \"TrunkInUse\", \"message\": \"Trunk 52190a49-381c-4a99-ae3e-02c34b4910c7 is currently in use.\", \"detail\": \"\"}}" "object"={"kind":"OpenStackMachine","namespace":"e2e-6ufzl7","name":"cluster-e2e-6ufzl7-control-plane-9hpqc","uid":"3c0c49c7-a42a-4986-be8e-d76ca0401a36","apiVersion":"infrastructure.cluster.x-k8s.io/v1beta1","resourceVersion":"7205"} "reason"="FailedDeleteTrunk"
`

**To Do:**
Wait for the port detachment to succeed and perform trunk deletion.

/kind bug

Xenwar:
This PR has two purposes:
- Enabling trunk support on the e2e infra
- fixing issue related to trunk deletion

Port deletion involves three steps.

1. Detaching port
2. Deleting trunk
3. Deleting port

The `Openstack api call` to `Detaching port` (1), returns immediately without waiting for the operation to complete. Due to this, it is possible that the `Deleting trunk `(2) be execute first, resulting in error.

This PR ensures `Deleting trunk` (2) is executed after a successful `Detaching port` (1). 

fixed #1124

### DIFF
--- a/hack/ci/cloud-init/controller.yaml.tpl
+++ b/hack/ci/cloud-init/controller.yaml.tpl
@@ -35,7 +35,8 @@
     ENABLED_SERVICES+=,g-api
 
     # Neutron
-    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3
+    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3,neutron-trunk
+
     ENABLED_SERVICES+=,neutron-metadata-agent,neutron-qos
     # Octavia
     ENABLED_SERVICES+=,octavia,o-api,o-cw,o-hm,o-hk,o-da
--- a/pkg/cloud/services/networking/trunk.go
+++ b/pkg/cloud/services/networking/trunk.go
@@ -95,6 +95,9 @@ func (s *Service) DeleteTrunk(eventObject runtime.Object, portID string) error {
 				record.Eventf(eventObject, "SuccessfulDeleteTrunk", "Trunk %s with id %s did not exist", trunkInfo[0].Name, trunkInfo[0].ID)
 				return true, nil
 			}
+			if capoerrors.IsConflict(err) {
+				return false, nil
+			}
 			if capoerrors.IsRetryable(err) {
 				return false, nil
 			}


hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -34,8 +34,10 @@
     # Glance
     ENABLED_SERVICES+=,g-api
 
-    # Neutron
-    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3
+    # Octavia-Neutron
+    Q_AGENT=linuxbridge
+    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3,neutron-trunk
```
jichenjc, line 39: curious we didn't have neutron-trunk before, so this means no trunk test before?

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -34,8 +34,10 @@
     # Glance
     ENABLED_SERVICES+=,g-api
 
-    # Neutron
-    ENABLED_SERVICES+=,neutron-api,neutron-agent,neutron-dhcp,neutron-l3
+    # Octavia-Neutron
+    Q_AGENT=linuxbridge
```
jichenjc, line 38: what's the reason to make it bridge instead of ovs?

hack/ci/cloud-init/controller.yaml.tpl
Xenwar: There was e2e trunk test before.

hack/ci/cloud-init/controller.yaml.tpl
Xenwar: Thanks, removed it.

pkg/cloud/services/compute/instance.go
```diff
@@ -639,6 +650,27 @@ func (s *Service) DeleteInstance(eventObject runtime.Object, openStackMachineSpe
 	return s.deleteInstance(eventObject, instanceStatus.InstanceIdentifier())
 }
 
+func (s *Service) ensureTrunkDeletion(eventObject runtime.Object, portID string) error {
```
seanschneeweiss, line 653: The `s.networkingService.DeleteTrunk` already uses `util.PollImmediate` to retry deletion.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/pkg/cloud/services/networking/trunk.go#L92-L104

What is the difference between the retry here and in the snippet above? Maybe it would be sufficient to add `capoerrors.IsConflict` to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8ae251502dbbcdd5685521272c7b4a5659152786/pkg/cloud/services/networking/trunk.go#L98 if needed.

Xenwar:
Test has passed repeatedly with the following input
```
metadata:
  name: ${CLUSTER_NAME}-control-plane
spec:
  template:
    spec:
      # Added for Testing purposes
      trunk: true
      ports:
      - description: "Port 1, trunk=true"
        trunk: true
      flavor: ${OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR}
```
Last successful build code: 1486668436183453696

Removing changes to the yaml file as it was added for testing purpose.

pkg/cloud/services/compute/instance.go
Xenwar: Thanks, will check that.

Xenwar: Changes made and waiting for the e2e test result

Xenwar:
@seanschneeweiss 
Thanks for the feedback. The test passed with changes you suggested.

Xenwar:
It does work with trunk enabled, i.e. when https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1125#issuecomment-1023290261 is in the pr. The moment it is removed the test start to fail. 
The previous test has failed due to that. 

Perhaps, there is another kind error (not did not exist, not conflict). Will check it on Monday.

Xenwar:
@seanschneeweiss Thanks for the suggestion. 
The fix works and tests have passed now.

Xenwar:
Tested the code with and without trunk enabled on the yaml files and tests have passed. please review.
@jichenjc
@hidekazuna
@macaptain

pkg/cloud/services/networking/trunk.go
```diff
@@ -95,6 +95,9 @@ func (s *Service) DeleteTrunk(eventObject runtime.Object, portID string) error {
 				record.Eventf(eventObject, "SuccessfulDeleteTrunk", "Trunk %s with id %s did not exist", trunkInfo[0].Name, trunkInfo[0].ID)
 				return true, nil
 			}
+			if capoerrors.IsConflict(err) {
+				return false, err
```
macaptain, line 99: `util.PollImmediate` ultimately calls this function in k8s apimachinery with `immediate` as `true`:

https://github.com/kubernetes/apimachinery/blob/86c0c0f8c8e2ad6601321a412cddb1f99ff6fae5/pkg/util/wait/wait.go#L579-L587

Since the `err` returned by `runConditionWithCrashProtectionWithContext` - our function right here - is non-nil in the case of a conflict, we will still treat the conflict as an error. The first argument as `false` is never considered. So there is no retry.

So this change looks like a no-op? Should the `err` be `nil`?

Xenwar: If `err` is set `nil`, then the user has to enable trunk in order for it to work. 
spec:
  template:
    spec:
      # Added for Testing purposes
      trunk: true
      ports:
      - description: "Port 1, trunk=true"
        trunk: true

macaptain: If trunk is enabled in the environment, but the port does not have a trunk configured, we still call `DeleteTrunk`, but I would expect that `len(trunkInfo) == 0` (where `trunkInfo` is the result of doing a list trunks against the OpenStack API) in the lines above:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c0819a496c023e905bb865e2c9907d642bb69070/pkg/cloud/services/networking/trunk.go#L88-L90

We should be returning `nil` and therefore never reach the code block that you have changed. Is this not happening somehow?

Xenwar: You were right. The failure seems to be flaky case. 
One test passed, and, nothing is logged, implying that the code did not reach that part. 
removing the log now.

seanschneeweiss:
Glad that it works now. Thank you for fixing it.
/lgtm

macaptain (approved):

hack/ci/cloud-init/controller.yaml.tpl
mdbooth: Yes, it's a long-standing oversight.

mdbooth (approved):
Is there any circumstance when a conflict isn't retryable? 🤔 This looks good.

Issue #1121, opened 2022-01-26
Change warning to Error for failing calls


Xenwar:
Currently, there are entries that log warning for something that fails. 
Example:

```
record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s
record.Warnf(eventObject, "FailedCreateServer", "Failed to create server %s
record.Warnf(eventObject, "FailedCreateTrunk", "Failed to create trunk %s
record.Warnf(eventObject, "FailedDeleteAttachInterface", "Failed to delete attach interface
record.Warnf(eventObject, "FailedDeletePort", "Failed to delete port %s with id %s
record.Warnf(openStackCluster, "FailedDeleteFloatingIP", "Failed to delete floating IP %s
record.Warnf(openStackCluster, "FailedDeleteLoadBalancer", "Failed to delete load balancer %s with id %s
record.Warnf(openStackCluster, "FailedDeleteNetwork", "Failed to delete network %s with id %s
record.Warnf(openStackCluster, "FailedDeleteRouter", "Failed to delete router %s with id %s
record.Warnf(openStackCluster, "FailedDeleteSecurityGroup", "Failed to delete security group %s with id %s
record.Warnf(openStackCluster, "FailedUpdateRouter", "Failed to update router %s with id %s
```

To Do:
Collect such entries and make appropriate changes. 
As shown here https://github.com/kubernetes/api/blob/bfe324c0d64befbab416bd4063c0a38a6cda9334/core/v1/types.go#L5650-L5655, there is Event type for Error. So, it needs to be introduced in `pkg/record/recorder.go `

/kind bug

jichenjc:
@Xenwar thanks for the issue

I am curious why the recorder.go itself doesn't have Error and let us add it ..
so if we update to that repo first then we consume it might be a reasonable way?

macaptain:
In the link above for Event types (https://github.com/kubernetes/api/blob/bfe324c0d64befbab416bd4063c0a38a6cda9334/core/v1/types.go#L5650-L5655), I only see `EventTypeNormal` and `EventTypeWarning`? It seems like these are the event severity types we're limited to. The current record.go is consistent with other providers / CAPI itself for these types (c. f. https://github.com/kubernetes-sigs/cluster-api/blob/58905fb5140a7844c3e24aa70d255fb8f9d0c491/util/record/recorder.go)

I will say that there are 'events' in the CAPO codebase which are better expressed as log messages. For example:

[...]

PR #1126, opened 2022-01-26
✨  Improve failing events with Error logs

Xenwar:
Due to the absence of [EventTypeError](https://github.com/kubernetes/api/blob/bfe324c0d64befbab416bd4063c0a38a6cda9334/core/v1/types.go#L5650-L5655), some part of the CAPO code is using `EventTypeWarning`  for failing events. 

This PR introduces `EventTypeError` constant and changes log entries of failing events to the correct type.

Fixes: #1121

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -320,7 +320,7 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 		return createdInstance.State() == infrav1.InstanceStateActive, nil
 	})
 	if err != nil {
-		record.Warnf(eventObject, "FailedCreateServer", "Failed to create server %s: %v", createdInstance.Name(), err)
+		record.Errorf(eventObject, "FailedCreateServer", "Failed to create server %s: %v", createdInstance.Name(), err)
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


mdbooth:
I don't think this would currently be correct, as Event.Type is documented to be either `Normal` or `Warning`: https://github.com/kubernetes/api/blob/b2d630a65cb2ef33faef611d793d551014defc0d/core/v1/types.go#L5698-L5700

Unfortunately I'm not familiar enough with the details of Event to know what issues setting something else might cause, if any. However, the problem described in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1121 does make sense, at least to me. I wonder if there's some other way users are intended to flag Events which are more than a warning?

Xenwar:
Good point. 
What if we start using `s.logger.Info, s.logger.Error .....` ?
of course, unless there is some compelling reason to the `Event.Types`.

mdbooth:
> Good point. What if we start using `s.logger.Info, s.logger.Error .....` ? of course, unless there is some compelling reason to the `Event.Types`.

See my comment in the issue. I'm still hoping somebody authoritative is going to step in with the expected behaviour here, but I wondered if we should set a Condition instead with the informative reason on the Condition. If it isn't something we would set a Condition for (and stop reconciling), it's probably not an error?

Xenwar:
Adding Error to Event.Type does not seem to be reasonable. And,  it is not clean what information will be lost when changing the logs to `s.logger.Info|Error|Warnning` format. 

The reason for creating this issue was ERROR was logged as WARNING. Now, I see that if an error occurs, the stack trace would provide more error related information. 
 
Therefore, closing this issue.

PR #1128, opened 2022-02-01
Add test for trunk feature

Xenwar:
Add test for trunk feature

Requires: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1125

### DIFF
--- a/test/e2e/suites/e2e/e2e_test.go
+++ b/test/e2e/suites/e2e/e2e_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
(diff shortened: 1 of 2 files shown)


test/e2e/suites/e2e/e2e_test.go
```diff
@@ -208,8 +209,13 @@ var _ = Describe("e2e tests", func() {
 
 			shared.Byf("Creating MachineDeployment with custom port options")
 			md3Name := clusterName + "-md-3"
+
+			tTrue := true
 			customPortOptions := &[]infrav1.PortOpts{
-				{Description: "primary"},
+				{
+					Description: "primary",
+					Trunk:       &tTrue,
```
mdbooth, line 217: nit: you could use `pointer.Bool(true)` here.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -208,8 +209,13 @@ var _ = Describe("e2e tests", func() {
...
+				},
```
mdbooth, line 218: Suggestion: we could have added a second port so the first would not be trunked and the second trunked.

mdbooth (approved):
Looks good. Some suggestions if you feel so inclined.

test/e2e/suites/e2e/e2e_test.go
Xenwar: Thanks, will do that.

test/e2e/suites/e2e/e2e_test.go
Xenwar: Thanks, will do that.

test/e2e/suites/e2e/e2e_test.go
Xenwar: Done, 
Tests have passed with the newly added trunked port as well.

mdbooth (approved):
Nice! Lets get an lgtm from somebody else.

PR #1137, opened 2022-02-09
WIP: bump CAPI to v1.1.0

jichenjc:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
give a try on v110 per discussion in CAPI #6080

jichenjc:
@mdbooth can you take a quick look? Thanks

go.mod
```diff
@@ -1,31 +1,120 @@
...
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	k8s.io/apiextensions-apiserver v0.23.0 // indirect
+	k8s.io/apiserver v0.23.0 // indirect
+	k8s.io/cluster-bootstrap v0.23.0 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
+	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
+	sigs.k8s.io/kind v0.11.1 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.0 // indirect
+)
```
mdbooth, line 118: Any idea where this appeared from?

pkg/cloud/services/networking/floatingip_test.go
```diff
@@ -77,7 +76,6 @@ func Test_GetOrCreateFloatingIP(t *testing.T) {
 			tt.expect(mockClient.EXPECT())
 			s := Service{
 				client: mockClient,
-				logger: logr.DiscardLogger{},
```
mdbooth, line 80: Why was this removed?

pkg/cloud/services/networking/port_test.go
```diff
@@ -439,7 +438,6 @@ func Test_GetOrCreatePort(t *testing.T) {
 			tt.expect(mockClient.EXPECT())
 			s := Service{
 				client: mockClient,
-				logger: logr.DiscardLogger{},
```
mdbooth, line 442: And again. Is there some behaviour change in logr which means calling methods on a nil logr doesn't error?

mdbooth (commented):
I'm a bit concerned by all the noise in go.mod, and by the appearance of a second require block. However, I'm guessing you didn't write that by hand! Any idea what's going on there?

go.mod
jichenjc: >>>I'm guessing you didn't write that by hand! Any idea what's going on there?

yes, I didn't do it manually...

pkg/cloud/services/networking/floatingip_test.go
jichenjc: check the commit msg  :)

the logr 0.4=>1.2 update makes this happen :)

go.mod
jichenjc: and I think it do no harm ? I don't know why update to 1.1.0 introduce so much changes

the //indirect is not something familiar to me , will do some home work..

jichenjc: https://github.com/kubernetes-sigs/cluster-api/blob/main/go.mod#L51
capi also has this kind of stuffs

not find the real solution yet, but seems other folks has same questions as well
https://stackoverflow.com/questions/61115111/how-to-avoid-indirect-dependencies-in-go-mod-file

Also curious why only this update (maybe go16->17?) trigger this...

mdbooth (approved):
I don't understand specifically why these new `requires` have turned up with this change, but having looked they're all clearly required. Suspect the answer is probably go 1.17, although that's purely a guess.

Also having the indirects in a separate requires is somewhat tidier. Despite my initial surprise I think I prefer it.

Incidentally, most of these deps seem to be pulled in via `sigs.k8s.io/cluster-api/test`, which is a separate module in CAPI. I wonder if we should also pull our e2e tests into a separate module to minimise the blast radius. @sbueringer do you happen to know if other projects are doing this?

/lgtm

PR #1144, opened 2022-02-21
🏃 devstack: Allow empty private SSH key

mdbooth:
This allows use of a private key stored on a hardware token.

/hold

### DIFF
--- a/hack/ci/aws-project.sh
+++ b/hack/ci/aws-project.sh
@@ -112,8 +112,8 @@ function get_mtu {
     echo 1300
 }
 
-function get_ssh_public_key {
-  cat "${SSH_PUBLIC_KEY_FILE}"
+function get_ssh_public_key_file {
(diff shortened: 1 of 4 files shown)


mdbooth:
Looks like I'm going to have to make this configurable.

jichenjc:
/lgtm

though I am not sure the use case for it, but no harm seems

mdbooth:
> /lgtm
> 
> though I am not sure the use case for it, but no harm seems

I need it to use the key on my Yubikey :) There's no private key file. To specify a particular key you specify the public key instead. Actual operations involving the private key happen directly on the hardware token, which SSH accesses via ssh-agent.

Yeah, it's pretty niche 😂

/hold cancel

jichenjc:
ok, thanks for the info , this make sense 
just involved in `Yubikey` discussion somewhere else so a good timing for me to do some homework :)

PR #1150, opened 2022-02-23
🐛 api/v1beta1/openstackmachine_types.go: change errorReason/errorMessage to failureReason/failureMessage

### LINKED ISSUE #1149: 🐛 align failureReason and failureMessage also for OpenStackMachines
/kind bug

Currently, we use `failureReason` and `failureMessage` status for all CAPO resources.
Only for the `OpenStackMachines` we use `errorReason` and `errorMessage` (even the variable names are correct).

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/642d851d3068f90864fedc1a28c283f9218e3850/api/v1alpha4/openstackmachine_types.go#L109
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/642d851d3068f90864fedc1a28c283f9218e3850/api/v1alpha4/openstackmachine_types.go#L128

This should be checked with all other resources to be consistent. 

Proposed for inclusion in 0.6: #1145

<sub>Tobias Giese <tobias.giese@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/mercedes-benz/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

tobiasgiese:

Fixes #1149

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

<sub>Tobias Giese <tobias.giese@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/mercedes-benz/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/api/v1beta1/openstackmachine_types.go
+++ b/api/v1beta1/openstackmachine_types.go
@@ -110,7 +110,7 @@ type OpenStackMachineStatus struct {
 	// +optional
 	InstanceState *InstanceState `json:"instanceState,omitempty"`
 
-	FailureReason *errors.MachineStatusError `json:"errorReason,omitempty"`
+	FailureReason *errors.MachineStatusError `json:"failureReason,omitempty"`
 
 	// FailureMessage will be set in the event that there is a terminal problem
 	// reconciling the Machine and will contain a more verbose string suitable
@@ -129,7 +129,7 @@ type OpenStackMachineStatus struct {
 	// can be added as events to the Machine object and/or logged in the
 	// controller's output.
 	// +optional
-	FailureMessage *string `json:"errorMessage,omitempty"`
+	FailureMessage *string `json:"failureMessage,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackmachines.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackmachines.yaml
@@ -1173,7 +1173,7 @@ spec:
                   - type
                   type: object
                 type: array
-              errorMessage:
+              failureMessage:
                 description: "FailureMessage will be set in the event that there is
                   a terminal problem reconciling the Machine and will contain a more
                   verbose string suitable for logging and human consumption. \n This
@@ -1188,7 +1188,7 @@ spec:
                   occur during the reconciliation of Machines can be added as events
                   to the Machine object and/or logged in the controller's output."
                 type: string
-              errorReason:
+              failureReason:
                 description: MachineStatusError defines errors states for Machine
                   objects.
                 type: string


jichenjc:
we are in beta1 already so I think we should not just do the update directly? 
instead, we should create another field something like

```
	DeprecatedFailureReason *errors.MachineStatusError `json:"errorReason,omitempty"`
	FailureReason *errors.MachineStatusError `json:"failureReason,omitempty"`
```

so we still honor both `errorReason` and `failureReason` ? 
this is my understanding of beta API version handling and happy to get any suggestions...

/hold

tobiasgiese:
> we are in beta1 already so I think we should not just do the update directly?

I think you're totally right. We already released a v1beta1 release with v0.5.0 and thus we have to deprecate the field instead of just removing/changing it.
Should we add a note that we want to remove the deprecated field in a future release? Maybe v0.8.0?

Meanwhile I'll test the behavior of the controller if the field has changed during an update.

tobiasgiese:
> we are in beta1 already so I think we should not just do the update directly?

I think you're totally right. We already released a v1beta1 release with v0.5.0 and thus we have to deprecate the field instead of just removing/changing it.
Should we add a note that we want to remove the deprecated field in a future release? Maybe v0.8.0?

Meanwhile I'll test the behavior of the controller if the field has changed during an update.

apricote:
> I think you're totally right. We already released a v1beta1 release with v0.5.0 and thus we have to deprecate the field instead of just removing/changing it.

I think the v1beta1 CRD was only included in the (broken) v0.5.1 release that was cut from `main` instead of the `release-0.5` branch, `v0.5.0` and `v0.5.2` did not include the new version.

See: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/release-0.5/api

tobiasgiese:
> I think the v1beta1 CRD was only included in the (broken) v0.5.1 release that was cut from `main` instead of the `release-0.5` branch, `v0.5.0` and `v0.5.2` did not include the new version.

Sure, that broken release totally confused me. So we made no v1beta1 release public so far. Thus our v1beta1 changes are still unstable and on our `main` branch. As this is the case we can easily change the API and add a conversion to handle the change  -- right?

mdbooth:
> > I think you're totally right. We already released a v1beta1 release with v0.5.0 and thus we have to deprecate the field instead of just removing/changing it.
> 
> I think the v1beta1 CRD was only included in the (broken) v0.5.1 release that was cut from `main` instead of the `release-0.5` branch, `v0.5.0` and `v0.5.2` did not include the new version.
> 
> See: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/release-0.5/api

Yep, our latest release still uses v1alpha4. We're using cluster-api v1beta1. I think now is the time to make this consistent, *before* we go beta.

jichenjc:
yes, saw the slack channel discussion, and yes, we should be able to make such change without deprecation, thanks for the discussion~

mdbooth:
Nice. Doesn't even require an explicit conversion because we're just changing the marshalling of the new type so it happens automatically.

/lgtm

seanschneeweiss (commented):
/lgtm

PR #1153, opened 2022-02-24
✨WIP: Refactor CreateInstance and CreateBastion

mdbooth:

***I have moved the following to #1191***

The primary purpose of this this PR is to cleanup the interface of `compute.CreateInstance` and make the separation of concerns between the machine controller (for Machines), the cluster controller (for the Bastion), and compute (for actual server creation) better defined.

The Bastion host is represented in the cluster spec as an `Instance`. An OpenStackMachine is represented by context in `OpenStackMachine` as well as `OpenStackCluster`. So while they both create a server in the same way, they both source the server's parameters in slightly different ways using different source objects.

At some point in history we also used `Instance` as the intermediate representation for an OpenStackMachine. That is, we combined parameters from an `OpenStackMachine` and an `OpenStackCluster` into an `Instance`, then passed that to CreateInstance. `Instance` is not ideal for this purpose as it is both Spec and Status. It contains fields which cannot be used as input parameters to CreateInstance. Therefore we refactored this into the internal-only `InstanceSpec`, which contains only spec fields.

This refactor takes this further. Firstly we ensure that the code previously contained directly in CreateBastion and CreateInstance is strictly data transformation and therefore very cheap. Anything expensive moves into the new ReconcileInstance. Secondly we move this code to the controller where it is relevant, and add new unit tests covering the transformation. This also reduces the scope of the former CreateInstance unit tests, making them simpler.

***The remainder in this PR which will need to be reworked is:***

In refactoring bastion creation, we also take the opportunity to make *all* bastion errors non-fatal to the cluster. We also ensure we update the cluster status to Ready before creating the Bastion, which means that control plane creation can run in parallel with bastion creation.

I don't intend to squash these commits. They are intended to be independent logical steps. You may find it easier to review this PR by commit rather than as a single change.

/hold


pkg/cloud/services/compute/instance.go
```diff
@@ -217,11 +157,21 @@ func (s *Service) createInstance(eventObject runtime.Object, clusterName string,
 		}
 
 		if err := s.deletePorts(eventObject, portList); err != nil {
-			s.logger.V(4).Error(err, "failed to clean up ports after failure", "cluster", clusterName, "machine", instanceSpec.Name)
+			s.logger.V(4).Error(err, "failed to clean up ports after failure")
```
seanschneeweiss, line 160:
```suggestion
			s.logger.V(4).Error(err, "Failed to clean up ports after failure")
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -393,26 +393,34 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
-		if err != nil {
-			return nil, err
-		}
+		// Don't validate an explicit UUID if we were given one
+		if sg.UUID != "" {
+			if isDuplicate(sgIDs, sg.UUID) {
+				continue
+			}
+			sgIDs = append(sgIDs, sg.UUID)
+		} else {
```
seanschneeweiss, line 402:
```suggestion
			sgIDs = append(sgIDs, sg.UUID)
			continue
		}
```

seanschneeweiss (commented):
This looks very, very promising. Great. Probably it would be good for me to review this a second time, just to understand the changes even better ;)

mdbooth:
@seanschneeweiss Thanks for an excellent review! It's going to take me a while to go through it.

controllers/openstackcluster_controller.go
```diff
@@ -126,9 +127,13 @@ func reconcileDelete(ctx context.Context, log logr.Logger, client client.Client,
 		return reconcile.Result{}, err
 	}
 
-	if err = deleteBastion(log, osProviderClient, clientOpts, cluster, openStackCluster); err != nil {
+	deleted, err := deleteBastion(log, osProviderClient, clientOpts, cluster, openStackCluster)
+	if err != nil {
 		return reconcile.Result{}, err
 	}
+	if !deleted {
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
chrischdi, line 135: Won't it make sense here to use a exponential backoff (`return reconcile.Result{}, nil`) to not "DDoS" the API?

controllers/openstackcluster_controller.go
```diff
@@ -294,14 +292,30 @@ func reconcileNormal(ctx context.Context, log logr.Logger, client client.Client,
 		}
 	}
 
-	openStackCluster.Status.Ready = true
 	openStackCluster.Status.FailureMessage = nil
 	openStackCluster.Status.FailureReason = nil
+	if !openStackCluster.Status.Ready {
+		openStackCluster.Status.Ready = true
+
+		// If we're setting Ready, return early to update status and
```
chrischdi, line 300: That's a great move :-)

pkg/cloud/services/networking/securitygroups.go
chrischdi: should not be necessary right?!

controllers/openstackcluster_controller.go
```diff
@@ -185,54 +190,51 @@ func contains(arr []string, target string) bool {
...
+			return false, err
 		}
 		addresses := instanceNS.Addresses()
 
 		for _, address := range addresses {
 			if address.Type == corev1.NodeExternalIP {
 				if err = networkingService.DeleteFloatingIP(openStackCluster, address.Address); err != nil {
-					handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete floating IP: %v", err))
-					return errors.Errorf("failed to delete floating IP: %v", err)
+					return false, errors.Errorf("failed to delete floating IP: %v", err)
```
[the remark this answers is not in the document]
mdbooth: I agree. Lets do this consistently as a separate cleanup, though.

controllers/openstackmachine_controller.go
```diff
@@ -242,7 +242,14 @@ func (r *OpenStackMachineReconciler) reconcileDelete(ctx context.Context, logger
...
-	if err := computeService.DeleteInstance(openStackMachine, &openStackMachine.Spec, openStackMachine.Name, instanceStatus); err != nil {
+	instanceSpec, err := machineToInstanceSpec(openStackCluster, machine, openStackMachine, "")
+	if err != nil {
+		err = errors.Errorf("machine spec is invalid: %v", err)
+		handleUpdateMachineError(logger, openStackMachine, err)
+		return ctrl.Result{}, err
+	}
+
+	if err := computeService.DeleteInstance(openStackMachine, instanceSpec, instanceStatus); err != nil {
 		handleUpdateMachineError(logger, openStackMachine, errors.Errorf("error deleting OpenStack instance %s with ID %s: %v", instanceStatus.Name(), instanceStatus.ID(), err))
```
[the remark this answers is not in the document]
mdbooth: Yes. However, apart from the Bastion stuff which I kinda did by accident, I'm intending to clean this stuff up as part of the remainder of this series. I agree this needs to go, but I'd like to do it in a follow-up.

pkg/cloud/services/compute/instance.go
mdbooth: Of course this is a log message at Error, not an actual error, so ST1005 wouldn't apply. Might as well fix it as we're already touching the line.

That said, I am hoping to entirely get rid of this 😁

pkg/cloud/services/networking/securitygroups.go
mdbooth: Yeah, that's nicer.

controllers/openstackmachine_controller.go
```diff
@@ -302,16 +309,21 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
...
+		err = errors.Errorf("machine spec is invalid: %v", err)
+		handleUpdateMachineError(logger, openStackMachine, err)
 		return ctrl.Result{}, err
 	}
 
-	// Set an error message if we couldn't find the instance.
+	instanceStatus, err := computeService.ReconcileInstance(openStackMachine, openStackCluster, instanceSpec, cluster.Name)
+	if err != nil {
+		err = errors.Errorf("openstack instance cannot be created: %v", err)
+		handleUpdateMachineError(logger, openStackMachine, err)
```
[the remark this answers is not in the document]
mdbooth: Again, I'd like to address the errors raised by instance creation separately.

Incidentally, I just wrote the initial draft of a discussion document defining the failure state. The eventual intended audience is CAPI, but I need to work on making it coherent first. The TL;DR is:

Before Ready:
Retryable errors are non-terminal
Non-retryable errors are terminal

After Ready:
All errors are non-terminal, except 404

In the meantime I think we can unilaterally implement something like this ourselves as it's currently the wild west. The problem we're going to have is reporting non-terminal serious errors which require urgent attention from the user.

controllers/openstackmachine_controller.go
```diff
@@ -383,21 +395,73 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, logger
 	return ctrl.Result{}, nil
 }
 
-func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, computeService *compute.Service, userData string) (*compute.InstanceStatus, error) {
-	instanceStatus, err := computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
-	if err != nil {
-		return nil, err
+func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, userData string) (*compute.InstanceSpec, error) {
+	if openStackMachine == nil {
+		return nil, fmt.Errorf("create Options need be specified to create instace")
```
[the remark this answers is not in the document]
mdbooth: Agree, but again in a separate series.

mdbooth:
@chrischdi On exponential backoff, yes absolutely! I'd added that as a placeholder for something cleverer and didn't change it. Must not commit this serious without addressing that, so adding a

/hold

In the meantime if controller-runtime does exp backoff it's probably good enough to just use that.

mdbooth:
The default backoff behaviour of controller-runtime is set here:

https://github.com/kubernetes-sigs/controller-runtime/blob/eb292e5d9bd6c59663fc6777ae2b99f901f386e4/pkg/controller/controller.go#L120-L122

and defined here:

https://github.com/kubernetes/client-go/blob/eb103e0abf6218751302bf984d01ac45aae2ac6b/util/workqueue/default_rate_limiters.go#L39-L45

To summarise, each individual object has a 5ms backoff growing with an exponent of 2 up to a maximum of 1000 seconds. There is also a global limit of 10 reconciles per second. So retries will happen with a backoff of:

```
0.005s
0.025s
0.125s
0.625s
3.125s
15.625s
78.125s
390.625s
1000s
...
```

I don't think this is great for us in practise, as most backoffs will be resource creation waits. We would likely almost always do the first 6 retries before success. We should probably customise it. e.g. a 10s backoff with an exponent of 1.5 up to 1 hour would look like:

```
10s
15s
22.5s
33.75s
50.625s
75.9375s
113.90625s
170.859375s
256.2890625s
384.4335938s
576.6503906s
864.9755859s
1297.463379s
1946.195068s
2919.292603s
3600s
```

I'd also be inclined to retain the global 10 reconciles/s limit.

However, for now I'm happy to use the default. We can revisit this.

controllers/openstackcluster_controller.go
mdbooth: I've updated everything to use the controller's rate limiter. I've also customised the rate limiter to have a higher base delay. Without writing my own exponential backoff rate limiter (which I wasn't keen on doing) we're stuck with an exponent of 2, so I've set the base delay at 2 seconds. We will back off at intervals of 2s, 4s, 8s... up to a max of 1000s.

seanschneeweiss (commented):
How would you feel about setting `openStackMachine.Status.Ready = false` after the following positions:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/27817ed4193cc4f04d5a68689cf17ecdf0af170f/controllers/openstackmachine_controller.go#L338
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/27817ed4193cc4f04d5a68689cf17ecdf0af170f/controllers/openstackmachine_controller.go#L346

main.go
```diff
@@ -199,19 +201,48 @@ func setupChecks(mgr ctrl.Manager) {
 }
 
 func setupReconcilers(ctx context.Context, mgr ctrl.Manager) {
+	// Based on workqueue.DefaultControllerRateLimiter with a higher baseDelay
```
chrischdi, line 204: Maybe to discuss this again here:

We not only have failures due to OpenStack API Calls.
I don't know if we really want to customize that.

When thinking about the "Happy Path" for CAPO I agree with you, it does not make sense.

But if we consider the existence of other controllers acting on the same resources as CAPO, the default is reasonable for doing retries to update the CR objects.

E.g. the reconcile fails at the end (when CAPO wants to apply its patch) because someone in the meantime updated the CR and by that increased the metadata.resourceVersion (maybe even only adding a label), this would result in using the increased backoff instead of using the default.

Also in the use-case of the PR / the discussion:
```
	if !deleted {
		return reconcile.Result{Requeue: true}, nil
	}
```
Would this even make use the rate-limiting or result in an immediate retry (making the custom rate-limited obsolete for this use-case) because it is not returning an error?

Maybe better suited to post that on the PR :smile:

I did not find that any other cloud provider does use something other than the default for their controller: https://cs.k8s.io/?q=RateLimiter&i=nope&files=&excludeFiles=vendor%2F.*&repos=kubernetes-sigs/cluster-api,kubernetes-sigs/cluster-api-operator,kubernetes-sigs/cluster-api-provider-aws,kubernetes-sigs/cluster-api-provider-azure,kubernetes-sigs/cluster-api-provider-digitalocean,kubernetes-sigs/cluster-api-provider-gcp,kubernetes-sigs/cluster-api-provider-ibmcloud,kubernetes-sigs/cluster-api-provider-kubemark,kubernetes-sigs/cluster-api-provider-kubevirt,kubernetes-sigs/cluster-api-provider-nested,kubernetes-sigs/cluster-api-provider-openstack,kubernetes-sigs/cluster-api-provider-packet,kubernetes-sigs/cluster-api-provider-vsphere

Note: I'm ok with the change (50:50), just wanted to raise some awareness about that :-)

MPV:
Will this change mean that a cluster that (for whatever reason) can't reconcile, can still reconcile its bastion?
Imagine in a case where you have a cluster created without a bastion, then the cluster fails, and you want CAPO to add a bastion so you can reach the cluster for troubleshooting. Curious about your thoughts on this.

mdbooth:
> How would you feel about setting `openStackMachine.Status.Ready = false` after the following positions:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/27817ed4193cc4f04d5a68689cf17ecdf0af170f/controllers/openstackmachine_controller.go#L338
> 
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/27817ed4193cc4f04d5a68689cf17ecdf0af170f/controllers/openstackmachine_controller.go#L346

I thought Ready was a one-way gate, i.e. we're not allowed to unset it once set? If that's not correct, though, I'm very interested in that.

I actually have a draft doc which I need to clean up and push somewhere with my thoughts on the failed state. I *do* think we need a status marker for non-terminal failure. These would be examples of non-terminal failure. If we could use Ready that way it would be very interesting, but without knowing for sure I suspect it would break assumptions. I think this is a CAPI discussion.

mdbooth:
> Will this change mean that a cluster that (for whatever reason) can't reconcile, can still reconcile its bastion? Imagine in a case where you have a cluster created without a bastion, then the cluster fails, and you want CAPO to add a bastion so you can reach the cluster for troubleshooting. Curious about your thoughts on this.

No. It won't start reconciling the bastion until the cluster is up. If the cluster never comes up it will never create the bastion.

There's no fundamental reason we couldn't do this, though, but it would take a fair amount of refactoring. This is essentially the same direction I'm trying to take the reconciliation of instances, i.e. independent resources can be reconciled concurrently.

main.go
mdbooth: Couple of things.

Firstly I agree that we should consider this carefully. I need to rebase this PR anyway, and when I do I'm going to remove the rate limiting. I'll propose it again as a separate PR and we can examine it in more detail there. I think this will work acceptably well without modifying the rate limiter, although with a bit more load on the OpenStack API than is necessary.

Secondly, the controller-runtime documentation on this feature sucks, so I resorted to RTFS 🙄 AFAICT the rate limiter is used in exactly 2 use cases:

* Reconcile returns error
* Reconcile returns Requeue: true

Specifically, AFAICT watches are *not* added to the queue with the rate limiting interface, which means they will be enqueued immediately. That is, if a controller touches a machine object, the machine controller will reconcile that machine object immediately without rate limiting. We could actually verify this manually by setting it to something ridiculously large and checking that Reconcile is called immediately, but lets do that on the new PR.

Also, if Reconcile returns RequeueAfter, that *also* doesn't use the rate limiter: it uses the value passed in.

So the rate limiter is only ever used when invoked by code in our controller returning one of the 2 values which uses it.

MPV:
> > Will this change mean that a cluster that (for whatever reason) can't reconcile, can still reconcile its bastion? Imagine in a case where you have a cluster created without a bastion, then the cluster fails, and you want CAPO to add a bastion so you can reach the cluster for troubleshooting. Curious about your thoughts on this.
> 
> No. It won't start reconciling the bastion until the cluster is up. If the cluster never comes up it will never create the bastion.

I'm referring to another scenario though, where a cluster has already been created successfully (but configured without a bastion). Then one might decide to add a bastion (for easier troubleshooting). But if, at that point, for whatever reason, the cluster can't reconcile, would the bastion still reconcile successfully on its own (so its added and we can troubleshoot the cluster)?

> There's no fundamental reason we couldn't do this, though, but it would take a fair amount of refactoring. This is essentially the same direction I'm trying to take the reconciliation of instances, i.e. independent resources can be reconciled concurrently.

Sounds promising. 🙏

mdbooth:
> > > Will this change mean that a cluster that (for whatever reason) can't reconcile, can still reconcile its bastion? Imagine in a case where you have a cluster created without a bastion, then the cluster fails, and you want CAPO to add a bastion so you can reach the cluster for troubleshooting. Curious about your thoughts on this.
> > 
> > 
> > No. It won't start reconciling the bastion until the cluster is up. If the cluster never comes up it will never create the bastion.
> 
> I'm referring to another scenario though, where a cluster has already been created successfully (but configured without a bastion). Then one might decide to add a bastion (for easier troubleshooting). But if, at that point, for whatever reason, the cluster can't reconcile, would the bastion still reconcile successfully on its own (so its added and we can troubleshoot the cluster)?

No, this isn't going to help in that case. That said, the bastion requires most of the cluster infrastructure to be up already, so I'm not sure how early we could usefully move it in the cluster reconciliation process anyway.

What's the use case, btw? I'd have thought that the bastion was most useful for debugging machines rather than the cluster? If we're failing to create networks/routers/security groups/load balancers that's surely all going to be debugged via object status/events/OpenStack API. Does the bastion help with any of that?

dulek:
I tested this on my env and it seems to work just fine.

controllers/openstackcluster_controller.go
```diff
@@ -185,54 +189,51 @@ func contains(arr []string, target string) bool {
...
 
 	if err = networkingService.DeleteBastionSecurityGroup(openStackCluster, fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)); err != nil {
-		handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete bastion security group: %v", err))
-		return errors.Errorf("failed to delete bastion security group: %v", err)
+		return false, errors.Errorf("failed to delete bastion security group: %v", err)
 	}
 	openStackCluster.Status.BastionSecurityGroup = nil
 
-	return nil
+	return true, nil
```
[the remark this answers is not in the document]
mdbooth: Currently yes, but for the same reason as updating reconcile instance I want to be able to return an incomplete result. i.e. There was no error, but delete is not finished yet.

seanschneeweiss:
> I thought Ready was a one-way gate, i.e. we're not allowed to unset it once set? If that's not correct, though, I'm very interested in that.
> 
> I actually have a draft doc which I need to clean up and push somewhere with my thoughts on the failed state. I _do_ think we need a status marker for non-terminal failure. These would be examples of non-terminal failure. If we could use Ready that way it would be very interesting, but without knowing for sure I suspect it would break assumptions. I think this is a CAPI discussion.

@mdbooth I had a look at [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/bf6ba4d18af38259d0690c147100ee49ba7eb74f/controllers/azuremachine_controller.go#L299), [CAPA](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/bf6ba4d18af38259d0690c147100ee49ba7eb74f/controllers/azuremachine_controller.go#L299), and [CAPIBM](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/1e2ed3a5c649550fee6c7ce0971b636b5bc86923/controllers/ibmpowervsmachine_controller.go#L213) to do so too.
CAPG doesn't change the ready status.

From the CRD and cluster-api docs I don't really identify whether to see this as a one-way gate or not. Definitely a question for CAPI. Personally I'd like to use it for non-terminal failure.

mdbooth:
I'll create a new PR when I'm working on this again.

PR #1165, opened 2022-03-08
🐛 Clean up removed AvailabilityZones from `OpenStackCluster.status.failureDomains`

### LINKED ISSUE #1163: `OpenStackCluster.status.failureDomains` is never cleaned up
/kind bug

- Have an existing `OpenStackCluster`
- Temporarily add a new Availability Zone in OpenStack
- Remove the Availability Zone again
- Availability Zone is still referenced in `OpenStackCluster.status.failureDomains` and never removed

I expected the temporary Availability Zone to be removed once it's not longer visible in OpenStack.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `v0.5.2`
- Cluster-API version: `v1.1.2`
- OpenStack version: Wallaby
- Minikube/KIND version: /
- Kubernetes version (use `kubectl version`): `v1.21.9`
- OS (e.g. from `/etc/os-release`): /

apricote:

Previously we only added to the List of `FailureDomains`, but this can cause issue in case any Availability Zone is removed in OpenStack, as that AZ will still be used by CAPI to deploy Control Plane machines. With this PR, the list of available FailureDomains/AZs is recreated on every reconcile and outdated entries are removed.

Fixes #1163

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -268,9 +268,9 @@ func reconcileNormal(ctx context.Context, log logr.Logger, client client.Client,
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if openStackCluster.Status.FailureDomains == nil {
-		openStackCluster.Status.FailureDomains = make(clusterv1.FailureDomains)
-	}
+	// Create a new list to remove any Availability
+	// Zones that have been removed from OpenStack
+	openStackCluster.Status.FailureDomains = make(clusterv1.FailureDomains)
 	for _, az := range availabilityZones {
 		// I'm actually not sure if that's just my local devstack,
 		// but we probably shouldn't use the "internal" AZ


mdbooth (approved):
This looks like a sensible change to me. As we're fetching AZs anyway there's no cost to it and it keeps the status up to date. I think it's also backportable.

If you're interested, as a follow up the for loop below could also do with a cleanup. Firstly we should drop the special case of the 'internal' AZ. There are no special AZ names in OpenStack, so this is a config thing.

The other thing is that the 9 locs below that can (and should) be simplified to 1:

```golang
  found := contains(openStackCluster.Spec.ControlPlaneAvailabilityZones, az.ZoneName)
```

MPV:
Possible candidate for backport `v0.5.3` / https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1166?

jichenjc:
>>>Firstly we should drop the special case of the 'internal' AZ

I think it's come from devstack test and added this workaround

normal user see this in a fresh devstack

```
+-----------+-------------+
| Zone Name | Zone Status |
+-----------+-------------+
| nova      | available   |
| nova      | available   |
+-----------+-------------+
```

admin see this 

```
+-----------+-------------+
| Zone Name | Zone Status |
+-----------+-------------+
| internal  | available   |
| nova      | available   |
| nova      | available   |
+-----------+-------------+
```

IIRC nova has this `internal` hard coded somewhere and technically it should not be 
useable in CAPI ?

jichenjc:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1168 for the internal AZ

/hold cancel

mdbooth:
> > > > Firstly we should drop the special case of the 'internal' AZ
> 
> I think it's come from devstack test and added this workaround
> 
> normal user see this in a fresh devstack
> 
> ```
> +-----------+-------------+
> | Zone Name | Zone Status |
> +-----------+-------------+
> | nova      | available   |
> | nova      | available   |
> +-----------+-------------+
> ```
> 
> admin see this
> 
> ```
> +-----------+-------------+
> | Zone Name | Zone Status |
> +-----------+-------------+
> | internal  | available   |
> | nova      | available   |
> | nova      | available   |
> +-----------+-------------+
> ```
> 
> IIRC nova has this `internal` hard coded inside nova somewhere and technically it should not be useable in CAPI ? @mdbooth @apricote

No, it's not hardcoded: it's configured by devstack. We shouldn't special case it because it's not special. If we don't want to use it in our tests we just shouldn't configure it.

jichenjc:
I have some additional findings https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1168, maybe you can check stuffs there ..

tobiasgiese:
> /cherry-pick release-0.5

This is only available for open PRs (like mentioned yesterday) 😞

apricote:
It should work with merged PRs according to the docs:

> Cherrypick a PR to a different branch. This command works both in merged PRs (the cherrypick PR is opened immediately) and open PRs (the cherrypick PR opens as soon as the original PR merges).

I think it does not work because I am not a member of kubernetes-sigs:

> Who can use?: Members of the trusted organization for the repo.

Both quotes from: https://prow.k8s.io/command-help#cherrypick

PR #1168, opened 2022-03-09
Remove `internal` AZ special handling

jichenjc:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -268,16 +268,11 @@ func reconcileNormal(ctx context.Context, log logr.Logger, client client.Client,
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
 	// Create a new list to remove any Availability
 	// Zones that have been removed from OpenStack
 	openStackCluster.Status.FailureDomains = make(clusterv1.FailureDomains)
 	for _, az := range availabilityZones {
-		// I'm actually not sure if that's just my local devstack,
-		// but we probably shouldn't use the "internal" AZ
-		if az.ZoneName == "internal" {
-			continue
-		}
-
 		found := true
 		// If Az given, then check whether it's in the allow list
 		// If no Az given, then by default put into allow list
--- a/docs/book/src/clusteropenstack/configuration.md
+++ b/docs/book/src/clusteropenstack/configuration.md
@@ -86,7 +86,7 @@ or [configure custom security groups](#security-groups) with rules allowing ingr
 ## OpenStack credential
 
 ### Generate credentials
-The [env.rc](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/env.rc) script sets the environment variables related to credentials.
+The [env.rc](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/env.rc) script sets the environment variables related to credentials. It's highly recommend to avoid using `admin` credential.
 
 ```bash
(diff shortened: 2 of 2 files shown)


apricote (approved):

mdbooth (changes requested):
Sorry for mis-remembering what this does! However, lets still hold fire on this.

2 things:
1. I *think* this is intended to be the name of a real AZ.
2. It's configurable.

So I still thing it's wrong to special case it, but I'm asking for clarification from Nova folks.

Bear in mind that this would be surprising behaviour for anybody who created an AZ called 'internal', which is not an unreasonable name for an AZ, and expected to be able to use it.

jichenjc:
Technically I think we should avoid show `internal` in the API response of openstack
but API change will be a big process ahead .. and yes, maybe you can get info through irc and 
see how those people react to us.. thanks~

mdbooth (approved):
/lgtm

Having thought about this some more I would still prefer to remove this code entirely. However, the new comment is definitely better than the old one so no reason not to merge this.

jichenjc:
>>>It's only visible to admin. We should never be using admin credentials, so we should never see it anyway.
If we do use admin credentials, we can specify valid failure domains in the cluster spec.

Ok, as I mentioned before, this caused some issue in previous test 
I am actually don't like the way hardware such restriction as well
but this can cause some trouble to the user who is new to CAPO
I will update this PR to include doc changes so they can refer to this doc if they encounter such issue

controllers/openstackcluster_controller.go
```diff
@@ -268,16 +268,15 @@ func reconcileNormal(ctx context.Context, log logr.Logger, client client.Client,
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	if len(openStackCluster.Spec.ControlPlaneAvailabilityZones) == 0 {
+		log.Info("There is openStackCluster.Spec.ControlPlaneAvailabilityZones provided, this can cause issue if the administrator credential is used because `internal` availability zone will be a candidate")
```
mdbooth, line 273: This is user-facing information and the logs are for devs so I'm not sure the right person would ever see this. I'd omit it personally. The docs message has it covered.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -109,6 +109,8 @@ Note: you need to set `clusterctl.cluster.x-k8s.io/move` label for the secret cr
 
 The availability zone names must be exposed as an environment variable `OPENSTACK_FAILURE_DOMAIN`.
 
+By default, if `Availability zone` is not given, all `Availability zone` that defined in openstack will be a candidate to provision from, If administrator credential is used then `internal` Availability zone which is internal only Availability zone inside `nova` will be returned and can cause potential problem, see [PR 1165](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1165) for further information. So we highly recommend to set `Availability zone` explicitly.
```
mdbooth, line 112: Nice. Do we document anywhere that admin credentials aren't recommended?

controllers/openstackcluster_controller.go
jichenjc: ok

docs/book/src/clusteropenstack/configuration.md
jichenjc: I don't think we do, I can find a place to update

mdbooth (approved):
/lgtm

Issue #1176, opened 2022-03-11
Floating IP events for machines reported against the cluster object


mdbooth:
/kind bug

**What steps did you take and what happened:**
The networking service floating IP operations all take an OpenStackCluster object as an argument. This object is used as an event target even when called by the machine controller.

**What did you expect to happen:**
Floating IP events created for a machine should target the machine

PR #1177, opened 2022-03-11
🐛 Fix event target of floating ip operations

mdbooth:
Floating IP operations were always reporting events against the cluster,
even when creating a floating IP on behalf of a machine.

Fixes: #1176 

/hold


seanschneeweiss (commented):
/lgtm

Issue #1146, opened 2022-02-23
Support application credentials


mdbooth:
/kind feature

CAPO doesn't currently support application credentials. From memory the issue relates to parsing project id from the cloud credentials. It's easily solved, as this can be obtained from the authentication response from keystone and is already exposed by Gophercloud.

Consider for inclusion in 0.6: #1145

apricote:
We currently use Application Credentials with CAPO, one can just specify the `project_id` together with the application credentials and CAPO figures out the right thing to do, but the `clouds.yaml` can not be used by the OpenStack CLI like this.

I would like to clean this up :+1: 

/assign

mdbooth:
I think the key is going to be https://pkg.go.dev/github.com/gophercloud/gophercloud#ProviderClient.GetAuthResult

However, I think it gets a bit messy here because the returned interface is insufficient so we may have to switch on the underlying type. Specifically I think we want to cast it to https://pkg.go.dev/github.com/gophercloud/gophercloud@v0.24.0/openstack/identity/v3/tokens#CreateResult, and call ExtractProject().

/cc @pierreprinetti @EmilienM

apricote:
We discussed this in Slack and decided on refactoring `provider.NewClient` to explicitly return the ProjectID and pass it around to the clients that need it.

https://app.slack.com/client/T09NY5SBT/CFKJB65G9/thread/CFKJB65G9-1645786779.195309

MPV:
I can also confirm that it's usable today, having a secret with a `clouds.yaml` that looks for example like this:

```yaml
clouds:
  name-of-cloud-here:
    identity_api_version: 3
    auth:
      auth_url: ${auth_url}
      application_credential_id: ${credential_id}
      application_credential_secret: ${credential_secret}
      project_id: ${project_id}
    auth_type: "v3applicationcredential"
```

PR #1178, opened 2022-03-14
🏃 Add Scopes to pass data to services

apricote:

This refactoring will make it easier to pass the `ProjectID` into the services that need it (see #1146). The Scopes pattern is also used by other CAPI Providers:

- CAPA: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/scope/cluster.go
- CAPG: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/cloud/scope/cluster.go
- CAPV (Called `Context`): https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/context/cluster_context.go

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/service.go
+++ b/pkg/cloud/services/loadbalancer/service.go
@@ -22,39 +22,39 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
-	"github.com/gophercloud/utils/openstack/clientconfig"
 
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/scope"
 )
 
 // Service interfaces with the OpenStack Neutron LBaaS v2 API.
 type Service struct {
 	projectID          string
+	scope              *scope.Scope
 	loadbalancerClient LbClient
 	networkingService  *networking.Service
-	logger             logr.Logger
 }
 
 // NewService returns an instance of the loadbalancer service.
-func NewService(client *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts, logger logr.Logger) (*Service, error) {
-	loadbalancerClient, err := openstack.NewLoadBalancerV2(client, gophercloud.EndpointOpts{
-		Region: clientOpts.RegionName,
+func NewService(scope *scope.Scope) (*Service, error) {
+	loadbalancerClient, err := openstack.NewLoadBalancerV2(scope.ProviderClient, gophercloud.EndpointOpts{
+		Region: scope.ProviderClientOpts.RegionName,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to create load balancer service client: %v", err)
 	}
 
-	networkingService, err := networking.NewService(client, clientOpts, logger)
+	networkingService, err := networking.NewService(scope)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/cloud/services/loadbalancer/service.go
```diff
@@ -22,39 +22,39 @@ import (
...
-	logger             logr.Logger
 }
 
 // NewService returns an instance of the loadbalancer service.
-func NewService(client *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts, logger logr.Logger) (*Service, error) {
-	loadbalancerClient, err := openstack.NewLoadBalancerV2(client, gophercloud.EndpointOpts{
-		Region: clientOpts.RegionName,
+func NewService(scope *scope.Scope) (*Service, error) {
+	loadbalancerClient, err := openstack.NewLoadBalancerV2(scope.ProviderClient, gophercloud.EndpointOpts{
+		Region: scope.ProviderClientOpts.RegionName,
```
mdbooth, line 41: I know we're already looking into project id, but it looks like region can also be extracted from the authentication response. Lets kill ProviderClientOpts entirely as soon as we can.

pkg/scope/scope.go
```diff
@@ -0,0 +1,65 @@
...
+	Logger logr.Logger
+	Client client.Client
+}
+
+type Params struct {
+	Client             *client.Client
+	ProviderClient     *gophercloud.ProviderClient
+	ProviderClientOpts *clientconfig.ClientOpts
+	Logger             *logr.Logger
+}
```
mdbooth, line 40: These are (almost) the same? Do we need a separate struct?

pkg/scope/scope.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Scope struct {
+	ProviderClient     *gophercloud.ProviderClient
+	ProviderClientOpts *clientconfig.ClientOpts
+
+	Logger logr.Logger
+	Client client.Client
```
mdbooth, line 32: If we're creating the objects in the calling scope and passing a pointer, these should probably remain pointers. Dereferencing these structs in the constructor could get weird.

mdbooth (commented):
I still like this in principal!

pkg/scope/scope.go
apricote: I adopted this from CAPA, where they used `interfaces` with getter methods for the actual scope, so they needed another `ScopeParams` struct to pass in the parameters to the `NewScope()` method.

You are right, it does not really make sense in this context.

pkg/scope/scope.go
apricote: The decision to make the `ProviderClient` members pointers and `Logger` and `Client` values was done based on how these were previously being passed around.

If we remove the `Params` struct, I can just make everything in `Scope` a value and then we pass around the `Scope` as a pointer.

pkg/cloud/services/loadbalancer/service.go
apricote: Sounds good to me, I never really understood why the `clientOpts` were necessary.

mdbooth: Should probably be a follow up, though! Lets leave it alone in for this PR.

jichenjc:
/lgtm

I like the idea, for the questions that need follow up, we can anyway create a new one

mdbooth (commented):
/approve

PR #1179, opened 2022-03-15
✨Export GetFloatingIP

mdbooth:
Although there are currently no internal users, this allows fetching a
floating IP without attempting to create one if it doesn't exist.

I'm actually adding this because we need it [in MAPO](https://github.com/openshift/machine-api-provider-openstack/blob/4adc3323d876799867e5bf66e6244fdcc99b415b/pkg/machine/actuator.go#L313). The use case is associating a specific floating IP to a machine which is required to exist already. We don't want to make any attempt to to create the floating IP if it doesn't exist.

/hold

### DIFF
--- a/pkg/cloud/services/networking/floatingip.go
+++ b/pkg/cloud/services/networking/floatingip.go
@@ -36,7 +36,7 @@ func (s *Service) GetOrCreateFloatingIP(eventObject runtime.Object, openStackClu
 	var fpCreateOpts floatingips.CreateOpts
 
 	if ip != "" {
-		fp, err = s.checkIfFloatingIPExists(ip)
+		fp, err = s.GetFloatingIP(ip)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


mdbooth:
AFAICT the netlify failure is a flake. Also this patch doesn't touch the docs.

jichenjc:
>>>Although there are currently no internal users, this allows fetching a
floating IP without attempting to create one if it doesn't exist.

just curious, you are only rename the function and no logic change, so how does this happen?
or you renamed the function so MAPO can use it?

mdbooth:
> > > > Although there are currently no internal users, this allows fetching a
> > > > floating IP without attempting to create one if it doesn't exist.
> 
> just curious, you are only rename the function and no logic change, so how does this happen? or you renamed the function so MAPO can use it?

It's just so MAPO can use it. We have a use case where we want to associate a floating IP which *must* already exist, and we don't have admin credentials so an attempt to create it (with a specified IP) will definitely fail. This just means that the `OrCreate` functionality is problematic.

mdbooth:
Incidentally I'm not aware of any way to make the netlify tests recheck. Don't know if tide will merge the change without them passing. If we can't re-run them, is there any way to mark them explicitly ignored?

jichenjc:
/hold cancel

seems the override doesn't take effect ...

mdbooth:
@jichenjc I triggered a full retest by changing the commit message 🤷

PR #1184, opened 2022-03-24
🐛 upgrade golangci-lint for go 1.18 support

apricote:

With versions lower than 1.45.0, golangci-lint returns a lot of invalid errors when running in Go 1.18+. Because of this, all PR pipeline jobs `pull-cluster-api-provider-openstack-test` fail.

This commit upgrades golangci-lint to the latest released version and fixes all linting errors that came up.

/hold


tobiasgiese:
/lgtm
/approve
/hold cancel

thanks @apricote 👌🏻

pkg/record/recorder.go
```diff
@@ -45,20 +46,20 @@ func InitFromRecorder(recorder record.EventRecorder) {
 
 // Event constructs an event from the given information and puts it in the queue for sending.
 func Event(object runtime.Object, reason, message string) {
-	defaultRecorder.Event(object, corev1.EventTypeNormal, strings.Title(reason), message)
+	defaultRecorder.Event(object, corev1.EventTypeNormal, cases.Title(language.Und).String(reason), message)
```
seanschneeweiss, line 49: Out of curiosity, what is the `language.Und` doing?

seanschneeweiss (commented):
/lgtm
Thank you very much for fixing this.

tobiasgiese:
/check-cla
/retest
/approve
/lgtm

I think prow had some issues today

jichenjc:
@apricote guess you forgot to run `go mod tidy` to update go.sum?

api/v1alpha3/openstackcluster_types.go
```diff
@@ -30,7 +30,6 @@ const (
 
 // OpenStackClusterSpec defines the desired state of OpenStackCluster.
 type OpenStackClusterSpec struct {
-
```
jichenjc, line 33: unrelated change..

pkg/record/recorder.go
jichenjc: same question here..

api/v1alpha3/openstackcluster_types.go
apricote: These are changes that `golangci-lint` requested after I upgrade it. In this case the structs were not formatted with gofumpt

pkg/record/recorder.go
apricote: I think this is for `Undefined` language, it's not really explained in the [docs]. Using this language tag returned the expected result, as you can see in this Go Playground example: https://go.dev/play/p/fNew-KrewPY

apricote:
> @apricote guess you forgot to run `go mod tidy` to update go.sum?

Looks like I forgot to run it a second time after fixing the linting issues. New commit is up

pkg/record/recorder.go
lentzi90: Makes sense. Maybe change it to `language.English` then?

apricote: I think that makes sense, considering the reasons should be in english so all consumers of them can understand them.

chrischdi (commented):
/lgtm

pierreprinetti:
pass! hurray!

apricote:
/cherry-pick release-0.5

third time's a charm

apricote:
/cherry-pick release-0.5

Third time also didn't work and I forgot about this afterwards, we are currently hitting this issue on #1204.

PR #1185, opened 2022-03-29
🐛 Install sshuttle using pip instead of from source

lentzi90:

We used to install sshuttle from source because we needed a commit that was not in the latest release. There is however a new release now that includes the commit. Upstream sshuttle also [dropped support for python 3.6 and 3.7 recently](https://github.com/sshuttle/sshuttle/commit/6d36916f48ae69abe175682f111f505923aba339) which means we can no longer install that, since we use python 3.7.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/hack/ci/create_devstack.sh
+++ b/hack/ci/create_devstack.sh
@@ -94,14 +94,7 @@ function wait_for_ssh {
 function start_sshuttle {
     if ! command -v sshuttle;
     then
-        # Install sshuttle from source because we need: https://github.com/sshuttle/sshuttle/pull/661
-        # TODO(sbueringer) install via pip after the next release after 1.0.5 via:
-        # pip3 install sshuttle
-        pushd /tmp
-        git clone https://github.com/sshuttle/sshuttle.git
-        cd sshuttle
-        pip3 install .
-        popd || exit 1
+        pip3 install sshuttle
     fi
 
     kill_sshuttle


tobiasgiese:
/ok-to-test
/lgtm

thanks 🙂

tobiasgiese:
```
ERROR: Package 'sshuttle' requires a different Python: 3.7.3 not in '>=3.8'
```

we need at least Python 3.8 to install sshuttle via pip

lentzi90:
> ```
> ERROR: Package 'sshuttle' requires a different Python: 3.7.3 not in '>=3.8'
> ```
> 
> we need at least Python 3.8 to install sshuttle via pip

Where do you see this? I think this is true for the *latest source code* but not for the latest release (v1.1.0 that was released in January). They dropped support for python 3.7 just a few hours ago. When installing from pip we should get the v1.1.0 release:

```
 + pip3 install sshuttle
Collecting sshuttle
  Downloading sshuttle-1.1.0-py2.py3-none-any.whl (70 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 70.6/70.6 KB 3.2 MB/s eta 0:00:00
Installing collected packages: sshuttle
Successfully installed sshuttle-1.1.0 
```

Issue #1105, opened 2021-12-28
✨ More config options for APIServer Loadbalancer (AvailabilityZone, Flavor, ...)


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

CPO has many LB related features
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#supported-features

this will be used in service create through `kubectl create svc`
but our LB for masters doesn't take those into considreations

something like `loadbalancer.openstack.org/availability-zone` etc might be helpful for CAPO.. 
let's use this issue to track..

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

mdbooth:
Can you give a worked example of how we'd use the annotation?

mdbooth:
@apricote was also interested in this.

apricote:
@jichenjc Is there a reason why you want to use annotations to expose these configuration values?

I would suggest to put these into the `OpenStackCluster` CRD instead:

```go
type OpenStackClusterSpec struct {
	// ... existing fields
   
	// Configure a managed loadbalancer for the API Server
	APIServerLoadBalancer Loadbalancer `json:"apiServerLoadBalancer,omitempty"`
}

type Loadbalancer struct {
	// Enabled replaces the current `OpenStackClusterSpec.ManagedAPIServerLoadBalancer`
	Enabled bool `json:"enabled,omitempty"`
	// AdditionalPorts replaces the current `OpenStackClusterSpec.APIServerLoadBalancerAdditionalPorts`
	AdditionalPorts []int `json:"additionalPorts,omitempty"`

	// New Options
	FlavorID         string `json:"flavorID,omitempty"`
	AvailabilityZone string `json:"availabilityZone,omitempty"`
}

jichenjc:
@apricote I am ok if you go with `Spec` :) ,doesn't matter if it's annotation or spec
just make it more flexible and fit for requirement is good enough ~

apricote:
Alright, I will present this during tomorrows office hours and get an approval for the CRD changes :+1:

apricote:
No objections to the general plan during office hours.

I am going to create PRs for this.

/retitle ✨ More config options for APIServer Loadbalancer (AvailabilityZone, Flavor, ...)

apricote:
[...]

PR #1187, opened 2022-03-31
⚠️ move loadbalancer opts to struct

apricote:

Modify the `v1beta1` CRD and move the loadbalancer fields `OpenStackClusterSpec.{ManagedAPIServerLoadBalancer,APIServerLoadBalancerAdditionalPorts}` into a struct at `OpenStackClusterSpec.APIServerLoadBalancer` to enable adding further customizations. This was previously discussed in #1105.

Related to #1105

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
Thanks for the PR, I assume you need update templates/*.yaml (e.g cluster-template.yaml) and maybe e2e settings 
in order to honor the new things introduced?

apricote:
I just pushed an update that contains:

- Unit tests for the conversions
- Fixed `OpenStackCluster` manifests in `templates/` and `test/data/infrastructure-openstack`
- updated generated conversions (`make generate`)

jichenjc:
/lgtm

thanks for the fix ~

mdbooth:
This is a really nice cleanup.

/approve

apricote:
Looks like a flaky test, was previously successful on the same commit.

/retest

api/v1alpha3/conversion.go
```diff
@@ -117,6 +119,10 @@ func Convert_v1alpha3_OpenStackClusterSpec_To_v1beta1_OpenStackClusterSpec(in *O
 			Name: in.CloudsSecret.Name,
 		}
 	}
+	out.APIServerLoadBalancer = infrav1.APIServerLoadBalancer{
+		Enabled:         in.ManagedAPIServerLoadBalancer,
+		AdditionalPorts: *(*[]int)(unsafe.Pointer(&in.APIServerLoadBalancerAdditionalPorts)),
```
chrischdi, line 124: What's the reason for using unsafe here? couldn't it be as easy as the following?:

```suggestion
		AdditionalPorts: in.APIServerLoadBalancerAdditionalPorts,
```

api/v1alpha3/conversion.go
```diff
@@ -139,6 +145,10 @@ func Convert_v1beta1_OpenStackClusterSpec_To_v1alpha3_OpenStackClusterSpec(in *i
 			Name: in.Bastion.Instance.IdentityRef.Name,
 		}
 	}
+
+	out.ManagedAPIServerLoadBalancer = in.APIServerLoadBalancer.Enabled
+	out.APIServerLoadBalancerAdditionalPorts = *(*[]int)(unsafe.Pointer(&in.APIServerLoadBalancer.AdditionalPorts))
```
chrischdi, line 150: What's the reason for using unsafe here? couldn't it be as easy as the following?:

```suggestion
	out.APIServerLoadBalancerAdditionalPorts = in.APIServerLoadBalancer.AdditionalPorts
```

apricote: I was too slow with answering, not its merged :(

I copied this from the previous autogenerated conversion.

The code is coming from `conversion-gen`: https://github.com/kubernetes/code-generator/blob/7e38d5771c5636f13555f0058e5936c521d2fe13/cmd/conversion-gen/generators/conversion.go#L929

chrischdi: So this whole part is auto generated? If so we have to keep it, otherwise I would prefer to clean it up in a follow-up PR :-)

apricote: Nope, my changes are not auto-generated, but I did not understand the semantics of the conversion code enough to remove the `unsafe.Pointer` logic provided by conversion-gen.

PR #1189, opened 2022-03-31
✨ Support Application Credential auth

### LINKED ISSUE #1146: Support application credentials

apricote:

We currently require an explicit `project_id` field in the clouds.yaml. This conflicts with the usual fields for Application Credentials, as they already include a scope, and setting another conflicts with that.

In this commit, we instead save the returned `project_id` from the initial auth api call, and pass it around to all services using the `Scope` struct.

Fixes #1146 

/hold

### DIFF
--- a/pkg/cloud/services/provider/provider.go
+++ b/pkg/cloud/services/provider/provider.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 	osclient "github.com/gophercloud/utils/client"
 	"github.com/gophercloud/utils/openstack/clientconfig"
 	corev1 "k8s.io/api/core/v1"
@@ -34,7 +35,6 @@ import (
 	"sigs.k8s.io/yaml"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
-	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
 )
 
 const (
@@ -42,7 +42,7 @@ const (
 	caSecretKey     = "cacert"
 )
 
-func NewClientFromMachine(ctx context.Context, ctrlClient client.Client, openStackMachine *infrav1.OpenStackMachine) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, error) {
+func NewClientFromMachine(ctx context.Context, ctrlClient client.Client, openStackMachine *infrav1.OpenStackMachine) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, string, error) {
 	var cloud clientconfig.Cloud
 	var caCert []byte
 
@@ -50,13 +50,13 @@ func NewClientFromMachine(ctx context.Context, ctrlClient client.Client, openSta
 		var err error
 		cloud, caCert, err = getCloudFromSecret(ctx, ctrlClient, openStackMachine.Namespace, openStackMachine.Spec.IdentityRef.Name, openStackMachine.Spec.CloudName)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, "", err
 		}
 	}
 	return NewClient(cloud, caCert)
 }
 
-func NewClientFromCluster(ctx context.Context, ctrlClient client.Client, openStackCluster *infrav1.OpenStackCluster) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, error) {
+func NewClientFromCluster(ctx context.Context, ctrlClient client.Client, openStackCluster *infrav1.OpenStackCluster) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, string, error) {
 	var cloud clientconfig.Cloud
 	var caCert []byte
 
@@ -64,13 +64,13 @@ func NewClientFromCluster(ctx context.Context, ctrlClient client.Client, openSta
 		var err error
 		cloud, caCert, err = getCloudFromSecret(ctx, ctrlClient, openStackCluster.Namespace, openStackCluster.Spec.IdentityRef.Name, openStackCluster.Spec.CloudName)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, "", err
 		}
 	}
 	return NewClient(cloud, caCert)
 }
 
-func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, error) {
+func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, string, error) {
 	clientOpts := new(clientconfig.ClientOpts)
 	if cloud.AuthInfo != nil {
 		clientOpts.AuthInfo = cloud.AuthInfo
@@ -80,13 +80,13 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 
 	opts, err := clientconfig.AuthOptions(clientOpts)
 	if err != nil {
-		return nil, nil, fmt.Errorf("auth option failed for cloud %v: %v", cloud.Cloud, err)
+		return nil, nil, "", fmt.Errorf("auth option failed for cloud %v: %v", cloud.Cloud, err)
 	}
 	opts.AllowReauth = true
 
 	provider, err := openstack.NewClient(opts.IdentityEndpoint)
 	if err != nil {
-		return nil, nil, fmt.Errorf("create providerClient err: %v", err)
+		return nil, nil, "", fmt.Errorf("create providerClient err: %v", err)
 	}
 
 	config := &tls.Config{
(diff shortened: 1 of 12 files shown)


pkg/cloud/services/provider/provider.go
```diff
@@ -169,34 +182,3 @@ type project struct {
 type projects struct {
 	Projects []project `json:"projects"`
 }
-
-func GetProjectID(client *gophercloud.ProviderClient, clientOpts *clientconfig.ClientOpts) (string, error) {
```
jichenjc, line 173: looks like we are going to drop the project (for normal case)? 

I understand we want to support application cred, but if this logic is removed, is it replaced by line 115-125 above?
but that code comment said it's for application cred only...

pkg/cloud/services/provider/provider.go
```diff
@@ -109,9 +109,22 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
...
-		return nil, nil, fmt.Errorf("providerClient authentication err: %v", err)
+		return nil, nil, "", fmt.Errorf("providerClient authentication err: %v", err)
 	}
-	return provider, clientOpts, nil
+
+	// Set ProjectID for consumers of the ProviderClient
+	// This is only needed when application credentials were used, because they
+	// do not specify a `project_id` in the clientconfig.Cloud.
+	var projectID string
+	if tokenResult, ok := provider.GetAuthResult().(tokens.CreateResult); ok {
```
mdbooth, line 119: @pierreprinetti Can I get your review on this dance specifically? 👆

pkg/cloud/services/provider/provider.go
```diff
@@ -109,9 +109,22 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
...
+		project, err := tokenResult.ExtractProject()
+		if err != nil {
+			return nil, nil, "", fmt.Errorf("unable to get project from createResult: %v", err)
+		}
+		projectID = project.ID
+	}
```
mdbooth, line 125: I think we want to return an error here if we failed to get a project from the authentication response. This corresponds to the error we would previously have raised if we didn't get a project from clientopts.

Because we have always required project to be specified in clientopts, we have always required project-scoped authentication, so IIUC project should always be available.

https://docs.openstack.org/api-ref/identity/v3/?expanded=#authentication-and-token-management

pkg/cloud/services/provider/provider.go
mdbooth: Yes, we should never have to pull project id from clientOpts. It should always be returned in the authentication response.

pkg/cloud/services/provider/provider.go
```diff
@@ -109,9 +109,22 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
...
 	err = openstack.Authenticate(provider, *opts)
 	if err != nil {
-		return nil, nil, fmt.Errorf("providerClient authentication err: %v", err)
+		return nil, nil, "", fmt.Errorf("providerClient authentication err: %v", err)
 	}
-	return provider, clientOpts, nil
+
+	// Set ProjectID for consumers of the ProviderClient
+	// This is only needed when application credentials were used, because they
+	// do not specify a `project_id` in the clientconfig.Cloud.
```
mdbooth, line 117: As noted by @jichenjc this is how we always get project id, not just for application credentials.

pkg/cloud/services/provider/provider.go
pierreprinetti: LGTM 👍

pierreprinetti: I’ll just say that if you want to protect against catastrophic Gophercloud changes, you could log something when !ok

apricote: Looking at the code in `github.com/gophercloud/gophercloud/openstack.Authenticate`, there are 4 auth scenarios that it seems to handle:

- [Identity v2 with `tokens`](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/openstack/client.go#L139)
- [Identity v3 with `ec2tokens`](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/openstack/client.go#L232)
- [Identity v3 with `oauth1`](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/openstack/client.go#L234)
- [Identity v3 with `tokens`](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/openstack/client.go#L236)

The `AuthResult` we get back is [of the type that was used for auth](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/provider_client.go#L230), but currently we only handle the `v3/tokens.CreateResult` type.

Looking at the generic [`AuthResult`](https://github.com/gophercloud/gophercloud/blob/321520e3be598fd242bb31cd2021d3b88912c42e/auth_result.go#L50) struct they even have an example of how one should handle this.

I am going to add handling for the other types of auth currently supported by Gophercloud and failure handling for unknown (new) types.

pierreprinetti: Apologies for the very newbie question, but are these scenarios supported in CAPO?

mdbooth: We can only support authentication mechanisms which can be entirely described by clouds.yaml, so I suspect that oauth at least wouldn't be implementable?

I would personally be ok with only supporting a single keystone auth mechanism as we currently do and just logging a useful error message. However, I'm also ok with handling all the response types we know about.

apricote: We do not specify which mechanisms we support: https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#openstack-credential

From the user perspective I would then assume that everything that can be describe in the clouds.yaml is fair game.

---

I checked if and where we can get the Project ID for the other mechanisms:

- Identity v2 with `tokens`: `result.ExtractToken().Tenant.ID`
- Identity v3 with `ec2tokens`: Seems to return a v3 `tokens.CreateResult`: https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/extensions/ec2tokens/requests.go#L291
- Identity v3 with `oauth1`: Also returns a v3 `tokens.CreateResult`: https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/extensions/oauth1/requests.go#L138

So the only additional case we would need to handle here is the Identity v2 `tokens` mechanism, which [was removed in Queens](https://docs.openstack.org/keystone/latest/contributor/http-api.html).

mdbooth: I would vote for keeping it simple and documenting that we require keystone v3. Lets return an error here if we haven't got that.

apricote: Okay :+1:

There is always the possibility about extending this in a later PR, if anyone wants the keystone v2 support.

pkg/cloud/services/provider/provider.go
apricote: @mdbooth is right, in a previous version of the patch I only added the new flow for application credentials, and this is where the comment is coming from.

Now that it's used for all auth mechanisms, the comment is outdated and I will remove it in the next push.

pkg/cloud/services/provider/provider.go
apricote: That makes sense. I will add an error return on the next push.

apricote:
/retest

Thanks for the extensive first round of review :) New commit with all requested changes is up.

pkg/cloud/services/provider/provider_test.xgo
```diff
@@ -0,0 +1,55 @@
+/*
```
tobiasgiese, line 1: Is the file type `xgo` correct? 🤔 
I think we must rename this file to `provider_test.go`

apricote: Nope.. It was too late in the work-day for me :sleeping:

I tried to write some unit tests for the function but failed to get it working with the `tokens.CreateResult` struct, which i couldn't mock. So if anyone has an idea how we could test this, feel free to teach me, otherwise I am going to remove this file.

tobiasgiese: Ah, so this is just a place holder / template test file. I understand.
Isn't it better to keep the correct file type but skip the test early?

apricote: I don't think the file is worth keeping around if we zero actual tests.

--- 

RE Mocking the `tokens.CreateResult`: We could copy the way `gophercloud` internally tests the `identity/v3/tokens` package, but that involves setting up a mock http server, so I am not sure if that is excessive.

apricote: I removed the file in the latest push.

seanschneeweiss:
Nice one.
/lgtm

jichenjc:
@apricote not sure it's eligible to add some test cases (with application ID) later on?

/approve

apricote:
> @apricote not sure it's eligible to add some test cases (with application ID) later on?

@jichenjc I had a unit test for the `getProjectIDFromAuthResult`, but couldn't figure out how to mock the `token.CreateResult` struct, if you have any guidance for that, I would gladly add a test case for this.

In general I would appreciate a test for this, that would probably mean to extend our e2e tests to use an addition clouds.yaml secret that uses application credentials. Are these supported by devstack?

PR #1191, opened 2022-04-01
✨Refactor CreateInstance and CreateBastion

mdbooth:
The primary purpose of this this PR is to cleanup the interface of `compute.CreateInstance` and make the separation of concerns between the machine controller (for Machines), the cluster controller (for the Bastion), and compute (for actual server creation) better defined.

The Bastion host is represented in the cluster spec as an `Instance`. An OpenStackMachine is represented by context in `OpenStackMachine` as well as `OpenStackCluster`. So while they both create a server in the same way, they both source the server's parameters in slightly different ways using different source objects.

At some point in history we also used `Instance` as the intermediate representation for an OpenStackMachine. That is, we combined parameters from an `OpenStackMachine` and an `OpenStackCluster` into an `Instance`, then passed that to CreateInstance. `Instance` is not ideal for this purpose as it is both Spec and Status. It contains fields which cannot be used as input parameters to CreateInstance. Therefore we refactored this into the internal-only `InstanceSpec`, which contains only spec fields.

This refactor takes this further. Firstly we ensure that creating an InstanceSpec from an OpenStackMachine or a cluster bastion is strictly data transformation and therefore very cheap. Anything expensive moves into the new CreateInstance. Secondly we move this code to the controller where it is relevant, and add new unit tests covering the transformation. This also reduces the scope of the CreateInstance unit tests, making them simpler.

I don't intend to squash these commits. They are intended to be independent logical steps. You may find it easier to review this PR by commit rather than as a single change.

This is the code motion portion of #1153 with functional changes to Bastion reconciliation removed.

/hold


apricote:
I like this refactoring a lot!

/lgtm

PR #1193, opened 2022-04-04
check ssh key pair before use it

### LINKED ISSUE #1103: check ssh key exist before take further actions
/kind feature

we allow ssh key as input but we don't check whether it exist or not 
just like Flavor 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/compute/instance.go#L206

we can add some check before really deploy a VM (createInstance)

jichenjc:

Fixes #1103

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/client.go
+++ b/pkg/cloud/services/compute/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	"github.com/gophercloud/utils/openstack/compute/v2/flavors"
@@ -43,6 +44,7 @@ type Client interface {
 	ListImages(listOpts images.ListOptsBuilder) ([]images.Image, error)
 
 	GetFlavorIDFromName(flavor string) (string, error)
+	GetSSHKeyPair(sshkey string) (*keypairs.KeyPair, error)
 	CreateServer(createOpts servers.CreateOptsBuilder) (*ServerExt, error)
 	DeleteServer(serverID string) error
 	GetServer(serverID string) (*ServerExt, error)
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/compute/client.go
```diff
@@ -87,6 +89,15 @@ func (s serviceClient) GetFlavorIDFromName(flavor string) (string, error) {
 	return flavorID, mc.ObserveRequest(err)
 }
 
+func (s serviceClient) GetSSHKeyPair(sshkey string) (*keypairs.KeyPair, error) {
+	mc := metrics.NewMetricPrometheusContext("keypair", "get")
```
seanschneeweiss, line 93: Shall we check for empty string in all the "get" functions?

pkg/cloud/services/compute/client_mock.go
```diff
@@ -141,6 +142,20 @@ func (mr *MockClientMockRecorder) GetFlavorIDFromName(arg0 interface{}) *gomock.
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetFlavorIDFromName", reflect.TypeOf((*MockClient)(nil).GetFlavorIDFromName), arg0)
 }
 
+// GetServer mocks base method.
```
seanschneeweiss, line 145:
```suggestion
// GetSSHKeyPair mocks base method.
```

pkg/cloud/services/compute/instance.go
```diff
@@ -142,6 +142,11 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 		return nil, fmt.Errorf("error getting flavor id from flavor name %s: %v", instanceSpec.Flavor, err)
 	}
 
+	_, err = s.computeService.GetSSHKeyPair(instanceSpec.SSHKeyName)
+	if err != nil {
+		return nil, fmt.Errorf("error geting ssh key with name %s: %v", instanceSpec.SSHKeyName, err)
```
seanschneeweiss, line 147:
```suggestion
		return nil, fmt.Errorf("error getting SSH key with name %s: %v", instanceSpec.SSHKeyName, err)
```

mdbooth:
What error message does Nova give us on creation if this is not valid? Hopefully it's good enough for user consumption? Can we instead pass that error to the user instead of duplicating this validation?

I'm wary of duplicate validations as they can become an additional source of errors. If we were going to add some additional functionality here, like pausing reconciliation until the SSH key is created, then that would be a different matter. But I think we're just catching an error which OpenStack should already catch and reporting it differently, right?

mdbooth:
> What error message does Nova give us on creation if this is not valid? Hopefully it's good enough for user consumption? Can we instead pass that error to the user instead of duplicating this validation?
> 
> I'm wary of duplicate validations as they can become an additional source of errors. If we were going to add some additional functionality here, like pausing reconciliation until the SSH key is created, then that would be a different matter. But I think we're just catching an error which OpenStack should already catch and reporting it differently, right?

I just tested this and nova says: `Invalid key_name provided.`. How would the UX be if the user saw an error in the machine: `Failed to create OpenStack server: Invalid key_name provided.`?

PR #1195, opened 2022-04-05
Remove unnecessary instanceStatus.State() call

Bo0km4n:
To remove unnecessary instanceStatus.State() call

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -339,13 +339,13 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 	addresses := instanceNS.Addresses()
 	openStackMachine.Status.Addresses = addresses
 
-	switch instanceStatus.State() {
+	switch state {
 	case infrav1.InstanceStateActive:
 		scope.Logger.Info("Machine instance is ACTIVE", "instance-id", instanceStatus.ID())
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
 		// Error is unexpected, thus we report error and never retry
-		handleUpdateMachineError(scope.Logger, openStackMachine, errors.Errorf("OpenStack instance state %q is unexpected", instanceStatus.State()))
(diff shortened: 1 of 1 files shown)


tobiasgiese:
I don't think that this change is needed nor correct.

We're just using the variable `state` as a pointer helper for `openStackMachine.Status.InstanceState`.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6ba04de45920c102886bdeeeb21bf1a1119c5967/controllers/openstackmachine_controller.go#L330-L331

Further, calling `instanceStatus.State()` just returns the string of the instance state.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6ba04de45920c102886bdeeeb21bf1a1119c5967/pkg/cloud/services/compute/instance_types.go#L91-L93

mdbooth:
It's a toss up, tbh. However, I think pulling state into a variable is less frequent than just calling it, so on balance I'm inclined to say calling State() is more idiomatic. I don't think either is 'wrong', though.

However, note that State() is *extremely* cheap. In fact godbolt suggests there's a good chance (compiler optimisations permitting) it's free here: https://godbolt.org/z/PbYE68Wof (note how neither switch statement generates code).

There may be an improvement we could make here, though. I said that we don't normally pull State() into a variable, and we wouldn't have to do it here if we did:

```golang
openStackMachine.Status.InstanceState = pointer.String(instanceStatus.State())
```

Is that easier on the eye? I don't think there's much in it.

What brought you to the code today, btw?

Bo0km4n:
I apprecieate your reviews.
If you guys don't think it's necessary, it doesn't have to be merged.
The reason I submitted this PR this time is simply because I thought it was a useless code.
I was not thinking of any compiler-level optimizations.

tobiasgiese:
I'll close this PR.
@Bo0km4n if you want to re-discuss this topic, feel free to reopen it 🙂

PR #1199, opened 2022-04-13
🏃 Add tests for cluster controller

lentzi90:

This adds some simple tests for the OpenStackCluster controller.

2. Note that there are 3 commit. I can make separate PRs for them or remove some of them if needed.
3. [This file](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1199/files#diff-9da8f7e78e11f549d7ff45cd4706d6946b664816033d9a29a518017b98a06a7a) was copied almost verbatim from CAPA. We could alternatively import the relevant package. What do you think would be better?
4. Please note the pending test. I'm not sure how to move forward with this and would appreciate any hints. If there are no obvious solutions I think maybe we should have an issue to discuss this further?

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller_test.go
+++ b/controllers/openstackcluster_controller_test.go
@@ -1,0 +1,224 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/utils/openstack/clientconfig"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/patch"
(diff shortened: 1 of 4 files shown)


controllers/openstackcluster_controller_test.go
```diff
@@ -0,0 +1,224 @@
...
+		req := createRequestFromOSCluster(testCluster)
+
+		result, err := reconciler.Reconcile(ctx, req)
+		// Expect error for getting OS clinet and empty result
+		Expect(err).ToNot(BeNil())
+		Expect(result).To(Equal(reconcile.Result{}))
+	})
+
+	// TODO: This test is set to pending (PIt instead of It) since it is not working.
+	PIt("should be able to reconcile when basition disabled", func() {
```
lentzi90, line 173: Please take a look at this test specifically. I'm not sure how to move forward here. The basic tests above does not require any interaction with the Openstack API, but this one does and I didn't find a good way to fake it. Am I missing something? Is it possible to get a fake/mock scope or dummy openstack API to use in tests like these?

controllers/suite_test.go
```diff
@@ -57,6 +59,11 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("..", "config", "crd", "bases"),
 		},
+		// Add fake CAPI CRDs that we reference
+		CRDs: []*apiextensionsv1.CustomResourceDefinition{
+			external.TestClusterCRD.DeepCopy(),
```
apricote, line 64: What is the benefit of faking the CRDs instead of including the actual ones? (I have not really worked with envtest before)

lentzi90: I think the main benefit is that we don't have to include the CAPI CRDs either in the repo or automatically download them to be able to run the tests.

apricote: Ah right, because this expects `CustomResourceDefinition` objects, but we currently only import the API types (from `sigs.k8s.io/cluster-api/api/*`.

That makes sense. Thank you :)

lentzi90:
I have split this up into two now. The first part is https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1239 and this is the second. I'll rebase this once the first is merged.

/hold

lentzi90:
@mdbooth this is now rebased and ready to go in I think. The only thing I'm wondering is if I should remove the disabled test for now or keep it there as a TODO. What do you think?

mdbooth:
> @mdbooth this is now rebased and ready to go in I think. The only thing I'm wondering is if I should remove the disabled test for now or keep it there as a TODO. What do you think?

I don't think it's important. It's a little messy but it's documented why it's messy and we have a plan to fix it up soon.

PR #1213, opened 2022-04-20
🏃Remove prankul88 as a reviewer

mdbooth:
@prankul88, if you're planning to stay around I'm very happy to leave you in! If not, this should stop spam from reviews being automatically assigned to you.

/hold


jichenjc:
I haven't see @prankul88 for a while, so I agree we can remove him and if he want to rejoin we are more than happy to do that

/lgtm
/hold cancel

Issue #1222, opened 2022-04-27
rootVolume is defined in instance.spec but Instance got created from Image


bavarianbidi:
/kind bug

**What steps did you take and what happened:**
We upgraded a self-managed management cluster from CAPO v0.5.3 to a pre-release build of v0.6.0 based on the main branch. To work around #1220, we created a new `OpenStackMachineTemplate` with following spec:

```yaml
spec:
  template:
    spec:
      cloudName: openstack
      flavor: n1.medium
      identityRef:
        kind: Secret
        name: openstack-cloud-config
      image: ubuntu-2004-kube-v1.22.5
      rootVolume:
        diskSize: 50
```

**What did you expect to happen:**

A new `OpenStackMachine` which get created from a volume (instead of an Image).

**Anything else you would like to add:**

Servers got created and joined the Kubernetes Cluster, but they aren't got created from a volume.

```
openstack server list | grep -E "ID|\-\-|galaxy"
+--------------------------------------+--------------------------------------+--------+---------------------------------------------------------------------------+--------------------------+-----------+
| ID                                   | Name                                 | Status | Networks                                                                  | Image                    | Flavor    |
+--------------------------------------+--------------------------------------+--------+---------------------------------------------------------------------------+--------------------------+-----------+
| 3273adb9-c52b-454e-98d0-232681719017 | galaxy-galaxy-mario-vc2fh            | ACTIVE | k8s-clusterapi-cluster-org-giantswarm-galaxy=10.6.0.226                   | ubuntu-2004-kube-v1.22.5 | n1.medium |
| 41cb7217-90d5-4643-8f9c-1119c92f2d79 | galaxy-galaxy-mario-xkbqx            | ACTIVE | k8s-clusterapi-cluster-org-giantswarm-galaxy=10.6.0.239                   | ubuntu-2004-kube-v1.22.5 | n1.medium |
| bb736526-f275-4f7d-9043-8292395a9265 | galaxy-control-plane-mario-pjrrk     | ACTIVE | k8s-clusterapi-cluster-org-giantswarm-galaxy=10.6.0.59                    | ubuntu-2004-kube-v1.22.5 | n1.medium |
[...]

PR #1225, opened 2022-04-27
🐛Don't set ImageRef on server when booting from volume

mdbooth:
We were setting image ID on the server when booting from volume. While this appears to work correctly, it doesn't appear to be the intended use of the Nova API.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1222

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -188,20 +188,6 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


apricote:
/lgtm

This now matches the process with a pre-created volume describe here: https://docs.openstack.org/nova/latest/user/launch-instance-from-volume.html#create-volume-from-image-and-boot-instance

apricote:
/lgtm

This now matches the process with a pre-created volume describe here: https://docs.openstack.org/nova/latest/user/launch-instance-from-volume.html#create-volume-from-image-and-boot-instance

PR #1236, opened 2022-05-06
Envtest mocks

mdbooth:

The purpose of this PR is to be able to run envtests using mock clients instead of real ones. This PR achieves that by enhancing Scope to return service clients. We add a new scope Factory interface to generate Scopes. We provide 2 implementations of Factory:

ProviderScopeFactory generates 'real' Scopes which expect to be able to make API calls to a running OpenStack cloud.
MockScopeFactory uses mock clients.

Because Scopes now return clients, the scope package must import those clients. However, clients are currently in their service packages. Services use scopes and must therefore import the scope package. This is a circular dependency. To avoid this we move clients into a separate package.

With the exception of the new factory interface, all other changes are mechanical. This PR does not intentionally introduce any functional changes.

Because of the large amount of code churn I have split the change into multiple commits. These commits each make a single refactor. I suggest that this PR will be simpler to review by commit, as the mechanical code churn in each individual commit is very similar.

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -47,7 +47,6 @@ import (
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/loadbalancer"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking"
-	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/provider"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/scope"
 )
 
@@ -112,7 +111,7 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}()
 
-	osProviderClient, clientOpts, projectID, err := provider.NewClientFromCluster(ctx, r.Client, openStackCluster, r.CaCertificates)
+	osProviderClient, clientOpts, projectID, err := scope.NewClientFromCluster(ctx, r.Client, openStackCluster, r.CaCertificates)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -59,6 +59,7 @@ type OpenStackClusterReconciler struct {
 	Client           client.Client
 	Recorder         record.EventRecorder
 	WatchFilterValue string
+	ScopeFactory     scope.Factory
 	CaCertificates   []byte // PEM encoded ca certificates.
 }
 
@@ -111,16 +112,9 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}()
 
-	osProviderClient, clientOpts, projectID, err := scope.NewClientFromCluster(ctx, r.Client, openStackCluster, r.CaCertificates)
+	scope, err := r.ScopeFactory.NewClientScopeFromCluster(ctx, r.Client, openStackCluster, r.CaCertificates, log)
 	if err != nil {
 		return reconcile.Result{}, err
-	}
-
-	scope := &scope.Scope{
-		ProviderClient:     osProviderClient,
-		ProviderClientOpts: clientOpts,
-		ProjectID:          projectID,
-		Logger:             log,
 	}
 
 	// Handle deleted clusters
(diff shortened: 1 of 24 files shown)


apricote:
Looks good to me! (adding lgtm after a rebase)

We discussed removing the `clientOpts` and passing the region through the `Scope` struct in #1178. At least now it is an implementation detail of the `ProviderScopeFactory`.

pkg/cloud/services/networking/trunk_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha6"
-	"sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking/mock_networking"
+	mock_clients "sigs.k8s.io/cluster-api-provider-openstack/pkg/clients/mock"
```
seanschneeweiss, line 27: Probably it is okay to not define any alias here - we could just use `mock.`
What do you think?
mock_clients is also fine.

controllers/openstackcluster_controller_test.go
```diff
@@ -160,9 +170,12 @@ var _ = Describe("OpenStackCluster controller", func() {
 		Expect(err).To(BeNil())
 		req := createRequestFromOSCluster(testCluster)
 
+		clientCreateErr := fmt.Errorf("Test failure")
+		mockScopeFactory.SetClientScopeCreateError(clientCreateErr)
+
 		result, err := reconciler.Reconcile(ctx, req)
 		// Expect error for getting OS clinet and empty result
```
seanschneeweiss, line 181:
```diff
- // Expect error for getting OS clinet and empty result
+ // Expect error for getting OS client and empty result
```

seanschneeweiss (commented):
The separate commits made it very nice to review. Looks very good.

pkg/cloud/services/networking/trunk_test.go
mdbooth: You're right, and this would be cleaner. I'll want to do it consistently, though. Let me try an update across the whole PR to see if 'mock' conflicts anywhere and if not I'll update it.

controllers/openstackcluster_controller_test.go
mdbooth: I'm inclined to leave this out of this patch as it's not strictly related. This PR has enough tentacles already!

stephenfin:
fwict, the CI failures are due to infra issues (quotas), not the change itself

pkg/cloud/services/networking/trunk_test.go
apricote: Looks like this request is still open @mdbooth

stephenfin: Good point. Done.

stephenfin:
This should be finally good to review again.

controllers/openstackcluster_controller.go
```diff
@@ -111,18 +111,11 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}()
 
-	osProviderClient, clientOpts, projectID, err := provider.NewClientFromCluster(ctx, r.Client, openStackCluster)
+	scope, err := r.ScopeFactory.NewClientScopeFromCluster(ctx, r.Client, openStackCluster, log)
```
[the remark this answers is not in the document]
seanschneeweiss: Made this comment a long time ago but didn't submit the review.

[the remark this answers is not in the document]
mdbooth: I just implemented this to play with it, but I don't think so.

My thinking is that in the controllers we retrieve a logger from controller runtime from the context, then we add context to it with a succession of `log = log.WithValues(...)`. The contract of this call is:

```
// WithValues returns a new Logger instance with additional key/value pairs.
```

I don't think we can assume that retrieving the logger again from the context would retrieve the logger customised for the current reconcile.

I'm prepared to be wrong on this, though. Either way, could we come back to it in a separate PR? It should be a fairly easy mechanical change if we decide to do this later.

lentzi90:
> Because of the large amount of code churn I have split the change into multiple commits. These commits each make a single refactor. I suggest that this PR will be simpler to review by commit, as the mechanical code churn in each individual commit is very similar.

Thank you! It was quite nice to review per commit! :slightly_smiling_face: 
This looks like a good improvement to me.
/lgtm

mdbooth:
Fixed merge conflict with #1451

lentzi90 (commented):
/lgtm
Let's get this merged!

PR #1246, opened 2022-05-20
Passing security groups by specifying more options in addition to UUIDs on ports

### LINKED ISSUE #1245: Specify security groups by name at Port level
As a user, I would like to specify security groups by name instance creation time.

As we have it now, it is possible to specify security groups at `port level `using uuids. However, the instance level setting is more flexible in that it allows specifying security groups using other parameters as shown below. 

```
DESCRIPTION:
FIELDS:
   filter       <Object>
   name <string>
     Security Group name
   uuid <string>
     Security Group UID
```
It would be useful if this feature is also added at port level. 
This can be archived by adding a new field in order to support backward compatibly. A sample template is shown below. 

```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha5
metadata:
  name: basic-1-md-0
  namespace: default
spec:
  template:
    spec:
      securityGroups:
      - name: capo-instance-level-sg
      ports:
      - description: "Port 1 with security group"
        securityGroups: # old field, left for backward compatiblity
        - "2f6584db-5138-453b-b47b-99696cb11f0f" 
        securityGroupsFilters: # the new field
        - name: "capo-port-level-sg" # Security Group name
      cloudName: openstack-1
      flavor: 2C-4GB-100GB
```

/kind feature

Xenwar:
With `OpenStackMachineTemplate` it is possible to specify security groups at both instance and port levels. However, both levels do not have same format, with the instance level having more options. This PR moves the port level format closer to that of the instance level, as shown below.

Specifying security groups at port level:

```
spec:
  template:
    spec:
      ports:
      - description: "Port 1 with security group in UUID format"
        securityGroups: 
        - "2f6584db-5138-453b-b47b-99696cb11f0f" 
```

Specifying security groups at port level:

```
spec:
  template:
    spec:
        securityGroups:
        - name: "capo-port-instance-level-sg"
           uuid: ....
           filter: .....
```

As can be seen above, the instance level setting has more options. This PR the `securityGroupsFilters` field. The original field `port.securityGroups` is kept for backward compatibility. However, the old field is redundant  due to the `securityGroupsFilters` already containing the `UUID` field. Therefore, removing it is more preferable. 

Fixes #1245

### DIFF
--- a/api/v1alpha4/conversion_test.go
+++ b/api/v1alpha4/conversion_test.go
@@ -244,6 +244,7 @@ func TestFuzzyConversion(t *testing.T) {
 						v1alpha5PortOpts.Network = nil
 					}
 				}
+				v1alpha5PortOpts.SecurityGroupFilters = nil
 			},
 			func(v1alpha5FixedIP *infrav1.FixedIP, c fuzz.Continue) {
 				c.FuzzNoCustom(v1alpha5FixedIP)
--- a/hack/ci/cloud-init/controller.yaml.tpl
+++ b/hack/ci/cloud-init/controller.yaml.tpl
@@ -152,7 +152,10 @@
 
     # Adjust the CPU quota
     openstack quota set --cores 32 demo
-    openstack quota set --secgroups 50 demo
+    openstack quota set --secgroups 200 demo
+    openstack quota set --secgroup-rules 1000 demo
+    openstack quota set --secgroups 100 admin
+    openstack quota set --secgroup-rules 1000 admin
 - path: /root/devstack.sh
   permissions: 0755
   content: |
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -89,11 +89,12 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 				MACAddress: ap.MACAddress,
 			})
 		}
-
-		securityGroups = portOpts.SecurityGroups
-
+		securityGroups, err = s.CollectPortSecurityGroups(eventObject, portOpts.SecurityGroups, portOpts.SecurityGroupFilters)
+		if err != nil {
+			return nil, err
+		}
 		// inherit port security groups from the instance if not explicitly specified
-		if securityGroups == nil {
+		if securityGroups == nil || len(*securityGroups) == 0 {
 			securityGroups = instanceSecurityGroups
 		}
 	}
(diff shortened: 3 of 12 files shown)


pkg/cloud/services/networking/port.go
```diff
@@ -91,9 +91,22 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
...
+		if securityGroupFilters != nil {
+			securityGroupFiltersByID, err := s.GetSecurityGroups(securityGroupFilters)
+			if err != nil {
+				return nil, fmt.Errorf("error getting security groups: %v", err)
+			}
+			for _, sg := range securityGroupFiltersByID {
+				if sg == "" {
+					continue
+				}
+				*securityGroups = append(*securityGroups, sg)
```
apricote, line 105: How will this handle duplicate security groups? E.g. if the same uuid is specified in `portOpts.SecurityGroups` and `portOpts.SecurityGroupFilters`

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -153,6 +153,8 @@
     # Adjust the CPU quota
     openstack quota set --cores 32 demo
     openstack quota set --secgroups 50 demo
+    openstack quota set --secgroups 100 demo
```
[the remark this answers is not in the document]
Xenwar: thanks, will do that after the current test.

pkg/cloud/services/networking/port.go
Xenwar: Tried duplicates, and it does fail, no check on that.
Thanks, will add the logic to remove the duplicates.

pkg/cloud/services/networking/port.go
```diff
@@ -89,11 +89,12 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 				MACAddress: ap.MACAddress,
 			})
 		}
-
-		securityGroups = portOpts.SecurityGroups
-
+		securityGroups, err = s.CollectPortSecurityGroups(eventObject, portOpts.SecurityGroups, portOpts.SecurityGroupFilters)
+		if err != nil {
+			return nil, fmt.Errorf("multiple ports found with name \"%s\"", portName)
```
jichenjc, line 94: should we only log the err itself here?

Xenwar: The error message is wrong, will remove the text. 
However, we need to return the error as it tells the absence of the security groups specified in the manifests.

pkg/cloud/services/networking/port.go
Xenwar: Done: now duplicates are removed
1. securityGroups field
2. securityGroupFilter field
3. A combination of the two.

hack/ci/cloud-init/controller.yaml.tpl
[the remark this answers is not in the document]
Xenwar: done

api/v1alpha4/conversion_test.go
```diff
@@ -244,6 +244,7 @@ func TestFuzzyConversion(t *testing.T) {
 						v1alpha5PortOpts.Network = nil
 					}
 				}
+				v1alpha5PortOpts.SecurityGroupFilters = nil
```
apricote, line 247: @bavarianbidi is introducing `utilconversion.(Un)MarshalData` from Cluster API in #1247 to avoid losing data on a round-trip conversion to v1alpha4 and back to alpha5.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1247/files#diff-9a53c6afb893cfd6bb4f50185dea035075cb0ef25fce713667c270d5c9668a31R41

I think it makes sense to also apply this here, though I am not sure about the best process for this, some options:

- wait until #1247 is merged and then rebase ontop
- extract the round-trip setup into a seperate PR and rebase this ontop
- continue as is and do this in a follow up PR once #1247  is merged

pkg/cloud/services/networking/port.go
```diff
@@ -257,3 +258,37 @@ func (s *Service) GarbageCollectErrorInstancesPort(eventObject runtime.Object, i
...
+		allSecurityGroupIDs = append(allSecurityGroupIDs, *portSecurityGroups...)
+	}
+	// generate unique values
+	uids := make(map[string]int)
+	for _, sg := range allSecurityGroupIDs {
+		if sg == "" {
+			continue
+		}
+		uids[sg] = 1
+		securityGroupCount++
```
apricote, line 283: `securityGroupCount` is initialized to the number of SGs returned from the filter (line 271). Here you increment it by one for every single security group returned by either the filters, or the explicit id list. In effect, you are double counting the security groups from the filter.

This can be fixed by initializing `securityGroupCount := 0` in line 271.

This does not introduce any impacting bugs, only the slice for `distinctSecurityGroupIDs` will have a backing array that is too large.

Xenwar: True, but even with `securityGroupCount := 0` at 271, we will still have extra space consumed. 
The reason is `allSecurityGroupIDs` could still contain duplicates. 

I will add an extra loop to count distinct values.

api/v1alpha4/conversion_test.go
Xenwar: I think doing it in a follow up PR could be good as I am planning the deprecate the `securityGroups` field. The reason being, the uuid it provides is also provided by the new introduced `securityGroupFilter` field.

pkg/cloud/services/networking/port.go
Xenwar: Added `securityGroupCount := 0 `at 271, change the incremental to conditional, thanks.

Xenwar:
/hold cancel
deprecating the `port.securityGroup` and comment https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1246#discussion_r881309276, will addressed on a separated PR.

api/v1alpha4/conversion_test.go
bavarianbidi: I'm also fine doing a separate PR as it seems that all conversation unit tests has to be refactored. Already though about but didn't had much time to start the discussion.

PR #1247, opened 2022-05-23
✨ Feature: restrict API Server LB access via IPs

### LINKED ISSUE #1045: Introduce an easy way to limit access to k8s api when ManagedAPIServerLoadBalancer: true
/kind feature

Hello,

I find having CAPO managing the ApiServerLB a very good feature that simplifies a lot of stuff in our case. That being said the k8s api is open from everywhere which is not ideal in our case. 
Currently the only way to have the ApiServerLB with limited access is to create it yourself, which means also creating network, subnet, router, pool etc etc which is a bit of extra work and adds an additional step in creation and deletion of clusters.
I'd like to have a way to limit access to the ApiServer without having to build it myself. I'm not sure if this is something you think that CAPO should support but im mentioning some potential approaches below. 

Possible approaches:
1. User can define a list of strings with allowed CIDRs on OpenStackCluster. CAPO creates an SG with these CIDRs and attaches it to the LB vip port after creation. This requires the least work from user perspective but adds quite extra complexity on CAPO.
2. User can define a list of strings with securityGroup IDs on OpenStackCluster. CAPO associates these SGs to the LB vip port after creation. A bit more work for the user, a bit less for CAPO.  
3. User can define LBPortID on OpenstackCluster. CAPO creates an LB that uses that port as the vip. This is possible only for octavia (not sure if LBaaS is supported instead of octavia). This adds quite some work for the user as he needs to create network,subnet, router, a port and his own SGs on it but is much easier for CAPO.
4. User can define a list with allowed CIDRs on OpenStackCluster. CAPO adds the [allowedCIDRs](https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/listeners/requests.go#L166) to the Listeners it creates on the LB. Only possible for [octavia > 2.12](https://github.com/openstack/octavia/blob/81134bdfec79c8019c26e7846d31a2b2952965a0/octavia/api/root_controller.py#L103) ( I think its Stein but i'll have to doublecheck).

- Let me know if this is something that should be added and if so which approach should be taken.
- I'd like to work on this one

bavarianbidi:
Access to API-Server can be restricted by setting allowedCidrs on `spec.APIServerLoadBalancer.AllowedCIDRs`.

## Behavior:

1. outgoing NAT Router-IPs is now part of the status: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9830f57ee854128f1f238f9142ea4f0dcc834c00/pkg/cloud/services/networking/router.go#L73-L83
1. once `spec.APIServerLoadBalancer.AllowedCIDRs` changed (create/update), CAPO will fetch all known required IP addresses from spec and status and will generate a new list of allowed CIDRs
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9830f57ee854128f1f238f9142ea4f0dcc834c00/pkg/cloud/services/loadbalancer/loadbalancer.go#L193-L207
1. after `allowed_cidrs` on OpenStack are updated on the corresponding listener, the list of applied CIDRs will be written back to the status https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9830f57ee854128f1f238f9142ea4f0dcc834c00/pkg/cloud/services/loadbalancer/loadbalancer.go#L113-L119

## Conclusion:

* only one additional request towards openstack - whenever the on OpenSTack applied CIDRs differes from the spec
* `spec.APIServerLoadBalancer.AllowedCIDRs` is now mutable

## Open questions:

1. **IP/CIDR validation during cluster-spec editing ?**
current implementation doesn't validate IPs during creation/editing of cluster-spec. IP validation will be done afterwards and non-valid IPs will get removed. 
Is this ok or should we already validate the CIDRs in the webhook?

Fixes #1045

With https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9830f57ee854128f1f238f9142ea4f0dcc834c00/api/v1alpha4/conversion.go#L36-L40 and https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9830f57ee854128f1f238f9142ea4f0dcc834c00/api/v1alpha4/conversion.go#L62-L65 we will now get the `cluster.x-k8s.io/conversion-data` Annotation if we query CAPO-related resources in a previous CAPO-Version.

This is currently missing but needed in general.
E.g. compare the output of `k get cluster.v1alpha4.cluster.x-k8s.io` and `k get cluster.v1beta1.cluster.x-k8s.io` where conversion is fine versus CAPO - e.g. `k get osc.v1alpha4.infrastructure.cluster.x-k8s.io` where we don't have the `cluster.x-k8s.io/conversion-data` annotation yet. 

xref: [SIG Cluster Lifecycle - ClusterAPI - API conversion code walkthrough](https://www.youtube.com/watch?v=Mk14N4SelNk)

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

/hold


bavarianbidi:
Open questions:

**IP/CIDR validation during cluster-spec editing**
current implementation doesn't validate IPs during creation/editing of cluster-spec. IP validation will be done afterwards and non-valid IPs will get removed

bavarianbidi:
> @bavarianbidi: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-cluster-api-provider-openstack-test	[12abd16](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/12abd16389db90a4313947939ac9f238cd346b02)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1247/pull-cluster-api-provider-openstack-test/1530130323419959296)	true	`/test pull-cluster-api-provider-openstack-test`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=cluster-api-provider-openstack&pr=1247). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Abavarianbidi). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues?q=is:issue+is:open) when you hit one in your PR.

wondering why this test doesn't fail on my local machine :shrug: will fix it on monday

jichenjc:
looks like this is Fuzzy test failure :) so need check the Convert function failure 
so local test might pass because of lucky ?

bavarianbidi:
> looks like this is Fuzzy test failure :) so need check the Convert function failure so local test might pass because of lucky ?

had to rebase again, as `openstackclustertemplate` conversion was added 5 days ago

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -175,6 +186,70 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
...
+		}
+
+		record.Eventf(openStackCluster, "SuccessfulUpdateListener", "Updated allowed_cidrs %s for listener %s with id %s", listener.AllowedCIDRs, listener.Name, listener.ID)
+
+		return nil
+	}
+	return nil
+}
+
+// validateIPs validate given IPs/CIDRs and remove non valid network objects.
```
seanschneeweiss, line 238:
```suggestion
// validateIPs validates given IPs/CIDRs and removes non valid network objects.
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -175,6 +186,70 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
...
+func validateIPs(definedCIDRs []string) []string {
+	marshaledCIDRs := []string{}
+
+	for _, v := range definedCIDRs {
+		if net.IsIPv4String(v) {
+			marshaledCIDRs = append(marshaledCIDRs, v+"/32")
+		} else if net.IsIPv4CIDRString(v) {
+			marshaledCIDRs = append(marshaledCIDRs, v)
+		}
```
seanschneeweiss, line 247: If one entry doesn't match the type ipv4 nor cidr then it will just be ignored without any warning. Should we catch cases where the string doesn't match?

pkg/utils/strings/strings.go
```diff
@@ -0,0 +1,29 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strings
+
+func Unique(s []string) []string {
```
seanschneeweiss, line 19: We could also use [github.com/mpvl/unique](https://github.com/mpvl/unique) but should be totally fine to just create the small function here.

seanschneeweiss (commented):
Very nice. Thanks a lot for the documentation too. This is a great feature.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -177,6 +178,57 @@ during an upgrade. When the API server has a floating IP, this role is fulfilled
...
+  apiServerLoadBalancer:
+    allowedCidrs:
+    - 192.168.10/24
+    - 10.10.0.0/16
+```
+
+All known IPs of the target cluster will be discovered dynamically (e.g. you don't have to take care of target Cluster own Router IP, internal CIDRs or any Bastion Host IP).
+**Note**: Please ensure, that at least the outgoing IP of your management Cluster is added to the list of allowed CIDRs. Otherwise CAPO can't reconcile the target Cluster correctly.
+
+All applied CIDRs (user defined + dynamically discovered) are written back into `status.network.apiServerLoadBalancer`
```
jichenjc, line 203: better add  `allowedCIDRs`

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -177,6 +178,57 @@ during an upgrade. When the API server has a floating IP, this role is fulfilled
...
+spec:
+  allowAllInClusterTraffic: true
+  apiServerLoadBalancer:
+    allowedCidrs:
+    - 192.168.10/24
+    - 10.10.0.0/16
+```
+
+All known IPs of the target cluster will be discovered dynamically (e.g. you don't have to take care of target Cluster own Router IP, internal CIDRs or any Bastion Host IP).
+**Note**: Please ensure, that at least the outgoing IP of your management Cluster is added to the list of allowed CIDRs. Otherwise CAPO can't reconcile the target Cluster correctly.
```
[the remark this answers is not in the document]
jichenjc: or I saw the code seems the IP is added in the code directly?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -175,6 +186,70 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
...
+
+		if err := s.waitForListener(listener.ID, "ACTIVE"); err != nil {
+			record.Warnf(openStackCluster, "FailedUpdateListener", "Failed to update listener %s with id %s: wait for listener active: %v", listener.Name, listener.ID, err)
+			return err
+		}
+
+		record.Eventf(openStackCluster, "SuccessfulUpdateListener", "Updated allowed_cidrs %s for listener %s with id %s", listener.AllowedCIDRs, listener.Name, listener.ID)
+
+		return nil
```
jichenjc, line 233: this is useless

pkg/utils/strings/strings.go
bavarianbidi: after this PR got merged, my idea is create a small PR to make use of this function on other places in the code-base where it might make sense (e.g. https://github.com/giantswarm/cluster-api-provider-openstack/blob/e4be0f25a1e8182ef41da77422b9826f2c713de1/controllers/openstackmachine_controller.go#L452-L463).

If wanted i can refactor to `mpvl/unique` as well.

docs/book/src/clusteropenstack/configuration.md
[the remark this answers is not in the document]
bavarianbidi: but how do we know the router IP of the management cluster? 
There might be so many different CAPO-deployment models (i guess) where it's not that easy to determine the outgoing IP of the management cluster (some of my thoughts where i don't had an easy answer to myself :-) ):
* Management Cluster might run somewhere else - how to determine the IP
* OpenstackCredentials during cluster provisioning are mostly scoped to target openstack tenant:
   even if the management cluster is self-managed by the same controller, how to tell in a generic way during reconciliation what's the `openstackcluster` object of the management cluster (by setting an annotation to the `openstackcluster` object of the management cluster?)
   
   for sure, it might be possible to implement something, my idea was to not implement to much "magic" for now and keep the feature as simple as possible. But i aggree with you, without setting the proper managmentcluster router ip, CAPO users will feel the pain. But before we start implementing this as feature, i guess we should discuss how the management-cluster-discovery could look like. If i can share/initiate the discussion.

pkg/cloud/services/loadbalancer/loadbalancer.go
bavarianbidi: if we stay on the current implementation (no input validation during Create/Update/Delete on `spec.APIServerLoadBalancer.AllowedCIDRs`) we should at least log if a `string` got dropped due invalid ipv4 or cidr.

But yes, i'm open for alternative ideas (but as we don't have any existing input validation i decided to go that way).

docs/book/src/clusteropenstack/configuration.md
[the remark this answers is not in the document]
jichenjc: yes, so we can follow  that pattern and implement here, then doc the requirements as well :)

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -177,6 +178,57 @@ during an upgrade. When the API server has a floating IP, this role is fulfilled
...
+        - 10.10.0.0/16 # user defined
+        - 192.168.10/24 # user defined
+        - 172.16.111.100/32 # bastion host floating IP
+        - 172.16.111.85/32 # router IP
+      internalIP: 10.6.0.144
+      ip: 172.16.111.159
+      name: k8s-clusterapi-cluster-<cluster-namespace>-<cluster-name>
+```
+
+If you looked out yourself or the CAPO management cluster, you can easily clear the `allowed_cidrs` field on OpenStack via
```
apricote, line 226:
```suggestion
If you locked out yourself or the CAPO management cluster, you can easily clear the `allowed_cidrs` field on OpenStack via
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -175,6 +186,68 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
...
+
+		if len(openStackCluster.Status.Network.Router.IPs) > 0 {
+			allowedCIDRs = append(allowedCIDRs, openStackCluster.Status.Network.Router.IPs...)
+		}
+	}
+
+	// validate CIDRs and convert any given IP into a CIDR
+	allowedCIDRs = validateIPs(allowedCIDRs)
+
+	// sort and remove duplicates
```
apricote, line 211: Is this actually sorting the slice?

```suggestion
	// remove duplicates
```

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -177,6 +178,57 @@ during an upgrade. When the API server has a floating IP, this role is fulfilled
 floating IP even if there is no load balancer. When the API server does not have a floating
 IP, the load balancer virtual IP on the cluster network is used.
 
+## Restrict Access to the API server
```
apricote, line 181: Please add the new heading to the Table of Content at the beginning of the file

pkg/utils/openstack/loadbalancer.go
```diff
@@ -0,0 +1,81 @@
...
+// IsOctaviaFeatureSupported returns true if the given feature is supported in the deployed Octavia version.
+// copied from https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L95-L148
+func IsOctaviaFeatureSupported(octaviaVer string, feature int, lbProvider string) bool {
+	currentVer, _ := version.NewVersion(octaviaVer)
+
+	switch feature {
+	case OctaviaFeatureVIPACL:
+		if lbProvider == lbProviderOVN {
+			return false
+		}
```
tobiasgiese, line 42: We can set the `lbProviderOVN` validation right before the switch-case to fail early (even if the compiler might do this anyway).

tobiasgiese (commented):
Thanks for this great contribution! 👌🏻

pkg/utils/openstack/loadbalancer.go
bavarianbidi: From the copied feature-matrix from https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L95-L148 tags are also supported on OVN

```go
	case OctaviaFeatureTags:
		verTags, _ := version.NewVersion("v2.5")
		if currentVer.GreaterThanOrEqual(verTags) {
			return true
		}
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -98,18 +130,26 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
 	openStackCluster.Status.Network.APIServerLoadBalancer = &infrav1.LoadBalancer{
-		Name:       lb.Name,
-		ID:         lb.ID,
-		InternalIP: lb.VipAddress,
-		IP:         lbFloatingIP,
+		Name:         lb.Name,
+		ID:           lb.ID,
+		InternalIP:   lb.VipAddress,
+		IP:           lbFloatingIP,
+		AllowedCIDRs: allowedCIDRs,
```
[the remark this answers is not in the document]
bavarianbidi: As `AllowedCIDRs` in the LoadBalancer status struct is configured with `omitempty`, k8s won't show the empty field:

```go
// LoadBalancer represents basic information about the associated OpenStack LoadBalancer.
type LoadBalancer struct {
	Name       string `json:"name"`
	ID         string `json:"id"`
	IP         string `json:"ip"`
	InternalIP string `json:"internalIP"`
	//+optional
	AllowedCIDRs []string `json:"allowedCIDRs,omitempty"`
}
```

pkg/utils/openstack/loadbalancer.go
tobiasgiese: Sure, totally missed this 👍🏻

PR #1255, opened 2022-06-05
bugfix: delete port left over

ruifaling:
This is an abnormal scenario: the virtual machine creation fails, but the port creation succeeds. In this case, the port will be left when deleting the cluster, because this port cannot be deleted according to the virtual machine. In this case, we can query the relevant port according to the subnet to delete it
Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


erkanerol:
@ruifaling is this bug fixed in another PR?

ruifaling:
> @ruifaling is this bug fixed in another PR?

I have commited more files,so I will create a new PR

ruifaling:
> @ruifaling is this bug fixed in another PR?

The new PR no is https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1260

PR #1263, opened 2022-06-15
🐛 Make failure domain optional for OpenStackMachine

mkjpryor:

The failure domain is an optional field for CAPI `Machine` objects. This patch makes sure that we can respect the case where a failure domain is not specified - currently this results in a failed machine.

This is implemented by omitting the availability zone from the server create if no failure domain is present on the CAPI machine. In this case, Nova will pick a suitable availability zone for the machine based on all other scheduling constraints.

This change was previously part of PR #1252.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -421,10 +421,6 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
 		return nil, fmt.Errorf("create Options need be specified to create instace")
 	}
 
-	if machine.Spec.FailureDomain == nil {
-		return nil, fmt.Errorf("failure domain not set")
-	}
-
 	instanceSpec := compute.InstanceSpec{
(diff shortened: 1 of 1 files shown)


mkjpryor:
@mdbooth 

As discussed in #1252, I have split out the changes to allow us to accept machines without a failure domain into a separate PR, which we consider a bug (as we are not respecting all of the CAPI interface).

Can I get an ok-to-test on this?

mkjpryor:
@mdbooth 

Looks like the tests were OK. Do we need anything else to merge this?

ruifaling:
which az will the vm to be launched if there was no failure domain?

mkjpryor:
@ruifaling Nova will pick one based on all the other scheduling constraints, falling back to the default if you have specified one.

Issue #1265, opened 2022-06-17
Machine controller permanently moves Floating IPs over controllers


zioc:
/kind bug

I was attemtping to use CAPI/CAPO on an openstack platform that doesn't provides loadbalancer API, and expected to use floatingIP 
to reach the API server. 

For that purpose I was using following parameters in OpenstackCluster:
  disableAPIServerFloatingIP: false
  apiServerLoadBalancer: {}

It turns out that openstackmachine controller is permanently moving floating IP overs the master nodes, causing API connectivity interruptions. Here is an extract from the capo controller logs: 
```
ubuntu@capo:~$ kubectl logs  -n capo-system capo-controller-manager-57dc99755b-xrpch |grep 'Associated floating'    
I0616 13:26:10.233715       1 recorder.go:103] events "msg"="Normal"  "message"="Associated floating IP 172.20.129.6 with port f2939e1b-e9c2-4e02-b1e9-da69087dd66e" "object"={"kind":"OpenStackCluster","namespace":"default","name":"management-cluster","uid":"26a135e7-f825-4e5a-8dd6-9c6fe6050fed","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha4","resourceVersion":"1460"} "reason"="SuccessfulAssociateFloatingIP"
I0616 13:27:41.963167       1 recorder.go:103] events "msg"="Normal"  "message"="Associated floating IP 172.20.129.6 with port 1c81e094-8470-472a-bb7f-2c0337cedc7d" "object"={"kind":"OpenStackCluster","namespace":"default","name":"management-cluster","uid":"26a135e7-f825-4e5a-8dd6-9c6fe6050fed","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha4","resourceVersion":"1460"} "reason"="SuccessfulAssociateFloatingIP"
I0616 13:28:00.712943       1 recorder.go:103] events "msg"="Normal"  "message"="Associated floating IP 172.20.129.6 with port f2939e1b-e9c2-4e02-b1e9-da69087dd66e" "object"={"kind":"OpenStackCluster","namespace":"default","name":"management-cluster","uid":"26a135e7-f825-4e5a-8dd6-9c6fe6050fed","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha4","resourceVersion":"1460"} "reason"="SuccessfulAssociateFloatingIP"
[...]

PR #1268, opened 2022-06-20
nit:make event to log for already associated FIP

jichenjc:

in reviewing https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1265
found we use event for the already associated FIP 
not remember whether this is valid as current understanding seems this should be log only as it might happen many times

```
# kubectl logs capo-controller-manager-5fcf678668-bf4sw -n capo-system | grep 'already associated with port' | wc
      7     112    3101
```

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


seanschneeweiss (commented):
Agree that the event doesn't make much sense here.
/lgtm

PR #1271, opened 2022-06-21
🌱 Generate cluster templates with kustomize

mdbooth:

This PR changes the way we generate our cluster templates for both release and e2e tests to use kustomize. We now have the following dependency structure between templates:

```mermaid
graph TB
  subgraph Release
    release/without-lb([Without LoadBalancer]) --> release/default([Default])
    release/external-cloud-provider([External Cloud Provider]) --> release/default
  end

  subgraph E2E Tests
    test/default([Default]) --> release/default
    test/external-cloud-provider([External Cloud Provider]) --> release/external-cloud-provider
    test/without-lb([Without LoadBalancer]) --> release/without-lb
    test/multi-az([Multi AZ]) --> test/default
    test/multi-network([Multi Network]) --> test/default
  end
```

Additionally, all the E2E templates now use a common set of patches to uniformly apply configuration we use in tests but not in release. This means that any change in the release templates is now reflected automatically in the E2E tests, which gives us a level of test coverage for changes to the release templates.

That said, in order to keep this PR simple the templates produced by this PR are *identical* to the previous templates except in the order of the objects and fields. The order difference is an artifact of using kustomize: it picks a consistent order according to its own rules. I have verified that the templates are identical by the following:

For every existing template in /templates and /test/e2e/data/infrastructure-openstack I generated a kustomization.yaml containing only a resources section specifying only the original template file, and then I built the kustomization. i.e. This is the original template whose only transformation is the re-ordering done by kustomize. I then generated templates from the new kustomize and verified that there were no diffs.

Release templates are generated by a new make target: `templates`.
E2E templates are generated by a new make target: `e2e-templates`.

I have created directories for v1alpha5 and v1alpha6, which are currently identical except for apiVersion, and we should consider adding a directory for v1alpha4 in the future as long as we support it. The intention here is to for reuse when adding tests for supported previous versions.

I have used strategic merge patches in the release templates because I find them more readable. However, in the e2e templates I have used a json patch everywhere we reference an OpenStack object. This allows the patches to be reusable across apiVersions.

/hold


Makefile
```diff
@@ -469,3 +497,6 @@ verify-gen: generate
 .PHONY: compile-e2e
 compile-e2e: ## Test e2e compilation
 	go test -c -o /dev/null -tags=e2e ./test/e2e/suites/conformance
+
+.PHONY: FORCE
+FORCE:
```
apricote, line 502: What is the purpose of this target?

Makefile
```diff
@@ -378,8 +395,19 @@ release-alias-tag: # Adds the tag to the last build tag.
 release-notes: $(RELEASE_NOTES) ## Generate release notes
 	$(RELEASE_NOTES) $(RELEASE_NOTES_ARGS)
 
+.PHONY: templates
+templates: templates/cluster-template.yaml \
+	   templates/cluster-template-without-lb.yaml \
+	   templates/cluster-template-external-cloud-provider.yaml
+
+templates/cluster-template.yaml: kustomize/v1alpha6/default FORCE
```
apricote, line 403: This fails for me with:

```
$ make templates
hack/tools/bin/kustomize build "kustomize/v1alpha6/default" > "templates/cluster-template.yaml"
bash: line 1: hack/tools/bin/kustomize: No such file or directory
make: *** [Makefile:404: templates/cluster-template.yaml] Error 127
make: *** Deleting file 'templates/cluster-template.yaml'
```

I would suggest we add the `$(KUSTOMIZE)` target as a dependency, so the binary is always available

```suggestion
templates/cluster-template.yaml: kustomize/v1alpha6/default FORCE $(KUSTOMIZE)
```

The same applies to targets `templates/cluster-template-%.yaml`, `$(E2E_TEMPLATES_DIR)/cluster-template.yaml` and `$(E2E_TEMPLATES_DIR)/cluster-template-%.yaml`

mdbooth: I've updated with this change.

Makefile
```diff
@@ -138,12 +138,29 @@ endif
 test: $(SETUP_ENVTEST) ## Run tests
 	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" go test -v ./... $(TEST_ARGS)
 
+E2E_TEMPLATES_DIR=test/e2e/data/infrastructure-openstack
+E2E_KUSTOMIZE_DIR=test/e2e/data/kustomize
+
+.PHONY: e2e-templates
+e2e-templates: $(addprefix $(E2E_TEMPLATES_DIR)/, \
```
tobiasgiese, line 145:
```suggestion
e2e-templates: $(addprefix $(E2E_TEMPLATES_DIR)/, \ ## Generate e2e templates
```

To have a proper usage info

Makefile
```diff
@@ -378,8 +395,19 @@ release-alias-tag: # Adds the tag to the last build tag.
 release-notes: $(RELEASE_NOTES) ## Generate release notes
 	$(RELEASE_NOTES) $(RELEASE_NOTES_ARGS)
 
+.PHONY: templates
+templates: templates/cluster-template.yaml \
```
tobiasgiese, line 399:
```suggestion
templates: templates/cluster-template.yaml \ ## Generate templates
```

same here

tobiasgiese (approved):
Also validated the changes via

```
git clone https://github.com/kubernetes-sigs/cluster-api-provider-openstack/ /tmp/cluster-api-provider-openstack
for template in templates/*yaml; do diff -u <(sort $template) <(sort /tmp/cluster-api-provider-openstack/$template); done
```

Noticed only some quoting and list syntax changes

Edit: btw, there is also a yaml differ called [dyff](https://github.com/homeport/dyff). But as there are multiple yamls inside the manifest dyff is not able to do its proper job 😿

tobiasgiese:
> Edit: btw, there is also a yaml differ called [dyff](https://github.com/homeport/dyff). But as there are multiple yamls inside the manifest dyff is not able to do its proper job

But there is [ksort](https://github.com/superbrothers/ksort/).
Combined I couldn't find any changes

```
$ for template in templates/*yaml; do dyff between --omit-header <(ksort -f $template) <(ksort -f /tmp/cluster-api-provider-openstack/$template); done
$
```

mdbooth:
> > Edit: btw, there is also a yaml differ called [dyff](https://github.com/homeport/dyff). But as there are multiple yamls inside the manifest dyff is not able to do its proper job
> 
> But there is [ksort](https://github.com/superbrothers/ksort/). Combined I couldn't find any changes
> 
> ```
> $ for template in templates/*yaml; do dyff between --omit-header <(ksort -f $template) <(ksort -f /tmp/cluster-api-provider-openstack/$template); done
> $
> ```

I did something like this against an unmodified repo to create a new directory (k) containing original templates which have been sorted by kustomize:

```bash
mkdir k
for i in cluster-template*.yaml; do
  echo -e "resources:\n- $i" > kustomization.yaml
  kustomize build . > k/$i
end
```

Then I diffed the new generated templates against the kustomize-sorted original templates. They are literally identical.

PR #1273, opened 2022-06-22
🌱 Add renovate configuration

### LINKED ISSUE #1093: A better way to sync with other community?
/kind feature

inspired by #1092 
we need update according to cluster-api, openstack from various places
not sure whether other community has such kind of sync mechanism ? 
as from process is much better than randomly check...

so hope to get some ideas from folks..

lentzi90:

This is to automatically bump versions through pull requests.

Fixes #1093 

2. Merging this PR is not enough to enable renovate. Someone with privileges to add GitHub apps also needs to follow the [onboarding steps](https://docs.renovatebot.com/getting-started/installing-onboarding/). The best way to do this is probably to *start* with the onboarding, then copy the configuration from this PR to the renovate onboarding PR and merge that instead
3. **I have tested the configuration in this PR in my fork. You can see the pull requests that renovate has created [here](https://github.com/lentzi90/cluster-api-provider-openstack/pulls).** Please check these to see if renovate is doing things the way we want.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,55 @@
+{
+  "extends":[
+    "config:base"
+  ],
+  "ignorePaths": [
+    "**/node_modules/**",
+    "**/bower_components/**",
+    "**/vendor/**",
+    "**/examples/**"
+  ],
+  "enabled": "true",
+  "baseBranches": ["main"],
+  "packageRules": [
+    {
+      "matchPackageNames": ["golang"],
+      "allowedVersions": "<=1.17"
+    },
+    {
+      "matchPackageNames": ["kubernetes/kubernetes"],
+      "allowedVersions": "<=1.23"
+    },
+    {
+      "extends": ["group:kubernetes"],
+      "allowedVersions": "<=0.23"
+    },
+    {
+      "matchPackageNames": ["sigs.k8s.io/cluster-api", "sigs.k8s.io/cluster-api/test"],
+      "groupName": "Cluster API"
+    }
+  ],
+  "prHourlyLimit": 5,
+  "prConcurrentLimit": 5,
+  "rangeStrategy": "bump",
+  "renovateFork": true,
+  "stabilityDays": 5,
+  "regexManagers": [
+    {
+      "fileMatch": ["^Makefile$"],
+      "matchStrings": [
+        "KUBEBUILDER_ENVTEST_KUBERNETES_VERSION .= (?<currentValue>.*?)\\n"
+      ],
+      "depNameTemplate": "kubernetes/kubernetes",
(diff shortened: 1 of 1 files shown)


apricote:
Looking at the created PRs in your fork I noticed that Renovate created an update PR "Update kubernetes packages to v0.24.2" ([link](https://github.com/lentzi90/cluster-api-provider-openstack/pull/3)). I think this should be kept in sync with `kubernetes/kubernetes`. Is there a way to configure this, or does this just not work because they use different major versions?

lentzi90:
Good point @apricote ! I can add another `packageRule` to keep it at 0.23. I don't think it can understand that it should be the same as the kubernetes/kubernetes version but not problem to have multiple package rules

renovate.json
```diff
@@ -0,0 +1,47 @@
...
+  ],
+  "ignorePaths": [
+    "**/node_modules/**",
+    "**/bower_components/**",
+    "**/vendor/**",
+    "**/examples/**"
+  ],
+  "enabled": "true",
+  "baseBranches": ["main"],
+  "packageRules": [
```
apricote, line 13: In your fork there are currently two PRs open for `cluster-api` and `cluster-api/test`. As those are always released together, we can create a group for them, so that one PR can update both:

```
"packageRules": [
    {
      "matchPackageNames": ["sigs.k8s.io/cluster-api", "sigs.k8s.io/cluster-api/test"],
      "groupName": "Cluster API"
    }
  ]
```

lentzi90:
Updated with CAPI package group and kubernetes packages version limit.

jichenjc:
@mdbooth maybe you want to take another look on this?  (so we can unhold)

/approve

renovate.json
```diff
@@ -0,0 +1,55 @@
+{
+  "extends":[
+    "config:base"
+  ],
+  "ignorePaths": [
+    "**/node_modules/**",
```
seanschneeweiss, line 6: Those paths do not exist in this repo. Should we therefore omit them and re-add when required?

lentzi90: Sorry I should have explained this actually but forgot about it. These are the defaults from `config:base` but with `**/test/**` removed since we do have code in that folder that I think we want to include. I'm open to reducing this list even more but didn't see any harm in keeping it.
Do you think we should remove it all? Or are there some folders we want to ignore?

seanschneeweiss: I'd agree to remove them all and add exclusions when we notice that they are helpful.

PR #1279, opened 2022-06-27
Add Age column for osc/osm

bavarianbidi:
Add the age column to openstackcluster and openstackmachine.

As openstackclustertemplate and opensackmachinetemplate don't have any custom `printcolumn` defined, per default only name of the object and Age will be printed - so we're already fine.

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha6/openstackcluster_types.go
+++ b/api/v1alpha6/openstackcluster_types.go
@@ -215,6 +215,7 @@ type OpenStackClusterStatus struct {
 // +kubebuilder:printcolumn:name="Subnet",type="string",JSONPath=".status.network.subnet.id",description="Subnet the cluster is using"
 // +kubebuilder:printcolumn:name="Endpoint",type="string",JSONPath=".spec.controlPlaneEndpoint.host",description="API Endpoint",priority=1
 // +kubebuilder:printcolumn:name="Bastion IP",type="string",JSONPath=".status.bastion.floatingIP",description="Bastion address for breakglass access"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of OpenStackCluster"
 
 // OpenStackCluster is the Schema for the openstackclusters API.
 type OpenStackCluster struct {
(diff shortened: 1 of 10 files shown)


api/v1alpha6/openstackcluster_types.go
```diff
@@ -215,6 +215,7 @@ type OpenStackClusterStatus struct {
 // +kubebuilder:printcolumn:name="Subnet",type="string",JSONPath=".status.network.subnet.id",description="Subnet the cluster is using"
 // +kubebuilder:printcolumn:name="Endpoint",type="string",JSONPath=".spec.controlPlaneEndpoint.host",description="API Endpoint",priority=1
 // +kubebuilder:printcolumn:name="Bastion IP",type="string",JSONPath=".status.bastion.floatingIP",description="Bastion address for breakglass access"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of OpenStackCluster"
```
jichenjc, line 218: seems this is not `Time duration since creation of OpenStackCluster`
it's the time stamp of the creation?

bavarianbidi: > seems this is not `Time duration since creation of OpenStackCluster` it's the time stamp of the creation?

if `type="date"` is set, (i guess) kubectl will parse this and calculate the Age of the object itself (xref: example from kubebuilder book- https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns)

tobiasgiese: I think this is just fine as it‘s the same text as in CAPI

https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta1/cluster_types.go

jichenjc: ok, learned something new , thanks for the info~

tobiasgiese: @bavarianbidi could you add this also for openstackclustertemplate and openstackmachinetemplate? Just to be consistent.

bavarianbidi: exactly, as soon as custom columns are defined, the default (object name + age) will be overwritten.

PR #1287, opened 2022-06-30
Fix generation of e2e-templates when running conformance tests

mdbooth:
Since https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1271 the e2e templates are now generated, but test-conformance-fast was not generating them.

**TODOs**:

- [x] Remove test-only change which runs conformance tests with e2e tests

/hold


tobiasgiese:
/lgtm

thanks!

PR #1257, opened 2022-06-06
Deprecating and/or replacing securityGroups field from OpenStackMachineTemplate

Xenwar:
Deprecating and replacing `PortOpts.SecurityGroups` of `[]string` format field from `OpenStackMachineTemplate` ports and replace is by `[]SecurityGroupParam` format.

fixes: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1251

PR #1291, opened 2022-07-04
⚠️ Remove PortOpts.SecurityGroups

### LINKED ISSUE #1251: Deprecating/Replacing securityGroups field from OpenStackMachineTemplate ports
/kind feature

With the PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1246, we have introduced a new field
`OpenStackMachineTemplate.spec.template.spec.ports.securityGroupsFilters.`

Before that, there was only one such field, OpenStackMachineTemplate` .spec.template.spec.ports.securityGroups` which is a lit of UUIDs.

I would like to deprecate the older field for two reasons. 

1. It is restrictive in that allows only specifying UUIDs
2. It is not the same as what is at intense level, shown below. 

```
DESCRIPTION:
FIELDS:
   filter       <Object>
   name <string>
     Security Group name
   uuid <string>
     Security Group UID
```

In favour of uniformity, I intend to 

- Remove `ports.SecurityGroups`
- Rename the new field `ports.securityGroupsFilters` to `ports.securityGroups`

I would like to know if this breaks existing deployments and if it is ok to start coding and If there are any side effect this may introduce.

lentzi90:
This replaces https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1257.

Deprecating and replacing PortOpts.SecurityGroups of *[]string format field from OpenStackMachineTemplate ports and replace is with *[]SecurityGroupParam format. The PortOpts.SecurityGroupFilters of type []SecurityGroupParam is removed.

Fixes #1251 

~~I opened a new PR instead of rebasing the existing since there are some issues and I'm unsure how to solve them. This way it will be easier to compare the rebased and original code side by side.~~

1. ~~We need to provide a conversion function for the `PortOpts.SecurityGroups` field that changes from `*[]string` in v1alpha5 to `*[]SecurityGroupParam` in v1alpha6.~~
2. ~~The same conversion is needed from v1alpha4 to v1alpha6 and this is where I'm having issues. The required name of the conversion function becomes the same for both v1alpha4 and v1alpha5 because it goes from `*[]string` which doesn't have a version.~~
   ```
   Convert_Slice_string_To_Slice_v1alpha6_SecurityGroupParam
   ```
   ~~If I add this conversion function to both v1alpha4 and v1alpha5, the generator complains that there is a duplicate.~~
3. ~~If I instead add the conversion in only one place (e.g. v1alpha5), then the generator will be happy and *import the conversion function in the other API version as well*! But the compiler does not like this at all. It gets an import error.~~
   ```
   could not import ./api/v1alpha4 (no required module provides package "./api/v1alpha4")
   ```

~~This wasn't a problem in the original PR since it only needed the conversion from v1alpha4 -> v1alpha5 (so no duplicate function name or import needed).
If you have any suggestions on how to solve this, please let me know.~~

Edit: I think I figured it out! By adding the conversion function in v1alpha6 instead of v1alpha5 and/or v1alpha4, I avoid the issue.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold


api/v1alpha5/zz_generated.conversion.go
```diff
@@ -24,6 +24,7 @@ package v1alpha5
 import (
 	unsafe "unsafe"
 
+	v1alpha4 "./api/v1alpha4"
```
lentzi90, line 27: One way to make the compiler happy would be to change this in the following way:
```suggestion
	v1alpha4 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha4"
```
Any clue how I can make code-gen understand this?

api/v1alpha6/types.go
```diff
@@ -117,11 +117,9 @@ type PortOpts struct {
 	FixedIPs  []FixedIP `json:"fixedIPs,omitempty"`
 	TenantID  string    `json:"tenantId,omitempty"`
 	ProjectID string    `json:"projectId,omitempty"`
-	// The uuids of the security groups to assign to the instance
-	SecurityGroups *[]string `json:"securityGroups,omitempty"`
-	// The names, uuids, filters or any combination these of the security groups to assign to the instance
-	SecurityGroupFilters []SecurityGroupParam `json:"securityGroupFilters,omitempty"`
-	AllowedAddressPairs  []AddressPair        `json:"allowedAddressPairs,omitempty"`
+	// The names of the security groups to assign to the port
+	SecurityGroups      *[]SecurityGroupParam `json:"securityGroups,omitempty"`
```
lentzi90, line 121: Please check this carefully.
I have copied it from https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1257 but I'm unsure if it should be `*[]SecurityGroupParam` or `[]SecurityGroupParam`. Note that the `SecurityGroupFilters` used to be `[]SecurityGroupParam` but now `SecurityGroups` is `*[]SecurityGroupParam`.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -393,6 +393,9 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
 func (s *Service) GetSecurityGroups(securityGroupParams []infrav1.SecurityGroupParam) ([]string, error) {
 	var sgIDs []string
 	for _, sg := range securityGroupParams {
+		if sg.UUID == "" {
+			continue
+		}
```
mdbooth, line 398: What's this doing? Doesn't this make it impossible to use the filter functionality?

mdbooth (changes requested):
This looks good to me with some caveats.
* We need to decide if it's useful to specify an explicitly empty list of SecurityGroups.
* Double check we didn't break selection by filter.

and a big one:

I would *really* like to see us test this with envtests rather than e2e. It's a perfect candidate.

pkg/cloud/services/networking/port.go
```diff
@@ -90,9 +90,15 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
...
+			portSecurityGroups, err := s.GetSecurityGroups(*portSecurityGroups)
+			if err != nil {
+				return nil, fmt.Errorf("error getting security groups: %v", err)
+			}
+			if len(portSecurityGroups) > 0 {
+				securityGroups = &portSecurityGroups
+			}
 		}
 		// inherit port security groups from the instance if not explicitly specified
 		if securityGroups == nil || len(*securityGroups) == 0 {
```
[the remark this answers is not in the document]
lentzi90: Not sure I understand. But I'll give it a go. So the `securityGroups == nil` is for checking if the portOps didn't have any security groups specified, right? Then the `len(*securityGroups) == 0` would mean that we check if the portOps had an explicit empty list of security groups?

If this is the case, then I agree that combining them is point less. And it makes the check do a different thing than the comment above implies. So the actual current behavior if to inherit the security groups from the instance, no matter if the portOpts had an empty list or nothing specified at all.
I would be happy to simplify this if we get consensus.

pkg/cloud/services/networking/securitygroups.go
lentzi90: This looks strange I agree. Will investigate.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -412,11 +413,7 @@ func (s *Service) GetSecurityGroups(securityGroupParams []infrav1.SecurityGroupP
 		if err != nil {
 			return nil, err
 		}
-
-		if len(SGList) == 0 {
-			return nil, fmt.Errorf("security group %s not found", sg.Name)
-		}
```
lentzi90, line 418: Note this change in behavior for GetSecurityGroups!
It used to return an error if none were found. Now it simply returns an empty slice.
If we want to keep the current behavior, we will need a way to tell the difference between "unable to get security groups" (e.g. network errors) and "these params doesn't match anything". Otherwise we have no way of knowing if we should create a new port or retry again later.

pkg/cloud/services/networking/securitygroups.go
lentzi90: This was definitely a mistake. Perhaps something left from debugging.

pkg/cloud/services/networking/port.go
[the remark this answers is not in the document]
lentzi90: A quick check revealed that there are no unit tests that check this at least. Could be something in the e2e tests of course but that seems unlikely.

lentzi90:
Replaced by #1516 which just drops `securityGroups` instead of renaming `securityGroupFilters` to `securityGroups` and dropping the old `securityGroups`.

PR #1288, opened 2022-06-30
✨ Conditions for OpenStackMachines

apricote:
**What this PR does / why we need it**:

Adds Conditions to the OpenStackMachineStatus.

See this example `clusterctl describe` output, that showcases the different conditions, and how they "bubble up" to the `Machine` resources:
screenshot-capo-machine-conditions

So far there are two conditions: 
- `InstanceReady` (related to the OpenStack Instance) and 
- `APIServerIngressReadyCondition` (for ControlPlane machines -> LoadBalancer Member or Floating IP attachment)

I set the severity based on:

- *Info*, if the reason is expected (missing bootstrap secret)
- *Warning*, if the condition is eventually retried (no call to `handleUpdateMachineError`)
- *Error* if `handleUpdateMachineError` was called and the machine is no longer retried/reconciled

The `Ready` condition is set to the aggregate of thhe two custom conditions, and the CAPI Machine resource "bubbles up" to Ready message, so any errors can be observed easily, without having to inspect the OpenStackMachine

**Which issue(s) this PR fixes**:
Fixes #832 
Replaces #972

**Open Tasks**:

- [ ] ~~Tests for OpenStackMachineController~~ Requires proper envtest mocks
- [x] Fix `APIServerIngressReadyCondition` on non-control-plane machines when using api loadbalancer (see #1294)

**Open Questions**:

*Should we add the conditions to all existing api versions (because its a backwards compatible change) or only to the most recent version (v1alpha6)?*

  - **Pro**: Makes conversion a no-op
  - **Pro**: Changes are backwards compatible
  - **Contra**: Changing old API versions feels wrong to me

/hold

PR #1294, opened 2022-07-11
🌱 only reconcile loadbalancer member if machine is control-plane

apricote:

IMO we should check whether it is required (machine == control-plane) before
calling `loadbalancer.Service#ReconcileLoadBalancerMember`. The service
should not be concerned with the type of machine it is called on.

This also fixes one of the open bugs in PR #1288.

-

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -359,32 +359,34 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 		return ctrl.Result{RequeueAfter: waitForInstanceBecomeActiveToReconcile}, nil
 	}
 
-	if openStackCluster.Spec.APIServerLoadBalancer.Enabled {
-		err = r.reconcileLoadBalancerMember(scope, openStackCluster, machine, openStackMachine, instanceNS, clusterName)
-		if err != nil {
-			handleUpdateMachineError(scope.Logger, openStackMachine, errors.Errorf("LoadBalancerMember cannot be reconciled: %v", err))
-			return ctrl.Result{}, nil
-		}
-	} else if util.IsControlPlaneMachine(machine) && !openStackCluster.Spec.DisableAPIServerFloatingIP {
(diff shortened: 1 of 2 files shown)


controllers/openstackmachine_controller.go
```diff
@@ -359,32 +359,34 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
-		if err != nil {
-			handleUpdateMachineError(scope.Logger, openStackMachine, errors.Errorf("Floating IP cannot be associated: %v", err))
-			return ctrl.Result{}, nil
+	if util.IsControlPlaneMachine(machine) {
+		if openStackCluster.Spec.APIServerLoadBalancer.Enabled {
+			err = r.reconcileLoadBalancerMember(scope, openStackCluster, machine, openStackMachine, instanceNS, clusterName)
+			if err != nil {
+				handleUpdateMachineError(scope.Logger, openStackMachine, errors.Errorf("LoadBalancerMember cannot be reconciled: %v", err))
+				return ctrl.Result{}, nil
+			}
```
tobiasgiese, line 368: Does this really makes any difference? `r.reconcileLoadBalancerMember` is a no-op as it returns early `nil` if it's no control plane machine.

tobiasgiese: Okay, I overlooked that we're calling the private method `reconcileLoadBalancerMember` and not `ReconcileLoadBalancerMember` the only difference is, that `loadbalancer.NewService()` could return an error.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/2eb85ef3b49a959cb56c56ec5399f3e5bbef8627/controllers/openstackmachine_controller.go#L500-L503

apricote: It does not make a huge difference right now, but I would like to set the condition `APIServerIngressReadyCondition` right after calling `r.reconcileLoadBalancerMember`, and that only makes sense if the machine actually was a control plane machine.

See #1288 for context of this change

PR #1303, opened 2022-07-19
✨ Add re-creation of bastion host on change

### LINKED ISSUE #1298: :sparkles: bastion should be mutable
/kind feature

**tl;dr:** changing values in `OpenStackCluster.Spec.Bastion` should be possible without doing the _disable_/_enable with new spec_ foo

We know the fact, that it's currently possible to e.g. change the image of the bastion host by disabling the host and afterwards re-enabling the host with the new spec again. 
But this circumstance doesn't really fit in a `git-ops` approach.

From a admin/user perspective It should be possible to change values in `OpenStackCluster.Spec.Bastion.Instance` + `OpenStackCluster.Spec.Bastion.AvailabilityZone` and rely on CAPO that the bastion host in the desired state got created.

From what i've seen so far:
* this is a feature which is not tracked in a separate issue (maybe this could used for that)
* that's something which is wanted but not implemented for some reasons
  * most of the discussion regarding the current implementation was done in #1070 and #1180 
  
    Matthew already proposed a possible solution in #1180 (at first glance i can't see anything bad about it):
    
    > Yes, I think I still stand by my older comment. If it doesn't work we should reflect that in the API. It's also user-hostile to pretend something works if it doesn't.
    > 
    > It would be best if it did work, of course. Do CAPA and CAPZ have a solution for this? There's always the solution of hashing the bastion spec and storing it in the status. If the status hash doesn't match the current hash you delete and recreate. Is there a neater solution?
  
    discussion in #1070 starts [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1070#discussion_r762952694)
    
    Matthew:
    
    > This is an interesting one. I note that in this unit test we're changing both the bastion spec and the enabled flag. This is fine: we will create a new Bastion with the new spec.
    > 
    > However, note we also permit changing the bastion spec without flipping enabled, but this won't work. It won't fail, but the bastion will not be updated until the enabled flag is flipped to false and the bastion is deleted.
    > 
    > What do we want to do about this? Options:
    > 
    > 1. Ignore it
    > 2. Document it
    > 3. Don't allow changing bastion spec if status.Bastion != nil
    > 4. Make it work
    > 
    > For now I feel like 2 + 3 would be user friendly and relatively simple.
    
    Tobias:
    
    > I totally agree that we should document this behavior. Regarding 3: do you think it's worth the effort to update our `reconcileBastion` to delete the bastion if sth has changed? Or is it enough to only allow changes if the enabled flag is false? Both are fine for me, I just think from the customer's perspective. WDYT?
    
    Matthew:
    
    > I feel like automatically deleting and recreating the bastion if something changed would be the most user friendly, but I should strongly caveat that with: I don't feel like I understand all the potential use cases here. Have we firmly set an expectation that we can blow the bastion away? Nobody's running anything on the bastion which is going to suffer data loss? I'd hope not.
    > 
    > I also think it would be fine to do 3 and do 4 later. I don't think they're incompatible: if the user writes orchestration to delete and recreate the bastion if they need to change it, that will continue to work even if we remove the requirement to do it. We're also being nice to them in the meantime by pointing out that it doesn't work.

tobiasgiese:
Signed-off-by: Tobias Giese <tobias.giese@mercedes-benz.com>

Thanks to @chrischdi for the discussion and hash annotation idea! 🙂

TODO:

- [x]  test bastion re-creation
- [x] add docs
- [x] ~~e2e test?~~ (we will add envtests after #1236 is merged)

Fixes #1298

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

/hold

<sub>Tobias Giese <tobias.giese@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -51,6 +51,10 @@ import (
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/scope"
 )
 
+const (
+	BastionInstanceHashAnnotation = "infrastructure.cluster.x-k8s.io/bastion-hash"
+)
+
 // OpenStackClusterReconciler reconciles a OpenStackCluster object.
 type OpenStackClusterReconciler struct {
 	Client           client.Client
@@ -242,6 +246,8 @@ func deleteBastion(scope *scope.Scope, cluster *clusterv1.Cluster, openStackClus
 	}
 	openStackCluster.Status.BastionSecurityGroup = nil
 
+	delete(openStackCluster.ObjectMeta.Annotations, BastionInstanceHashAnnotation)
+
 	return nil
 }
 
@@ -313,20 +319,35 @@ func reconcileBastion(scope *scope.Scope, cluster *clusterv1.Cluster, openStackC
 		return err
 	}
 
+	instanceSpec := bastionToInstanceSpec(openStackCluster, cluster.Name)
+	bastionHash, err := compute.HashInstanceSpec(instanceSpec)
+	if err != nil {
+		return errors.Wrap(err, "failed computing bastion hash from instance spec")
+	}
+
 	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, fmt.Sprintf("%s-bastion", cluster.Name))
 	if err != nil {
 		return err
 	}
 	if instanceStatus != nil {
-		bastion, err := instanceStatus.APIInstance(openStackCluster)
-		if err != nil {
+		if !bastionHashHasChanged(bastionHash, openStackCluster.ObjectMeta.Annotations) {
+			bastion, err := instanceStatus.APIInstance(openStackCluster)
+			if err != nil {
+				return err
+			}
+			// Add the current hash if no annotation is set.
+			if _, ok := openStackCluster.ObjectMeta.Annotations[BastionInstanceHashAnnotation]; !ok {
+				annotations.AddAnnotations(openStackCluster, map[string]string{BastionInstanceHashAnnotation: bastionHash})
+			}
+			openStackCluster.Status.Bastion = bastion
+			return nil
+		}
+
+		if err := deleteBastion(scope, cluster, openStackCluster); err != nil {
 			return err
 		}
-		openStackCluster.Status.Bastion = bastion
-		return nil
 	}
 
-	instanceSpec := bastionToInstanceSpec(openStackCluster, cluster.Name)
 	instanceStatus, err = computeService.CreateInstance(openStackCluster, openStackCluster, instanceSpec, cluster.Name)
 	if err != nil {
 		return errors.Errorf("failed to reconcile bastion: %v", err)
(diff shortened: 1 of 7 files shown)


controllers/openstackcluster_controller.go
```diff
@@ -313,20 +314,31 @@ func reconcileBastion(scope *scope.Scope, cluster *clusterv1.Cluster, openStackC
 		return err
 	}
 
+	instanceSpec := bastionToInstanceSpec(openStackCluster, cluster.Name)
+	bastionHash, err := compute.ComputeHashInstanceSpec(instanceSpec)
+	if err != nil {
+		return errors.Wrap(err, "failed computing bastion hash from instannce spec")
```
seanschneeweiss, line 324:
```suggestion
		return errors.Wrap(err, "compute bastion hash from instance spec")
```

controllers/openstackcluster_controller.go
```diff
@@ -360,6 +376,7 @@ func reconcileBastion(scope *scope.Scope, cluster *clusterv1.Cluster, openStackC
 	}
 	bastion.FloatingIP = fp.FloatingIP
 	openStackCluster.Status.Bastion = bastion
+	annotations.AddAnnotations(openStackCluster, map[string]string{compute.BastionInstanceHashAnnotation: bastionHash})
```
seanschneeweiss, line 379: Can this be moved to line 326? Right after computing the hash?

seanschneeweiss (commented):
Lovely implementation. Thanks.
/lgtm

controllers/openstackcluster_controller.go
```diff
@@ -313,20 +319,35 @@ func reconcileBastion(scope *scope.Scope, cluster *clusterv1.Cluster, openStackC
...
 	}
 	if instanceStatus != nil {
-		bastion, err := instanceStatus.APIInstance(openStackCluster)
-		if err != nil {
+		if !bastionHashHasChanged(bastionHash, openStackCluster.ObjectMeta.Annotations) {
+			bastion, err := instanceStatus.APIInstance(openStackCluster)
+			if err != nil {
+				return err
+			}
+			// Add the current hash if no annotation is set.
```
jichenjc, line 338: just curious, we should never enter such 
unless admin update the openstack cluster annonation directly?

jichenjc (commented):
/lgtm

controllers/openstackcluster_controller.go
tobiasgiese: And during the update to this feature. Each enabled bastion host will get this annotation.
IMHO it's totally fine to re-add it each time someone removed the annotation.

tobiasgiese:
had to rebase @seanschneeweiss / @jichenjc

PR #1319, opened 2022-08-02
:bug: Add security groups to ports only

### LINKED ISSUE #1316: Do not set security groups on instances
/kind feature

Currently, the managed security groups are added to the instance, not to individual ports. This prevents clusters that have managed security groups from having one NIC with security group rules and one NIC without, as the instance security groups are applied to the second NIC even if the port itself specifies no security groups  

Nova itself also discourages use of this API - https://docs.openstack.org/nova/latest/contributor/project-scope.html#no-more-api-proxies. 

Instead, all security groups should be applied at the port level, with the managed groups being used when port security is enabled and no explicit groups are set for the port. 

mkjpryor:

This PR removes the used of a [deprecated API](https://docs.openstack.org/nova/latest/contributor/project-scope.html#no-more-api-proxies) by removing the setting of security groups directly on instances.

Fixes #1316 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -235,7 +235,6 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 		AvailabilityZone: instanceSpec.FailureDomain,
 		Networks:         portList,
 		UserData:         []byte(instanceSpec.UserData),
-		SecurityGroups:   securityGroups,
 		Tags:             instanceSpec.Tags,
 		Metadata:         instanceSpec.Metadata,
 		ConfigDrive:      &instanceSpec.ConfigDrive,
--- a/pkg/cloud/services/compute/instance_test.go
+++ b/pkg/cloud/services/compute/instance_test.go
@@ -582,9 +582,6 @@ func TestService_ReconcileInstance(t *testing.T) {
 				"imageRef":          imageUUID,
 				"flavorRef":         flavorUUID,
 				"availability_zone": failureDomain,
-				"security_groups": []map[string]interface{}{
-					{"name": workerSecurityGroupUUID},
-				},
 				"networks": []map[string]interface{}{
 					{"port": portUUID},
 				},


pkg/cloud/services/compute/instance.go
```diff
@@ -235,7 +235,6 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 		AvailabilityZone: instanceSpec.FailureDomain,
 		Networks:         portList,
 		UserData:         []byte(instanceSpec.UserData),
-		SecurityGroups:   securityGroups,
```
jichenjc, line 238: I assume you have follow up PR to separate the sec group into port level?

mkjpryor: @jichenjc 

All the security groups are actually already propagated to the ports! Prior to this patch they are just duplicated on the instance as well.

jichenjc: ah.. thanks for the info
yes, the port definition has recently been updated to add the sec group info already

but do we want to update `instanceSpec.SecurityGroups` since we don't honor 
it at all?

mkjpryor: @jichenjc 

`instanceSpec` is actually used to propagate the "instance default security groups" to the ports, as in the secgroups that should be used if the port does not specify its own (or disable port security).

See the following chain of usage:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6859acb599b3c358e682da790d2fe59174fb1290/controllers/openstackmachine_controller.go#L465
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6859acb599b3c358e682da790d2fe59174fb1290/pkg/cloud/services/compute/instance.go#L162
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6859acb599b3c358e682da790d2fe59174fb1290/pkg/cloud/services/compute/instance.go#L176
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6859acb599b3c358e682da790d2fe59174fb1290/pkg/cloud/services/networking/port.go#L97-L100

So the security groups in `instanceSpec` actually end up being propagated to the ports in the case where the ports don't set their own secgroups, which is what we want. If you think of `instanceSpec` as an overall container for what we want the instance to look like it makes sense. I would probably prefer to leave it as it is TBH.

mkjpryor:
@jichenjc 

It looks like the remaining test failure is due to lack of capacity in the test project. What do we do about that?

jichenjc:
looks like it's due to lack of resource of gcp cloud
someone already reported such issue recently ..not sure the progress though.

pkg/cloud/services/compute/instance.go
jichenjc: yes, I admin I didn't dig into such detail logic but I checked the code and I agree it's what you proposed way
so let port use sg if given otherwise from instanceSpec..

mkjpryor:
@jichenjc @apricote 

Is there anything else to do before this can be merged?

Issue #1274, opened 2022-06-23
Support different CNI plugin 


jichenjc:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

we use calico by default now , e.g the security group is set to calico now 
as we envolve, support multiple CNI is reasonable way to move forward

at least, I am thinking about add 
https://github.com/kubeovn/kube-ovn for now ...

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

mkjpryor:
@jichenjc 

Don't forget this flag that I added: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1alpha6/openstackcluster_types.go#L113

I added that so that I could use Cilium

jichenjc:
ok, great ~ Thanks for the reminder~

jichenjc:
the idea is to e.g make different yaml for rules like this 

e.g for default rule create default-rule.yaml
for all in, create all-in-rule.yaml
for calico , apply calico.yaml 

so we have the flexibility to update and modify 

```
infrav1.SecurityGroupRule{
                        Description:  "Kubernetes API",
                        Direction:    "ingress",
                        EtherType:    "IPv4",
                        PortRangeMin: 6443,
                        PortRangeMax: 6443,
                        Protocol:     "tcp",
                },
```

jichenjc:
#1323 seems not best solution as it introduce configuration into code 

the goal of this feature is to split the CNI settings from code to some configurations, either configmap or CRD (yaml)
- we need provide CNI settings in cluster definition 
```
     -  CNI:  calico
```
   if provided CNI is not recognizable, there is no no CNI setting at all ,it will be named `all`
- we create multiple pre-defined configmap (stored in json) for calico, clium (first) , the name has naming conventions like secgroup-cni-calico etc, for `all`, it's `secgroup-cni-all` means all traffic allowed
like this example https://stackoverflow.com/questions/61653284/does-kubernetes-take-json-format-as-input-file-to-create-configmap-and-secret
- when start, the cluster read the setting and load the config then honor it and create sec group 
- user is eligible to create additional security group and create config map like secgroup-cni-xxxx then it will be load during init process

jichenjc:
@mdbooth can you help review above to see whether you have any comments? Thanks

PR #1323, opened 2022-08-10
WIP: Add CNI selection for security groups

jichenjc:

part of #1274 

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


api/v1alpha6/openstackcluster_types.go
```diff
@@ -112,6 +112,13 @@ type OpenStackClusterSpec struct {
 	// +optional
 	AllowAllInClusterTraffic bool `json:"allowAllInClusterTraffic"`
 
+	// SecurityGroupForCNI is used when you deploy different CNI
+	// AllowAll means allow all traffic
+	// Calico creates security group rules for Calico CNI and it's the default
+	// not supported value will print a warning and no security group applied
+	// +optional
+	SecurityGroupForCNI string `json:"securityGroupForCNI"`
```
seanschneeweiss, line 120: Here we can set the name of the CNI to use, e.g. calico?
So actually the value represents the CNI selection to decide, which security group rules to apply.
I'd propose to name it "CNIForSecurityGroup" or simply "CNI".

What does "AllowAll" mean in this description - the description could be misleading, as "AllowAll" looks like it is a valid value.
```suggestion
	// CNIForSecurityGroup defines the CNI to set the appropriate security group rules.
	// The default CNI used is "calico" and creates security group rules for Calico CNI.
	// Unsupported values create warnings and no security groups are applied.
	// +optional
	CNIForSecurityGroup string `json:"cniForSecurityGroup"`
```

pkg/cloud/services/networking/securitygroups_rules.go
```diff
@@ -20,6 +20,11 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha6"
 )
 
+var (
+	CALICOCNI  = "calico"
+	DEFAULTCNI = CALICOCNI
```
seanschneeweiss, line 25: Not sure if the default should be defined here or better in the code within the switch case.
```suggestion
	CNICalico  = "calico"
```

pkg/cloud/services/networking/securitygroups_rules.go
```diff
@@ -283,16 +288,45 @@ func GetSGWorkerAllowAll(remoteGroupIDSelf, secControlPlaneGroupID string) []inf
 	}
 }
 
-func GetSGControlPlaneGeneral(remoteGroupIDSelf, secWorkerGroupID string) []infrav1.SecurityGroupRule {
+func getSGControlPlaneCNI(openStackCluster *infrav1.OpenStackCluster, remoteGroupIDSelf, secWorkerGroupID string) []infrav1.SecurityGroupRule {
+	cni := openStackCluster.Spec.SecurityGroupForCNI
+	if len(cni) == 0 {
+		cni = DEFAULTCNI
```
seanschneeweiss, line 294:
```suggestion
		cni = CNICalico
```

pkg/cloud/services/networking/securitygroups_rules.go
```diff
@@ -283,16 +288,45 @@ func GetSGWorkerAllowAll(remoteGroupIDSelf, secControlPlaneGroupID string) []inf
...
+	}
+
+	switch cni {
+	case CALICOCNI:
+		return getSGControlPlaneCalico(remoteGroupIDSelf, secWorkerGroupID)
+	default:
+		return []infrav1.SecurityGroupRule{}
```
seanschneeweiss, line 301: The CRD description describes that a warning is printed when the value is unknown.
I think this warning could be added here.

jichenjc:
@seanschneeweiss can you help check this again? thanks

mdbooth:
I'll be honest: I didn't actually realise we have this stuff in code. I don't think any of this should be in code: it's configuration. I don't think we should be doubling down on it by adding support for additional CNIs. I would prefer to remove it entirely and replace it with configuration. Maybe a CRD or a ConfigMap.

jichenjc:
um... this is from beginning , yes , I think it's reasonable to change to CRD if possible ..
and it might take some time :(

mdbooth:
Lets grandfather what's there, but not add to it. If we're not adding to it, I don't think we need to specify a CNI in the api. If we need to add to it (which it sounds like we do), lets create a new interface for that.

jichenjc:
>>>lets create a new interface for that.

yes, I will draft some design and then put into the original issue
after that ,we can discuss and see whether that's acceptable to everyone

and I agree with you, move to configmap or CRD is reasonable

pierreprinetti:
> @pierreprinetti I wonder if security groups are another candidate for a multi-machine treatment?

Good question. There may be some value in dynamically opening ports for NodePort for example, on a [group-of-]machine basis; but that may be a complex endeavour for little added value. All in all, I think the answer is no, I don't think I have heard of any practical use-case for per-[group-of]machine security group [rules].

Issue #1282, opened 2022-06-28
Cluster API v1.2


apricote:
Cluster API v1.2 [beta.0] and [beta.1] have already been released, the normal release is scheduled to happen within the next month. We should migrate CAPO to use CAPI v1.2(.0-beta.1), by following the [migration guide for providers](https://release-1-2.cluster-api.sigs.k8s.io/developer/providers/v1.1-to-v1.2.html).

I think it makes sense to split this into three (mostly independant) PRs:

- [x] Update dependencies and implement breaking changes: #1283
- [x] Bump to `rc.0` to fix our conformance tests: #1295 
- [x] Add Server Side Apply markers to our CRDs: #1325
- [x] Migrate logging to `component-base/logs`: #1296 
- [x] Bump to CAPI v1.2 (once released): #1302

/kind feature
/assign
/milestone v0.7
/area dependency

apricote:
All required migration steps are implemented.

/close

PR #1325, opened 2022-08-16
🌱 add merge strategy markers

apricote:

This commit adds merge strategy markers to our v1alpha6 API. This is required for Server-Side Apply (SSA) to work correctly. SSA is used in Cluster API v1.2, and not having these markers might cause issues when using ClusterClass patches.

This only adds a subset of markers, where it was possible. I skipped adding markers for two cases:

- We have many slices that do not have any proper map keys, as all fields are optional. This includes mostly "filter" structs, for example `NetworkParam` and `SecurityGroupParam`.
- The selected map key may not be optional (`omitempty`). If it is, it can not be used as the map key. This applied to structs `AddressPair` and `ExternalRouterIPParam`.

Further documentation:

- CAPI v1.2 provider migration: https://cluster-api.sigs.k8s.io/developer/providers/v1.1-to-v1.2.html#required-api-changes-for-providers
- SSA Markers: https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

Related to #1282

/hold

### DIFF
--- a/api/v1alpha6/openstackcluster_types.go
+++ b/api/v1alpha6/openstackcluster_types.go
@@ -48,6 +48,7 @@ type OpenStackClusterSpec struct {
 	// DNSNameservers is the list of nameservers for OpenStack Subnet being created.
 	// Set this value when you need create a new network/subnet while the access
 	// through DNS is required.
+	// +listType=set
 	DNSNameservers []string `json:"dnsNameservers,omitempty"`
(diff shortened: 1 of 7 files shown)


tobiasgiese:
/approve
/lgtm

Thanks for implementing the SSA 👍🏻 one step further towards CAPI v1.2

PR #1333, opened 2022-09-01
🌱 remove apricote from reviewers

apricote:
**What this PR does / why we need it**:

As I announced at the Office Hours a few weeks ago, tomorrow will be my last day at $CURRENT_JOB and I won't be involved with CAPO at $NEW_JOB. Because of this I step down from my reviewers position. Thanks for the great time :)

PR #1336, opened 2022-09-05
🌱 remove apricote from reviewers

apricote:

Manual cherry-pick of #1333 onto `release-0.6`


jichenjc:
I don't think we need a CI for this

PR #1340, opened 2022-09-09
🐛 fix nilpointer during clusterctl move

### LINKED ISSUE #1339: nil pointer if `osc.status` is empty (clusterctl move) and apiserver access is restricted
/kind bug

The current unreleased CAPO code contains a bug (introduced by myself :see_no_evil: ) which only occurs during `clusterctl move` and maybe during some `CR` restore procedure

If a cluster has restricted API access configured (introduced in PR #1247 ) and the `osc.status` is empty, the reconciliation might fail due a nil-pointer, caused by the following lines https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/loadbalancer/loadbalancer.go#L231-L237

No nil-pointer

Possible implementations:
1. check if `osc.status` is empty and not calling [`getOrUpdateAllowedCIDRS`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/loadbalancer/loadbalancer.go#L221)
2. ignore `openStackCluster.Status.Network.Subnet.CIDR` and `openStackCluster.Status.Network.Router.IPs` during the reconciliation with empty `status` (will lead to the following behavior: `routerIP` and `SubnetCIDR` is configured on the `allowedAddressPair` field in Openstack, will be removed during the `clusterctl move` and than added back again)

Option 1 seems to me as the best approach. As it doesn't change anything on Openstack side during the first reconciliation after `clusterctl move` was triggered. Once the `osc.status` got filled up again, it should be ok to update the `allowedAddressPair` field in Openstack.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `058f560669885cdec0487ce043bdb0408d559818`
- Cluster-API version: `v1.1.2`
- OpenStack version: 
- Minikube/KIND version:

bavarianbidi:

If .status.network is empty (e.g. during clusterctl move) the reconciliation will fail due a nil pointer.

This commit skips the first reconciliation of
openStackCluster.Spec.APIServerLoadBalancer.AllowedCIDRs after a clusterctl move.

Fixes #1339

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -134,10 +134,13 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 		}
 
 		if allowedCIDRsSupported {
-			if err := s.getOrUpdateAllowedCIDRS(openStackCluster, listener); err != nil {
-				return err
+			// Skip reconciliation if network status is nil (e.g. during clusterctl move)
+			if openStackCluster.Status.Network != nil {
+				if err := s.getOrUpdateAllowedCIDRS(openStackCluster, listener); err != nil {
+					return err
+				}
+				allowedCIDRs = listener.AllowedCIDRs
 			}
-			allowedCIDRs = listener.AllowedCIDRs
 		}
 	}
 


tobiasgiese:
/approve
/hold cancel

Thanks for the fix 👍🏻 🙂

PR #1346, opened 2022-09-24
:sparkles: Add Cilium as CNI Option

modzilla99:
Signed-off-by: modzilla99 <justin.lamp96@gmail.com>

With that PR you'll be able to deploy Cilium with SecurityGroups enabled.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/networking/securitygroups_rules.go
+++ b/pkg/cloud/services/networking/securitygroups_rules.go
@@ -114,6 +114,66 @@ func getSGControlPlaneCalico(remoteGroupIDSelf, secWorkerGroupID string) []infra
 	}
 }
 
+// Permit traffic for cilium.
+func getSGControlPlaneCilium(remoteGroupIDSelf, secWorkerGroupID string) []infrav1.SecurityGroupRule {
+	return []infrav1.SecurityGroupRule{
+		{
+			Description:   "HealthChecks (cilium)",
+			Direction:     "ingress",
+			EtherType:     "IPv4",
+			PortRangeMin:  4240,
+			PortRangeMax:  4240,
+			Protocol:      "tcp",
+			RemoteGroupID: remoteGroupIDSelf,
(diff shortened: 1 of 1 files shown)


jichenjc:
thanks for the PR

I am currently actually working on https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1323

so this might be follow up of that ?

modzilla99:
Yes that would indeed be a great follow up! Will wait for your pr to be merged.

modzilla99:
/remove-lifecycle rotten

@jichenjc hey there, sorry to bother or are there any news on different cnis?

wwentland:
My preference would be for users to be able to configure security groups in a generic way, rather than providing specific configuration for every CNI that could be supported.

The AWS provider defines a `CNIIngressRule` which could very well serve as inspiration here.

wwentland:
My preference would be for users to be able to configure security groups in a generic way, rather than providing specific configuration for every CNI that could be supported.

The AWS provider defines a `CNIIngressRule` which could very well serve as inspiration here for providing similar configuration.

jichenjc:
>>>My preference would be for users to be able to configure security groups in a generic way, rather than providing specific configuration for every CNI that could be supported.

yes, I Think it's similar to #1274 , provide a config map or define a CRD ,then predefine some well known CNI like calico or clium
then let's provide options for customer to define their own as well..

mdbooth:
Yes. Same as the last time this came up I remain opposed to merging this. We shouldn't be embedding this configuration in code. If we need to add support for another CNI, now is the time to add configuration support for that.

EmilienM:
You'll be covered by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1826

PR #1347, opened 2022-09-27
tolerate no volume endpoint

jichenjc:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/service.go
+++ b/pkg/cloud/services/compute/service.go
@@ -30,6 +30,7 @@ type Service struct {
 	scope             *scope.Scope
 	computeService    Client
 	networkingService *networking.Service
+	noVolumeService   bool
 }
 
 /*
@@ -63,8 +64,12 @@ func NewService(scope *scope.Scope) (*Service, error) {
 	volumeClient, err := openstack.NewBlockStorageV3(scope.ProviderClient, gophercloud.EndpointOpts{
 		Region: scope.ProviderClientOpts.RegionName,
 	})
+	noVolumeService := false
 	if err != nil {
-		return nil, fmt.Errorf("failed to create volume service client: %v", err)
+		// failed to create volume service client, we should tolerate
+		scope.Logger.Info("failed to create volume service client, no volume related operations allowed")
+		volumeClient = nil
+		noVolumeService = true
 	}
 
 	computeService := serviceClient{computeClient, imagesClient, volumeClient}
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/compute/service.go
```diff
@@ -63,8 +64,12 @@ func NewService(scope *scope.Scope) (*Service, error) {
 	volumeClient, err := openstack.NewBlockStorageV3(scope.ProviderClient, gophercloud.EndpointOpts{
 		Region: scope.ProviderClientOpts.RegionName,
 	})
+	noVolumeService := false
 	if err != nil {
-		return nil, fmt.Errorf("failed to create volume service client: %v", err)
+		// failed to create volume service client, we should tolerate
+		scope.Logger.Info("failed to create volume service client, no volume related operations allowed")
```
mdbooth, line 70: We should check what this error is, first. This would be very weird to debug if it just happened to fail with an ephemeral keystone error.

I don't have a system available without cinder. Can anybody confirm what this returns if there is no cinder?

pkg/cloud/services/compute/service.go
```diff
@@ -63,8 +64,12 @@ func NewService(scope *scope.Scope) (*Service, error) {
...
+		volumeClient = nil
```
mdbooth, line 71: This isn't strictly necessary as we already set it to nil when returning an error.

pkg/cloud/services/compute/service.go
```diff
@@ -63,8 +64,12 @@ func NewService(scope *scope.Scope) (*Service, error) {
...
+		noVolumeService = true
```
mdbooth, line 72: This isn't strictly necessary, as we can infer it from the nil volumeClient.

mdbooth (changes requested):
Lets get this fixed, like this if necessary, but nil pointers floating around give me the heebie jeebies. They're landmines. I wonder if we could do better than a nil pointer. How about if we changed this it to do lazy initialisation, then we wouldn't need to work out what non-existent cinder looks like. It would also be more efficient if we don't do any volume or glance operations 🤔 I might knock something up to see what it looks. like.

Incidentally, I lumped the volume client into the the compute service because of a blindspot: I've never encountered a cloud without cinder. That was a mistake, sorry.

mdbooth:
FYI I posted this as an alternative. It does lazy initialisation of all the clients: #1350

RakeshAMD:
Hi, any update on this issue? openstack without cinder.

jichenjc:
@RakeshAMD have you tried latest CAPO ?(from master)? 
#1350  from @mdbooth should handle that already

Issue #1351, opened 2022-10-03
E2E failure due to openstack cli python deps?


mdbooth:
/kind bug

**What steps did you take and what happened:**
This job failed: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1350/pull-cluster-api-provider-openstack-e2e-test/1576886172808908800/build-log.txt

The relevant trace from the build log looks like:

```
++ openstack versions show
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/stevedore/_cache.py", line 159, in _get_data_for_path
    with open(filename, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/python-entrypoints/aa1e52575b0b33cd4df0aed2af56d2b5dfa248c667d0055ed45042f5ce9d68b9'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/openstack", line 5, in <module>
    from openstackclient.shell import main
  File "/usr/local/lib/python3.7/dist-packages/openstackclient/shell.py", line 26, in <module>
    from openstackclient.common import clientmanager
  File "/usr/local/lib/python3.7/dist-packages/openstackclient/common/clientmanager.py", line 198, in <module>
    'openstack.cli.base',
  File "/usr/local/lib/python3.7/dist-packages/openstackclient/common/clientmanager.py", line 149, in get_plugin_modules
    mgr = stevedore.ExtensionManager(group)
  File "/usr/local/lib/python3.7/dist-packages/stevedore/extension.py", line 136, in __init__
    verify_requirements)
  File "/usr/local/lib/python3.7/dist-packages/stevedore/extension.py", line 218, in _load_plugins
    for ep in self.list_entry_points():
  File "/usr/local/lib/python3.7/dist-packages/stevedore/extension.py", line 207, in list_entry_points
    eps = list(_cache.get_group_all(self.namespace))
  File "/usr/local/lib/python3.7/dist-packages/stevedore/_cache.py", line 179, in get_group_all
    data = self._get_data_for_path(path)
  File "/usr/local/lib/python3.7/dist-packages/stevedore/_cache.py", line 162, in _get_data_for_path
    data = _build_cacheable_data(path)
  File "/usr/local/lib/python3.7/dist-packages/stevedore/_cache.py", line 110, in _build_cacheable_data
[...]

PR #1352, opened 2022-10-03
ci: Pin to the yoga version of openstackclient

mdbooth:
Works round an issue caused by an incompatible change to the importlib-metadata python library which breaks us on python 3.7.

Fixes: #1351 

/hold

### DIFF
--- a/hack/ci/create_devstack.sh
+++ b/hack/ci/create_devstack.sh
@@ -76,12 +76,32 @@ function retry {
 function ensure_openstack_client {
     if ! command -v openstack;
     then
-        apt-get install -y python3-dev
-        # install PyYAML first because otherwise we get an error because pip3 doesn't upgrade PyYAML to the correct version
-        # ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which
-        # files belong to it which would lead to only a partial uninstall.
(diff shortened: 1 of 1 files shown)


hack/ci/create_devstack.sh
```diff
@@ -76,12 +76,32 @@ function retry {
...
+        # the build constraints.
+        apt-get install -y python3-virtualenv curl
+        python3 -m virtualenv -p $(which python3) /tmp/openstack-venv
+        VIRTUAL_ENV_DISABLE_PROMPT=1 source /tmp/openstack-venv/bin/activate
+
+        # openstackclient has never actually supported python 3.7, only 3.6 and
+        # 3.8. Here we download the yoga constraints file and modify all the
+        # 3.8 constraints to be 3.7 constraints.
+        curl -L https://releases.openstack.org/constraints/upper/yoga -o /tmp/yoga-constraints
+        sed -i "s/python_version=='3.8'/python_version=='3.7'/" /tmp/yoga-constraints
```
stephenfin, line 99: A huge hack...but it should work :angel:

stephenfin (approved):

hack/ci/create_devstack.sh
mdbooth: 🤞

seanschneeweiss (approved):

PR #1353, opened 2022-10-04
🐛 Don't require cinder when not using volumes

### LINKED ISSUE #1345: Cluster creation fails when OpenStack does not have volume support
/kind bug

Running CAPO against an OpenStack cluster that does *not* have ceph, cinder, swift or any other persistent volume support configured. (This is intentional -- this is an OpenStack deployment designed for ephemeral workloads only).

Been using a number of different CAPO provider versions in the past, up to v0.4.0, without issues.

Recently upgraded to v0.6.3, and we are no longer able to create workload clusters at all. CAPO logs show:

```
E0923 22:32:19.906429       1 controller.go:317] controller/openstackcluster "msg"="Reconciler error" "error"="failed to create volume service client: No suitable endpoint could be found in the service catalog." "name"="..." "namespace"="..." "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster"
```

It looks like https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/f5b623047780c0dd0a3c509896ef6737e3fa8970 added OpenStack volume support, but in doing so, made the volume service a hard requirement -- and would now fail to bind to the compute service if there was no volume endpoint.

I expected to be able to deploy workload clusters using provider version v0.6.3 in the same way as I was able to deploy them with earlier versions :-)

It's clear that the features implemented in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1030 require volume support. However, it'd be great if CAPO could continue to function *without* a volume OpenStack endpoint when volumes are not being used.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.6.3
- Cluster-API version: v1.2.1
- OpenStack version: Xena
- Minikube/KIND version: N/A
- Kubernetes version (use `kubectl version`): v1.22

mdbooth:

The addition of multi-AZ volume support inadvertently added a hard requirement for Cinder. This change removes that requirement by:

* Splitting ComputeClient into separate Compute, Image, and Volume clients
* Making initialisation of each client lazy on first use

If a machine definition does not use volumes the VolumeClient will not be initialised, meaning Cinder is no longer a requirement.

Fixes #1345 

It might be simpler to review these 2 commits separately. I pulled the first one out of one of my other outstanding PRs #1236. The separation of clients which I did in that PR for other reasons works very nicely for this, too.

The first commit is mostly code motion. The second commit implements the change.

/hold


pkg/clients/volume.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (c volumeClient) GetVolume(volumeID string) (*volumes.Volume, error) {
+	mc := metrics.NewMetricPrometheusContext("volume", "get")
+	volume, err := volumes.Get(c.client, volumeID).Extract()
+	return volume, mc.ObserveRequestIgnoreNotFound(err)
+}
+
+type volumeErrorClient struct{ error }
+
+// NewVolumeErrorClient returns a VolumeClient in which every method returns the given error.
```
jichenjc, line 80: I like the idea , create such error client should help abstract the error case

mdbooth: I wanted to create a proper monad, but that's not ergonomic in Go. This seemed like a fairly neat alternative.

mdbooth:
@jichenjc Had to fix a couple of merge conflicts, but nothing consequential. Appreciate it if you could have a look and restore your lgtm.

jichenjc:
/lgtm
/hold

lgtm, hold this wait for CI and please unhold when you think it's ok after CI pass @mdbooth

tobiasgiese:
> lgtm, hold this wait for CI and please unhold when you think it's ok after CI pass @mdbooth

Prow will not merge until the required tests are green 🙂 thus, there is no need to add the hold to wait for CI to pass. Just FYI

tobiasgiese:
> lgtm, hold this wait for CI and please unhold when you think it's ok after CI pass @mdbooth

@jichenjc Prow will not merge until the required tests are green 🙂 thus, there is no need to add the hold to wait for CI to pass. Just FYI

jichenjc:
@tobiasgiese thanks :)

jichenjc:
plan to cherry pick this for 0.6 but encounter many conflict ...

PR #1364, opened 2022-10-25
🌱 CI: Clean openstack volumes

lentzi90:

The hack/ci/openstack.sh script creates servers that boot from volumes. These volumes were not cleaned up before. With this commit the volumes are created explicitly with the same name as the servers and then deleted together with them.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: -

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/hack/ci/openstack.sh
+++ b/hack/ci/openstack.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# hack script for preparing GCP to run cluster-api-provider-openstack e2e
+# hack script for preparing Openstack to run cluster-api-provider-openstack e2e
 
 set -x -o errexit -o nounset -o pipefail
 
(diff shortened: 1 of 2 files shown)


hack/ci/openstack.sh
```diff
@@ -157,6 +168,9 @@ function cloud_cleanup {
         if ! openstack server delete --wait "$name"; then
             openstack server show "$name" && exit 1
         fi
+        if ! openstack volume delete "$name"; then
+            openstack volume show "$name" && exit 1
```
jichenjc, line 172: just curious , this is CI nested KVM env, should be deleted (the GCP VM) 

so the reason for explicit delete volume is ? for local test?

lentzi90: No this is for setting up the devstack on Openstack (instead of GCP). I'm running this against an Openstack cloud when developing e2e tests. It creates the servers for running devstack that is then used for the e2e tests.

I use it like this:
```bash
# In the normal CI we have RESOURCE_TYPE=gce-project
export RESOURCE_TYPE=openstack
./hack/ci/create_devstack.sh
make test-e2e
./hack/ci/create_devstack.sh cleanup
# After this the volumes for the devstack servers were still there.
```

PR #1366, opened 2022-10-26
Fix accesIPv4 checking

dulek:
Seems like the code got moved around, while checking for a condition
haven't, this commit makes sure we error out when instance we're
creating has no AccessIPv4 address candidate.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -132,12 +132,7 @@ func (s *Service) CreateInstance(eventObject runtime.Object, openStackCluster *i
 
 func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluster *infrav1.OpenStackCluster, instanceSpec *InstanceSpec, clusterName string, retryInterval time.Duration) (*InstanceStatus, error) {
 	var server *clients.ServerExt
-	accessIPv4 := ""
 	portList := []servers.Network{}
-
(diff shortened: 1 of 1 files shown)


tobiasgiese:
I don't think that we should remove it!
The validation was moved to the top of this func with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1094. We should rather fix it instead of removing it. WDYT @jichenjc?

dulek:
> I don't think that we should remove it! The validation was moved to the top of this func with #1094. We should rather fix it instead of removing it. WDYT @jichenjc?

Ah, true, the condition probably still makes sense, it just needs to go before `getOrCreateRootVolume()` invocation, right?

tobiasgiese:
> /ok-to-test

We don't need the `/ok-to-test` as the removal is not reasonable

mdbooth:
@tobiasgiese Ah, true! I added it on the basis that I trust @dulek before reading the issue.

@dulek You need to join the SIG! Then your oversights can warm the planet without delay like the rest of us 😉

dulek:
Alright, I updated this by moving the `if` clause to what I think is a proper place. Please tell me if this makes sense.

tobiasgiese:
Fair for me 👍🏻 

/approve

jichenjc:
@dulek is the commit message still valid? others lgtm

PR #1369, opened 2022-10-28
🌱 Synchronize versions between Makefile, Containerfile, e2e and go.mod

lentzi90:

I noticed a warning in the e2e tests when trying to load the cert-manager images. It is because we have different versions in the [e2e_conf.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/data/e2e_conf.yaml#L24-L29) and the [Makefile](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/Makefile#L177-L179).
I also bumped to go version in go.mod since we have 1.18 already in the [Containerfile](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/Containerfile#L27), and bumped the kubernetes version for the envtest. (It would be nice to automatically get the same version that we use for the e2e tests if someone can come up with a solution for that.)

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: -

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90:
Ah the cert-manager version should be synced the other way... There is no v1.8.3 tag, only v1.8.2. Fixing...

tobiasgiese:
@lentzi90 Jason is no active contributor of this project 🙂 even if Prow is mentioning him.

/approve

PR #1370, opened 2022-10-31
:sparkles:  Add RouterName for use existing router

### LINKED ISSUE #1322: Use existing router
/kind feature

I tried to run the "Quick start" guide to setup cluster api on openstack but ran into some issues.
I'm using a openstack setup that only allows 1 router per project.
This doesn't work since the CAPO controller manager tries to create a new one. I see that if the router is named exactly as the code assumes it should be named it will reuse it. However I don't know if that's a recommended way of doing it and it's still not possible to specify the name of the router it should look for.

These are the logs from the CAPO controller manager when it's trying to reconcile it:

```
I0809 12:36:43.308778       1 network.go:177] controller/openstackcluster "msg"="Reconciling subnet" "cluster"="cluster-api-test-1" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster" "name"="k8s-clusterapi-cluster-default-cluster-api-test-1"
I0809 12:36:43.771114       1 recorder.go:103] events "msg"="Warning"  "message"="Failed to create router k8s-clusterapi-cluster-default-cluster-api-test-1: Expected HTTP response code [] when accessing [POST https://ops.elastx.cloud:9696/v2.0/routers], but got 409 instead\n{\"NeutronError\": {\"type\": \"OverQuota\", \"message\": \"Quota exceeded for resources: ['router'].\", \"detail\": \"\"}}" "object"={"kind":"OpenStackCluster","namespace":"default","name":"cluster-api-test-1","uid":"f3200d67-8f5d-46b2-a7b9-49c66d8fda8d","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha5","resourceVersion":"16131"} "reason"="Failedcreaterouter"
E0809 12:36:43.781927       1 controller.go:317] controller/openstackcluster "msg"="Reconciler error" "error"="failed to reconcile router: Expected HTTP response code [] when accessing [POST https://ops.elastx.cloud:9696/v2.0/routers], but got 409 instead\n{\"NeutronError\": {\"type\": \"OverQuota\", \"message\": \"Quota exceeded for resources: ['router'].\", \"detail\": \"\"}}" "name"="cluster-api-test-1" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster" 
```

My suggested solution would be to add a new field in the `OpenStackCluster` object so you could specify an already existing router name/ID.

dhnikolas:
Now you can use an existing router using the RouterName field. Cluster initialization does not create a new router. And when deleting a cluster with an existing router, it is not deleted

Fixes #1322

2. I am ready to get feedback. And make changes :) thx
**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [ ] adds unit tests

/hold
/test pull-cluster-api-provider-openstack-conformance-test


pkg/cloud/services/networking/router.go
```diff
@@ -190,7 +199,20 @@ func (s *Service) setRouterExternalIPs(openStackCluster *infrav1.OpenStackCluste
 }
 
 func (s *Service) DeleteRouter(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
-	router, subnet, err := s.getRouter(clusterName)
+	routerName := getRouterName(clusterName)
```
jichenjc, line 202: how about inject this logic into getRouter function 
but in that function we do the check ? seems it's readable? thoughts?

pkg/cloud/services/networking/router.go
```diff
@@ -45,6 +45,11 @@ func (s *Service) ReconcileRouter(openStackCluster *infrav1.OpenStackCluster, cl
 	}
 
 	routerName := getRouterName(clusterName)
+	existingRouter := false
+	if openStackCluster.Spec.RouterName != "" {
+		routerName = openStackCluster.Spec.RouterName
+		existingRouter = true
+	}
 	s.scope.Logger.Info("Reconciling router", "name", routerName)
```
jichenjc, line 53: guess this line can be put into existing = false part?

pkg/cloud/services/networking/router.go
dhnikolas: If you inject this logic inside a function, then the check (RouterName != "") will have to be done twice. Inside and outside the getRouter function. Also, the getRouter function will have two arguments routerName and clusterName, which is not very obvious. Since routerName can be empty. And if it's not empty then it will override the clusterName argument. It seems that the current version is more readable. what do you think about it?

pkg/cloud/services/networking/router.go
```diff
@@ -213,6 +235,10 @@ func (s *Service) DeleteRouter(openStackCluster *infrav1.OpenStackCluster, clust
 		}
 	}
 
+	if existingRouter {
+		return nil
```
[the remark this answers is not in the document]
dhnikolas: done

pkg/cloud/services/networking/router.go
dhnikolas: done

pkg/cloud/services/networking/router.go
jichenjc: ok, I am fine with easier logic, thanks for the more info

jichenjc:
/lgtm

and some doc update will be helpful

dhnikolas:
> /lgtm
> 
> and some doc update will be helpful

dhnikolas:
docs updated

dhnikolas:
@jichenjc can you approve this PR? Or do I need to invite other reviewers?

jichenjc:
@seanschneeweiss do you want to take a look on this ? Thanks

seanschneeweiss:
Thank you for the improvements around re-using an existing router.

The spec `routerName` might be a bit misleading as it suggests, that even if no router exists, the specified router Name would be used for creating the cluster's network router.

A simple approach for re-using an existing router could be, to just rename the existing router to the name, that CAPO would use by default.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c9cca1d0eba4bcec8e5d3364d795b562cd30d121/pkg/cloud/services/networking/router.go#L284-L286

So `openstack router set --name <capo-name>` would be sufficient.
I understand that maybe not everyone has access for renaming the existing router, but would this help for your case?

If this doesn't help - I guess it should be also possible to define a routerName to create a new router using that specific name (rather than just creating one if it is empty). WDYT?

dhnikolas:
Thank you

In our case, the option using the current router name template is not very good. Since we want to add capi to an existing infrastructure that has many tenants. And it will not work to rename routers in these tenants. And in general, the use of such a name looks more like a hack than a feature.
I have 2 ideas.
1. We can rename the current key RouterName -> ExistingRouterName. Thus the key will not have a double meaning. And in the future we can add the RouterName key as a name for the created router
Or
2. RouterName will use the existing router. But if there is no router with that name. Then a new router will be created using this name

lukasmrtvy:
any progress on this ?

jichenjc:
@tobiasgiese @mdbooth may you want to take a look as well?

lukasmrtvy:
@dhnikolas should not be `OPENSTACK_EXTERNAL_NETWORK_ID` now optional ? With `disableAPIServerFloatingIP: true` it should not be needed.

seanschneeweiss:
@dhnikolas 

Sorry for the delay.

> RouterName will use the existing router. But if there is no router with that name. Then a new router will be created using this name

Let's keep the `RouterName` key. However, there is probably no benefit for pre-defining a certain name if CAPO manages the router.
So if `RouterName` is set, this indicates that an existing router should be used. And the code should return an error, if the router cannot be found.

api/v1alpha6/openstackcluster_types.go
```diff
@@ -50,6 +50,9 @@ type OpenStackClusterSpec struct {
 	// through DNS is required.
 	// +listType=set
 	DNSNameservers []string `json:"dnsNameservers,omitempty"`
+	// RouterName is the name of an existing router.
+	// If specified, no router will not be created.
```
seanschneeweiss, line 54:
```suggestion
	// If specified, the router is used and no new router will be created.
```

dhnikolas:
> @dhnikolas
> 
> Sorry for the delay.
> 
> > RouterName will use the existing router. But if there is no router with that name. Then a new router will be created using this name
> 
> Let's keep the `RouterName` key. However, there is probably no benefit for pre-defining a certain name if CAPO manages the router. So if `RouterName` is set, this indicates that an existing router should be used. And the code should return an error, if the router cannot be found.
> 
> So exactly as you already implemented it 😃

@jichenjc @seanschneeweiss  great! thx!  Can you approve?

seanschneeweiss:
@dhnikolas 
Would you mind squashing all commits into a single commit? That way we could omit the `Merge Branch ...` commits.

api/v1alpha6/openstackcluster_types.go
seanschneeweiss: @dhnikolas 
Would be great if you could rephrase this part, as right now it is double negative: `no router will not be created`.

dhnikolas:
> @dhnikolas Would you mind squashing all commits into a single commit? That way we could omit the `Merge Branch ...` commits.
> 
> Please also check the unresolved conversation. Thank you.

done)

dhnikolas:
@seanschneeweiss Hi) please could you review this?

drew-viles:
I'd love to see this implemented. 

I have a use case which requires this exact feature!

mdbooth:
> Functionally this is awesome, but I wonder if it shouldn't use UUID instead of name? I understand how name is easier to implement, but UUIDs feel much more reliable in OpenStack.

Came here to say something very similar. We have an existing pattern of <Foo>Filter which we should probably copy here. How about:

```go
    ....
    Router RouterFilter
    ...

type RouterFilter struct {
    UUID string
    Name string

    // For completeness if we thing it's useful, or leave it and maybe add later:
    Tags
    TagsAny
    NotTags
    NotTagsAny
}
```

The behaviour would be that if UUID is specified we don't do any lookup. If not we lookup according to the other parameters. As for other lookups, it would be an error to specify a filter which returns more than 1 router.

In practise this would look like:

```yaml
    router:
        uuid: aaaa-bbbb-ccccc-dddd
```

or:

```yaml
    router:
        name: my-router
```

dhnikolas:
> > Functionally this is awesome, but I wonder if it shouldn't use UUID instead of name? I understand how name is easier to implement, but UUIDs feel much more reliable in OpenStack.
> 
> Came here to say something very similar. We have an existing pattern of Filter which we should probably copy here. How about:
> 
> ```go
>     ....
>     Router RouterFilter
>     ...
> 
> type RouterFilter struct {
>     UUID string
>     Name string
> 
>     // For completeness if we thing it's useful, or leave it and maybe add later:
>     Tags
>     TagsAny
>     NotTags
>     NotTagsAny
> }
> ```
> 
> The behaviour would be that if UUID is specified we don't do any lookup. If not we lookup according to the other parameters. As for other lookups, it would be an error to specify a filter which returns more than 1 router.
> 
> In practise this would look like:
> 
> ```yaml
>     router:
>         uuid: aaaa-bbbb-ccccc-dddd
> ```
> 
> or:
> 
> ```yaml
>     router:
>         name: my-router
> ```

Looks good. I got it

pkg/cloud/services/networking/router.go
```diff
@@ -44,18 +44,28 @@ func (s *Service) ReconcileRouter(openStackCluster *infrav1.OpenStackCluster, cl
...
 		return err
 	}
 
+	if existingRouter && len(routerList) == 0 {
+		return fmt.Errorf("router not found by routerFilter ")
+	}
+
 	if len(routerList) > 1 {
-		return fmt.Errorf("found %d router with the name %s, which should not happen", len(routerList), routerName)
+		return fmt.Errorf("found %d router, which should not happen", len(routerList))
```
dulek, line 68: Should be "found %d routers".

pkg/cloud/services/networking/router.go
```diff
@@ -190,7 +200,26 @@ func (s *Service) setRouterExternalIPs(openStackCluster *infrav1.OpenStackCluste
...
 func (s *Service) DeleteRouter(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
-	router, subnet, err := s.getRouter(clusterName)
+	listOpts := routers.ListOpts{}
+	existingRouter := false
+	if openStackCluster.Spec.Router != nil {
+		listOpts = routers.ListOpts(*openStackCluster.Spec.Router)
+		existingRouter = true
+	} else {
+		listOpts.Name = getRouterName(clusterName)
+	}
```
dulek, line 210: Instead of else clause you could have put `Name` directly on `listOpts` initialization.

dulek (commented):
Looks pretty good! Minor comments inline. My bigger concern is if we should add new option to both v1alpha6 and v1alpha7 or just to v1alpha7?

dhnikolas:
@dulek @jichenjc @seanschneeweiss  hi! can you review this?

api/v1alpha7/types.go
```diff
@@ -104,6 +104,25 @@ type SubnetFilter struct {
...
 }
 
+type RouterFilter struct {
+	ID           string `json:"id,omitempty"`
+	Name         string `json:"name,omitempty"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up,omitempty"`
+	Distributed  *bool  `json:"distributed,omitempty"`
+	Status       string `json:"status,omitempty"`
+	TenantID     string `json:"tenantId,omitempty"`
```
mdbooth, line 114: Please don't add TenantID here: just ProjectID.

api/v1alpha7/types.go
```diff
@@ -104,6 +104,25 @@ type SubnetFilter struct {
...
+	ProjectID    string `json:"projectId,omitempty"`
+	Limit        int    `json:"limit,omitempty"`
+	Marker       string `json:"marker,omitempty"`
+	SortKey      string `json:"sortKey,omitempty"`
+	SortDir      string `json:"sortDir,omitempty"`
```
mdbooth, line 119: Definitely don't add any of these (Limit-SortDir inclusive)!

api/v1alpha7/types.go
```diff
@@ -104,6 +104,25 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type RouterFilter struct {
+	ID           string `json:"id,omitempty"`
+	Name         string `json:"name,omitempty"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up,omitempty"`
+	Distributed  *bool  `json:"distributed,omitempty"`
+	Status       string `json:"status,omitempty"`
```
mdbooth, line 113: We should not add Status here as it would not be deterministic.

api/v1alpha7/types.go
```diff
@@ -104,6 +104,25 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type RouterFilter struct {
+	ID           string `json:"id,omitempty"`
+	Name         string `json:"name,omitempty"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up,omitempty"`
+	Distributed  *bool  `json:"distributed,omitempty"`
```
mdbooth, line 112: I'm less sure about `Distributed`, but I'm inclined to say it's not relevant to what we're filtering on here, so we should omit it.

mdbooth (changes requested):

api/v1alpha7/types.go
```diff
@@ -104,6 +104,25 @@ type SubnetFilter struct {
 	NotTagsAny      string `json:"notTagsAny,omitempty"`
 }
 
+type RouterFilter struct {
+	ID           string `json:"id,omitempty"`
+	Name         string `json:"name,omitempty"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up,omitempty"`
```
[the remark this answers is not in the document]
dulek: Yeah, that doesn't seem like a good fit for a filter, you're right. I completely missed this.

api/v1alpha7/types.go
dulek: I think this is an immutable property, but it's hard to imagine case when filtering using this makes sense.

api/v1alpha7/types.go
```diff
@@ -104,6 +104,20 @@ type SubnetFilter struct {
...
 }
 
+type RouterFilter struct {
+	ID          string `json:"id,omitempty"`
+	Name        string `json:"name,omitempty"`
+	Description string `json:"description,omitempty"`
+	ProjectID   string `json:"projectId,omitempty"`
+	Limit       int    `json:"limit,omitempty"`
+	Marker      string `json:"marker,omitempty"`
+	SortKey     string `json:"sortKey,omitempty"`
```
mdbooth, line 114: I think you missed these 3 (Limit-SortKey).

dhnikolas: Oh, yeah. I'll fix that now

dulek:
I keep my question about which API versions should be affected by this. I'm totally not sure what is allowed in older API versions.

dhnikolas:
> I keep my question about which API versions should be affected by this. I'm totally not sure what is allowed in older API versions.

Now I only add changes to version 7.  In older versions I made changes only for conversions and tests. Its ok?

dulek:
To me this looks good, someone else would need to take a look at the conversions and their correctness.

mdbooth:
> > I keep my question about which API versions should be affected by this. I'm totally not sure what is allowed in older API versions.
> 
> Now I only add changes to version 7. In older versions I made changes only for conversions and tests. Its ok?

I think it's ok to only add this to v1alpha7.

api/v1alpha6/conversion_test.go
```diff
@@ -44,11 +44,13 @@ func TestFuzzyConversion(t *testing.T) {
 				c.FuzzNoCustom(v1alpha7Cluster)
 
 				v1alpha7Cluster.Spec.APIServerLoadBalancer.Provider = ""
+				v1alpha7Cluster.Spec.Router = nil
 			},
 			func(v1alpha7ClusterTemplate *infrav1.OpenStackClusterTemplate, c fuzz.Continue) {
 				c.FuzzNoCustom(v1alpha7ClusterTemplate)
 
 				v1alpha7ClusterTemplate.Spec.Template.Spec.APIServerLoadBalancer.Provider = ""
+				v1alpha7ClusterTemplate.Spec.Template.Spec.Router = nil
```
mdbooth, line 53: I didn't expect this to be necessary, but I can see that the test fails without it. I am investigating this.

mdbooth: I've just submitted https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1517 to remove the other use of FuzzerFuncs here. Please could you look at that PR for how to do this in the converter rather than ignoring this in the test.

Please could you rebase on top of that PR? It's probably simplest to wait for it to merge first. Sorry about this!

pkg/cloud/services/networking/router.go
```diff
@@ -45,17 +45,26 @@ func (s *Service) ReconcileRouter(openStackCluster *infrav1.OpenStackCluster, cl
...
 
 	routerName := getRouterName(clusterName)
-	s.scope.Logger().Info("Reconciling router", "name", routerName)
+	routerListOpts := routers.ListOpts{Name: routerName}
+	existingRouter := false
+	if openStackCluster.Spec.Router != nil {
+		routerListOpts = openStackCluster.Spec.Router.ToListOpt()
+		existingRouter = true
+	} else {
+		s.scope.Logger().Info("Reconciling router", "name", routerName)
```
mdbooth, line 54: Can you move this back up to the top so it's unconditional? I would like to see this log message even when we're going to re-use an existing router: it's a useful marker.

pkg/cloud/services/networking/router.go
```diff
@@ -229,26 +257,8 @@ func (s *Service) getRouterInterfaces(routerID string) ([]ports.Port, error) {
...
-	}
-
-	return router, subnet, nil
-}
-
-func (s *Service) getRouterByName(routerName string) (routers.Router, error) {
-	routerList, err := s.client.ListRouter(routers.ListOpts{
-		Name: routerName,
-	})
+func (s *Service) getRouterByFilter(opts routers.ListOpts) (routers.Router, error) {
```
mdbooth, line 260: nit: This function has only 1 caller and it's not super complicated. Personally I'd either re-use it in the create flow, which would have the benefit of de-deduplicating the error message if there's more than one matching router, or inline it in the delete flow.

This is just a nit. Not important to change if you don't want.

mdbooth (commented):
This is looking great. Very nearly good to go. Please can we keep the "Reconciling router" log message as noted below? The language fix would also be good in the docs.

The nit is at your discretion.

pkg/cloud/services/networking/router.go
dhnikolas: We can add a log at the top, but without specifying the name of the router. Like this
s.scope.Logger().Info("Reconciling router", "cluster", clusterName)
Or we can display the log after receiving the real name of the router by filter. What will be better?

dhnikolas: I still think the first option is better. Since the message with the real id and the name of the router is already below
`s.scope.Logger().V(6).Info(fmt.Sprintf("Reuse existing Router %s with id %s", router.Name, router.ID))`

Issue #1363, opened 2022-10-20
Additional e2e tests for various scenarios


lentzi90:
/kind feature

**Describe the solution you'd like**

Currently, the [e2e tests](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/suites/e2e/e2e_test.go) mainly check that it is possible to create (and delete) clusters with various configurations.
However, we have no  tests for things like upgrading CAPO from older releases or [pivoting into a self hosted cluster](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/self_hosted.go) (`clusterctl move`).

Luckily, CAPI has already [implemented some test specs](https://github.com/kubernetes-sigs/cluster-api/tree/main/test/e2e) for these scenarios and more.
They can be easily reused in CAPO for validating the functionality.
I would like to add some of these tests to CAPO.

Now the question is when and how to run the tests.
I think it may be too much to ask to run them automatically on all PRs, so perhaps a periodic or manually triggered jobs would be better?
For example we could make sure to run it before every release.

**Anything else you would like to add:**

I have started working on this already but thought it would be better to discuss it a bit in an issue first.
Will open a PR once I have something more substantial.

jichenjc:
>>>However, we have no tests for things like upgrading CAPO from older releases or [pivoting into a self hosted cluster](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/self_hosted.go) (clusterctl move).

FYI, I proposed this before but folks think it's a CAPI function , should not be part of CAPO
but I am still in favor of add such 

>>>For example we could make sure to run it before every release.

our internal dev for other projects has weekly run on one time effort job to avoid put too much effort in every PR
not sure this is doable here

lentzi90:
Adding here some findings from working on clusterctl upgrade tests.

1. The CAPI [clusterctl upgrade test spec](https://github.com/kubernetes-sigs/cluster-api/blob/release-1.2/test/e2e/clusterctl_upgrade.go) is a bit special since it creates not just a workload cluster, but a secondary management cluster also. The flow goes like this: Bootstrap cluster -> create secondary management cluster with older CAPI/CAPO -> create workload cluster -> clusterctl upgrade -> scale workload cluster. The reason for this is to avoid any issues with parallel jobs requiring different versions so the bootstrap cluster can be shared.
[...]

PR #1371, opened 2022-11-03
🌱 Add e2e clusterctl upgrade tests

lentzi90:

This adds e2e clusterctl upgrade tests from the CAPI e2e framework. Having a clusterctl upgrade test is very important for verifying upgrades between API versions.
The CAPI [clusterctl upgrade test spec](https://github.com/kubernetes-sigs/cluster-api/blob/release-1.2/test/e2e/clusterctl_upgrade.go) is a bit special since it creates not just a workload cluster, but a secondary management cluster also. The flow goes like this:
1. From the bootstrap cluster, create secondary management cluster with older CAPI/CAPO
2. From the secondary management cluster, create a workload cluster to check that the older versions are working correctly
3. Run clusterctl upgrade for the secondary management cluster (this requires a way to get the e2e image to the node where there upgraded controller will run)
4. Scale the workload cluster to verify that the upgraded versions are working.

The reason for the secondary management cluster is to avoid any issues with parallel jobs requiring different versions, so the bootstrap cluster can be shared as normal.

Part of #1363 but I would want to add some more tests before closing it.

1. There are 2 commits in this PR for making it easier to discuss and review until we agree on the implementation:
    1. Adding the tests
    2. WIP upload the e2e image to the devstack controller and pre-pull it on the nodes.
        Note that this is only needed if we want to test the image built by CI (for example on PRs). If it is ok to just take the `main` tagged image, then we can drop this.
2. I added tests only for v0.6 -> current as agreed in the office hours. The work required to add older versions would make things more complicated in a few different ways:
   1. They don't all support the latest kubernetes version, so we need to set the version separately for them
   2. We get new combinations of CAPI/CAPO API versions, which means that the current way to just divide by the CAPO API version [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/main/kustomize) would not be enough. (We would have CAPI v1alpha4 + CAPO v1alpha4 and CAPI v1beta1 + CAPO v1alpha4.)
   3. Already for v0.5 we would have the issue that it does not support Openstack yoga, so we would need to run the devstack with some older version.
3. Because of the secondary management cluster, we get quite a lot of VMs, which means we cannot run 2 of these tests in parallel as we currently do. We get a total of 1 bastion + 1 KCP + 1 worker for the management cluster + 1 bastion + 1 KCP + 2 workers (when scaling to test) in the workload cluster. If we run in parallel we are already above the devstack default quota of 10 instances.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 ################################################################################
-# usage: ci-conformance.sh
-#  This program runs the clusterctl conformance e2e tests.
+# usage: ci-e2e.sh
+#  This program runs the e2e tests.
 ################################################################################
 
 set -x
@@ -83,6 +83,20 @@ if [ -n "${BOSKOS_HOST:-}" ]; then
 fi
 
 "hack/ci/create_devstack.sh"
+
+# Upload image for e2e clusterctl upgrade tests
+source "${REPO_ROOT}/hack/ci/${RESOURCE_TYPE}.sh"
+CONTAINER_ARCHIVE="${ARTIFACTS}/capo-e2e-image.tar"
+SSH_KEY="$(get_ssh_private_key_file)"
+SSH_ARGS="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -o PasswordAuthentication=no"
+CONTROLLER_IP=${CONTROLLER_IP:-"10.0.3.15"}
+
+make e2e-image
+docker save -o "${CONTAINER_ARCHIVE}" gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+scp -i "${SSH_KEY}" ${SSH_ARGS} "${CONTAINER_ARCHIVE}" "cloud@${CONTROLLER_IP}:capo-e2e-image.tar"
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo chown root:root capo-e2e-image.tar
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo chmod u=rw,g=r,o=r capo-e2e-image.tar
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo mv capo-e2e-image.tar /var/www/html/capo-e2e-image.tar
 
 export OPENSTACK_CLOUD_YAML_FILE
 OPENSTACK_CLOUD_YAML_FILE="$(pwd)/clouds.yaml"
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -100,10 +100,8 @@ providers:
     - sourcePath: "./infrastructure-openstack/cluster-template-external-cloud-provider.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template-without-lb.yaml"
     replacements:
-    # TODO: We should use e2e here instead of main, but we need a way to get it into the workload cluster
-    # for upgrade tests for that to work.
     - old: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:dev
-      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:main
+      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
     - old: "imagePullPolicy: Always"
       new: "imagePullPolicy: IfNotPresent"
     - old: "--v=2"
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -35,6 +35,7 @@ providers:
   - name: v1.2.4
     value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.2.4/core-components.yaml"
     type: url
+    contract: v1beta1
     files:
     - sourcePath: "../data/shared/v1beta1/metadata.yaml"
     replacements:
@@ -48,6 +49,7 @@ providers:
   - name: v1.2.4
     value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.2.4/bootstrap-components.yaml"
     type: url
+    contract: v1beta1
     files:
     - sourcePath: "../data/shared/v1beta1/metadata.yaml"
     replacements:
@@ -61,6 +63,7 @@ providers:
   - name: v1.2.4
     value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.2.4/control-plane-components.yaml"
     type: url
+    contract: v1beta1
     files:
     - sourcePath: "../data/shared/v1beta1/metadata.yaml"
     replacements:
@@ -71,17 +74,36 @@ providers:
 - name: openstack
   type: InfrastructureProvider
   versions:
+  # This is only for clusterctl upgrade tests
+  - name: v0.6.3
+    value: "https://github.com/kubernetes-sigs/cluster-api-provider-openstack/releases/download/v0.6.3/infrastructure-components.yaml"
+    type: url
+    contract: v1beta1
+    files:
+    - sourcePath: "../data/shared/v1beta1_provider/metadata.yaml"
+    - sourcePath: "./infrastructure-openstack/cluster-template.yaml"
+    replacements:
+    - old: "imagePullPolicy: Always"
+      new: "imagePullPolicy: IfNotPresent"
+    - old: "--v=2"
+      new: "--v=4"
+    - old: "--leader-elect"
+      new: "--leader-elect=false\n        - --sync-period=1m"
   - name: v0.6.99
     value: ../../../config/default
-    contract: v1beta1
+    # This is the upcoming version.
+    # Specify no contract so that upgrade tests that start from a specific contract won't pick it up.
+    # contract: v1beta1
     files:
     - sourcePath: "../data/shared/v1beta1_provider/metadata.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template-external-cloud-provider.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template-without-lb.yaml"
     replacements:
+    # TODO: We should use e2e here instead of main, but we need a way to get it into the workload cluster
+    # for upgrade tests for that to work.
     - old: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:dev
-      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:main
     - old: "imagePullPolicy: Always"
       new: "imagePullPolicy: IfNotPresent"
     - old: "--v=2"
@@ -116,6 +138,7 @@ variables:
   OPENSTACK_VOLUME_TYPE_ALT: "test-volume-type"
   CONFORMANCE_WORKER_MACHINE_COUNT: "5"
   CONFORMANCE_CONTROL_PLANE_MACHINE_COUNT: "1"
+  INIT_WITH_KUBERNETES_VERSION: "v1.25.0"
 
 intervals:
   conformance/wait-control-plane: ["30m", "10s"]
--- a/test/e2e/data/kustomize/upgrade-patches/ci-hack-kcp.yaml
+++ b/test/e2e/data/kustomize/upgrade-patches/ci-hack-kcp.yaml
@@ -1,0 +1,31 @@
+---
+# Workaround for https://github.com/kubernetes-sigs/cluster-api/issues/7457
+# There is a small but important difference between these two:
+# path: /a/b/c
+# *creates* the c array, overwriting anything that was there before
+# path: /a/b/c/-
+# *adds* to the c array and does not work if the array is missing
+#
+# We add to the postKubeadmCommands (instead of pre*) since we need the CI artifacts
+# script to run first. Without this, the container images are not imported properly.
+- op: add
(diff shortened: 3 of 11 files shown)


lentzi90:
NOTE: The e2e clusterctl test will not pass with the WIP commit (I couldn't get it to import the container image properly so it gets stuck in ImagePullBackOff). I have however verified that it is working when using the main tag instead of e2e.

scripts/ci-e2e.sh
```diff
@@ -84,6 +84,24 @@ fi
...
+# This is just a quick hack to test the process
+make e2e-image
+make hack/tools/bin/gojq
+docker save -o ${ARTIFACTS}/capo-e2e-image.tar gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+export OS_CLOUD=capo-e2e
+VIRTUAL_ENV_DISABLE_PROMPT=1 source /tmp/openstack-venv/bin/activate
+openstack container create --public capo-e2e
+openstack object create capo-e2e ${ARTIFACTS}/capo-e2e-image.tar --name capo-e2e-image.tar
+ACCOUNT=$(openstack object show capo-e2e capo-e2e-image.tar -f json | hack/tools/bin/gojq -r .account)
+URL=http://10.0.3.15:8080/v1/${ACCOUNT}/capo-e2e/capo-e2e-image.tar
```
jichenjc, line 97: I guess we should use this variable?
```
hack/ci/create_devstack.sh:CONTROLLER_IP=${CONTROLLER_IP:-"10.0.3.15"}
```

test/e2e/data/e2e_conf.yaml
```diff
@@ -71,17 +107,21 @@ providers:
...
     files:
     - sourcePath: "../data/shared/v1beta1_provider/metadata.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template-external-cloud-provider.yaml"
     - sourcePath: "./infrastructure-openstack/cluster-template-without-lb.yaml"
     replacements:
+    # TODO(lentzi90): We should use the e2e tag here instead of main
     - old: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:dev
-      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+      new: gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:main
```
jichenjc, line 124: is there any reason we didn't use e2e with current PR?

scripts/ci-e2e.sh
```diff
@@ -84,6 +84,24 @@ fi
...
+# TODO: Upload image to swift for e2e clusterctl upgrade tests
+# This is just a quick hack to test the process
+make e2e-image
+make hack/tools/bin/gojq
+docker save -o ${ARTIFACTS}/capo-e2e-image.tar gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+export OS_CLOUD=capo-e2e
+VIRTUAL_ENV_DISABLE_PROMPT=1 source /tmp/openstack-venv/bin/activate
+openstack container create --public capo-e2e
+openstack object create capo-e2e ${ARTIFACTS}/capo-e2e-image.tar --name capo-e2e-image.tar
+ACCOUNT=$(openstack object show capo-e2e capo-e2e-image.tar -f json | hack/tools/bin/gojq -r .account)
```
jichenjc, line 96: I am not against swift, so double check here whether we have other simpler way?
e.g glance can store files , httpd(apache) can provide data through wget .. 
so just a question here

test/e2e/suites/e2e/clusterctl_upgrade_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo"
+	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+
+	"sigs.k8s.io/cluster-api-provider-openstack/test/e2e/shared"
+)
+
+// TODO: Do we want v0.3 -> current and v0.4 -> current also?
```
jichenjc, line 33: I don't think we need so as we have changed a lot since there
but we can ask in slack if they do ..

test/e2e/data/e2e_conf.yaml
lentzi90: It is just to see that the test works. I could not get it to properly import the e2e image for some reason so testing with this until I have solved it

test/e2e/suites/e2e/clusterctl_upgrade_test.go
lentzi90: I took this up at the office hours. We agreed to just test v0.6 -> current to start with and then add more versions when they are releases. So no testing of the older versions.

scripts/ci-e2e.sh
lentzi90: Thank you for this! I was not aware of the httpd server running on the controller. This was indeed much easier to use.

scripts/ci-e2e.sh
```diff
@@ -84,6 +84,25 @@ fi
...
+SSH_KEY="$(get_ssh_private_key_file)"
+SSH_ARGS="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -o PasswordAuthentication=no"
+CONTROLLER_IP=${CONTROLLER_IP:-"10.0.3.15"}
+
+make e2e-image
+docker save -o "${CONTAINER_ARCHIVE}" gcr.io/k8s-staging-capi-openstack/capi-openstack-controller:e2e
+scp -i "${SSH_KEY}" ${SSH_ARGS} "${CONTAINER_ARCHIVE}" "cloud@${CONTROLLER_IP}:capo-e2e-image.tar"
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo chown root:root capo-e2e-image.tar
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo chmod u=rw,g=r,o=r capo-e2e-image.tar
+ssh -i "${SSH_KEY}" ${SSH_ARGS} "cloud@${CONTROLLER_IP}" -- sudo mv capo-e2e-image.tar /var/www/html/capo-e2e-image.tar
```
lentzi90, line 100: Any suggestions on how to improve this?
The point of this is to build the e2e image, save it as an archive and upload it to the controller where it can be fetched for the secondary management cluster. (In the KinD bootstrap cluster, we simply inject the image directly, but we cannot do this for the secondary management cluster that it used for the upgrade test.)

The main issue I have with it is that we build the e2e image twice. Once here and once when running `make test-e2e` below. For the upgrade test it may be okay but it is completely unnecessary for all the other tests.

scripts/ci-e2e.sh
```diff
@@ -84,6 +84,25 @@ fi
...
+
+# TODO: This is just to test the clusterctl upgrade tests
+export E2E_GINKGO_PARALLEL="1"
+export E2E_GINKGO_ARGS="-focus='\[clusterctl-upgrade\]'"
```
lentzi90, line 104: The idea with this would be to pass it in directly in test-infra, similar to how it is [done for CAPI](https://github.com/kubernetes/test-infra/blob/245873efc4d80d737143d7ebf62d5493495cba0a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml#L140-L141).

test/e2e/data/kustomize/upgrade-patches/ci-hack-kcp.yaml
```diff
@@ -0,0 +1,31 @@
...
+# *creates* the c array, overwriting anything that was there before
+# path: /a/b/c/-
+# *adds* to the c array and does not work if the array is missing
+#
+# We add to the postKubeadmCommands (instead of pre*) since we need the CI artifacts
+# script to run first. Without this, the container images are not imported properly.
+- op: add
+  path: /spec/kubeadmConfigSpec/postKubeadmCommands
+  value:
+  - /usr/local/bin/ci-artifacts-openstack.sh
```
mdbooth, line 14: Where does this run? Is it on the test runner or on the target? I assume the latter.

test/e2e/suites/e2e/clusterctl_upgrade_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Note: There isn't really anything fixing this to the v0.6 release. The test will
+// simply pick the latest release with the correct contract from the E2EConfig, as seen here
+// https://github.com/kubernetes-sigs/cluster-api/blob/3abb9089485f51d46054096c17ccb8b59f0f7331/test/e2e/clusterctl_upgrade.go#L265
+// When new minor releases are added (with the same contract) we will need to work on this
+// if we want to continue testing v0.6.
+var _ = Describe("When testing clusterctl upgrades (v0.6=>current) [clusterctl-upgrade]", func() {
+	ctx := context.TODO()
+	shared.SetEnvVar("USE_CI_ARTIFACTS", "true", false)
+	shared.SetEnvVar("DOWNLOAD_E2E_IMAGE", "true", false)
```
mdbooth, line 39: Isn't this setting an environment variable on the test runner? If so, how does that affect `/usr/local/bin/ci-artifacts-openstack.sh` which I'm assuming (possibly incorrectly) is executing on the target?

mdbooth (commented):
Some questions quite possibly just for my own understanding.

Can we sort out the commit message of that second commit? I don't want to merge something with WIP in the title!

test/e2e/data/e2e_conf.yaml
```diff
@@ -116,6 +136,7 @@ variables:
   OPENSTACK_VOLUME_TYPE_ALT: "test-volume-type"
   CONFORMANCE_WORKER_MACHINE_COUNT: "5"
   CONFORMANCE_CONTROL_PLANE_MACHINE_COUNT: "1"
+  INIT_WITH_KUBERNETES_VERSION: "v1.25.0"
```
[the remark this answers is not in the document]
lentzi90: This is for the [secondary management cluster](https://github.com/kubernetes-sigs/cluster-api/blob/e032bec7c04c45b42d92b7bb9bd4f16765218181/test/e2e/clusterctl_upgrade.go#L73). It will be used when rendering the cluster-template [here](https://github.com/kubernetes-sigs/cluster-api/blob/e032bec7c04c45b42d92b7bb9bd4f16765218181/test/e2e/clusterctl_upgrade.go#L196-L209).
I also noticed that the installation image is stuck on 1.18, but we use the CI artifacts script to actually upgrade before kubeadm is run.

lentzi90:
I expect the e2e tests to fail now on this PR until https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1390 and https://github.com/kubernetes/test-infra/pull/28082 are merged. This is because of resource constraints. We cannot run the clusterctl-upgrade test in parallel with other tests.

test/e2e/suites/e2e/clusterctl_upgrade_test.go
lentzi90: Indeed! It is used when rendering the cluster-template. Both these variables are used in the cluster-template that we use in the tests. You can see the rendered result in `_artifacts/templates`. Here is the one from the previous (successful) e2e test run on this PR: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1371/pull-cluster-api-provider-openstack-e2e-test/1594601712893562880/artifacts/templates/cluster-template-ci-artifacts.yaml (and [here is the complete _artifacts folder](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1371/pull-cluster-api-provider-openstack-e2e-test/1594601712893562880/artifacts/) from that run for reference).

test/e2e/data/e2e_conf.yaml
```diff
@@ -71,9 +74,26 @@ providers:
...
+    files:
+    - sourcePath: "../data/shared/v1beta1_provider/metadata.yaml"
+    - sourcePath: "./infrastructure-openstack/cluster-template.yaml"
+    replacements:
+    - old: "imagePullPolicy: Always"
+      new: "imagePullPolicy: IfNotPresent"
+    - old: "--v=2"
+      new: "--v=4"
+    - old: "--leader-elect"
+      new: "--leader-elect=false\n        - --sync-period=1m"
```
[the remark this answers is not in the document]
lentzi90: Anything in particular that you are wondering about? My understanding of this is that it is used to [create a repository](https://github.com/kubernetes-sigs/cluster-api/blob/e032bec7c04c45b42d92b7bb9bd4f16765218181/test/framework/clusterctl/repository.go#L83) that clusterctl is then configured to use as the source of truth instead of reaching out to github to check what releases we have. This is mostly useful for not-yet-released versions, but since the same config is also used for upgrade tests, we need to explicitly list the versions that should be "available" to the test.

I didn't think much about the additional config here (replacements and files) but I think they make sense and follow what we already have as well as what CAPI is doing.

lentzi90:
> Can we sort out the commit message of that second commit? I don't want to merge something with WIP in the title!

Done :heavy_check_mark:

mdbooth:
I'd like to wait until I get my head round the `DOWNLOAD_E2E_IMAGE` thing. I'm sure it's fine, it's just eluding me right now.

/hold

test/e2e/suites/e2e/clusterctl_upgrade_test.go
mdbooth: I'm going to have to look at the template renderer. I didn't think it would support bash syntax like `${DOWNLOAD_E2E_IMAGE:=false}`.

If that question itself reveals my fundamental misunderstanding, please can you  enlighten me?

mdbooth: The substitution language is linked from the clusterctl documentation here: https://cluster-api.sigs.k8s.io/clusterctl/commands/generate-yaml.html

PR #1375, opened 2022-11-06
🐛 fix: improve load balancer health checks

### LINKED ISSUE #1374: Hard-coded health monitors (downtime during upgrades)
/kind bug

The health checks for the load balancer that is setup at the moment have hard-coded values, in this case:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/046d2e37ac3ffe2cdf113c99f6cc9dafe0a6b7c4/pkg/cloud/services/loadbalancer/loadbalancer.go#L338

There is a 30 second delay between every health check, during a control plane upgrade, this can actually result in a large majority of the nodes going to `NotReady` since they lost comms with the control plane.

The ability to control the health monitors to be a bit more aggressive to kick systems out of the pool faster to avoid this issue.  In this case, it's almost 1m30s before it gets removed.

I went over other providers and this is what I determined:

AWS provider (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4dda2debcf94a2f9d6cbe0b862bfcd6296178832/pkg/cloud/services/elb/loadbalancer.go#L437-L443):
- 10s interval, 5s timeout, 3 unhealthy threshold, so kicked out after 30s

Azure (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/a2ce2a2943d290307e08fe77233b462bc82bf869/azure/services/loadbalancers/spec.go#L260-L261):
- 15s interval, 4 probes (so I assume 4 failures = kick out), so 60s to get kicked out

DigitalOcean (https://github.com/kubernetes-sigs/cluster-api-provider-digitalocean/blob/29d4bd4993291ff7943f3b1a19b56c0d53bfb069/api/v1beta1/types.go#L176-L202):
- dynamically configured, defaults to 10s interval, 5s timeout, 3 unhealthy threshold, so kicked out after 30s by default

GCP (https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/01061d433f1ef7003d75515d72e49e56ca0892a0/cloud/scope/cluster.go#L316-L319):
- 10s interval, 5s timeout, 3 unhealthy threshold, so kick out after 30s

I propose that we do the following:
- `delay`: 10s (The time, in seconds, between sending probes to members.  )
- `max_retries`: 5 (The number of successful checks before changing the operating status of the member to ONLINE. A valid value is from 1 to 10. )
- `max_retries_down`: 3 (The number of allowed check failures before changing the operating status of the member to ERROR. A valid value is from 1 to 10. The default is 3.)
- `timeout`: 5s (The maximum time, in seconds, that a monitor waits to connect before it times out. This value must be less than the delay value.  )

I would have liked to change the following too:

- `type`: HTTPS
- `url_path`: `/healthz`

This will give us actual health checks that involve the API endpoint telling us that it's happy,  and it should be OK to accept the unknown cert since https://github.com/openstack/octavia/blob/8c793f2e8ff7857a97cc68a4706bdcdfdff1fde5/octavia/common/jinja/haproxy/split_listeners/templates/macros.j2#L178-L182 is OK with it however OVN doesnt support this so we can keep TCP for now.

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

mnaser:

This lower the delay between checks on the health
monitors which should reduce the recovery time
during an upgrade for example.

Fixes #1374

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -332,12 +332,13 @@ func (s *Service) getOrCreateMonitor(openStackCluster *infrav1.OpenStackCluster,
 	s.scope.Logger.Info(fmt.Sprintf("Creating load balancer monitor for pool %q", poolID), "name", monitorName, "lb-id", lbID)
 
 	monitorCreateOpts := monitors.CreateOpts{
-		Name:       monitorName,
-		PoolID:     poolID,
-		Type:       "TCP",
-		Delay:      30,
-		Timeout:    5,
-		MaxRetries: 3,
+		Name:           monitorName,
+		PoolID:         poolID,
+		Type:           "TCP",
+		Delay:          10,
+		MaxRetries:     5,
+		MaxRetriesDown: 3,
+		Timeout:        5,
 	}
 	monitor, err = s.loadbalancerClient.CreateMonitor(monitorCreateOpts)
 	if err != nil {


jichenjc:
thanks for the detailed info in the issue, it's very helpful info

/lgtm

mdbooth (approved):
There's a fair bit to unpack here. Firstly I wonder if this should be configurable with good defaults, but lets leave that aside for now.

Secondly we're doing a TCP check rather than hitting the heackcheck endpoint, which seems suboptimal.

Lastly, here's a really good internal OpenShift doc for platform implementors with justified recommendations: https://github.com/openshift/installer/blob/master/docs/dev/kube-apiserver-health-check.md

Approving this because it's clearly better than what's currently there, but I think there's more to do here.

mnaser:
> There's a fair bit to unpack here. Firstly I wonder if this should be configurable with good defaults, but lets leave that aside for now.

Agreed

> Secondly we're doing a TCP check rather than hitting the heackcheck endpoint, which seems suboptimal.

I was going to propose making the change to HTTPS (see issue) however OVN provider for Octavia doesn't support it, so I figured that would be a bit hard for the RH side of things :)

> Lastly, here's a really good internal OpenShift doc for platform implementors with justified recommendations: https://github.com/openshift/installer/blob/master/docs/dev/kube-apiserver-health-check.md

That's a really neat document.

> Approving this because it's clearly better than what's currently there, but I think there's more to do here.

Issue #1372, opened 2022-11-05
capo-controller crashloop -  invalid memory address or nil pointer dereference


fmoctezuma:
/kind bug

**What steps did you take and what happened:**
1.- We created a new cluster successfully, 
2.- We tried to create a second cluster and capo-controller is now on crashloop, with the following error

```
I1105 13:45:07.697460       1 controller.go:178] controller/openstackcluster "msg"="Starting EventSource" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster" "source"="kind source: *v1beta1.Cluster"
I1105 13:45:07.697968       1 controller.go:186] controller/openstackcluster "msg"="Starting Controller" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster"
I1105 13:45:07.697274       1 controller.go:178] controller/openstackmachine "msg"="Starting EventSource" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" "source"="kind source: *v1alpha5.OpenStackMachine"
I1105 13:45:07.698349       1 controller.go:178] controller/openstackmachine "msg"="Starting EventSource" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" "source"="kind source: *v1beta1.Machine"
I1105 13:45:07.698379       1 controller.go:178] controller/openstackmachine "msg"="Starting EventSource" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" "source"="kind source: *v1alpha5.OpenStackCluster"
I1105 13:45:07.698407       1 controller.go:178] controller/openstackmachine "msg"="Starting EventSource" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine" "source"="kind source: *v1beta1.Cluster"
I1105 13:45:07.698425       1 controller.go:186] controller/openstackmachine "msg"="Starting Controller" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackMachine"
I1105 13:45:07.801556       1 controller.go:220] controller/openstackcluster "msg"="Starting workers" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="OpenStackCluster" "worker count"=10
[...]

PR #1376, opened 2022-11-09
check InfrastructureRef  before usage

jichenjc:

potential fix for #1372 ,from the call stack looks like the infrastructureRef might be nil
and lead to that issue, do no harm if we add such check ..

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -116,7 +116,7 @@ func (r *OpenStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, nil
 	}
 
-	infraCluster, err := r.getInfraCluster(ctx, cluster, openStackMachine)
+	infraCluster, err := r.getInfraCluster(ctx, log, cluster, openStackMachine)
 	if err != nil {
 		return ctrl.Result{}, errors.New("error getting infra provider cluster")
 	}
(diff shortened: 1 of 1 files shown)


mdbooth (commented):
This fixes the panic, but it's just a line in the controller logs so it doesn't let the user know what they need to fix. This needs to be user-visible somehow.

How does the InfrastructureRef get set? It's in the spec so I'd *hope* it's by the user, at least indirectly, but that's not always the case. If it's by the user, why would it not be set? Could we catch this in the cluster controller instead and set an appropriate condition there?

jichenjc:
@mdbooth based on https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html
the Spec.InfrastructureRef is set by CAPI, I am wondering whether it's a timing issue
so with proposed logic, we will enter `OpenStackCluster is not ready yet` logic then we will retry which seems fit for 
fixing timing issue

seanschneeweiss (commented):
CAPD seems to have a different approach.
They filter the watch to only add dockermachines to the list that reference machines with having set the infrastructureRef.

See https://github.com/kubernetes-sigs/cluster-api/blob/5bc3b40be412f22e86d0ad6297f5f9764e33324b/test/infrastructure/docker/internal/controllers/dockermachine_controller.go#L436-L439
and
https://github.com/kubernetes-sigs/cluster-api/blob/5bc3b40be412f22e86d0ad6297f5f9764e33324b/test/infrastructure/docker/internal/controllers/dockermachine_controller.go#L476-L478

What do you think of doing something similar to that?

mdbooth:
@jichenjc Shall we reopen this? I doubt we've fixed it any other way since.

jichenjc:
/reopen 

will see whether I can find sometime to continue to work on this soon..

PR #1381, opened 2022-11-15
🐛 Improve `env.rc` and `create_cloud_conf.sh` templates

skatsaounis:

It fixes the following erroneous situations:
1. Scripts are run from macOS where base64 supports `--break` instead of `--wrap`
2. For security reasons OpenStack password is present in `OS_PASSWORD` environment variable rather than clouds.yaml

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/templates/env.rc
+++ b/templates/env.rc
@@ -70,10 +70,22 @@ yqNavigating(){
         fi
 }
 
+b64encode(){
+  # Check if wrap is supported. Otherwise, break is supported.
+  if echo | base64 --wrap=0 &> /dev/null; then
+    base64 --wrap=0 $1
+  else
+    base64 --break=0 $1
+  fi
+}
+
 # Just blindly parse the cloud.yaml here, overwriting old vars.
 CAPO_AUTH_URL=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yqNavigating - clouds.${CAPO_CLOUD}.auth.auth_url)
 CAPO_USERNAME=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yqNavigating - clouds.${CAPO_CLOUD}.auth.username)
 CAPO_PASSWORD=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yqNavigating - clouds.${CAPO_CLOUD}.auth.password)
+if [[ "$CAPO_PASSWORD" = "" || "$CAPO_PASSWORD" = "null" ]]; then
+  CAPO_PASSWORD="${OS_PASSWORD}"
+fi
 CAPO_REGION=$(echo "$CAPO_OPENSTACK_CLOUD_YAML_CONTENT" | yqNavigating - clouds.${CAPO_CLOUD}.region_name)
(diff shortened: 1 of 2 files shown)


templates/env.rc
```diff
@@ -63,17 +63,30 @@ CAPO_OPENSTACK_CLOUD_YAML_CONTENT=$(cat "${CAPO_CLOUDS_PATH}")
 
 CAPO_YQ_VERSION=$(yq -V)
 yqNavigating(){
-        if [[ ${CAPO_YQ_VERSION} == *"version 1"* || ${CAPO_YQ_VERSION} == *"version 2"* || ${CAPO_YQ_VERSION} == *"version 3"* ]]; then
-                yq r $1 $2
-        else
-                yq e .$2 $1
-        fi
+  if [[ ${CAPO_YQ_VERSION} == *"version 1."* || ${CAPO_YQ_VERSION} == *"version 2."* || ${CAPO_YQ_VERSION} == *"version 3."* ]]; then
```
mdbooth, line 66: Is there a reason for adding the dots here? If so, is there any reason not to do it elsewhere?

mdbooth (changes requested):
This looks ok to me, and presumably very useful if you're running it on a Mac!

Please could you remove all the whitespace and incidental changes? They make the change itself much harder to see and will also make future git spelunking harder by adding non-functional noise. If the whitespace changes are important we could consider them in a separate PR, though.

Thanks!

seanschneeweiss:
@skatsaounis Would be great if you could address the changes proposed by @mdbooth.
Let us know if something is unclear of if you need assistance for completing this PR.

templates/env.rc
skatsaounis: No, not really. I just reverted the change

skatsaounis:
Hi both @seanschneeweiss and @mdbooth! Sorry for such a huge delay in addressing the PR review comments.

Basically, I just removed what is not related to the changes that are fixing the issue with macOS and fetching the OS_PASSWORD from an environment variable.

I think now the PR is much more clear and ready to be merged if you agree to do so.

Once again, sorry for the delay and thank you for taking time to consider it and review it 🍻

templates/env.rc
```diff
@@ -70,10 +70,23 @@ yqNavigating(){
         fi
 }
 
+b64encode(){
+  # Check if wrap is supported. Otherwise, break is supported.
+  echo | base64 --wrap=0 > /dev/null 2>&1
+  if [ $? -eq 0 ]; then
```
tobiasgiese, line 76: nit: we can simplify this :)

```suggestion
  if echo | base64 --wrap=0 &> /dev/null; then
```

tobiasgiese (approved):
One nit, but all in all LGTM

skatsaounis:
@tobiasgiese thanks for the suggestion, which it makes sense. I just applied with a separate commit through the UI.

tobiasgiese:
> I just applied with a separate commit through the UI.

Can you please squash the commits? Prow will rebase this branch into master during merge, thus commits must be squashed :)

mdbooth (commented):
/lgtm

mdbooth (approved):

PR #1385, opened 2022-11-17
 🐛 Fix Octavia versions endpoint

### LINKED ISSUE #1384: Incorrect detection of Octavia features?
/kind bug

We have [this code for determining of allowedCIDRs is supported](https://github.com/Nordix/cluster-api-provider-openstack/blob/38027fbd81c44a1263eb3bf027dbe2005b76649f/pkg/cloud/services/loadbalancer/loadbalancer.go#L107-L116):

```golang
	octaviaVersions, err := s.loadbalancerClient.ListOctaviaVersions()
	if err != nil {
		return err
	}
	// The current version is always the last one in the list.
	octaviaVersion := octaviaVersions[len(octaviaVersions)-1].ID
	if openstackutil.IsOctaviaFeatureSupported(octaviaVersion, openstackutil.OctaviaFeatureVIPACL, lbProvider) {
		allowedCIDRsSupported = true
	}
```

However [loadbalancerClient.ListOctaviaVersions](https://github.com/Nordix/cluster-api-provider-openstack/blob/38027fbd81c44a1263eb3bf027dbe2005b76649f/pkg/clients/loadbalancer.go#L259-L267) appears to be hitting a compute endpoint, we're passing an ID as a version, and [the Octavia documentation](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=show-load-balancer-details-detail,show-provider-flavor-capabilities-detail#show-provider-flavor-capabilities) doesn't mention an apiVersions API. I suspect this is legacy from before Octavia was split from Nova and no longer has any meaning. 

Perhaps we're supposed to be using [Provider capabilities](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=show-load-balancer-details-detail,show-provider-flavor-capabilities-detail#show-provider-flavor-capabilities)?

Don't know, but I suspect the current code is meaningless.

mdbooth:
We were hitting the compute endpoint instead of the octavia endpoint when fetching octavia versions.

This is used to detect if Octavia supports allowed CIDRs, so presumably that feature is broken.

Fixes #1384

/hold

### DIFF
--- a/pkg/clients/loadbalancer.go
+++ b/pkg/clients/loadbalancer.go
@@ -21,7 +21,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
-	"github.com/gophercloud/gophercloud/openstack/compute/apiversions"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/apiversions"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
--- a/pkg/clients/mock/loadbalancer.go
+++ b/pkg/clients/mock/loadbalancer.go
@@ -24,7 +24,7 @@ import (
 	reflect "reflect"
 
 	gomock "github.com/golang/mock/gomock"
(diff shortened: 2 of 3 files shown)


jichenjc:
I think no need fix gophercloud CPO works fine on this logic
so fix the UT error then we should be good to go

mdbooth:
> https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/apiversions/results.go#L8

I think gophercloud is ok. The struct is incomplete, but shouldn't be a problem for our purposes.

I've fixed the reference in the unit test.

mnaser (approved):

PR #1390, opened 2022-11-21
🌱 Tag current e2e tests as PR-Blocking

lentzi90:

This is preparation for adding more e2e tests which we may not want to run on all PRs. Also adds E2E_GINKGO_FOCUS and E2E_GINKGO_SKIP as convenience variables, and made E2E_GINKGO_PARALLEL configurable

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests


lentzi90:
This is where the inspiration is coming from:
- https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/quick_start_test.go#L27
- https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml#L141

CAPI is using tags like `[PR-Blocking]` for their CI and ginkgos focus/skip functionality to decide what tests to run.

Makefile
```diff
@@ -120,10 +120,18 @@ LDFLAGS := $(shell source ./hack/version.sh; version::ldflags)
 ## --------------------------------------
 
 # The number of ginkgo tests to run concurrently
-E2E_GINKGO_PARALLEL=2
+E2E_GINKGO_PARALLEL ?= 2
 
 E2E_ARGS ?=
 
+E2E_GINKGO_FOCUS ?=
+E2E_GINKGO_SKIP ?=
```
jichenjc, line 128: just to learn this 
what's the FOCUS means ? 100% ratio?

Makefile
```diff
@@ -120,10 +120,18 @@ LDFLAGS := $(shell source ./hack/version.sh; version::ldflags)
...
+
+# to set multiple ginkgo skip flags, if any
+ifneq ($(strip $(E2E_GINKGO_SKIP)),)
+_SKIP_ARGS := $(foreach arg,$(strip $(E2E_GINKGO_SKIP)),-skip="$(arg)")
```
jichenjc, line 132: for now ,there is no SKIP test , you will add the E2E full test into this skip list (so it will not run every PR
but should be daily/weekly or manual trigger?)

Makefile
lentzi90: Focus means "run only this, skip everything else". So it is for when you want to run only one or a few tests. :slightly_smiling_face:

Makefile
lentzi90: Actually my plan was to use focus to run the default tests (`E2E_GINKGO_FOCUS="\[PR-Blocking\]"`). Then for the e2e full test we would skip PR-Blocking since it would anyway run on all PRs. This is the same way that CAPI has done it.

Here CAPI focus on PR-Blocking for their default e2e test: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml#L120-L150
And here they skip PR-Blocking and other tests that would anyway run separately: https://github.com/kubernetes/test-infra/blob/b27527630f6d4c0f2787f0362efd54c4cab5fbd6/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml#L222-L256

Makefile
```diff
@@ -163,7 +171,10 @@ e2e-prerequisites: $(GINKGO) e2e-templates e2e-image test-e2e-image-prerequisite
 E2E_GINKGO_ARGS ?=
 .PHONY: test-e2e ## Run e2e tests using clusterctl
 test-e2e: e2e-prerequisites ## Run e2e tests
-	time $(GINKGO) --failFast -trace -progress -v -tags=e2e --nodes=$(E2E_GINKGO_PARALLEL) $(E2E_GINKGO_ARGS) ./test/e2e/suites/e2e/... -- -config-path="$(E2E_CONF_PATH)" -artifacts-folder="$(ARTIFACTS)" --data-folder="$(E2E_DATA_DIR)" $(E2E_ARGS)
+	time $(GINKGO) --failFast -trace -progress -v -tags=e2e --nodes=$(E2E_GINKGO_PARALLEL) \
+			--focus="$(E2E_GINKGO_FOCUS)" $(_SKIP_ARGS) $(E2E_GINKGO_ARGS) ./test/e2e/suites/e2e/... -- \
```
mdbooth, line 175: Does this error if `E2E_GINKGO_FOCUS` is empty? Do we need the same dance as on lines 131-133 above for skip?

If not, can we also omit the dance for `E2E_GINKGO_SKIP`?

lentzi90: It works even if `E2E_GINKGO_FOCUS` is empty. I actually stole it from [CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L777-L778). They also have the same `_SKIP_ARGS` and I think this is because an empty string would match everything. So empty focus is no problem, we just run all the tests. Skipping everything on the other hand would not be great.

seanschneeweiss (commented):
/lgtm

Thank you for taking such a close look at how CAPI solves the same challenge and adapting to their implementation.

lentzi90:
Some kind of temporary issue with the devstack from what I understand

jichenjc:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1397 tracking

lentzi90:
Rebased to get the CI fix. Tests are passing :tada:

erkanerol:
@lentzi90 THANKS A LOT!!!

PR #1391, opened 2022-11-21
✨ Add OpenStack simulator for running envtests

mdbooth:
This PR builds on the work done in the (unmerged) #1236. That PR added the ability to substitute implementations of the service Scope and introduced a Mock implementation.

This change goes a step further and introduces a simulator. The simulator intends to provide realistic responses to OpenStack API calls. For example, if a controller creates a Router while reconciling, listing routers on a subsequent reconciliation will include the created Router in the response until it is deleted.

Tests are able to inspect the internal state of the simulator to determine what OpenStack changes were made in response to API changes. Additionally, all OpenStack calls have Pre- and Post- hooks.

Pre-hook callbacks receive the arguments to the OpenStack call and return the expected response types with the addition of a 'handled' boolean. If Pre-hook callback sets handled then the simulator implementation will be called. The intended usage is custom behaviour modification, including error injection. 

Post-hook callbacks receive the arguments the method was called with in addition to the values returned by the simulator. Post-hooks will not be called if a pre-hook set handled.

The simulator is not complete, and likely contains a lot of bugs and inaccuracies. Where the simulator is incomplete it will explicitly `panic` with a message indicating the API call which is not yet implemented. However, it seems to be complete enough to to write useful tests today, and we can fix and extend it over time.

Commits in this PR up to `Correct typo` by @stephenfin are part of  #1236. If we push this PR it will obsolete the other one. Alternatively we could approve #1236 separately and I'll rebase this on top. My preference would be the latter.

This PR also current contains  #1386, which is now approved. I will rebase when that lands.

To understand what this change is about, I recommend starting with the last commit. This includes some tests that actually use the simulator. The implementation is contained in the prior commit.

**TODOs**:

- [x] Rebase on top of #1386
- [ ] Rebase on top of #1236 (if we decide to do that)

/hold

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -170,6 +170,12 @@ issues:
     - stylecheck
     text: "ST1003: should not use underscores in Go names;"
     path: .*(api|types)\/.*\/.*conversion.*\.go$
+  # Disable goconst in the simulator package due to the large number of false
+  # positives: fields with the same name in different structs should not be
+  # combined with a common const definition.
+  - linters:
+    - goconst
+    path: pkg/clients/simulator/.*\.go$
 
 run:
   timeout: 10m
--- a/Makefile
+++ b/Makefile
@@ -226,12 +226,15 @@ $(SETUP_ENVTEST): # Build setup-envtest from tools folder.
 ##@ Linting
 ## --------------------------------------
 
+# e2e-templates are a dependency of lint and lint-update to validate the
+# go:embed directive which reads them.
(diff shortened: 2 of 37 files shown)


mnaser:
wow, this is really neat.  could be super handy as a test toolkit library.

mdbooth:
This was an interesting experiment which I hope to come back to, but I think it would be better implemented in gophercloud.

PR #1393, opened 2022-11-22
✨ Add Finalizer to Identity Secret

### LINKED ISSUE #1392: Deprovisioning Improvements
/kind feature

I may be missing something here, so please let me know!

Typically when you provision things it's a case of `clusterctl generate cluster | kubectl apply -f -`, and to deprovision you'd expect just to just do the same in reverse.  It's a similar kind of model with Helm, create all the things, then delete all the things.

One thing I noted was that doing so deletes the openstack credentials secret, no secret means the CAPO controller cannot talk to openstack to remove the openstack resources, thus you are in a deadlock situation.

My initial thought here would be to add a finalizer to the secret and release it when all the openstack* resources are done with.  At the moment I kinda hack it by saying "delete all the things, except the secret", which is functional, but not really idiomatic.

Quite happy to author this and drive it if/when a technical solution can be agreed upon.

spjmurray:

To support a create cluster/delete cluster workflow (a la helm), we need to ensure the identity secret is alive for the duration of OS resource provisioning and deprovisioning, so add a finalizer that gets added on cluster/machine creation, and removed when their finalizers are done so.

Fixes #1392 

This is a PoC (so not tested, just to get a finger in the wind), head to the issue for discussion!

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


mnaser:
I like this idea!  Can you look into the tests?

spjmurray:
Thanks!  It's not a perfect idea by any means, we'll need to discuss this in the larger CAPI forum but, from my testing:

* as the finalizers are attached to a referred to resource, you cannot replace the reference, or you get drift and deadlocks, but you can rotate the data no problem
* updating the finalizers is a concurrency problem, so you need some ugly retries in there

That said, it's actually a quite succinct fix, and should be - through the magic of interfaces - made to be generic across all providers.  Anyway over to the main forum...

mdbooth:
I wonder if we should revisit this.

Issue #1397, opened 2022-11-25
gate (temp?) failure due to devstack


jichenjc:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

keep seeing following error  e.g

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1396/pull-cluster-api-provider-openstack-e2e-test/1595934082498301952/artifacts/devstack/controller-devstack.log

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1390/pull-cluster-api-provider-openstack-e2e-test/1595931786162999296/artifacts/devstack/controller-devstack.log

so need check whether devstack issue or our own configuration need update

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built):
- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

lentzi90:
> oslo_config.cfg.NoSuchOptError: no such option ovn in group [DEFAULT]

Seems like a real problem. I wonder if it is caused by https://opendev.org/openstack/devstack/commit/8d299efa4b6346ccfc3c6fcf9cf011b3c884bebc
I imagine that there is some kind of python version / python lib incompatibility.

jichenjc:
I knew community usually do some update when they notice it 
but we are actually on yoga and zed stable branch.. so maybe wait 1-2 days and give another try
if still problem , let's check inside our code ..

nikParasyr:
Hello,

From a quick check i saw that you are pinning to stable/yoga version. Also it seems that ovn config is barking. I think this change on neutron on branch stable/yoga could be triggering the issue: https://github.com/openstack/neutron/commit/44ec1f150338ec69a811e1b2cfa4214c07a9af83

I might have a more in depth look later on

jichenjc:
yes, both yoga and zed has issue..

mdbooth:
In the face of it this looks like a regression in stable yoga. I suspect (hope!) they'll want to address it very quickly if we report it as such.

mdbooth:
[...]

PR #1399, opened 2022-11-25
🌱Ensure we capture early devstack logs

mdbooth:
Hoping this might help us debug #1397 and similar future issues.

/hold


ralonsoh:
> I understand @ralonsoh has enough to debug (thanks @stephenfin!) so I don't feel the need to push this while CI is failing. From the looks of the output it's working as intended, though.

Hi @mdbooth, please check https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1397#issuecomment-1327444355

PR #1402, opened 2022-11-28
🌱 Bump Gophercloud to v1.1.0

pierreprinetti:
Update Gophercloud to the latest available version. This also means updating indirect dependencies, among which notably golang.org/x/crypto and golang.org/x/text.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests


seanschneeweiss:
Thank you for your pull request.

v0.25.0 and v1.0.0 of gophercloud don't seem to introduce any breaking change that are of interest for CAPO.
This should be safe to upgrade then.

/lgtm

pierreprinetti:
@seanschneeweiss 
Unless we screw up big time, [there won't be breaking changes until v2](https://github.com/gophercloud/gophercloud#backwards-compatibility-guarantees); this is one major goal of the project. We have put in place tooling that should help with that, alongside with [gorelease](https://pkg.go.dev/golang.org/x/exp/cmd/gorelease).

PR #1412, opened 2022-12-12
🌱 Remove It block from clusterctl upgrade

lentzi90:
**What this PR does / why we need it**:

The ClusterctlUpgradeSpec contains a BeforeEach block so it cannot be nested in an It block.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #1417, opened 2022-12-12
🌱 e2e: Change logging to avoid use of By

### LINKED ISSUE #1414: e2e: Allow logging outside of It blocks
/kind feature

As seen in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1412, the current e2e logging solution does not handle cases when we want to use it outside of `It` blocks. To improve this I suggest adopting a different log solution instead of the `Byf` that we currently use. It seems like [CAPZ has already solved this](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/00d08a3b63cf822474548a1046768a4a1bbe01f0/test/e2e/log.go#L35-L42) so we could copy what they do.

The relevant code is here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/43c27cd0246347e603ba82bef64a8442272010ec/test/e2e/shared/common.go#L210-L218

lentzi90:

Changes the e2e logging solution to avoid `By`, since `By` can only be used inside `It` blocks and similar.
Reverted the [workaround for clusterctl upgrade](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1412) also.

Fixes #1414

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/shared/common.go
+++ b/test/e2e/shared/common.go
@@ -40,7 +40,7 @@ import (
 )
 
 func SetupSpecNamespace(ctx context.Context, specName string, e2eCtx *E2EContext) *corev1.Namespace {
-	Byf("Creating a namespace for hosting the %q test spec", specName)
+	Logf("Creating a namespace for hosting the %q test spec", specName)
 	namespace, cancelWatches := framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
 		Creator:   e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
 		ClientSet: e2eCtx.Environment.BootstrapClusterProxy.GetClientSet(),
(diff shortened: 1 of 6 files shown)


lentzi90:
Checking that this actually solves the issue.

test/e2e/shared/common.go
```diff
@@ -207,13 +207,13 @@ func dumpSpecResources(ctx context.Context, e2eCtx *E2EContext, namespace *corev
 	})
 }
 
-func Byf(format string, a ...interface{}) {
-	By("[" + time.Now().Format(time.RFC3339) + "] " + fmt.Sprintf(format, a...))
+func Logf(format string, a ...interface{}) {
+	fmt.Fprintf(GinkgoWriter, "["+time.Now().Format(time.RFC3339)+"] "+format+"\n", a...)
```
jichenjc, line 211: this is the key thing we refer from CAPZ? 
isntead of By, we use `fmt.Fprintf(GinkgoWriter`, correct?

lentzi90: Correct!

PR #1428, opened 2022-12-20
🌱 Add e2e self hosted test

lentzi90:

Adds e2e self hosted test. This is a test spec [implemented in CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/self_hosted.go) that checks the `clusterctl move` functionality, i.e. that the controllers can be moved from the bootstrap cluster to a workload cluster. It also checks that the controllers are functioning in the new cluster, and can handle node rotations, by doing a k8s upgrade.

Part of #1363

2. This would not be part of the normal e2e tests, only e2e-full.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -116,6 +116,10 @@ variables:
   # used to ensure we deploy to the correct management cluster
   KUBE_CONTEXT: "kind-capo-e2e"
   KUBERNETES_VERSION: "v1.25.0"
+  KUBERNETES_VERSION_UPGRADE_FROM: "v1.24.9"
+  KUBERNETES_VERSION_UPGRADE_TO: "v1.25.0"
+  ETCD_VERSION_UPGRADE_TO: "3.5.4-0"
+  COREDNS_VERSION_UPGRADE_TO: "v1.9.3"
   CNI: "../../data/cni/calico.yaml"
   CCM: "../../data/ccm/cloud-controller-manager.yaml"
   EXP_CLUSTER_RESOURCE_SET: "true"
@@ -148,3 +152,5 @@ intervals:
   default/wait-worker-nodes: ["30m", "10s"]
   default/wait-delete-cluster: ["5m", "10s"]
   default/wait-alt-az: ["20m", "30s"]
+  default/wait-machine-upgrade: ["30m", "10s"]
+  default/wait-nodes-ready: ["15m", "10s"]
--- a/test/e2e/suites/e2e/self_hosted_test.go
+++ b/test/e2e/suites/e2e/self_hosted_test.go
@@ -1,0 +1,52 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


test/e2e/data/e2e_conf.yaml
```diff
@@ -115,7 +115,11 @@ providers:
 variables:
   # used to ensure we deploy to the correct management cluster
   KUBE_CONTEXT: "kind-capo-e2e"
-  KUBERNETES_VERSION: "v1.25.0"
+  KUBERNETES_VERSION: "v1.25.3"
```
jichenjc, line 118: latest is 1.25.5
https://kubernetes.io/releases/
curious we didn't use it ?due to consistent to existing test version?

jichenjc:
assume it's a full test scope? 

test/e2e/data/e2e_conf.yaml
lentzi90: Good point! I just copied from CAPI. I'll see if I can get it working with v.25.5

test/e2e/suites/e2e/self_hosted_test.go
```diff
@@ -0,0 +1,43 @@
...
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
```
seanschneeweiss, line 26: As in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/4526db5675ddaf7e62ab8017af7cb1ff46d1b201/test/e2e/shared/cluster.go#L29
```suggestion
	capie2e "sigs.k8s.io/cluster-api/test/e2e"
```

lentzi90:
Test is failing due to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1429
Will wait for that to be fixed until I continue working on this.

test/e2e/suites/e2e/self_hosted_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"sigs.k8s.io/cluster-api-provider-openstack/test/e2e/shared"
+)
+
+var _ = Describe("When testing Cluster API provider Openstack working on [self-hosted] clusters", func() {
```
seanschneeweiss, line 31:
```suggestion
var _ = Describe("Create a Cluster API Provider OpenStack cluster, pivote to a self-hosted cluster", func() {
```

lentzi90: This is the style used in CAPI and also in the [clusterctl upgrade test in CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/576a13f19f5fc52a063c039d6734b72fb3d62ecd/test/e2e/suites/e2e/clusterctl_upgrade_test.go#L36). I'm open to change it but would like to explain why it is expressed in this way first.
The top level `Describe` text is followed by `It` and `By` blocks in the test, so this would read something like

```
[Describe] When testing Cluster API provider Openstack working on [self-hosted] clusters
[It] Should pivot the bootstrap cluster to a self-hosted cluster
[By] Creating a workload cluster
...
```

lentzi90:
Seems like this will need a bit more work. Marking it as WIP.

mdbooth:
Works, only runs in `-full-test`, executes in a bit over 10 minutes. It would be good to fix the upgrade issue, but this is ok without.

/approve

PR #1432, opened 2023-01-04
🌱Be more robust when checking gophercloud errors in IsNotFound

mdbooth:

As [discussed here](https://github.com/gophercloud/gophercloud/issues/2279) gophercloud is not consistent in whether it returns `error` or `*error`. [IsNotFound](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ba962696a8affa2fde65232ed4d28bf241b5ddab/pkg/utils/errors/errors.go#L35-L36) currently assumes that the error it has been given is not a pointer, and will incorrectly return false if it is a pointer.

Rather than audit CAPO to determine all the possible sources of errors we pass to IsNotFound and checking gophercloud to ensure they are not pointers, this change simply checks both.

/hold
/cc @pierreprinetti

### DIFF
--- a/pkg/utils/errors/errors.go
+++ b/pkg/utils/errors/errors.go
@@ -33,13 +33,15 @@ func IsRetryable(err error) bool {
 }
 
 func IsNotFound(err error) bool {
+	// Gophercloud is not consistent in how it returns 404 errors. Sometimes
+	// it returns a pointer to the error, sometimes it returns the error
+	// directly.
+	// Some discussion here: https://github.com/gophercloud/gophercloud/issues/2279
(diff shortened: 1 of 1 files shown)


mdbooth:
@pierreprinetti you need to join kubernetes-sigs!

In the meantime, please could you review this?

pierreprinetti:
Sadly, this looks correct. I hope we can be more consistent on the Gophercloud side in v2.

Thank you!

/lgtm

tobiasgiese (approved):
/lgtm

thanks for fixing this :+1:

PR #1434, opened 2023-01-06
 🐛 Add support for clouds.yaml cacert file

### LINKED ISSUE #1433: Support clouds.yaml cacert file
/kind bug

`clouds.yaml` has a `cacert` key referencing a path to a ca certificate file.
For now, the provider doesn't use this ca certificate when defined.

Creating a cluster with the following `clouds.yaml` file should work if file provided in `clouds.openstack.cacert` exists. 
```
clouds:
  openstack:
    auth:
      auth_url: "HIDDEN"
      username: "HIDDEN"
      password: "HIDDEN"
      project_id:"HIDDEN"
      project_name: "HIDDEN"
      user_domain_name: "Default"
    cacert: /etc/ssl/certs/ca-certificates.crt
    region_name: "HIDDEN"
    interface: "public"
    identity_api_version: 3
 ```

The field is supported by gophercloud: https://pkg.go.dev/github.com/gophercloud/utils/openstack/clientconfig#Cloud

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.6.4
- Cluster-API version: v1.3.1
- OpenStack version: N/A
- Kubernetes version (use `kubectl version`): 1.25
- OS (e.g. from `/etc/os-release`): ubuntu 22.04

### PR DESCRIPTION
**Pull request by alexandrevilain**

Fixes #1433

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/provider/provider.go
+++ b/pkg/cloud/services/provider/provider.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/http"
+	"os"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
@@ -96,8 +97,17 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
 	if cloud.Verify != nil {
 		config.InsecureSkipVerify = !*cloud.Verify
 	}
+
 	if caCert != nil {
 		config.RootCAs.AppendCertsFromPEM(caCert)
+	}
+
+	if cloud.CACertFile != "" {
+		certfile, err := os.ReadFile(cloud.CACertFile)
+		if err != nil {
+			return nil, nil, "", fmt.Errorf("can't open cloud.yaml cacert file: %w", err)
+		}
+		config.RootCAs.AppendCertsFromPEM(certfile)
 	}
 
 	provider.HTTPClient.Transport = &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}

### CODE REVIEW

**Comment by seanschneeweiss** [comment]
/ok-to-test

Thank you for your contribution.

File: pkg/cloud/services/provider/provider.go [hunk 1]
```diff
@@ -96,10 +97,19 @@ func NewClient(cloud clientconfig.Cloud, caCert []byte) (*gophercloud.ProviderCl
...
 	}
 
+	if cloud.CACertFile != "" {
+		certfile, err := os.ReadFile(cloud.CACertFile)
+		if err != nil {
+			return nil, nil, "", fmt.Errorf("can't open cloud.yaml cacert file: %w", err)
+		}
+		config.RootCAs.AppendCertsFromPEM(certfile)
+	}
+
```
jichenjc, line 112: we do have a cacert setup in line 172 of this file, which is ,we get the cacert from the secret we created 
what's the difference between those code we added with previous logic?

alexandrevilain: Openstack’s clouds.yaml has a cacert key referencing a path to a ca certificate file. 
For now CAPO only uses the ca certificate provided in the secret alongside the clouds.yaml file.
Here it adds support for reading file referenced in clouds.yaml.
It allows users to use containers’ ca certificates or ca certificates mounted from a configmap or a secret in CAPO pod.

alexandrevilain: I provided an example in #1433

mdbooth:
Interesting. So the user is reponsible for providing openstack credentials for their own tenant. This includes a bunch of boilerplate. Taking an example from my clouds.yaml:

```yaml
    cloud29_openshift:
        auth:
            auth_url: https://10.1.39.224:13000
            password: ****************
            project_domain_name: Default
            project_name: openshift
            user_domain_name: Default
            username: openshift
        cacert: /home/mbooth/.config/openstack/cloud29-ca.crt
        identity_api_version: '3'
        interface: public
        region_name: regionOne
        volume_api_version: '3'
```

We could pull out cacert, but except that it's a path, what makes it different from e.g. auth_url in terms of who owns the data? I'm assuming the infrastructure operator concern here is that you might rotate the CA cert and you don't want to have to update all user secrets when you do it? What about changing other aspects of the credentials like endpoints or api versions?

I don't want that to come across as negative, btw. This is an interesting idea. We have other areas in OpenShift where we've effectively removed config options from the user because there's only 1 possible correct value and anything else will break. Could that apply here? What would it look like? Or is a solution covering only cacert sufficient in practise?

alexandrevilain:
>  I'm assuming the infrastructure operator concern here is that you might rotate the CA cert and you don't want to have to update all user secrets when you do it?

That could be a good example.

> because there's only 1 possible correct value and anything else will break

You're right. Maybe a custom dev can be done to inject certificates in secrets to simplify tenants experience.

mdbooth:
Coming back to *this* solution (using a path), though. How would the infrastructure provider use this to provide a common CACert to users in practise?

mdbooth:
I also wondered about supporting referencing a secret in another namespace.

How are operators granting permissions to multi-user management clusters in practise? Would this get complicated with the required RBAC?

alexandrevilain:
> Coming back to this solution (using a path), though. How would the infrastructure provider use this to provide a common CACert to users in practise? 

The use case would be really easy as we're using public endpoints. We would ask the end users to set this value to `/etc/ssl/certs/ca-certificates.crt` which is the path where distroless store ca certificates.

> I also wondered about supporting referencing a secret in another namespace. 

This idea came to my mind too. But looks like a hack to me. 
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  labels:
    capo.x-k8s.io/ca-cert: capo-system/public-ca
type: Opaque
data:
  clouds.yaml: [HIDDEN– 
 ```
 
 Or we can add a field in API:
 ```yaml
 identityRef:
  name: cloud-config
  kind: Secret
caRef:
  name: ca-public
  namespace: capo-system
  kind: Secret
 ```

**Comment by jichenjc** [comment]
CPO actually use (search ca-file key word)
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md

not sure we want to replicate here

**Comment by alexandrevilain** [comment]
Their integration looks fine. That would be great to be able to do the same for CAPO.

But I understand the issue "1 possible correct value and anything else will break". CPO is one time configuration and CAPO is per-cluster configuration.

**Comment by jichenjc** [comment]
>>> CPO is one time configuration and CAPO is per-cluster configuration.

yes, you made a good point here

**Comment by alexandrevilain** [comment]
So the thing I can purpose to you is a flag that infrastructure operators could set at controller's level ?

```go
fs.StringVar(&cacerts, "ca-certs", "", "The path to a ca certificate file to use for each requests")
```

WDYT @jichenjc  & @mdbooth  ?

jichenjc:
Just another thought ..if we don't support multiple openstack in one CAPO , I think we are fine 
not sure we have any case that let's say we have 2 OPenstack, different endpoint so different ca?
if we don't have this kind of use case, I think set ca is good

alexandrevilain:
@jichenjc 

As CAPO already provides `cacert` key alongside `clouds.yaml` in cloud-config secret it already covers the multiple openstacks use case.

Adding the `ca-certs` flag will allow infrastructure operators to support single openstack but many clusters without requiring the `cacert` duplication in each secrets.

jichenjc:
@alexandrevilain ok, works for me ,maybe @mdbooth have further comments

mdbooth:
I like the ca-certs flag. I think its behaviour would be to supply a default cacert if the user doesn't explicitly supply one.

alexandrevilain:
Thanks for you feedback, closing in favor of #1440

PR #1440, opened 2023-01-11
✨ Add ca-cert flag to supply a default ca certificate for each requests

### LINKED ISSUE #1439: Add `--ca-cert` flag to supply default ca certificates in requests.
/kind feature

After discussions on #1434 it has been stated that adding a `--ca-cert` flag could be a good idea.

Actually using CAPO, users have to add the `cacert` key in each secrets for each clusters. It can becomes complex in the case where many clusters are created by many tenants on the same (or public) openstack endpoints. 

Adding the `--ca-cert` will allow infrastructure operators to set default ca certificates to supply to each requests when no `cacert` key is provided in the cloud-config secret alongside the `clouds.yaml`. 

You can follow the whole discussions in #1434 to get more context and most of our reflexions. 
This solution looks better than solution described in #1433

alexandrevilain:

Fixes #1439

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,6 +69,7 @@ var (
 	webhookCertDir              string
 	healthAddr                  string
 	lbProvider                  string
+	caCertsPath                 string
 	logOptions                  = logs.NewOptions()
 )
 
@@ -133,6 +134,8 @@ func InitFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&lbProvider, "lb-provider", "amphora",
 		"The name of the load balancer provider (amphora or ovn) to use (defaults to amphora).")
+
+	fs.StringVar(&caCertsPath, "ca-certs", "", "The path to a PEM-encoded CA Certificate file to supply as default for each request.")
 }
 
 func main() {
@@ -159,6 +162,15 @@ func main() {
 	cfg, err := config.GetConfigWithContext(os.Getenv("KUBECONTEXT"))
 	if err != nil {
 		setupLog.Error(err, "unable to get kubeconfig")
+	}
+
+	var caCerts []byte
+	if caCertsPath != "" {
+		caCerts, err = os.ReadFile(caCertsPath)
+		if err != nil {
+			setupLog.Error(err, "unable to read provided ca certificates file")
+			os.Exit(1)
+		}
 	}
 
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
@@ -191,7 +203,7 @@ func main() {
 	record.InitFromRecorder(mgr.GetEventRecorderFor("openstack-controller"))
 
 	setupChecks(mgr)
-	setupReconcilers(ctx, mgr)
+	setupReconcilers(ctx, mgr, caCerts)
 	setupWebhooks(mgr)
 
 	// +kubebuilder:scaffold:builder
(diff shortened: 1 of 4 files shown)


jichenjc:
/lgtm

follow up we may guide user how to utilize this flag, e.g some doc update etc..

alexandrevilain:
Thanks @jichenjc 

For sure, I'll create a documentation PR about ca-certificate behavior as soon as possible :)

mdbooth:
This is great, thanks! Just trying to think of a way to test it; it doesn't fit into any of our existing test patterns. Any ideas?

jichenjc:
I think we don't have https enabled in CI? maybe we can create devstack with https enabled, and find one of full test (not for every PR) to run https , thus we first cover https test along with this PR (with the flag)?

alexandrevilain:
So you want me to implement e2e test for this feature ?
In a follow up PR alongside documentation ?

jichenjc:
>>>In a follow up PR alongside documentation ?

I think we can use follow up PR , not a must have for this PR

main.go
```diff
@@ -161,6 +164,15 @@ func main() {
 		setupLog.Error(err, "unable to get kubeconfig")
 	}
 
+	var cacerts []byte
+	if cacertsPath != "" {
```
seanschneeweiss, line 168:
```suggestion
	var caCerts []byte
	if caCertsPath != "" {
```

main.go
```diff
@@ -133,6 +134,8 @@ func InitFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&lbProvider, "lb-provider", "amphora",
 		"The name of the load balancer provider (amphora or ovn) to use (defaults to amphora).")
+
+	fs.StringVar(&cacertsPath, "ca-certs", "", "The path to a ca certificate file to supply as default cacert for each requests.")
```
seanschneeweiss, line 138: Might be good to document here, if the secret takes precedence over the default path provided here.
```suggestion
	fs.StringVar(&cacertsPath, "ca-certs", "", "The path to a PEM-encoded CA Certificate file to supply as default for each request.")
```

pkg/cloud/services/provider/provider.go
```diff
@@ -67,6 +72,11 @@ func NewClientFromCluster(ctx context.Context, ctrlClient client.Client, openSta
 			return nil, nil, "", err
 		}
 	}
+
+	if caCert == nil && defaultCACert != nil {
```
seanschneeweiss, line 76: Considerung `defaultCACert == nil` and  `caCert == nil`, it shouldn't matter if we set `caCert = nil` or leave it `nil`.
```suggestion
	if caCert == nil {
```

seanschneeweiss (commented):
Thank you for that contribution. Following the discussion of #1434, providing a default CaCert rather than having to set it in each cluster's secret seems reasonable.

alexandrevilain:
Thanks for the review @seanschneeweiss !
I applied all your comments.

alexandrevilain:
/retest

I don't understand why tests are failing ...

alexandrevilain:
e2e are ok since my rebase 🎉

mdbooth:
/approve

Thanks! This should just need lgtm from somebody now.

PR #1451, opened 2023-01-18
check flavor and reject CPUs less than 2

### LINKED ISSUE #1450: Handle scenario when flavour is too small
/kind bug

When trying to launch a cluster with little number of vCPUs, CAPO does not validate it and therefore the cluster won't build because of `kubeadm` failing with this:

```
[2023-01-17 18:01:50] [preflight] Running pre-flight checks
[2023-01-17 18:01:51] error execution phase preflight: [preflight] Some fatal errors occurred:
[2023-01-17 18:01:51] 	[ERROR NumCPU]: the number of available CPUs 1 is less than the required 2
```

Since this is an expected failure case and the only way to find out that this was broken is by SSH-ing into the system and looking at cloud-init logs, it would be nice for CAPO to simply refuse to spawn the system if the flavour has 1 CPU only.

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

jichenjc:

check flavor and reject flavor with VCPUs < 2 to align with kubeadm checks 

Fixes #1450 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -26,6 +26,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/schedulerhints"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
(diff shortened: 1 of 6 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -126,11 +127,23 @@ func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster,
...
 }
 
-func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluster *infrav1.OpenStackCluster, instanceSpec *InstanceSpec, clusterName string, retryInterval time.Duration) (*InstanceStatus, error) {
+func (s *Service) getAndValidateFlavor(flavorName string, isBastion bool) (*flavors.Flavor, error) {
+	f, err := s.getComputeClient().GetFlavorFromName(flavorName)
+	if err != nil {
+		return nil, fmt.Errorf("error getting flavor from flavor name %s: %v", flavorName, err)
+	}
+	if !isBastion && f.VCPUs <= 1 {
+		return nil, fmt.Errorf("error create machines with flavor's vCPU less than 2, this will cause kubeadm fail to start, update your flavor setting to make VCPUS equal to or bigger than 2")
```
lentzi90, line 140: I found this a bit hard to understand, would suggest something like this instead:
```suggestion
		return nil, fmt.Errorf("kubeadm requires a minimum of 2 vCPUs, pick a flavor with at least 2 vCPUs")
```

jichenjc: ok, will update

lentzi90 (commented):
/lgtm

mnaser (approved):

PR #1460, opened 2023-01-27
🌱 Bump ginkgo to v2.7.0 and sync with hack/tools

lentzi90:

Ginkgo had different versions in the root go.mod and the hack/tools/go.mod. These are now synchronized and bumped to v2.7.0.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


api/v1alpha6/openstackmachinetemplate_webhook_test.go
```diff
@@ -150,6 +150,7 @@ func TestOpenStackMachineTemplate_ValidateUpdate(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
lentzi90, line 153: I don't understand why this showed up now. Make lint was complaining, but for me it was complaining also on `main`, but I don't see it complaining on other PRs. :shrug:

jichenjc: um... this is weird .. but should not block our merge for this PR?

lentzi90: Yes it is fine now I was just unsure why we didn't see it before when running `make lint`. Anyway it is fixed in this PR. :slightly_smiling_face:

mdbooth (approved):

PR #1461, opened 2023-01-27
✨ add explicit securitycontext to controller

tuminoid:
Adding explicit securitycontext ensures the CAPO controller will run as non-root, without special capabilities. Those are often also the defaults but being explicit avoids reliance on fallback values, and enables further hardening by making it possible to apply strict Pod Security Standards for CAPO namespace etc.

In addition, adding seccompProfile of RuntimeDefault adds runtime specific syscall filtering (mostly off-limit by not having capability in the first place) but also couple other, non-namespaced syscalls. In general, very safe set to be filtered.

There is good discussion and reference links in similar CAPI PR at: https://github.com/kubernetes-sigs/cluster-api/pull/7831

Also, reindent the command block properly.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -17,28 +17,40 @@ spec:
         control-plane: capo-controller-manager
     spec:
       containers:
-      - command:
-        - /manager
-        args:
-        - "--leader-elect"
-        - "--v=2"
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

PR #1464, opened 2023-02-01
🐛 Return from reconciler after adding finalizer

mdbooth:
This avoids an extremely surprising write-after-read consistency bug when later returning after making subsequent object changes.

The issue stems from the fact that we don't write to the same place we read from. We read from a cache: a shared informer which is populated from a watch running against the apiserver. However we write directly to the apiserver rather than writing through the cache.

If we patch the watched object without returning, while we continue to execute the change will be written to the api server and will propagate to our own shared informer. This change will result in a new reconcile being queued for this version of the object which now has a Finalizer but no other changes.

Taking the case of the machine controller we will then go on to create a server and we will patch the object with a providerID and return. On return we will write this change to the apiserver which will eventually propagate to our shared informer and result in another reconcile. However our previous patch has already propagated, so we will be *immediately* called with the *old version* of the object. If we've been careful this will hopefully be no more than inefficient, but can lead to very hard to debug errors.

The safe way to patch objects when reconcile is called from an informer is to always return after patching an object which will directly result in another reconcile. We also *must not* set Requeue in the returned Result as this will result in being called again with the stale object. When patching an object we must return and wait for the informer to see the change and call the reconciler again.

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -119,14 +119,14 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// Handle deleted clusters
 	if !openStackCluster.DeletionTimestamp.IsZero() {
-		return reconcileDelete(ctx, scope, patchHelper, cluster, openStackCluster)
+		return reconcileDelete(scope, cluster, openStackCluster)
 	}
 
 	// Handle non-deleted clusters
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/lgtm

pierreprinetti:
Thank you for the fix and even more for the explanation.

lentzi90:
Should we backport to release-0.7? I think it would be nice to get this in a release before the next minor :slightly_smiling_face:

PR #1467, opened 2023-02-06
📖 Add documentation about --ca-cert flag

alexandrevilain:

This PR is a follow up for: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1440 
It adds the documentation about the `--ca-cert` flag added in this PR. 

**Which issue(s) this PR fixes** *

Fixes: #1439

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/docs/book/src/clusteropenstack/configuration.md
+++ b/docs/book/src/clusteropenstack/configuration.md
@@ -8,6 +8,9 @@
   - [SSH key pair](#ssh-key-pair)
   - [OpenStack credential](#openstack-credential)
     - [Generate credentials](#generate-credentials)
+  - [CA certificates](#ca-certificates)
+    - [Per cluster](#per-cluster)
+    - [Global configuration](#global-configuration)
   - [Availability zone](#availability-zone)
   - [DNS server](#dns-server)
   - [Machine flavor](#machine-flavor)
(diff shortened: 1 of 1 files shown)


docs/book/src/clusteropenstack/configuration.md
```diff
@@ -107,6 +110,29 @@ Note: Only the [external cloud provider](https://cluster-api-openstack.sigs.k8s.
 
 Note: you need to set `clusterctl.cluster.x-k8s.io/move` label for the secret created from `OPENSTACK_CLOUD_YAML_B64` in order to successfully move objects from bootstrap cluster to target cluster. See [bug 626](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/626) for further information.
 
+## CA certificates
+
+When using an `https` openstack endpoint, providing CA certificates is mandatory.
```
mdbooth, line 115: nit: You can actually disable it in clouds.yaml with `verify: False` so strictly speaking:

`s/mandatatory/required unless verification is explicitly disabled/`.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -107,6 +110,29 @@ Note: Only the [external cloud provider](https://cluster-api-openstack.sigs.k8s.
...
+To use the same ca certificate for all clusters you can use the `--ca-certs` flag.
+When reconciling a cluster, if no `cacert` is set in the cluster's cloud-config secret, CAPO will use the certicates provided with this flag.
+
+For instance, to use CAPO's docker image ca certificates:
+
+```bash
+kubectl patch deployment capo-controller-manager -n capo-system \
+  --type='json' \
+  -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--ca-certs=/etc/ssl/certs/ca-certificates.crt"}]'
+```
```
mdbooth, line 134: Nice! I wonder if there's a good way to integrate this into our existing templates to make this simpler 🤔

docs/book/src/clusteropenstack/configuration.md
alexandrevilain: Thanks! This is now fixed :)

docs/book/src/clusteropenstack/configuration.md
alexandrevilain: That would be nice! Maybe I'm wrong but as far as I know we can't do conditionals in templates ?

mdbooth: Yeah, I can't think of a way to do it off the top of my head.

PR #1471, opened 2023-02-09
Remove the resource and machine tickers from e2e tests

mdbooth:
The resource and machine tickers are responsible for fetching logs every 5 and 10 seconds respectively during a test run. Each fetch overwrites the previous one. These are problematic because they cause failure state to be overwritten by logs captured during deletion. That is, if a test ends in failure, we continue to capture these logs regularly during cleanup. This causes the test failure state to be overwritten by the state during cleanup.

We already capture these resources after each test regardless of success or failure, so regular capture during the test is redundant. This change simply removes the tickers. We still get the same logs when we execute DumpSpecResourcesAndCleanup() after each test.

Additionally, if there is an error during cleanup we optionally capture that state too, but to a separate directory.

/hold

### DIFF
--- a/test/e2e/shared/common.go
+++ b/test/e2e/shared/common.go
@@ -57,18 +57,34 @@ func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, namespace
 	Logf("Running DumpSpecResourcesAndCleanup for namespace %q", namespace.Name)
 	// Dump all Cluster API related resources to artifacts before deleting them.
 	cancelWatches := e2eCtx.Environment.Namespaces[namespace]
-	dumpSpecResources(ctx, e2eCtx, namespace)
 
-	dumpOpenStack(ctx, e2eCtx, e2eCtx.Environment.BootstrapClusterProxy.GetName())
+	dumpAllResources := func(directory ...string) {
+		dumpSpecResources(ctx, e2eCtx, namespace, directory...)
+		dumpOpenStack(ctx, e2eCtx, e2eCtx.Environment.BootstrapClusterProxy.GetName(), directory...)
 
-	Logf("Dumping all OpenStack server instances in the %q namespace", namespace.Name)
-	dumpMachines(ctx, e2eCtx, namespace)
+		Logf("Dumping all OpenStack server instances in the %q namespace", namespace.Name)
+		dumpMachines(ctx, e2eCtx, namespace, directory...)
+	}
+
+	dumpAllResources()
 
 	if !e2eCtx.Settings.SkipCleanup {
-		framework.DeleteAllClustersAndWait(ctx, framework.DeleteAllClustersAndWaitInput{
-			Client:    e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
-			Namespace: namespace.Name,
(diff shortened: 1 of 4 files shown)


jichenjc:
should we make some error on purpose to prove those files are correctly dumped and get a sample output
then put those files into dev guide for further reference?

lentzi90 (commented):
Nice! The logs are much cleaner with this!
One thing I have noticed (and I don't think it is anything new) is that we don't seem to be able to dump the openstack server instances. Maybe we could fix that at the same time?

Logs look like this:
```
   [2023-02-09T18:25:48Z] Dumping all OpenStack server instances in the "e2e-pe41vo" namespace
  error getting internal ip for server cluster-e2e-dajzgq-control-plane-7kf5p: internal ip doesn't exist (yet)
  error getting internal ip for server cluster-e2e-dajzgq-bastion: internal ip doesn't exist (yet) 
```
Perhaps there is something wrong with how we try to get the IP here
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6645eb9bb4d450c4ea719f8c774b2719a20ac5db/test/e2e/shared/openstack.go#L468-L470

Any clues?

mdbooth:
> should we make some error on purpose to prove those files are correctly dumped and get a sample output then put those files into dev guide for further reference?

Yes, definitely. It's on my todo list to prove that errors during delete are captured. I haven't even tested it manually yet.

mdbooth:
> Nice! The logs are much cleaner with this! One thing I have noticed (and I don't think it is anything new) is that we don't seem to be able to dump the openstack server instances. Maybe we could fix that at the same time?
> 
> Logs look like this:
> 
> ```
>    [2023-02-09T18:25:48Z] Dumping all OpenStack server instances in the "e2e-pe41vo" namespace
>   error getting internal ip for server cluster-e2e-dajzgq-control-plane-7kf5p: internal ip doesn't exist (yet)
>   error getting internal ip for server cluster-e2e-dajzgq-bastion: internal ip doesn't exist (yet) 
> ```
> 
> Perhaps there is something wrong with how we try to get the IP here
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6645eb9bb4d450c4ea719f8c774b2719a20ac5db/test/e2e/shared/openstack.go#L468-L470
> 
> Any clues?

I've also wondered about this. It's on my TODO list to investigate properly.

mdbooth:
We correctly dumped the error on deletion in:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1471/pull-cluster-api-provider-openstack-e2e-test/1626172319116300288

Note that there is an additional directory created for the deletion failure:

https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1471/pull-cluster-api-provider-openstack-e2e-test/1626172319116300288/artifacts/clusters/bootstrap/resources/deletion-failure/e2e-ptzr4v/

Incidentally, I deliberately added the failure after cleanup had succeeded, so the contents of that directory are 'leaked' objects.

lentzi90 (commented):
/lgtm

PR #1470, opened 2023-02-09
Download golangci-lint instead of building it

mdbooth:
Works round some go mod dependency issues upgrading to v1.51.1, and is in line with what CAPI does.

Fixes #1469

PR #1473, opened 2023-02-15
Fix boilerplate linter

mdbooth:
Fixes a bug introduced by #1470 which meant that the boilerplate linter was no longer checking anything.

/hold


tobiasgiese (approved):
/lgtm

PR #1474, opened 2023-02-16
Fix "internal ip doesn't exist (yet)" in e2e logs

mdbooth:
We see lots of instances of log messages in e2e logs of the form:
```
error getting internal ip for server cluster-e2e-l7xvyz-control-plane-hnj96: internal ip doesn't exist (yet)
```

These are because GetOpenStackServers(), which is called after every test completes, fetches *all* OpenStack servers, not just those for the current cluster. The above log message is displayed when it doesn't find an interface for the server on the current cluster's internal network.

This change cuts down the noise by passing in an explicit list of machine names to return. We still fetch all servers, but we filter them after fetching.

/hold

### DIFF
--- a/test/e2e/shared/common.go
+++ b/test/e2e/shared/common.go
@@ -31,6 +31,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/cluster-api/util"
@@ -111,7 +112,12 @@ func dumpMachines(ctx context.Context, e2eCtx *E2EContext, namespace *corev1.Nam
 		_, _ = fmt.Fprintf(GinkgoWriter, "cannot dump machines, could not get machines: %v\n", err)
 		return
 	}
-	srvs, err := GetOpenStackServers(e2eCtx, cluster)
+
+	machineNames := sets.NewString()
+	for _, machine := range machines.Items {
+		machineNames.Insert(machine.Name)
+	}
+	srvs, err := GetOpenStackServers(e2eCtx, cluster, machineNames)
 	if err != nil {
 		_, _ = fmt.Fprintf(GinkgoWriter, "cannot dump machines, could not get servers from OpenStack: %v\n", err)
 		return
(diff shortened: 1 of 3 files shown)


mdbooth:
This is a dump of a server which isn't being correctly parsed:
```
{
    ID:256891de-868d-4522-be51-3f1e63186c3c
    TenantID:c333267e6c4b4111a0365bfadd4b728f
    UserID:9fe773355d3b4790b3b375aca840091d
    Name:cluster-e2e-6ejowb-control-plane-ntn22
    Updated:2023-02-16 11:52:53 +0000 UTC
    Created:2023-02-16 11:52:51 +0000 UTC
    HostID:b0e53a68fa576c6d20fcec49ca23c60bc7d3cef9c67128023d6202fa
    Status:ACTIVE
    Progress:0
    AccessIPv4:
    AccessIPv6:
    Image:map[id:453675fd-016d-42f0-b313-a02afe0bb901 links:[map[href:http://10.0.3.15/compute/images/453675fd-016d-42f0-b313-a02afe0bb901 rel:bookmark]]]
    Flavor:map[id:3 links:[map[href:http://10.0.3.15/compute/flavors/3 rel:bookmark]]]
    Addresses:map[
        k8s-clusterapi-cluster-e2e-6ejowb-cluster-e2e-6ejowb:[
                map[OS-EXT-IPS-MAC:mac_addr:fa:16:3e:bc:94:99 OS-EXT-IPS:type:fixed addr:10.6.0.57 version:4]
                map[OS-EXT-IPS-MAC:mac_addr:fa:16:3e:bc:94:99 OS-EXT-IPS:type:floating addr:172.24.4.30 version:4]
        ]
    ]
    Metadata:map[]
    Links:[map[href:http://10.0.3.15/compute/v2.1/servers/256891de-868d-4522-be51-3f1e63186c3c rel:self] map[href:http://10.0.3.15/compute/servers/256891de-868d-4522-be51-3f1e63186c3c rel:bookmark]]
    KeyName:cluster-api-provider-openstack-sigs-k8s-io
    AdminPass:
    SecurityGroups:[map[name:k8s-cluster-e2e-6ejowb-cluster-e2e-6ejowb-secgroup-controlplane]]
    AttachedVolumes:[]
    Fault:{Code:0 Created:0001-01-01 00:00:00 +0000 UTC Details: Message:}
    Tags:<nil>
    ServerGroups:<nil>
}
```

mdbooth:
@lentzi90 They're servers from the other cluster, so they don't have an interface on the right network! It's just noise. Lets try to suppress it by filtering the returned servers.

lentzi90:
> @lentzi90 They're servers from the other cluster, so they don't have an interface on the right network! It's just noise. Lets try to suppress it by filtering the returned servers.

Thanks for checking and good find!

lentzi90 (approved):

PR #1479, opened 2023-02-17
:bug: Fix Provisioning to Unavailable AZs

### LINKED ISSUE #1477: AZ Availabilty Not Being Checked
/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/02a31730ef1ae7962f1cc36c090b624025a2f5bc/pkg/cloud/services/compute/availabilityzone.go#L25 or https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/f4e61b626eec3c5eb82e5f3328de0da6fe1d540d/pkg/clients/compute.go#L79 don't filter out unavailable availability zones.

It filters unavailable zones out so the control plane etc. gets scheduled on an available AZ.

We could/should work around this problem by explicitly stating the AZs in the OSC resource, but that's not available to us as we punt everything through a single Helm chart, adding the field gets rejected because of the validating webhook for existing clusters.  If we try to switch the OSC out with a new one, that gets really broken as it tries to allocate a new VIP on the load balancer, fails as the old one is in place, uses up all the floating IPs on the platform in the process.  Disassociating the VIP to lend a helping hand is even worse because everything loses the ability to talk to the API.  Non fun 😸 

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

spjmurray:
CAPO uses any ZA it finds, rather than checking whether the ZA is actually available, meaning you can get into a state where it's impossible to provision a cluster.  This allows clients to filter based on AZ availability before scheduling.

Fixes #1477

Discussion in #1477 

Fixes #

**TODOs**:

- [x ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha6/openstackcluster_webhook.go
+++ b/api/v1alpha6/openstackcluster_webhook.go
@@ -115,6 +115,10 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		r.Spec.APIServerLoadBalancer.AllowedCIDRs = []string{}
 	}
 
+	// Allow changes to the availability zones.
+	old.Spec.ControlPlaneAvailabilityZones = []string{}
+	r.Spec.ControlPlaneAvailabilityZones = []string{}
+
 	if !reflect.DeepEqual(old.Spec, r.Spec) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
 	}
--- a/pkg/cloud/services/compute/availabilityzone.go
+++ b/pkg/cloud/services/compute/availabilityzone.go
@@ -28,5 +28,13 @@ func (s *Service) GetAvailabilityZones() ([]availabilityzones.AvailabilityZone, 
 		return nil, fmt.Errorf("error extracting availability zone list: %v", err)
 	}
 
-	return availabilityZoneList, nil
+	available := make([]availabilityzones.AvailabilityZone, 0, len(availabilityZoneList))
+
+	for _, az := range availabilityZoneList {
+		if az.ZoneState.Available {
+			available = append(available, az)
+		}
+	}
+
+	return available, nil
 }
--- a/api/v1alpha6/openstackcluster_webhook_test.go
+++ b/api/v1alpha6/openstackcluster_webhook_test.go
@@ -175,6 +175,65 @@ func TestOpenStackCluster_ValidateUpdate(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name: "Adding OpenStackCluster.Spec.ControlPlaneAvailabilityZones is allowed",
+			oldTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+				},
+			},
+			newTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+					ControlPlaneAvailabilityZones: []string{
+						"alice",
+						"bob",
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "Modifying OpenStackCluster.Spec.ControlPlaneAvailabilityZones is allowed",
+			oldTemplate: &OpenStackCluster{
+				Spec: OpenStackClusterSpec{
+					CloudName: "foobar",
+					ControlPlaneAvailabilityZones: []string{
+						"alice",
+						"bob",
+					},
+				},
+			},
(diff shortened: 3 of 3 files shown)


lentzi90 (commented):
/ok-to-test

pkg/cloud/services/compute/availabilityzone.go
```diff
@@ -30,3 +30,21 @@ func (s *Service) GetAvailabilityZones() ([]availabilityzones.AvailabilityZone,
 
 	return availabilityZoneList, nil
 }
+
+// GetAvailableAvailabilityZones does as it says, but only returns usable AZs.
+func (s *Service) GetAvailableAvailabilityZones() ([]availabilityzones.AvailabilityZone, error) {
+        availabilityZoneList, err := s.getComputeClient().ListAvailabilityZones()
```
lentzi90, line 36: Maybe slightly better to use the function above. Just a suggestion
```suggestion
        availabilityZoneList, err := s.GetAvailabilityZones()
```

api/v1alpha6/openstackcluster_webhook.go
```diff
@@ -115,6 +115,10 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) error {
 		r.Spec.APIServerLoadBalancer.AllowedCIDRs = []string{}
 	}
 
+	// Allow changes to the availability zones.
+	old.Spec.ControlPlaneAvailabilityZones = []string{}
+	r.Spec.ControlPlaneAvailabilityZones = []string{}
+
```
mdbooth, line 121: Whoops! That's quite an oversight.

Please could you add unit tests for these to `openstackcluster_webhook_test.go`?

pkg/cloud/services/compute/availabilityzone.go
```diff
@@ -30,3 +30,21 @@ func (s *Service) GetAvailabilityZones() ([]availabilityzones.AvailabilityZone,
 
 	return availabilityZoneList, nil
 }
+
+// GetAvailableAvailabilityZones does as it says, but only returns usable AZs.
+func (s *Service) GetAvailableAvailabilityZones() ([]availabilityzones.AvailabilityZone, error) {
```
mdbooth, line 35: Couple of things:

Firstly there's only 1 caller of GetAvailabilityZones() and this PR replaces it with a call to this new function, leaving the old one as dead code. We should just modify GetAvailabilityZones() to do what we want.

Secondly, I'm wondering what the impact of removing the FD from cluster.Status.FailureDomains will be in the context of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1466. In that PR I'm modifying the machine controller to actually look up the failure domain on first use. With this change, if a MachineDeployment referenced an AZ which became unavailable, the cluster controller would eventually delete it. Prior to that the machine would be created but would be unschedulable and enter the ERROR state. After that the machine controller would refuse to create it in the first place and enter a backoff loop waiting for it to exist.

Meh, don't worry about the second one: it's not in scope of your PR. In general, it points out that we have some baked assumptions about AZs being static that aren't necessarily true. We should likely ensure that the cluster controller runs periodically even without k8s watch events. Perhaps every 10 minutes or so.

mdbooth (changes requested):
Looks good in principal. Just need to update the unit tests for the webhook change and to avoid creating dead code in availabilityzones.go.

pkg/cloud/services/compute/availabilityzone.go
spjmurray: I figured it best to leave both versions, then the caller man make the decision whether they want to consider available ones, or not.  Combining the two views you can derive either {available, not available, non-existent}.  But I'm happy to clean up.

I thought the AZ mutability would possibly raise some eye-brows 😄  I figured, seeing as it's implicitly refreshed every reconcile, then technically speaking it's already mutable, which as you say reacts nicely to AZs vanishing.

api/v1alpha6/openstackcluster_webhook.go
spjmurray: roger!

pkg/cloud/services/compute/availabilityzone.go
spjmurray: Followed Matt's advice below.

mdbooth:
/approve

Thanks! I'll leave lgtm to another reviewer.

lentzi90 (commented):
/lgtm

mdbooth:
@spjmurray If you cancel the hold this will merge.

PR #1483, opened 2023-02-17
Output junit from ginkgo

mdbooth:
/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,8 @@ E2E_GINKGO_ARGS ?=
 .PHONY: test-e2e ## Run e2e tests using clusterctl
 test-e2e: e2e-prerequisites ## Run e2e tests
 	time $(GINKGO) -fail-fast -trace -timeout=3h -show-node-events -v -tags=e2e -nodes=$(E2E_GINKGO_PARALLEL) \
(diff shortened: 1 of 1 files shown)


mdbooth:
That didn't work. Presumably there's another piece 🤔

mdbooth:
@lentzi90 This is subterranean priority, but if you happen to know what pieces need to go together for this I'm interested to know. I'm on a slow journey towards making the e2e test output more readable!

lentzi90:
Not sure why that didn't work. Maybe it just ends up outside of the _artifacts folder? Worth a try with `_artifacts/report.xml` I guess :shrug: 

Locally I get reports in the repo root with this:

```
export E2E_GINKGO_ARGS="--json-report=report.json --junit-report=report.xml"
make test-e2e
```

lentzi90:
I think that's it actually! I can see the `capo-e2e-image.tar` in the artifacts (which maybe we should not put there) and how it ends up there is simply this:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ff4c1de60c31e27f548529289b0c2f876f43a1ee/scripts/ci-e2e.sh#L89

PR #1482, opened 2023-02-17
🐛 uplift golang and x/net

tuminoid:
Uplift golang to 1.19.6 and x/net to 0.7.0 due https://osv.dev/vulnerability/GO-2023-1571

PR #1486, opened 2023-02-21
🐛 [release-0.7] uplift go and x/net

lentzi90:

Backport of #1482 and #1470


lentzi90:
Ok now it found the same linting issue that suddenly appeared #1460. Fixing...

api/v1alpha6/openstackmachinetemplate_webhook_test.go
```diff
@@ -150,6 +150,7 @@ func TestOpenStackMachineTemplate_ValidateUpdate(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
mdbooth, line 153: This is fine, but backporting a PR which changes api/ is surprising.

It would be good to move most code out of api which isn't type definitions at some point.

PR #1497, opened 2023-02-28
⚠️ Add v1alpha7

mdbooth:
Add a new v1alpha7 API version in anticipation of breaking API changes.

/hold


api/v1alpha6/conversion.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
```
lentzi90, line 2: Not sure where this came from. Maybe change it to 2023 instead?

lentzi90 (commented):
Looks good!
One nit below, and one question: Should we change the clusterctl upgrade test to initialize with v1alpha6 instead of v1alpha5 now?
Ref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1e1847406a99b35f5002f8c3a00aa85401eee434/test/e2e/suites/e2e/clusterctl_upgrade_test.go#L52

api/v1alpha6/conversion.go
mdbooth: That will be after I deleted it and then copied it from elsewhere after realising that was a mistake 😂

mdbooth:
> Looks good! One nit below, and one question: Should we change the clusterctl upgrade test to initialize with v1alpha6 instead of v1alpha5 now? Ref:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1e1847406a99b35f5002f8c3a00aa85401eee434/test/e2e/suites/e2e/clusterctl_upgrade_test.go#L52

I would defer this to you. We could:
* Leave it alone (test v1alpha5->v1alpha7)
* Update it to v1alpha6 (test v1alpha6->v1alpha7)
* Add a second test and do both

lentzi90:
> I would defer this to you. We could:
> 
>     * Leave it alone (test v1alpha5->v1alpha7)
> 
>     * Update it to v1alpha6 (test v1alpha6->v1alpha7)
> 
>     * Add a second test and do both

Probably the nicest thing would be to add a second test and test both. But I would say it is more important to have v1alpha6 -> v1alpha7.

lentzi90 (commented):
/lgtm

lentzi90:
Ah missed that the e2e-full test was failing...
It is because the e2e config is missing the v0.7.1 release that it is trying to use now. Basically the v0.6.4 section can be replaced with v0.7.1 here
https://github.com/shiftstack/cluster-api-provider-openstack/blob/28533e5a2b06d703781678f606507f3f64c66bf7/test/e2e/data/e2e_conf.yaml#L77-L91

mdbooth:
> Ah missed that the e2e-full test was failing... It is because the e2e config is missing the v0.7.1 release that it is trying to use now. Basically the v0.6.4 section can be replaced with v0.7.1 here https://github.com/shiftstack/cluster-api-provider-openstack/blob/28533e5a2b06d703781678f606507f3f64c66bf7/test/e2e/data/e2e_conf.yaml#L77-L91

Thanks! I've added a second section. Is that going to work?

lentzi90:
Having both v0.6.4 and v0.7.1 is perfectly fine, but now I see two sections with v0.7.99. Not sure how that will go :thinking:

mdbooth:
> Having both v0.6.4 and v0.7.1 is perfectly fine, but now I see two sections with v0.7.99. Not sure how that will go 🤔

🤦

lentzi90 (commented):
/lgtm

furkatgofurov7 (commented):
This looks good to me, only one nit I could notice is missing the API types update in the doc: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/clusteropenstack/configuration.md

mdbooth:
I'll let this -full test complete before pushing a docs update. Assuming everything is ok I won't bother running it again.

furkatgofurov7:
> I'll let this -full test complete before pushing a docs update. Assuming everything is ok I won't bother running it again.

@mdbooth makes sense 🙂 
 Although I see `pull-cluster-api-provider-openstack-e2e-test ` is run in every and each PR which does not really makes sense to me (i.e for only doc update changes in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1500). 
We could add `run_If_changed` on at least ^ e2e presubmit job (example: https://github.com/kubernetes/test-infra/blob/464e4bbc73d0934491fb285eca8719e22a173adc/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-release-1-3.yaml#L125) to ensure it is not run if only e.g. documentation is changed.

mdbooth:
Only change is updating the api version refs in the docs.

lentzi90 (commented):
Thanks!
/lgtm

furkatgofurov7 (commented):
/lgtm

Not sure if the repo has requirements for the PR labels, but this is generally should be labelled as ⚠️ change to me.

PR #1468, opened 2023-02-07
WIP: ✨ Use NewServiceClient to create clients

lentzi90:
**What this PR does / why we need it**:

WIP: Trying to rely more on gophercloud for config parsing.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #1498, opened 2023-03-01
🐛 fix: allow using multiattach volume types

### LINKED ISSUE #1496: VMs fail to go up with mutliattach backends
/kind bug

CAPO fails to boot up VMs when a backend supports multi-attach, this is because the API microversion that's set in CAPO is 2.53.  2.60 allows the use of multiattach backends and it's the maxmium in Queens, a release that was EOL'd last month.

I expected CAPO to be able to bring up a VM just fine.

I think we should be able to safely bump the microversion to 2.60 without any impact but an advantage.

https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#maximum-in-queens
https://releases.openstack.org/

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

mnaser:
This bumps the minimum supported micro-version to one that would support mulit-attach, but also happens to be the most recent EOL release of OpenStack

Fixes #1496

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Each version of Cluster API for OpenStack will attempt to support two Kubernetes
 **NOTE:** As the versioning for this project is tied to the versioning of Cluster API, future modifications to this
 policy may be made to more closely aligned with other providers in the Cluster API ecosystem.
 
-**NOTE:** The minimum microversion of CAPI using nova is `2.53` now due to `server tags` support, see [code](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c052e7e600f0e5ebddc839c08746bb636e79be87/pkg/cloud/services/compute/service.go#L38) for additional information.
+**NOTE:** The minimum microversion of CAPI using nova is `2.60` now due to `server tags` support as well permitting `multiattach` volume types, see [code](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c052e7e600f0e5ebddc839c08746bb636e79be87/pkg/cloud/services/compute/service.go#L38) for additional information.
 
 **NOTE:** We require Keystone v3 for authentication.
 
--- a/pkg/clients/compute.go
+++ b/pkg/clients/compute.go
@@ -33,14 +33,15 @@ import (
 
 /*
 NovaMinimumMicroversion is the minimum Nova microversion supported by CAPO
-2.53 corresponds to OpenStack Pike
+2.60 corresponds to OpenStack Queens
 
 For the canonical description of Nova microversions, see
 https://docs.openstack.org/nova/latest/reference/api-microversion-history.html
 
 CAPO uses server tags, which were added in microversion 2.52.
+CAPO supports multiattach volume types, which were added in microversion 2.60.
 */
-const NovaMinimumMicroversion = "2.53"
+const NovaMinimumMicroversion = "2.60"
 
 // ServerExt is the base gophercloud Server with extensions used by InstanceStatus.
 type ServerExt struct {


mnaser:
looks like it works, the main issue behind the failure was a linter fail :)

lentzi90:
While I'm happy to see that this works (we also need it in some places!), I'm not so happy about upping the required version. There are ancient openstack instances out there still. :disappointed: 

I have a draft PR (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1468) where I have been experimenting with allowing gophercloud to pick the version instead of setting it directly. The good thing with this approach is that we would always get the highest possible supported version. The drawback is that we cannot really set a minimum version then, rather we can check what gophercloud returned and error out if it was too old. A side effect is that the user can set a desired version in `clouds.yaml`.

mnaser:
unfortunately, i don't think i have the bandwidth to help with that even though i'd love to, and we're stuck in a situtation where all cluster api functionality was broken by enabling multiattach...

jichenjc:
>>>While I'm happy to see that this works (we also need it in some places!), I'm not so happy about upping the required version. There are ancient openstack instances out there still

I think we can do this in a separate PR :) 

/lgtm

mdbooth:
> While I'm happy to see that this works (we also need it in some places!), I'm not so happy about upping the required version. There are ancient openstack instances out there still. 😞
> 
> I have a draft PR (#1468) where I have been experimenting with allowing gophercloud to pick the version instead of setting it directly. The good thing with this approach is that we would always get the highest possible supported version. The drawback is that we cannot really set a minimum version then, rather we can check what gophercloud returned and error out if it was too old. A side effect is that the user can set a desired version in `clouds.yaml`.

Note that we definitely don't want to negotiate the latest available microversion in CAPO. This is like using a `:latest` image in production: the interface can change at any time. This could be a case of fields being added or removed, or more subtly semantic changes. It's just asking for trouble.

We need to be deliberate in our selection of microversions, but this doesn't mean we can't support more than one.

lentzi90:
> Note that we definitely don't want to negotiate the latest available microversion in CAPO. This is like using a `:latest` image in production: the interface can change at any time. This could be a case of fields being added or removed, or more subtly semantic changes. It's just asking for trouble.
> 
> We need to be deliberate in our selection of microversions, but this doesn't mean we can't support more than one.

Thanks for the heads up! I got the impression that the client should always try to use the latest version it knows about, from reading [this](https://docs.openstack.org/openstacksdk/latest/user/microversions.html). But I guess it just means we should define some version (as high as possible) and use that?
Anyway, I don't want to hijack the thread here. I'll come back with some questions and suggestions on the draft PR when I have time to work on it again :slightly_smiling_face:

mdbooth:
@jichenjc @lentzi90 Without new development work this feels like a trade-off between supporting current versions of OpenStack and EOL versions of OpenStack. I feel like we should prioritise current versions.

How about if we approved this and looked at adding support for EOL OpenStack if somebody squeals about it? But if folks are OK with not upgrading, I guess they can always run an old version of CAPO.

I'm inclined to approve a microversion bump, but I'll hold off if there are strong objections.

jichenjc:
>>>I'm inclined to approve a microversion bump, but I'll hold off if there are strong objections.

yes, I think we should go new microversion for now and if someone comes to us ( at least I was asked in CPO several times)
we can consider a fix or a workaround for them to continue use CAPO

lentzi90 (approved):
I'm happy to take it in as is also. As mentioned, we need it! :smile: 
But I will keep working on a proposal for how to support lower versions. We have a use case for even lower versions than what is currently supported. :grimacing:

mnaser:
> I'm happy to take it in as is also. As mentioned, we need it! :smile: 
> 
> But I will keep working on a proposal for how to support lower versions. We have a use case for even lower versions than what is currently supported. :grimacing: 

Ouch, sorry to hear that! 😅

PR #1504, opened 2023-03-13
⚠️Remove OpenStackMachineSpec.Subnet

mdbooth:
If set, OpenStackMachineSpec.Subnet must correspond to the subnet of at least 1 port on the machine. The IP in this subnet will be set as AccessIPv4 on the OpenStack server.

There are two principal problems with this API:
* Users don't appear to be using it
* It doesn't support IPv6

Supporting it adds complexity in code I would like to remove, so I propose removing it entirely.

/hold


mdbooth:
Tests blocked by https://github.com/kubernetes/test-infra/issues/29003

dulek:
How do we achieve the same thing that this API offered without the field?

mdbooth:
> How do we achieve the same thing that this API offered without the field?

We don't: the functionality is removed. However, the premise of this change is that the functionality isn't useful. Do we know of anybody using it?

dulek:
What will be set as `AccessIPv4` by default then? What effect does it have on the server? It's used by `openstack server ssh`?

mdbooth:
> What will be set as `AccessIPv4` by default then? What effect does it have on the server? It's used by `openstack server ssh`?

I don't think anybody's using it in practise. Even if they are it's an unnecessarily complex way to specify it and it doesn't support dual stack. I doubt we would bother, but if we did I'd prefer to implement a simpler way of achieving it in v1beta1.

dulek:
Okay, so https://codesearch.opendev.org/?q=access_ip_v4&i=nope&literal=nope&files=.*py&excludeFiles=&repos= tells me that the field has no real logic in Nova besides saving data. It has some in openstacksdk, but I don't think we care too much.

/lgtm

api/v1alpha6/conversion.go
```diff
@@ -17,24 +17,52 @@ limitations under the License.
...
+	// Store the original source object in the dest object's annotations.
+	if err := utilconversion.MarshalData(src, dst); err != nil {
+		return false, err
 	}
 
-	// Manually restore data.
-	restored := &infrav1.OpenStackCluster{}
-	if ok, err := utilconversion.UnmarshalData(r, restored); err != nil || !ok {
+	return restored, nil
+}
```
mdbooth, line 51: @lentzi90 This is the change to the conversion functions I was talking about.

With this in place I don't think we will have to ignore any fields in the fuzzer tests because all fields should either be losslessly convertible or restored from the stashed value.

api/v1alpha6/conversion_test.go
```diff
@@ -19,29 +19,98 @@ package v1alpha6
...
-		Hub:   &infrav1.OpenStackMachineTemplate{},
-		Spoke: &OpenStackMachineTemplate{},
+		Hub:              &infrav1.OpenStackMachineTemplate{},
+		Spoke:            &OpenStackMachineTemplate{},
+		HubAfterMutation: ignoreDataAnnotation,
 	}))
 }
+
+// Test that mutation of a converted object survives a subsequent conversion.
+func TestMutation(t *testing.T) {
```
mdbooth, line 64: @lentzi90 This is the other test I was talking about which ensures that we don't simply overwrite an object with the stashed version on conversion.

api/v1alpha6/conversion.go
lentzi90: Interesting! The first use of golang generics I have come across in the wild :smile:

mdbooth:
@dulek This version just fixes api conversion from the version you previously reviewed.

PR #1511, opened 2023-03-22
🌱 Un-pointer PortOpts.SecurityGroups

mdbooth:
This changes `PortOpts.SecurityGroups` from `*[]string` to `[]string`. Note that this make no changes to the CRDs. As it only affects Go marshalling it's effectively an internal change.

We make the change to all historic API versions to simplify conversion. It isn't a change to those APIs because their CRDs remain unchanged.

/hold

### DIFF
--- a/api/v1alpha4/types.go
+++ b/api/v1alpha4/types.go
@@ -132,7 +132,7 @@ type PortOpts struct {
 	FixedIPs            []FixedIP     `json:"fixedIPs,omitempty"`
 	TenantID            string        `json:"tenantId,omitempty"`
 	ProjectID           string        `json:"projectId,omitempty"`
-	SecurityGroups      *[]string     `json:"securityGroups,omitempty"`
+	SecurityGroups      []string      `json:"securityGroups,omitempty"`
 	AllowedAddressPairs []AddressPair `json:"allowedAddressPairs,omitempty"`
 	// Enables and disables trunk at port level. If not provided, openStackMachine.Spec.Trunk is inherited.
 	Trunk *bool `json:"trunk,omitempty"`
--- a/api/v1alpha4/zz_generated.conversion.go
+++ b/api/v1alpha4/zz_generated.conversion.go
@@ -1474,7 +1474,7 @@ func autoConvert_v1alpha4_PortOpts_To_v1alpha7_PortOpts(in *PortOpts, out *v1alp
 	}
 	out.TenantID = in.TenantID
 	out.ProjectID = in.ProjectID
-	out.SecurityGroups = (*[]string)(unsafe.Pointer(in.SecurityGroups))
+	out.SecurityGroups = *(*[]string)(unsafe.Pointer(&in.SecurityGroups))
 	out.AllowedAddressPairs = *(*[]v1alpha7.AddressPair)(unsafe.Pointer(&in.AllowedAddressPairs))
 	out.Trunk = (*bool)(unsafe.Pointer(in.Trunk))
 	out.HostID = in.HostID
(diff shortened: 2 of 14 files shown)


mdbooth:
I'm hitting this in the networks removal PR. For PortOpts.SecurityGroups the fuzzer creates `nil` and `&nil` separately. However when marshalled to json and back these both become the same value: `nil`. This breaks the test, but it also highlights that we can't actually use this datatype in the API. Rather than create complexity in the test, this PR removes complexity in the type marshalling.

mdbooth:
This would conflict with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1291, but as it's also simpler it might be ok to merge it first.

jichenjc:
seems test need update as it's compare pointer and nil now .. fix it then we are good to go

lentzi90:
Makes sense to me!

dulek:
Make sense, but out of curiosity - was there any argument to do `*[]string` in the first place?

mdbooth:
> Makes sense, but out of curiosity - was there any argument to do `*[]string` in the first place?

Gophercloud takes *[]string in PortOpts, so I suspect it was just copying that.

mdbooth:
Tests are now passing. As discussed above, the API type changes are just mechanical: I changed all the types and re-ran `make generate`. This updated deepcopy and conversion, and verified that there was no CRD change.

The code changes require close inspection.

In compute/instance.go we're changing the argument passed to GetOrCreatePort from &securityGroups to securityGroups. Note that:
* This is the only caller of GetOrCreatePort
* The value could previously never have been nil (but it could have been &nil).

In networking/port.go we're making the argument type change to GetOrCreatePort, which we now know could never have been nil so was already equivalent to []string.

Next is the call to collectPortSecurityGroups. Firstly I made it internal as it has no external callers. Note that it previously couldn't return nil except on error. The 'least' it could previously return was `&[]`, so not even `&nil`. I added a shortcut to the top of the method which means it can now return nil. Perhaps I shouldn't have done that. Opinions appreciated.

This means that in GetOrCreatePort when using the return from collectPortSecurityGroups we now check for nil before inheriting instanceSecurityGroups. This means that if the user passes an explicitly empty list of either SecurityGroups or SecurityGroupFilters then the port will be created with no SecurityGroups. If they specify nothing, we inherit. Previously, I think it was impossible to specify no security groups. Again, possibly mixing an unwanted functional change here, but I was already looking at the code...

Lastly, to preserve the previous value passed to gophercloud we translate a nil slice into an empty pointer.

mdbooth:
Never mind, writing that rubber-ducked me out of the idea. I've removed the change in behaviour such that it is still not possible to specify a port with no security groups. I'm not sure a port with no security groups makes sense anyway, tbh.

lentzi90 (commented):
This looks good to me!
/lgtm
I would like to point out that the changes to the older API versions still requires some caution though. It is true that it won't affect the CRDs, but if someone is using the CAPO go modules it would affect them and require code changes for new releases even when using older API versions.
From our side it is good to go, but I'll add hold just in case there are more comments from others.
/hold

mdbooth:
> This looks good to me! /lgtm I would like to point out that the changes to the older API versions still requires some caution though. It is true that it won't affect the CRDs, but if someone is using the CAPO go modules it would affect them and require code changes for new releases even when using older API versions. From our side it is good to go, but I'll add hold just in case there are more comments from others. /hold

This is true, but I suspect the only people crazy enough to do such a thing are my team at Red Hat 😉 We also won't be broken except when we do a version bump, and we expect a certain amount of breakage doing that.

lentzi90:
> This is true, but I suspect the only people crazy enough to do such a thing are my team at Red Hat wink We also won't be broken except when we do a version bump, and we expect a certain amount of breakage doing that.

Same here! :wink:

mdbooth:
> > This is true, but I suspect the only people crazy enough to do such a thing are my team at Red Hat wink We also won't be broken except when we do a version bump, and we expect a certain amount of breakage doing that.
> 
> Same here! 😉

I had no idea! How much of a problem do you think this is? I could undo it: the conversion functions should be trivial. I just didn't want to write them if nobody was likely to benefit from it.

lentzi90:
It is fine for us! We all agreed this is a good change, just go ahead :slightly_smiling_face:

PR #1513, opened 2023-03-23
📖 Docs: Update link to NetworkParam codebase.

dalees:

The existing link to NetworkParam struct in Network Filters documentation section is broken, as there is no v1beta1 path in the codebase.

We could consider permalink - but likely to get outdated.

**TODOs**:

- [ x] squashed commits
- if necessary:
  - [ x] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/docs/book/src/clusteropenstack/configuration.md
+++ b/docs/book/src/clusteropenstack/configuration.md
@@ -279,7 +279,7 @@ openstack loadbalancer listener unset --allowed-cidrs <listener ID>
 
 ## Network Filters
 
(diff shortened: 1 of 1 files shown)


dulek:
Hm, it's strongly sub-optimal to update the reference each time we create a new API version. I would just add JSON structure of the filter with an example instead.

mdbooth:
Note that I'm in the process of removing NetworkParam from the API in v1alpha7!

Please use ports instead.

dalees:
Closing, this isn't the right change, and `NetworkParam` is being removed.

Thank you to the reviewers for the feedback.

PR #1514, opened 2023-03-27
⚠️Deprecate In-tree Cloud Provider

ching-kuo:

In-tree cloud provider has been marked deprecated in Kubernetes for a long time. In this commit, we make external cloud provider default and provide a seperate template if users would like to use in-tree cloud provider.

Fixes #

1. This change will change the default cluster-template to use external cloud provider, while e2e test will also change to test external cloud provider as default.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


jichenjc:
thanks for the PR, this is right direction
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/topics/external-cloud-provider.md
there are some manual steps (RBAC, CNI etc) , does this handled in the test as well?

ching-kuo:
There's already a e2e test for external cloud provider which includes CNIs and CCM installation before I made the changes, the changes I made is to change the references for the kustomize files.

I haven't test the e2e test locally yet, that's why I've marked it as WIP

jichenjc:
yes, I knew, but I am just curious whether we should setup those steps in order to make it a full functional cluster..

ching-kuo:
As far as I know, you will need to install CNIs and CCM to make the cluster ready in `clusterctl decribe cluster <cluster_name>` output.

CCM will be needed as CNIs will not schedule until CCM is installed when cloud provider is set as external.

dulek:
I thought that legacy cloud-provider-openstack is completely gone since 1.26 [1]. Shouldn't we just drop it's support completely?

[1] https://github.com/kubernetes/legacy-cloud-providers/tree/release-1.26

ching-kuo:
Yes, it's completely gone in 1.26. I'm also ok to completely drop support with it. I would like to know how other users thinks, I believe most of the users are already using CCM though.

ching-kuo:
Should we still need to do e2e test for in-tree cloud providers? Or should we just remove it in v1alpha7?

ching-kuo:
@dulek Can you review if you have time if there is anything missing here? Thanks!

test/e2e/data/kustomize/default/kustomization.yaml
```diff
@@ -1,8 +1,21 @@
 ---
 resources:
 - ../../../../../kustomize/v1alpha7/default
+- ccm.yaml
 
 components:
 - ../common-patches/cni
 - ../common-patches/containerd
-- ../upgrade-patches
```
mdbooth, line 8: @lentzi90 Can you check this out? It doesn't feel right to remove this.

test/e2e/data/kustomize/without-lb/patch-ccm-cloud-config.yaml
mdbooth, line 1: Would it be possible to move all of these into `test/e2e/data/kustomize/common-patches`?

mdbooth (commented):
Thanks for this!

test/e2e/data/kustomize/without-lb/patch-ccm-cloud-config.yaml
lentzi90: +1 for that!

test/e2e/data/kustomize/default/kustomization.yaml
lentzi90: Yes this is needed for the clusterctl upgrade test (part of the e2e-full suite). I guess what happened here was that we only use the default template in those tests, so the external-cloud-provider kustomizations didn't have the upgrade-patches and that is why it was removed.
Please add it back and the e2e-full test should hopefully pass :slightly_smiling_face:

lentzi90 (commented):
Nice work on this!
I agree with @mdbooth that it would be nice to move the ccm stuff to the common-patches since it is now needed almost everywhere. The upgrade-patches must also be added back for the e2e-full test to work.

test/e2e/data/kustomize/without-lb/patch-ccm-cloud-config.yaml
ching-kuo: I just tried, I think it's not possible to have `patchesStrategicMerge` and `patches` not under the same folder, it will get error.

Please let me know if there's other way doing it.

```
Error: accumulating resources: accumulation err='accumulating resources from '../default': '/home/ubuntu/cluster-api-provider-openstack/test/e2e/data/kustomize/default' must resolve to a file': recursed accumulation of path '/home/ubuntu/cluster-api-provider-openstack/test/e2e/data/kustomize/default': trouble configuring builtin PatchTransformer with config: `
path: ../common-patches/ccm/patch-allow-all-in-cluster-traffic.yaml
target:
  kind: OpenStackCluster
  name: \${CLUSTER_NAME}
`: security; file '/home/ubuntu/cluster-api-provider-openstack/test/e2e/data/kustomize/common-patches/ccm/patch-allow-all-in-cluster-traffic.yaml' is not in or below '/home/ubuntu/cluster-api-provider-openstack/test/e2e/data/kustomize/default'
```

lentzi90: Indeed, the patch files cannot be outside of the kustomization itself. The way to achieve it is to turn the CCM patches into a `component`. Then you can use it the same way as the `cni`:

```yaml
components:
- ../common-patches/cni
- ../common-patches/ccm
```

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -143,47 +143,6 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
-			machineStatusInput := framework.WaitForMachineStatusCheckInput{
-				Getter:       e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
-				Machine:      &workerMachines[0],
-				StatusChecks: statusChecks,
-			}
-			framework.WaitForMachineStatusCheck(ctx, machineStatusInput, e2eCtx.E2EConfig.GetIntervals(specName, "wait-machine-status")...)
-
-			workloadCluster := e2eCtx.Environment.BootstrapClusterProxy.GetWorkloadCluster(ctx, namespace.Name, clusterName)
-
-			waitForDaemonSetRunning(ctx, workloadCluster.GetClient(), "kube-system", "openstack-cloud-controller-manager")
```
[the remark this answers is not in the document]
lentzi90: Oh good point. I think we should add this to the default test at the top since it is now using the external cloud provider.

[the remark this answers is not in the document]
ching-kuo: Fixed

ching-kuo:
Fixed all comments and ready for review again.

ching-kuo:
Fixed some comments on `docs/book/src/clusteropenstack/configuration.md`. Ready for final review

lentzi90 (commented):
/lgtm

mdbooth (approved):
I think there might still be some more kustomize minimisation we can do here, but lets get this landed. Thanks for sticking with it!

PR #1516, opened 2023-03-27
⚠️ Remove PortOpts.SecurityGroups

### LINKED ISSUE #1251: Deprecating/Replacing securityGroups field from OpenStackMachineTemplate ports

lentzi90:

SecurityGroupFilters can be used in all cases where SecurityGroups were used and supports additional use cases.
This removes SecurityGroups from the PortOpts in v1alpha7.
SecurityGroups are converted to SecurityGroupFilters automatically on up conversion. On down conversion, any previously set SecurityGroups or SecurityGroupFilters will be restored as they were before the up conversion.

Fixes #1251 and replaces #1291 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

/hold


docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -33,3 +33,25 @@ The OpenStackMachine spec previously contained a `subnet` field which could used
...
 used, difficult to use, and could not be extended to support IPv6. It is
 removed without replacement.
+
+#### Removal of Port SecurityGroups
+
+The Port field of the OpenStackMachine spec previously contained both `securityGroups` and `securityGroupFilters`.
+As `securityGroups` can easily be replaced with `securityGroupFilters`, that can express the same and more, `securityGroups` has now been removed.
+CAPO can automatically convert `securityGroups` to `securityGroupFilters` when upgrading.
+Downgrading will however *not* give back `securityGroups` but instead retain the `securityGroupFilters`.
+
```
mdbooth, line 43: I think this is no longer true, right? Should probably also update the PR description.

api/v1alpha6/conversion.go
```diff
@@ -86,6 +86,11 @@ func restorev1alpha6MachineSpec(previous *OpenStackMachineSpec, dst *OpenStackMa
 	// Subnet is removed from v1alpha7 with no replacement, so can't be
 	// losslessly converted. Restore the previously stored value on down-conversion.
 	dst.Subnet = previous.Subnet
+	// SecurityGroups is removed from v1alpha7 with no replacement.
+	// Restore the previously stored value on down-conversion.
+	for i := range dst.Ports {
+		dst.Ports[i].SecurityGroups = previous.Ports[i].SecurityGroups
```
mdbooth, line 92: The fuzzer tests don't seem to like this, btw. Didn't we decided that as ports are immutable it'd be easier just to replace the whole ports slice?

lentzi90: Yes, true! I will put this back in WIP mode until I have time to fix this

lentzi90:
I think this one is ready for review now @mdbooth

api/v1alpha6/conversion.go
```diff
@@ -94,6 +94,20 @@ func restorev1alpha6MachineSpec(previous *OpenStackMachineSpec, dst *OpenStackMa
 	dst.Ports = previous.Ports
 }
 
+func restorev1alpha6ClusterStatus(previous *OpenStackClusterStatus, dst *OpenStackClusterStatus) {
+	// PortOpts.SecurityGroups have been removed in v1alpha7
+	// We restore the whole PortOpts/Networks since they are anyway immutable.
+	if previous.ExternalNetwork != nil {
```
mdbooth, line 100: Couple of interesting notes:

Firstly, it looks like ExternalNetwork might be mutable! I doubt this is intentional, though.

Secondly, the only reason you have to do this at all is because the Network type is used in too many places and doesn't make sense in most/all of them. Cleaning it up may be the largest remaining v1beta1 task.

Given the above I suspect it may be ok to merge this anyway. I'll try to get back to it properly later: I haven't had time to follow these thoughts through fully yet.

mdbooth: No, it's not mutable: the webhook prevents changes. The second point still stands, but not relevant to this PR.

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -111,3 +111,24 @@ The OpenStackMachine spec previously contained a `subnet` field which could used
...
+# securityGroups are available in v1alpha6
+securityGroups:
+- 60ed83f1-8886-41c6-a1c7-fcfbdf3f04c2
+- 0ddd14d9-5c33-4734-b7d0-ac4fdf35c2d9
+- 4a131d3e-9939-4a6b-adea-788a2e89fcd8
+# securityGroupFilters are available in both v1alpha6 and v1alpha7
+securityGroupFilters:
+- UUID: 60ed83f1-8886-41c6-a1c7-fcfbdf3f04c2
+- UUID: 0ddd14d9-5c33-4734-b7d0-ac4fdf35c2d9
+- UUID: 4a131d3e-9939-4a6b-adea-788a2e89fcd8
```
mdbooth, line 133: I think this should be:
```suggestion
- id: 60ed83f1-8886-41c6-a1c7-fcfbdf3f04c2
- id: 0ddd14d9-5c33-4734-b7d0-ac4fdf35c2d9
- id: 4a131d3e-9939-4a6b-adea-788a2e89fcd8
```

Incidentally, we should consider how to cleanup inconsistent use of id and uuid for OpenStack resources throughout the API. We should probably pick one and add it as an option everywhere it's not used, and mark the other one as deprecated.

pkg/cloud/services/networking/port.go
```diff
@@ -303,39 +305,3 @@ func (s *Service) GarbageCollectErrorInstancesPort(eventObject runtime.Object, i
...
-	}
-	distinctSecurityGroupIDs := make([]string, 0, securityGroupCount)
-	// collect distict values
-	for key := range uids {
-		if key == "" {
-			continue
-		}
-		distinctSecurityGroupIDs = append(distinctSecurityGroupIDs, key)
-	}
-	return distinctSecurityGroupIDs, nil
```
mdbooth, line 340: Do we need to retain this code which removes duplicates?

mdbooth (commented):
I think the docs need a fix. Undecided on the duplicates removal. Why was that added in the first place? Do those reasons apply to filters as well as uuids? I would have thought this kind of functionality makes more sense if we were merging security groups from the port and the instance, but we're not doing that.

Two more thoughts: we should probably (separately) also change OpenStackMachineSpec.SecurityGroups to be SecurityGroupFilter instead of SecurityGroupParam. The latter looks entirely redundant.

And related, for consistency with OpenStackMachineSpec I wonder if we should rename SecurityGroupFilters to SecurityGroups in the port spec after all 🤔

Given where we are, though, I think this can merge without the additionals. We can continue to evolve the API in v1alpha7 in separate PRs if we choose to.

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
lentzi90: Good catch! I got so confused by this that I'm going to put what I found here for future reference.
The `PortOpts.SecurityGroupFilters` is `[]SecurityGroupParam`. This struct has the `UUID` field that translates to uuid in json/yaml, so that is what I should have put in the docs.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e35f1ca854b8bd7c3a06df8ff406aeb5f6c79d82/api/v1alpha7/types.go#L32-L39
To make things more confusing, the struct also contains a `Filter` of type `SecurityGroupFilter`. This is the one that has `ID` instead of `UUID`. I guess this is what you were talking about also? We could probably just use `SecurityGroupFilter` everywhere and get rid of the `SecurityGroupParam` struct.

pkg/cloud/services/networking/port.go
lentzi90: Good question. From my understanding it was added to ensure that `securityGroupFilters` and `securityGroups` when merged didn't produce duplicates. Since we now remove the `securityGroups` it should no longer be needed. The `securityGroupFilters` are already de-duplicated [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e35f1ca854b8bd7c3a06df8ff406aeb5f6c79d82/pkg/cloud/services/networking/securitygroups.go#L196).

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
lentzi90: Updated `UUID -> uuid`

mdbooth:
/lgtm

Up to you: do you want to pull the removal of SecurityGroupParam into this PR? Now that I've seen it I don't want to bring it forward into v1beta1.

mdbooth:
> /lgtm
> 
> Up to you: do you want to pull the removal of SecurityGroupParam into this PR? Now that I've seen it I don't want to bring it forward into v1beta1.

Actually lets merge this. I'll open an issue for removal of SecurityGroupParam. I think the only impact on this PR will be that we'll have to remember to update the docs again.

mdbooth:
I opened https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1548 to track a follow-up.

PR #1501, opened 2023-03-09
✨Add Octavia OVN Provider Support

ching-kuo:
**What this PR does / why we need it**:

This commit adds support for Octavia OVN Provider. Changes includes adding provider option in apiServerLoadBalancer and when it is set to OVN, use SOURCE_IP_PORT as load balancing algorithm.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1472 

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

PR #1517, opened 2023-03-28
Restore APIServerLoadBalancer.Provider on up-conversion

mdbooth:
From v1alpha6 onwards we have the facility to losslessly preserve values during conversion which were added or removed in the new version, so we should never require the FuzzerFuncs argument to the fuzzer conversion tests.

/hold

### DIFF
--- a/api/v1alpha6/conversion.go
+++ b/api/v1alpha6/conversion.go
@@ -88,14 +88,23 @@ func restorev1alpha6MachineSpec(previous *OpenStackMachineSpec, dst *OpenStackMa
 	dst.Subnet = previous.Subnet
 }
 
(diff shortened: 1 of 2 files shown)


mdbooth:
@ching-kuo @dulek My apologies, I missed this in review of #1501. In prior api versions we effectively added an exception to the fuzzer tests for values which were added or removed, but as we're now using CAPI's 'previous version' annotation we can now manually restore these values and we don't need this any more.

Incidentally, we might eventually consider retrofitting this to v1alpha5, but it's not worth adding to prior versions.

/cc @lentzi90 @tobiasgiese

lentzi90 (commented):
/lgtm

PR #1518, opened 2023-03-28
⚠️ Remove Networks

### LINKED ISSUE #1154: Remove Networks from the Machine spec
/kind feature

Ports should have a superset of the functionality of Networks, with one exception. Networks allow the creation of multiple ports from a single filter definition, whereas ports do not. I propose that was never intended behaviour, nobody's actually relying on this, and the most likely outcome of it is user surprise if their filter accidentally has more than one match.

I would like to simply remove Networks.

`NetworkParam` is currently defined as:

```golang
type NetworkParam struct {
	// Optional UUID of the network.
	// If specified this will not be validated prior to server creation.
	// Required if `Subnets` specifies a subnet by UUID.
	// A fixed IPv4 address for the NIC.
	FixedIP string `json:"fixedIP,omitempty"`
	// Filters for optional network query
	Filter NetworkFilter `json:"filter,omitempty"`
	// Subnet within a network to use
	Subnets []SubnetParam `json:"subnets,omitempty"`
}
```

This translates as:

`NetworkParam.UUID` -> `PortOpts.Network.ID`
`NetworkParam.FixedIP` -> `PortOpts.FixedIPs[0].IPAddress`
`NetworkParam.Filter` -> `PortOpts.Network`
`NetworkParam.Subnets[]` -> `PortOpts.FixedIPs[].Subnet`

The only functional difference would be that if any of the network or subnet filters return multiple results it will be an error. This is an excellent argument for doing this before v1beta1.

Proposed for 0.6: #1145

mdbooth:
Removes Networks from the OpenStackMachine spec.

Fixes #1154

The parts requiring the most focus are:
* Networks -> Ports conversion (api/v1alpha6/conversion.go)
* Unit tests for the above
* Documentation (docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md)

/hold


lentzi90 (commented):
Well explained and nice to read!
/lgtm

mdbooth:
I wonder if I could get some user-focussed review on this. Perhaps @mnaser and/or @mkjpryor? Reviewing the docs should be enough.

mnaser:
This seems okay, how does this work with the cluster topology? I assume it'll just work(tm)

mkjpryor:
This looks good to me. We only really use ports anyway as the only reason we do machine-specific networking is to add SR-IOV ports on a secondary VLAN.

mdbooth:
> This seems okay, how does this work with the cluster topology? I assume it'll just work(tm)

I have no reason to think it'll impact cluster topologies at all: it's just a change in the machine spec.

PR #1519, opened 2023-03-29
✨Infer port network from subnet

mdbooth:
NetworkID is a required field when creating a neutron port. We previously passed on this requirement in the Ports API. However we didn't have this restriction in the Networks API and inferred the network from a subnet if one was defined. This change eases the transition from Networks to Ports by removing this restriction for Ports.

**TODOs**:
- [x] tests

/hold

### DIFF
--- a/docs/book/src/clusteropenstack/configuration.md
+++ b/docs/book/src/clusteropenstack/configuration.md
@@ -358,9 +358,6 @@ spec:
   ports:
   - network:
       id: <your-network-id>
-    nameSuffix: <your-port-name>
-    description: <your-custom-port-description>
-    vnicType: normal
     fixedIPs:
     - subnet:
         id: <your-subnet-id>
@@ -370,6 +367,9 @@ spec:
         tags:
           - tag1
           - tag2
+    nameSuffix: <your-port-name>
+    description: <your-custom-port-description>
+    vnicType: normal
     securityGroups:
     - <your-security-group-id>
     profile:
@@ -379,7 +379,70 @@ spec:
 
 Any such ports are created in addition to ports used for connections to networks or subnets.
 
-Also, `port security` can be applied to specific port to enable/disable the `port security` on that port; When not set, it takes the value of the corresponding field at the network level.
+### Port network and IP addresses
+
+Together, `network` and `fixedIPs` define the network a port will be created on, and the addresses which will be assigned to the port on that network.
+
+`network` is a filter which uniquely describes the Neutron network the port will be created be on. Machine creation will fail if the result is empty or not unique. If a network `id` is specified in the filter then no separate OpenStack query is required. This has the advantages of being both faster and unambiguous in all circumstances, so it is the preferred way to specify a network where possible.
+
+The available fields are described in [the CRD](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api-provider-openstack/infrastructure.cluster.x-k8s.io/OpenStackMachine/v1alpha6@v0.7.1#spec-ports-network).
+
+If `network` is not specified at all, it may be possible to infer the network from any uniquely defined subnets in `fixedIPs`. As this may result in additional OpenStack queries and the potential for ambiguity is greater, this is not recommended.
+
+`fixedIPs` describes a list of addresses from the target `network` which will be allocated to the port. A `fixedIP` is either a specific `ipAddress`, a `subnet` from which an ip address will be allocated, or both. If only `ipAddress` is specified, it must be valid in at least one of the subnets defined in the current network. If both are defined, `ipAddress` must be valid in the specified subnet.
+
+`subnet` is a filter which uniquely describe the Neutron subnet an address will be allocated from. Its operation is analogous to `network`, described above.
+
+`fixedIPs`, including all fields available in the `subnet` filter, are described in [the CRD](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api-provider-openstack/infrastructure.cluster.x-k8s.io/OpenStackMachine/v1alpha6@v0.7.1#spec-ports-fixedIPs).
+
+If no `fixedIPs` are specified, the port will get an address from every subnet in the network.
+
+#### Examples
+
+A single explicit network with a single explicit subnet.
+```yaml
+ports:
+- tags:
+  - control-plane
+  network:
+    id: 0686143b-f0a7-481a-86f5-cc1f8ccde692
+  fixedIPs:
+  - subnet:
+      id: a5e50a9c-58f9-4b6f-b8ee-2e7b4e4414ee
+```
+
+No network or fixed IPs: the port will be created on the cluster default network, and will get a single address from the cluster default subnet.
+```yaml
+ports:
+- tags:
+  - control-plane
+```
+
+Network and subnet are specified by filter. They will be looked up. Note that this is not as efficient or reliable as specifying the network by `id`.
+```yaml
+ports:
+- tags:
+  - storage
+  network:
+    name: storage-network
+  fixedIPs:
+  - subnet:
+      name: storage-subnet
+```
+
+No network, but a fixed IP with a subnet. The network will be inferred from the network of the subnet. Note that this is not as efficient or reliable as specifying the network explicitly.
+```yaml
+ports:
+- tags:
+  - control-plane
+  fixedIPs:
+  - subnet:
+      id: a5e50a9c-58f9-4b6f-b8ee-2e7b4e4414ee
+```
+
+### Port Security
+
+`port security` can be applied to specific port to enable/disable the `port security` on that port; When not set, it takes the value of the corresponding field at the network level.
 
 ```yaml
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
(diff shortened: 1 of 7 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -46,57 +46,150 @@ const (
...
+						return "", err
+					}
+					for len(subnets) > 1 {
+						s.scope.Logger().V(4).Info("Can't infer network from subnet %d: multiple subnets match filter", i)
+						continue
+					}
+					for _, subnet := range subnets {
+						s.scope.Logger().V(4).Info("Inferred network %s from subnet %s", subnet.NetworkID, subnet.ID)
+						return subnet.NetworkID, nil
+					}
```
dulek, line 110: Feels like a pattern we often have, I'd at least extract it into a named function for code clarity.

pkg/cloud/services/compute/instance.go
```diff
@@ -46,57 +46,150 @@ const (
...
-					ID: openStackCluster.Status.Network.Subnet.ID,
-				},
-				PortOpts: port,
-			})
+
+			port.Network = &infrav1.NetworkFilter{
+				ID: networkID,
+			}
+
+		// Network is defined by ID
```
dulek, line 124: Given the condition it should say "Network is defined by filter"?

pkg/cloud/services/compute/instance.go
```diff
@@ -46,57 +46,150 @@ const (
...
+	// trunk support is required if the instance or any port has trunk enabled
+	portUsesTrunk := func() bool {
+		for _, port := range instanceSpec.Ports {
+			if port.Trunk != nil && *port.Trunk {
+				return true
+			}
+		}
+		return false
+	}
+	if instanceSpec.Trunk || portUsesTrunk() {
```
dulek, line 186: If you have `instanceSpec.Trunk=true`, but all the ports override that with `false`, then I guess we don't need trunk extension. We could only check `portUsesTrunk()` here.

dulek (changes requested):
It looks quite complicated for a simple change, a few ideas inline.

pkg/cloud/services/compute/instance.go
mdbooth: Nice edge case! This is actually here because we create the default port down below based on instanceSpec.Trunk, so neither is quite correct. I'll have a think.

mdbooth:
> It looks quite complicated for a simple change, a few ideas inline.

I agree, but there are a few things going on here.

Firstly is the stated change itself. This is itself more complex than it sounds because:
* There may or may not be subnets
* There may be arbitrarily many subnets
* There may be fixed ips which don't have subnets
So you have to loop over it until you find one which works. Incidentally, documenting this behaviour isn't going to be fun, and I still need to do that ☹️ If it wasn't for the Network->Ports upgrade driver I probably wouldn't bother and just continue to require that network is defined.

Secondly I've refactored it a bit because it's complex enough to require its own unit tests, which I haven't written yet. Hence pulling out the `normalizePorts` function.

Lastly there's a slightly hidden agenda here. I'm writing this in the context of removing Networks, so I'm going to be throwing away the non-Ports part of `constructNetworks` in another PR. You'll notice, however, that this function returns `[]infrav1.Network`. That struct doesn't make any sense any more, and I want to throw it away entirely and just pass Ports around. So in the new world order you normalize your ports, then use them. You don't have to wrap them in a different pointless struct first. Some of the code motion in this PR is with that in mind.

I want to backport this PR to version 0.7, though, so I can't fully do that here. This is to help anybody who needs to manually update their ports in v0.7 before upgrading to v0.8.

mdbooth:
> Alright, let's see the completed version then!

That isn't going to come in this PR, btw. I need to remove Networks first, which I'm doing in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1518.

However, this PR now has tests and docs. Hopefully ready to roll.

pkg/cloud/services/compute/instance.go
mdbooth: I started pulling this out, btw. I kept the non-related parts in a separate commit.

mdbooth:
I'm working on this in parallel with the Networks removal PR, btw, which is why I keep updating it. I don't want to move this code twice, so I'm effectively pulling Network removal requirements back into this PR.

I'm now at the point where I might move the entire 'use PortOpts natively' thing into this PR instead. The defaulting behaviour of ports is currently spread untidily across `networking/port.go` and `compute/instance.go` with very little in the way of pattern. I'm going to try to consolidate it.

pkg/cloud/services/compute/instance.go
```diff
@@ -46,58 +48,149 @@ const (
 	timeoutInstanceDelete       = 5 * time.Minute
 )
 
-// constructNetworks builds an array of networks from the network, subnet and ports items in the instance spec.
-// If no networks or ports are in the spec, returns a single network item for a network connection to the default cluster network.
-func (s *Service) constructNetworks(openStackCluster *infrav1.OpenStackCluster, instanceSpec *InstanceSpec) ([]infrav1.Network, error) {
-	trunkRequired := false
+// normalizePortTarget ensures that the port has a network ID.
+func (s *Service) normalizePortTarget(port *infrav1.PortOpts, openStackCluster *infrav1.OpenStackCluster, portIdx int) error {
```
dulek, line 52: `normalizePortTarget()` sounds a bit confusing if it's just `normalizePortNetwork()`.

dulek (approved):
Looks great, tiny suggestion inline.

dulek:
/lgtm

It has hold, so /lgtm is safe.

docs/book/src/clusteropenstack/configuration.md
```diff
@@ -387,7 +387,70 @@ spec:
...
+
+If `network` is not specified at all, it may be possible to infer the network from any uniquely defined subnets in `fixedIPs`. As this may result in additional OpenStack queries and the potential for ambiguity is greater, this is not recommended.
+
+`fixedIPs` describes a list of addresses from the target `network` which will be allocated to the port. A `fixedIP` is either a specific `ipAddress`, a `subnet` from which an ip address will be allocated, or both. If only `ipAddress` is specified, it must be valid in at least one of the subnets defined in the current network. If both are defined, `ipAddress` must be valid in the specified subnet.
+
+`subnet` is a filter which uniquely describe the Neutron subnet an address will be allocated from. Its operation is analogous to `network`, described above.
+
+`fixedIPs`, including all fields available in the `subnet` filter, are described in [the CRD](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api-provider-openstack/infrastructure.cluster.x-k8s.io/OpenStackMachine/v1alpha6@v0.7.1#spec-ports-fixedIPs).
+
+If no `fixedIPs` are specified, the port will get an address from every subnet in the network.
```
jichenjc, line 406: `every subnet ` => is this first subnet of the network? This seems to be the way I provide my openstack to my end user..

dulek: I tested this few days ago, if network has multiple subnets, creating a port specifying just the network gives you a port with address in every of the network's subnets:

A POST with:
```
{
  "port": {
    "admin_state_up": true,
    "name": "test",
    "network_id": "6acb08e4-86a1-47fa-a806-2f79a0b48abb"
  }
}
```

Will get you:
```
{
  "port": {
    <SNIP>
    "fixed_ips": [
      {
        "ip_address": "192.168.25.127",
        "subnet_id": "4b73762e-8fbb-4555-9df6-fcd2b8fe2734"
      },
      {
        "ip_address": "2001:db8::2c8",
        "subnet_id": "d808f460-99b0-4356-b158-92c3b6f97efd"
      }
    ],
    "id": "f5d15252-9f61-4d95-885b-9c3346fa2608",
    <SNIP>
    "name": "test",
    "network_id": "6acb08e4-86a1-47fa-a806-2f79a0b48abb",
    <SNIP>
  }
}
```

mdbooth:
I'm going to rework this a bit because it was starting to combine a refactor of the port creation code with a behavioural change. I'm going to try to separate them into 2 separate PRs.

mdbooth:
> I'm going to rework this a bit because it was starting to combine a refactor of the port creation code with a behavioural change. I'm going to try to separate them into 2 separate PRs.

I've changed my mind on this. I want to backport this change, but it was getting too invasive and too big. I'd like to land this version after all, and I'll do any refactoring on top of it in main.

pkg/cloud/services/compute/instance.go
mdbooth: I would update this, but I plan to refactor this code a fair bit anyway after backporting this to 0.7.

dulek:
/lgtm

Question inline, feel free to unhold this if that TODO is supposed to stay there.

pkg/cloud/services/compute/instance.go
```diff
@@ -46,58 +48,149 @@ const (
...
+	if err != nil {
+		return err
+	}
+
+	// TODO: These are spec errors: they should set the machine to failed
+	if len(netIDs) > 1 {
+		return fmt.Errorf("network filter for port %d returns more than one result", portIdx)
+	} else if len(netIDs) == 0 {
+		return fmt.Errorf("network filter for port %d returns no networks", portIdx)
+	}
```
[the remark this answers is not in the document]
mdbooth: It's already the case: I'm just adding the TODO because I noticed it. However, fixing these requires the refactor I'm trying to avoid here. I want to split the validation of ports such that everything which needs to be looked up is done for all ports before we start to create any of them. It just involves a ton of code motion.

PR #1521, opened 2023-04-05
:seedling: pull cluster-api@v1.4.1

tormath1:

This PR pulls cluster-api@v1.4.1 (including 1.3.2, 1.3.3, 1.3.4, 1.3.5, 1.3.6 and 1.4.0) - this follow a mention from the office hours

1. Not so much to report regarding the upgrade as the commits speak by themselves.
2. BIG QUESTION MARK on the following commit: `test/e2e/suites: update remediation flavor`
3. `sigs.k8s.io/cluster-api/api/v1alpha{3,4}` are deprecated - we currently ignore the warning before agreeing on what to do
4. Got a weird issue with the CI: cloud-init was failing (but don't know why), cloud-init validation was complaining about permissions not being a string so I fixed this in a commit.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


hack/ci/cloud-init/common.yaml.tpl
```diff
@@ -20,13 +20,13 @@ packages:
 package_upgrade: true
 write_files:
 - path: /etc/sysctl.d/devstack.conf
-  permissions: 0644
+  permissions: "0644"
```
jichenjc, line 23: not sure whether it's because latest cloud-init has this kind of restriction? 
https://github.com/canonical/cloud-init/pull/1360/files#diff-c18cfd88ec0bd5f1077576fee00a0b2e87cf0b30aa547ad1eaacfd5742e4a757R1870

jichenjc:
>>>BIG QUESTION MARK on the following commit: test/e2e/suites: update remediation flavor

seems one additional test added here... so what's the big question?

test/e2e/suites/e2e/mhc_remediations_test.go
```diff
@@ -29,18 +29,34 @@ import (
...
-			E2EConfig:             e2eCtx.E2EConfig,
-			ClusterctlConfigPath:  e2eCtx.Environment.ClusterctlConfigPath,
-			BootstrapClusterProxy: e2eCtx.Environment.BootstrapClusterProxy,
-			ArtifactFolder:        e2eCtx.Settings.ArtifactFolder,
-			SkipCleanup:           false,
-			KCPFlavor:             pointer.String(shared.FlavorKCPRemediation),
-			MDFlavor:              pointer.String(shared.FlavorMDRemediation),
-		}
+var _ = Describe("When testing unhealthy machines remediation", func() {
+	Describe("[mhc-remediations]", func() {
```
mdbooth, line 33: @lentzi90 This change passes my smell test: `KCPFlavor` and `MDFlavor` have been removed in favour of just `Flavor`, but I'd appreciate if you could take a closer look.

mdbooth (commented):
/lgtm

I'd like to wait for @lentzi90's take on the test changes before approving, but hopefully this is good to go. Thanks!

test/e2e/suites/e2e/mhc_remediations_test.go
lentzi90: Umm I may be missing something, but I don't see `KCPRemediationSpec`? I see two times `MachineDeploymentRemediationSpec` with different flavors.

mdbooth:
Can you squash the commits before removing the hold?

/approve

lentzi90:
Ow the KCPRemediationSpec is a bit more involved. I didn't realize they changed so much. It is now a much more thorough test but it means we need some changes to make it work. Previously these remediation tests were almost the same, just one for the control plane and one for the MachineDeployment. But now the KCP test is also checking remediation of the very first node that fails to initialize.

They have put some comments in the code for what is needed: https://github.com/kubernetes-sigs/cluster-api/blob/9ad6d519caed6755da688225172d8999be6c7015/test/e2e/kcp_remediations.go#L52-L77
If I understand it correctly we would need the following:

1. Define wait intervals for `wait-cluster`, `wait-machines`, `check-machines-stable` and `wait-machine-provisioned` (if they don't already exist) [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8e8d359ab21b7b206307fe998521f3bb5d70a240/test/e2e/data/e2e_conf.yaml#L163-L174).
2. Ensure that the flavor we use have 3 control plane nodes and 0 workers. I think this is already the case.
3. The control plane must be configured with `preKubeadmCommands` that queries for a well-known ConfigMap *on the management cluster*, holding up bootstrap until a signal is passed via the config map. I'm not even sure if this is possible in our CI since it requires the workload cluster to access the management cluster...
4. The MachineHealthCheck itself that we already have (may need some tweak though, I haven't checked the exact details of it).

Point 3 is the big one. It may be a bit too much to add in this PR. I would be fine with dropping the KCP test for now since it is basically a completely new test after the changes they made. I will anyway investigate if we can add it later when I have time.

PR #1531, opened 2023-04-27
🌱 Remove creation of 2 Events for the same event

dulek:
`record.Infof()` and `record.Warnf()` both create an event, so there's no need to use both at the same time. This commit removes one such occurrence.

**TODOs**:
- [X] squashed commits

/hold


lentzi90 (commented):
/lgtm

mdbooth (approved):

PR #1536, opened 2023-05-05
Update external approvers

mdbooth:
For a long time, in addition to the CAPO maintainers we have also allowed approvals from:
- sig-cluster-lifecycle-leads
- cluster-api-admins
- cluster-api-maintainers

However, we have to maintain these lists manually and our record of doing this has not been good. In practise there is also significant overlap in these groups.

As far as I am aware this facility has never been used, but it is easy to imagine a circumstance where it might be useful, e.g. a centrally coordinated CVE affecting multiple providers urgently. Allowing commits from CAPI maintainers should cover this use case in practise, and removing the other 2 lists reduces the amount of copy/paste we need to maintain from other projects.

/hold

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -7,23 +7,19 @@
 # - https://github.com/kubernetes/k8s.io/blob/main/groups/sig-cluster-lifecycle/groups.yaml
 
 aliases:
-  # correct as of 2021/03/03
+  # sig-cluster-lifecycle-leads are included as approvers in case it's ever
+  # operationally expedient:
+  # Source:
+  # https://github.com/kubernetes/org/blob/main/config/kubernetes/sig-cluster-lifecycle/teams.yaml
+  # correct as of 2023/05/10
   sig-cluster-lifecycle-leads:
+    members:
+    - CecileRobertMichon
+    - fabriziopandini
     - justinsb
     - neolit123
-    - timothysc
-    - fabriziopandini
-  # correct as of 2021/03/03
-  cluster-api-admins:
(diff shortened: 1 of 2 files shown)


fabriziopandini:
I don't have context, but I think it is a practice of SIG contrib-ex when they create subprojects; I'm not aware of automation to keep those list in sinc, but worth to ask to SIG contrib-ex directly.

Also, even if it is a maintenance burden, SIG leads is pretty stable and changes are usually broadcasted to the mailing list (and I don't see warm in catching up with some delay). I agree that CAPI admin overlaps with CAPI maintaners so, I think dropping CAPI admin is a good cleanup

mdbooth:
I've asked in #sig-contribex here: https://kubernetes.slack.com/archives/C1TU9EB9S/p1683627445012439

jichenjc:
/lgtm
/hold cancel

I don't know the detail but seems reasonable to me , an always update if find anything need

OWNERS_ALIASES
```diff
@@ -7,23 +7,19 @@
...
+  # sig-cluster-lifecycle-leads are included as approvers in case it's ever
+  # operationally expedient:
+  # Source:
+  # https://github.com/kubernetes/org/blob/main/config/kubernetes/sig-cluster-lifecycle/teams.yaml
+  # correct as of 2023/05/10
   sig-cluster-lifecycle-leads:
-    - justinsb
-    - neolit123
-    - timothysc
+    members:
```
BenTheElder, line 16: this line shouldn't be here

Issue #1545, opened 2023-05-11
Remove Bastion.Networks


mdbooth:
Part of: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1544

PR #1546, opened 2023-05-11
🌱 Reduce the Bastion status to only fields which are used

mdbooth:
The Instance type's only remaining use was to report the status of the Bastion. Most of the fields weren't used, but were still polluting the API. This change makes the intended use of the type explicit, and reduces it to only the fields which are actually used.

While this is an API change, it doesn't change any fields which have ever actually been set.

Fixed: #1545

/hold


dulek:
/lgtm

Nice, I like this.

PR #1550, opened 2023-05-15
🌱Remove PortOpts from Network

### LINKED ISSUE #1549: Remove PortOpts from Network
/kind feature

Part of: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1544

Network.PortOpts is not used in the external API. Its only use is internal, in createInstance().

`constructNetworks()` returns `[]infrav1.Network`, which is then passed to `GetOrCreatePort()`. The only required information is passed in the contained `PortOpts`. This `PortOpts` is not required for any other use of `infrav1.Network`, which means we are adding it to the external API for a purely internal use.

mdbooth:
This is an API change, but the removed field is not used by the API. This can be verified by looking at the second commit: it removes the field from the API but does not touch any code outside of api/.

Fixes #1549 

Easiest to review as 2 separate commits.

/hold


lentzi90 (commented):
/lgtm

lentzi90:
Not sure about the error. Checking if it was flake.

Issue #1552, opened 2023-05-15
ExternalNetwork should not have Subnet and APIServerLoadBalancer


mdbooth:
/kind feature

OpenStackClusterStatus.ExternalNetwork currently shares a type with OpenStackClusterStatus.Network, but the uses are not similar. It should have its own type.

The only field which is currently used is ID. However, Name and Tags have been present for a long time and could reasonably be informational.

ExternalNetwork should be reduced to:
```go
	Name string `json:"name"`
	ID   string `json:"id"`

	//+optional
	Tags []string `json:"tags,omitempty"`
```

PR #1555, opened 2023-05-15
🌱Remove unused fields from ExternalNetwork

mdbooth:
OpenStackClusterStatus.ExternalNetwork currently shares a type with OpenStackClusterStatus.Network, but the uses are not similar. It should have its own type.

The only field which is currently used is ID. However, Name and Tags have been present for a long time and could reasonably be informational.

Fixes: #1552 

/hold


api/v1alpha7/types.go
```diff
@@ -197,6 +197,15 @@ type Network struct {
 	APIServerLoadBalancer *LoadBalancer `json:"apiServerLoadBalancer,omitempty"`
 }
 
+// NetworkStatus contains basic information about an existing neutron network.
+type NetworkStatus struct {
```
dulek, line 201: Is it really a `Status`? Feels more like a `Spec` to me.

mdbooth: It's definitely a Status: it has an ID apart from anything else. In the Spec we use NetworkFilter.

seanschneeweiss (commented):
/lgtm

api/v1alpha7/openstackcluster_types.go
```diff
@@ -159,8 +159,8 @@ type OpenStackClusterStatus struct {
 	// It includes Subnets and Router.
 	Network *Network `json:"network,omitempty"`
 
-	// External Network contains information about the created OpenStack external network.
-	ExternalNetwork *Network `json:"externalNetwork,omitempty"`
+	// externalNetwork contains information about the external network used for default ingress and egress traffic.
```
jichenjc, line 162: s/externalNetwork /ExternalNetwork/

PR #1557, opened 2023-05-17
⚠️ Replace SecurityGroupParam with SecurityGroupFilter

### LINKED ISSUE #1548: Remove SecurityGroupParam
/kind feature

SecurityGroupParam contains a SecurityGroupFilter object which already contains both of the other fields of SecurityGroupParam. We should replace SecurityGroupParam with SecurityGroupFilter.

SecurityGroupParam is used in PortOpts.SecurityGroupFilters and OpenStackMachineSpec.SecurityGroups.

mdbooth:
SecurityGroupParam was:
```golang
type SecurityGroupParam struct {
       // Security Group UID
       // Security Group name
       Name string `json:"name,omitempty"`
       // Filters used to query security groups in openstack
       Filter SecurityGroupFilter `json:"filter,omitempty"`
}
```

However, the SecurityGroupFilter is contains also has ID and Name fields, making this wrapper redundant.

This is a breaking change due the removal of nesting of `filter` and also because the ID field is now called `ID` instead of `UUID`.

Fixes #1548 

- [x] update docs

/hold


docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -112,6 +112,41 @@ to set the `accessIPv4` field on Nova servers. This feature was not widely
...
+```
+becomes:
+```yaml
+securityGroups:
+  tags: my-security-group-tag
+```
+
+#### Type change to Port SecurityGroupFilters
+
+The same change is made to `securityGroupFilters` in `ports` as is made to `securityGroups` in the machine spec.
```
seanschneeweiss, line 148: Shall we link to the above heading?

mdbooth: If I can work out how!

lentzi90 (commented):
Nice!
/lgtm

jichenjc:
/lgtm

maybe squash the commit then we should be good to go

Issue #1554, opened 2023-05-15
Remove SubnetParam


mdbooth:
/kind feature

Similar to SecurityGroupParam, SubnetParam contains a Filter which already includes an ID parameter. We should replace all uses of SubnetParam with SubnetFilter.

PR #1559, opened 2023-05-18
⚠️ Replace SubnetParam with SubnetFilter

mdbooth:
SubnetParam was:
```golang
type SubnetParam struct {
        // Optional UUID of the subnet.
        // If specified this will not be validated prior to server creation.
        // If specified, the enclosing `NetworkParam` must also be specified by UUID.

        // Filters for optional subnet query
        Filter SubnetFilter `json:"filter,omitempty"`
}
```

However SubnetFilter already has an ID field, so the additional nesting was redundant.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1554

/hold


lentzi90 (commented):
/lgtm

PR #1560, opened 2023-05-19
:warning: Transform `Profile` into an interface

### LINKED ISSUE #1186: Transform `Profile` into an interface
/kind feature

Right now, when creating a Neutron port with specific binding capabilities (e.g. `switchdev` for OVS Hardware Offload), the syntax would be:
```
profile:
  capabilities: '[''switchdev'']'
```

Because our interface is:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ae4101718f0ab5e949e74dd692b55e83b55213e9/api/v1beta1/types.go#L134

While in Gophercloud, it's an interface:
```go
Profile map[string]interface{} `json:"binding:profile,omitempty"`
```
https://github.com/gophercloud/gophercloud/blob/4885c347dcf48fc518798d37fb93ffa7bdaa43e9/openstack/networking/v2/extensions/portsbinding/requests.go#L23

Providing the list of capabilities via a string isn't clean.

Convert `Profile` to `Profile map[string]interface{}`, so we'll be able to do:
```
profile:
  capabilities:
  - switchdev
```

Note: this feature is difficult to test against a real OpenStack cloud, for example `switchdev` required specific hardware (Mellanox Connect-X) for OVS HW offload. When implementing this change, we should also add functional testing with a basic profile that would work with current devstack setup.

EmilienM:
We previously allowed to use the Neutron `binding:profile` via `ports.bindingProfile`
but this API is not declared as stable from
the [Neutron API description](https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#create-port).

Instead, we now explicitly support two use cases:
* OVS Hardware Offload
* Trusted Virtual Functions (VF)

Here is an example on how to use `ports.bindingProfile` for enabling OVS Hardware Offload:

```yaml
bindingProfile:
  OVSHWOffload: true
```

Here is an example on how to use `ports.bindingProfile` for enabling "trusted-mode" to the VF:

```yaml
bindingProfile:
  TrustedVF: true
```

Convertion is supported from both v1alpha5 & v1alpha6 to v1alpha7.

Fixes #1186


api/v1alpha7/types.go
```diff
@@ -108,10 +108,11 @@ type PortOpts struct {
...
 
-	// A dictionary that enables the application running on the specified
-	// host to pass and receive virtual network interface (VIF) port-specific
-	// information to the plug-in.
-	Profile map[string]string `json:"profile,omitempty"`
+	// Profile is a set of key-value pairs that are used for binding details.
+	// We intentionally don't expose this as a map[string]string because we only want to enable
+	// the users to set the values of the keys that are known to work in OpenStack Networking API.
+	// See https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#create-port
+	Profile BindingProfile `json:"bindingProfile,omitempty"`
```
mdbooth, line 115: The public-facing name of this field is the json name, which I wasn't anticipating we would change. Was there a specific reason to change it from `profile` to `bindingProfile`?

If we do decide to change it to `bindingProfile`, please could you also update the Go field name to match?

api/v1alpha7/types.go
```diff
@@ -108,10 +108,11 @@ type PortOpts struct {
 	// The virtual network interface card (vNIC) type that is bound to the neutron port.
 	VNICType string `json:"vnicType,omitempty"`
 
-	// A dictionary that enables the application running on the specified
-	// host to pass and receive virtual network interface (VIF) port-specific
-	// information to the plug-in.
-	Profile map[string]string `json:"profile,omitempty"`
+	// Profile is a set of key-value pairs that are used for binding details.
```
mdbooth, line 111: The docstring should use the json name, which I'm expecting should be `profile` unless we decide to follow the change below and use `bindingProfile`.

We haven't been consistent with this to date, but I'm trying to be consistent with new docstrings. We'll need a cleanup at some point.

api/v1alpha6/conversion.go
```diff
@@ -25,6 +27,8 @@ import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha7"
 )
 
+var trueValue = "true"
```
mdbooth, line 30:
```suggestion
const trueString = "true"
```

api/v1alpha6/conversion.go
```diff
@@ -401,6 +405,18 @@ func Convert_v1alpha6_PortOpts_To_v1alpha7_PortOpts(in *PortOpts, out *infrav1.P
 	for i := range in.SecurityGroups {
 		out.SecurityGroupFilters = append(out.SecurityGroupFilters, infrav1.SecurityGroupFilter{ID: in.SecurityGroups[i]})
 	}
+
+	// BindingProfile is now an interface in v1alpha7.
```
mdbooth, line 409:
```suggestion
	// BindingProfile is now a struct in v1alpha7.
```

api/v1alpha6/conversion.go
```diff
@@ -401,6 +405,18 @@ func Convert_v1alpha6_PortOpts_To_v1alpha7_PortOpts(in *PortOpts, out *infrav1.P
...
+	for k, v := range in.Profile {
+		if k == "capabilities" {
+			if strings.Contains(v, "switchdev") {
+				out.Profile.OVSHWOffload = true
+			}
+		}
+		if k == "trusted" && v == trueValue {
+			out.Profile.TrustedVF = true
+		}
+	}
```
mdbooth, line 419: nit

```suggestion
	if strings.Contains(in.Profile["capabilities"], "switchdev") {
		out.Profile.OVSHWOffload = true
	} else if in.Profile["trusted"] == trueValue {
		out.Profile.TrustedVF = true
	}
```

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -175,6 +175,28 @@ securityGroupFilters:
 - id: 4a131d3e-9939-4a6b-adea-788a2e89fcd8
 ```
 
+### Change to ports.bindingProfile
+
+We previously allowed to use the Neutron `binding:profile` via `ports.bindingProfile` but this API is not declared as stable from the [Neutron API description](https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#create-port).
```
mdbooth, line 180: It used to be called `port.profile`, not `ports.bindingProfile`. If we do decide to change the name we should document that clearly here.

We should also be explicit that the conversion is lossy, and if anyone was relying on anything other than the supported behaviour it will be lost.

mdbooth (changes requested):
I'm neutral on changing the name of the field from `profile` to `bindingProfile`, but whether we make the change or not we should do it consistently and there are a couple of things to clean up there.

My barely informed thoughts on a name change:
* Pro: `profile` is a bit generic?
* Pro: We're drastically changing the type, so it's no major hardship for users to also update the field name.
* Con: We have vnicType, not bindingVNICType

Interested in thoughts on a field name change from @dulek and @mkjpryor.

Conversion and new type lgtm.

api/v1alpha7/types.go
EmilienM: Let's stick to `Profile`. I'll update everything.

EmilienM: Also if you don't mind, I'll let the `BindingProfile` for the new struct.

api/v1alpha5/conversion.go
```diff
@@ -227,7 +243,19 @@ func Convert_v1alpha7_APIServerLoadBalancer_To_v1alpha5_APIServerLoadBalancer(in
 
 func Convert_v1alpha5_PortOpts_To_v1alpha7_PortOpts(in *PortOpts, out *infrav1.PortOpts, s conversion.Scope) error {
 	// SecurityGroups have been removed in v1alpha7.
-	return autoConvert_v1alpha5_PortOpts_To_v1alpha7_PortOpts(in, out, s)
+	err := autoConvert_v1alpha5_PortOpts_To_v1alpha7_PortOpts(in, out, s)
+	if err != nil {
+		return err
+	}
+
+	// BindingProfile is now a struct in v1alpha7.
```
mdbooth, line 251: nit: Profile

api/v1alpha6/conversion.go
```diff
@@ -401,6 +405,14 @@ func Convert_v1alpha6_PortOpts_To_v1alpha7_PortOpts(in *PortOpts, out *infrav1.P
 	for i := range in.SecurityGroups {
 		out.SecurityGroupFilters = append(out.SecurityGroupFilters, infrav1.SecurityGroupFilter{ID: in.SecurityGroups[i]})
 	}
+
+	// BindingProfile is now a struct in v1alpha7.
```
mdbooth, line 409: nit: Profile

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -175,6 +176,31 @@ securityGroupFilters:
 - id: 4a131d3e-9939-4a6b-adea-788a2e89fcd8
 ```
 
+### Change to ports.bindingProfile
```
mdbooth, line 179: `ports.profile`. This should also be `####`. Wouldn't have noticed if the indentation on the doctoc wasn't off, so thanks for updating that!

```suggestion
#### Change to ports.profile
```

mdbooth (commented):
Just the docs update.

mdbooth:
The doctoc needs to be updated again, but as we need to do a general cleanup on that anyway I'm personally happy to land this as is. Feel free to update it if you're feeling keen, though.

/approve

mdbooth (approved):

EmilienM:
@jichenjc  when you have a chance to review it, thanks a lot!

jichenjc:
/lgtm
/hold cancel

Thanks for the great PR

EmilienM:
Thanks for the reviews, I learned a bunch and I'm looking forward to my next PR :-)

PR #1563, opened 2023-05-24
⚠️Remove last uses of TenantID and pagination filters

mdbooth:
TenantID is replaced by ProjectID.
Limit, Marker, SortKey, and SortDir should never have have been included.

**TODOs**:

- [x] Documentation

/hold


stephenfin:
> @dulek @stephenfin I've removed `tenantId` from PortOpts in favour of `projectId`. But it occurs to me: this is a spec parameter. Is there ever a case when specifying even `projectId` on a port would make sense during creation?

It would make sense if you were creating a port for another project besides your own. This would require admin permissions, however. For our sake, it's likely redundant: neutron will pull this information from the auth info. For reference, running `openstack --debug port create --network private` against a DevStack deployment yields the following request:

```
REQ: curl -g -i -X POST http://10.0.109.180:9696/networking/v2.0/ports -H "Content-Type: application/json" -H "User-Agent: openstacksdk/1.0.1 keystoneauth1/5.1.2 python-requests/2.28.2 CPython/3.11.3" -H "X-Auth-Token: {SHA256}8ff6b7234c2b5a0c5e651cd4169f2a51bdec0fd8e3c89fe863f80d456da03914" -d '{"port": {"network_id": "01bd28f2-9841-4b30-b2cd-e87b6d4d5ab3", "name": "test-port", "admin_state_up": true}}'
```

Not shown: the request headers including auth info.

Running `openstack --debug port create --network private --project admin` results in the following:

```
REQ: curl -g -i -X POST http://10.0.109.180:9696/networking/v2.0/ports -H "Content-Type: application/json" -H "User-Agent: openstacksdk/1.0.1 keystoneauth1/5.1.2 python-requests/2.28.2 CPython/3.11.3" -H "X-Auth-Token: {SHA256}248b586da4656ff4bd7c46a59c16048d206241b4539818c743882126e01d3483" -d '{"port": {"project_id": "c4fdc1bcf30541048ca086d7332a57e3", "network_id": "01bd28f2-9841-4b30-b2cd-e87b6d4d5ab3", "name": "test-port", "admin_state_up": true}}'
```

Attempting to create port using a project field 

It also results in the eventual failure of the request, however

```
openstack.exceptions.BadRequestException: BadRequestException: 400: Client Error for url: http://10.0.109.180:9696/networking/v2.0/ports, Specifying 'project_id' or 'tenant_id' other than the authenticated project in request requires admin or advsvc privileges
```

which kind of proves it is unnecessary/potentially misleading.

mdbooth:
Ok. I think I'll submit a separate PR to remove it because it has a more complicated feel to it than this one, which is basically just mechanical.

Issue #1448, opened 2023-01-17
Support configurable microversions for optional features like multi-attachable volumes


lentzi90:
/kind feature

Specifically we have a use case for running CAPO in an OpenStack environment where all volumes are multi-attachable. This requires Nova microversion >2.73 if I'm not mistaken. (Currently CAPO is hard coded to 2.53.)

**Describe the solution you'd like**

Ideally, I think CAPO should work similar to other OpenStack clients in that users should be able to specify the default microversions used. This is normally done either with environment variables or in the `clouds.yaml`:

> A user of the REST layer may set the default microversion by setting `{service_type}_default_microversion` in clouds.yaml or `OS_{service_type|upper}_DEFAULT_MICROVERSION` environment variable.
https://docs.openstack.org/openstacksdk/rocky/user/microversions.html

However, it seems that gophercloud is currently lacking support for this (https://github.com/gophercloud/utils/issues/178).
But this is the way I would like to see things work. We would add support for it in gophercloud and then use the specified version (if any) in CAPO, instead of hard coding it.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/02a31730ef1ae7962f1cc36c090b624025a2f5bc/pkg/clients/compute.go#L42
If no default version is specified, we could fall back to the minimum supported version in CAPO. Although it would be nice to try to detect the latest version that works instead.

jichenjc:
>>>Ideally, I think CAPO should work similar to other OpenStack clients in that users should be able to specify the default microversions used

yes, I think it's reasonable approach 

>>>But this is the way I would like to see things work. We would add support for it in gophercloud and then use the specified version (if any) in CAPO, instead of hard coding it.

yes, we should add such , guess there are 2 use cases so far 
- we need minimum because mandatory feature like 2.53 in CAPO is for tag which is required 
- there are optional cases like https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/util/openstack/loadbalancer.go#L100 , which is if current version support , we enbale the feature ..

>>>If no default version is specified, we could fall back to the minimum supported version in CAPO. Although it would be nice to try to detect the latest version that works instead.

based on my formal microversion experience, I think it's not that easy to detect the latest version ? 
as if we want to compare, we still have to know which version add the support and that's still hard code

mdbooth:
I think I've said this elsewhere, but just in case: a brief recap.

[...]

PR #1565, opened 2023-05-26
📖 Proposal for microversion support

lentzi90:

This is a proposal related to #1448 for how to support more microversions in CAPO.

This is just a proposal so it won't contain the actual implementation.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/docs/proposals/20230525-openstack-microversion-support.md
+++ b/docs/proposals/20230525-openstack-microversion-support.md
@@ -1,0 +1,141 @@
+# Configurable/multi microversion support
+
+OpenStack is diverse.
+Multiple versions and configurations exists that can impact what microversions and features that are supported.
+The user (of CAPO) can also to some extent choose what features to make use of.
+The cross section of this is where CAPO is useful:
+
+## Motivation
+
+A CAPO cluster can be configured in many ways and depending on this configuration, certain features may or may not be needed.
+This makes it necessary to know and detect if a feature is used that requires a certain microversion.
+
+It is not enough to check if the user has asked for a specific configuration that would require a certain microversion.
+The requirement can come from the OpenStack instance itself.
+As an example, we have seen environments where all volumes are multi-attachable.
+I.e. it is not possible to create a volume that is not multi-attachable, and thus multi-attach support is a must to be able to do anything in this environment, even if the user doesn't explicitly ask for it.
+
+Currently CAPO has a fixed microversion set.
+It is a take it or leave it situation.
+Either this version is supported by the OpenStack environment or you cannot use CAPO.
+
+This proposal is to make CAPO more flexible in what it supports.
+For example, there could be other versions that doesn't really impact CAPO in any way or there could be cross sections between what the OpenStack environment and CAPO supports, that could be useful to the user if certain features are avoided.
+Here is a simple diagram to try to summarize what this proposal aims to achieve:
+
+```text
+ ┌────────────────────────────────────┐
+ │                                    │
+ │                                    │
+ │    Versions supported by           │
+ │    OpenStack instance              │
+ │                                    │
+ │    ┌─────────────────────────┐     │
+ │    │                         │     │
+ │    │  Versions available to  │     │
+ │    │  the user               │     │
+ └────┼─────────────────────────┼─────┘
+      │                         │
+      │  Versions supported     │
+      │  by CAPO                │
+      │                         │
+      └─────────────────────────┘
+```
+
+"Versions" could be changed to "features" in the diagram also, but a version difference does not always expose a difference in features at the CAPO level.
+Therefore "versions" is used in the diagram to highlight the need to support multiple versions even though that may not give any new features.
+
+### Goals
+
+1. Detect user requested features and their version needs.
+
+2. Detect server version support.
+
+### Non-goals
+
+- To allow the user to set the versions freely.
+
+### CAPO microversions history
+
+Initially no microversions were specified (only versions).
+In OpenStack, there is support for microversions in Nova, Cinder, Manila and Barbican.
+Neutron uses extensions and Keystone uses neither.
+So far only Nova (compute) has any microversion specified in CAPO.
+
+The compute client has been bumped from the original v2 version in order to add support for specific features:
+
+- 2.53 is needed for server tags (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/924)
+- 2.60 is needed for multi-attach volumes (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1498)
+
+## Proposal
+
+### User stories
+
+#### Story 1
+
+As a user of CAPO, I would like to create a cluster in an older OpenStack environment that doesn't support `currently hard coded microversion` and I am fine with limiting the features available to me based on this older version.
+
+#### Story 2
+
+As a user of CAPO, I would like to create a cluster in an OpenStack environment that requires a higher microversion than `currently hard coded microversion`.
+
+*Note: The requirement could be because of configuration. For example there could be only multi-attach volume types configured.*
+
+#### Story 3
+
+As developer/maintainer of CAPO, I want to ensure that the code is executed (and tested) with a specific microversion so that the result is deterministic.
+
+#### Story 4
+
+As a developer/maintainer of CAPO, I want to keep the code readable and limit the burden of testing and maintaining support for multiple microversions.
+
+### Implementation details and notes
+
+There is already one part of CAPO that actually has dynamic versioning, thanks to Gophercloud.
+The [authentication](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/904381c7f31f8fdda83586f208339cd2aab53e78/pkg/scope/provider.go#L173) flow makes use of [this function](https://github.com/gophercloud/gophercloud/blob/6e1423b668969548d46fd862202a3bf9623b52e0/openstack/client.go#L92-L112) that chooses the version from a list of supported versions, provided by the caller, and the range of versions supported by the server.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,95 @@
...
+
+This proposal is to make CAPO more flexible in what it supports.
+For example, there could be other versions that doesn't really impact CAPO in any way or there could be cross sections between what the OpenStack environment and CAPO supports, that could be useful to the user if certain features are avoided.
+Here is a simple diagram to try to summarize the situation:
+
+```text
+  ┌────────────────────────────────┐
+  │                                │
+  │                                │
+  │     OpenStack support          │
```
mdbooth, line 30: Had to look at this a couple of times. Might be better to highlight that this refers to a *specific* OpenStack rather than OpenStack in general, which would presumably encompass everything.

Maybe:
* Features supported by user's OpenStack
* Features supported by CAPO
* Features available to user

docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Detect user requested features and their version needs.
+
+- Detect server version support.
+
+### Non-goals
+
+- To allow the user to set the versions freely.
+
+### CAPO mocroversions history
```
mdbooth, line 60:
```suggestion
### CAPO microversions history
```

docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,95 @@
...
+
+### User stories
+
+#### Story 1
+
+As a user of CAPO, I would like to create a cluster in an older OpenStack environment that doesn't support `currently hard coded microversion` and I am fine with limiting the features available to me based on this older version.
+
+#### Story 2
+
+As a user of CAPO, I would like to create a cluster in an OpenStack environment that requires a higher microversion than `currently hard coded microversion`.
```
mdbooth, line 79: I don't think this use case exists or is ever likely to exist. I *think* microversions are basically forever.

@stephenfin are you able to confirm/deny?

lentzi90: Isn't this exactly why we had https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1498?
We had an OpenStack environment that did not support 2.53 so we had to bump to 2.60 to make CAPO work with it.

mdbooth: No. 2.53 was supported, but because multi-attach support was only added in 2.60 (not carefully checking my version numbers here, btw), if you created a volume using microversion 2.53 you continued to get the old behaviour: i.e. no multi-attach. The microversion bump was required to get the new behaviour, but the old one still worked fine.

lentzi90: Yes but that is assuming the old behavior was available in the OpenStack environment. I'm saying we saw an environment that straight out rejected 2.53. All volumes *had* to be multi-attach because that was the only thing supported. In other words, it was impossible to use CAPO at all in this environment until the version was bumped to 2.60.

lentzi90: Maybe we had different reasons. :shrug:  The message I got on our side was at least that there was some environment where it was impossible to create volumes that wasn't multi-attach, i.e. CAPO was impossible to use without a version bump.

stephenfin: I think you're both correct. While there was talk in the past of removing support for older microversions (Ironic were looking into it), I think the conclusion was that there weren't enough benefits to justify it. As such, you should be able to request even the oldest microversion for as long as OpenStack remains a thing.

With that said, if you want to take advantage of newer features then you might have to rely on newer microversions. Typically this would be something service-specific like the ability to indicate a reason when locking a server (introduced in [compute microversion 2.73](https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#id66)) but it could also be something inter-service. A volume supports multi-attach [if the volume type it was created with has the `multiattach` property/extra-spec](https://docs.openstack.org/cinder/latest/admin/volume-multiattach.html). To use such volumes, you must use compute microversion 2.60 or later. If you don't, it'll fail. For example:

```
❯ openstack volume type create --property multiattach='<is> True' mutiattach-demo
❯ openstack volume create --type mutiattach-demo --size 10 my-volume
❯ openstack server create ... --block-device uuid=e3f8fd3f-a380-469f-abde-d0172e54366e,source_type=volume test-server
Multiattach volumes are only supported starting with compute API version 2.60. (HTTP 400) (Request-ID: req-5ffd9c45-ae0e-421e-b89c-0510c85cb3d4)
```

So yes, assuming the default volume type (or perhaps all volume types) has multi-attach enabled then you would be forced to use compute microversion 2.60 to be able to use volumes.

docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,133 @@
...
+2. Detect server version support.
+
+### Non-goals
+
+- To allow the user to set the versions freely.
+
+### CAPO microversions history
+
+Initially no microversions were specified (only versions).
+This is still true for all clients except compute.
```
dulek, line 61: I think only Cinder, Manila and Barbicna support microversions beside Nova. Neutron uses extensions, Keystone does neither. It's probably worth mentioning that here.

docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,133 @@
...
+As developer/maintainer of CAPO, I want to ensure that the code is executed (and tested) with a specific microversion so that the result is deterministic.
+
+#### Story 4
+
+As a developer/maintainer of CAPO, I want to keep the code readable and limit the burden of testing and maintaining support for multiple microversions.
+
+### Implementation details and notes
+
+There is already one part of CAPO that actually has dynamic versioning, thanks to Gophercloud.
+The [authentication](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/904381c7f31f8fdda83586f208339cd2aab53e78/pkg/scope/provider.go#L173) flow makes use of [this function](https://github.com/gophercloud/gophercloud/blob/6e1423b668969548d46fd862202a3bf9623b52e0/openstack/client.go#L92-L112) that chooses the version from a list of supported versions, provided by the caller, and the range of versions supported by the server.
```
dulek, line 90: That's not really a microversion. I get the point, but this example is super simple. I believe this is a better example: https://github.com/kubernetes/cloud-provider-openstack/blob/35ce5a0c59da23795fc48b292e3736f8cb0bb10f/pkg/util/openstack/loadbalancer.go#L101-L162

Please note this is not really a microversion, Octavia does regular API versioning and you cannot request an older microversion.

dulek (commented):
I like this, this is exactly how I believe microversions should be consumed.

docs/proposals/20230525-openstack-microversion-support.md
```diff
@@ -0,0 +1,133 @@
...
+
+In this alternative the user would be able to override the version set in CAPO.
+The obvious consequence is that the user could very easily break things by using versions that are not tested or supported.
+On the other hand, it gives maximum flexibility.
+The user can use any version, but the responsibility to ensure it is working also falls on them.
+
+The `clouds.yaml` configuration file already has support for specifying the version so this should be familiar to many users.
+However, Gophercloud does not support reading all the versions from this file as of now.
+The main reason for avoiding this alternative, though, is the lack of control that CAPO has.
+With a well defined list of supported and tested versions, perhaps it could still be considered.
```
seanschneeweiss, line 133: A well defined list of supported and tested versions would be nice.
Howerver, I don't think it should be expected that our list is a complete list of all available microversions of OpenStack Cinder, Manila, Barbican, Nova and wether we tested or not.

seanschneeweiss (commented):
Thanks for putting the effort into microversion handling.

docs/proposals/20230525-openstack-microversion-support.md
lentzi90: Updated!

docs/proposals/20230525-openstack-microversion-support.md
lentzi90: Added a note to explain the situation.

docs/proposals/20230525-openstack-microversion-support.md
lentzi90: Thanks! Added this. Could you double check if it makes sense?

docs/proposals/20230525-openstack-microversion-support.md
lentzi90: Thanks! I added the linked example but I also kept the original since I think it can still be useful for the implementation.

docs/proposals/20230525-openstack-microversion-support.md
lentzi90: Very true. I'm mainly keeping this here as a reminder of where the discussion started. That is, I wanted to let the user pick the version (with all the consequences that it brings). So if we created a list of known good versions for CAPO (the list could have a single version even) then maybe we could still consider this implementation. The user could check that the version they set is supported, but in theory they could then still set an unsupported version (with the risks that it would bring).

mdbooth:
I like this as a problem statement. However, to merge it I think we'd want a discussion of the proposed implementation. There was one at some point. Do you want to add it to the doc?

mdbooth:
I knew you'd previously created a design doc repo! Lets merge this and I'll update mine to correspond.

/lgtm

PR #1566, opened 2023-05-26
:seedling: Rebase k8s to 1.27

odvarkadaniel:
This PR updates the repository to use the 1.27 k8s packages as upstream `cluster-api` repository has already done that.


odvarkadaniel:
/hold
did something wrong

odvarkadaniel:
/hold cancel
my bad, fixed now i think!

go.mod
```diff
@@ -11,24 +11,24 @@ require (
...
-	k8s.io/client-go v0.26.3
-	k8s.io/component-base v0.26.3
-	k8s.io/klog/v2 v2.90.1
-	k8s.io/utils v0.0.0-20230313181309-38a27ef9d749
-	sigs.k8s.io/cluster-api v1.4.1
-	sigs.k8s.io/cluster-api/test v1.4.1
-	sigs.k8s.io/controller-runtime v0.14.6
+	k8s.io/api v0.27.2
+	k8s.io/apiextensions-apiserver v0.27.2
+	k8s.io/apimachinery v0.28.0-alpha.0.0.20230524221503-b43878988803
```
mdbooth, line 24: Any idea what this is about?

go.mod
```diff
@@ -11,24 +11,24 @@ require (
...
+	k8s.io/client-go v0.27.2
+	k8s.io/component-base v0.27.2
+	k8s.io/klog/v2 v2.100.1
+	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2
+	sigs.k8s.io/cluster-api v1.4.0-beta.2.0.20230526073652-9b1ead8d49ee
+	sigs.k8s.io/cluster-api/test v1.4.0-beta.2.0.20230526073652-9b1ead8d49ee
```
mdbooth, line 30: These all seem to have gone backwards.

pkg/cloud/services/compute/instance.go
```diff
@@ -300,7 +301,7 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 
 	// Wait for volume to become available
 	if volume != nil {
-		err = util.PollImmediate(retryIntervalInstanceStatus, instanceCreateTimeout, func() (bool, error) {
+		err = wait.PollUntilContextTimeout(context.TODO(), retryIntervalInstanceStatus, instanceCreateTimeout, true, func(context.Context) (bool, error) {
```
mdbooth, line 304: Nice! We should probably start passing contexts around.

odvarkadaniel: I am not familiar with this code base at all, so these are just temp fixes.
Would you prefer to keep it as is for this PR so the code gets compiled and you can fix these up in separate PR or else?

go.mod
odvarkadaniel: This should be IMO ok - this is just the main branch of `cluster-api`; `sigs.k8s.io/cluster-api@main`

go.mod
odvarkadaniel: I needed to get the master branch version of `k8s.io/apimachinery` in order to get the latest changes to the polling function (util.PollImmediate is deprecated).

go.mod
lentzi90: Yeah the "version" looks weird because of how go resolves them. It looks for the latest tag in the git history. So since the later tags (e.g. v1.4.2) is on a release branch, it will not find it. Instead it finds the latest tag that sits in the main branch.

That said, I'm not sure we should pull in main here. It makes more sense to stick to actual releases.

odvarkadaniel: This is needed to be done because the newest cluster-api version is v1.4.2 which doesn't contain the bumps to 1.27 (including big changes in controller-runtime package).
The release including this change might take some time as well.

lentzi90: Why not then wait for it to land?

pkg/cloud/services/compute/instance.go
dulek: I think this can be addressed in a separate PR, `context.TODO()` is good here.

dulek (commented):
To me it looks good, but I also think that we might want to wait for the actual release of CAPI.

seanschneeweiss (commented):
Looks good to me, and I'm also in favor waiting for the CAPI release.

mdbooth:
@odvarkadaniel I just submitted https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1600 which, it now occurs to me, is almost identical to this PR. I knew when I was making the controller-runtime fixes that I'd seen this before somewhere 🤦

We should only keep one of them. Would you like to maintain this one, or close it in favour of the new one?

PR #1567, opened 2023-05-29
✨ Flexible Nova microversions

### LINKED ISSUE #1448: Support configurable microversions for optional features like multi-attachable volumes

lentzi90:
This is related to the proposal in #1565. 

Use a list of supported versions instead of a single hard coded one. The list can be filtered based on specific feature requirements (e.g. usage of server tags). The version to use is then picked from the intersection of this list and what the server supports.

Fixes #1448

**TODOs**:

- [x] squashed commits
- [x] support nova 2.38 (no tagging)
- [x] support multiattach volumes
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/pkg/clients/compute.go
+++ b/pkg/clients/compute.go
@@ -30,19 +30,30 @@ import (
 	"github.com/gophercloud/utils/v2/openstack/clientconfig"
 
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
+	openstackutil "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/openstack"
 )
 
 /*
-NovaMinimumMicroversion is the minimum Nova microversion supported by CAPO
-2.60 corresponds to OpenStack Queens
+Constants for specific microversion requirements.
+2.60 corresponds to OpenStack Queens and 2.53 to OpenStack Pike,
+2.38 is the maximum in OpenStack Newton.
 
 For the canonical description of Nova microversions, see
 https://docs.openstack.org/nova/latest/reference/api-microversion-history.html
 
-CAPO uses server tags, which were added in microversion 2.52.
+CAPO uses server tags, which were first added in microversion 2.26 and then refined
+in 2.52 so it is possible to apply them when creating a server (which is what CAPO does).
+We round up to 2.53 here since that takes us to the maximum in Pike.
+
 CAPO supports multiattach volume types, which were added in microversion 2.60.
+
+2.38 was chosen as a base level since it is reasonably old, but not too old.
 */
-const NovaMinimumMicroversion = "2.60"
+const (
+	MinimumNovaMicroversion = "2.38"
+	NovaTagging             = "2.53"
+	NovaMultiAttachVolume   = "2.60"
+)
 
 type ComputeClient interface {
 	ListAvailabilityZones() ([]availabilityzones.AvailabilityZone, error)
@@ -57,9 +68,14 @@ type ComputeClient interface {
 	DeleteAttachedInterface(serverID, portID string) error
 
 	ListServerGroups() ([]servergroups.ServerGroup, error)
+	WithMicroversion(required string) (ComputeClient, error)
 }
 
-type computeClient struct{ client *gophercloud.ServiceClient }
+type computeClient struct {
+	client     *gophercloud.ServiceClient
+	minVersion string
+	maxVersion string
+}
 
 // NewComputeClient returns a new compute client.
 func NewComputeClient(providerClient *gophercloud.ProviderClient, providerClientOpts *clientconfig.ClientOpts) (ComputeClient, error) {
@@ -70,9 +86,25 @@ func NewComputeClient(providerClient *gophercloud.ProviderClient, providerClient
 	if err != nil {
 		return nil, fmt.Errorf("failed to create compute service client: %v", err)
 	}
-	compute.Microversion = NovaMinimumMicroversion
 
-	return &computeClient{compute}, nil
+	// Find the minimum and maximum versions supported by the server
+	serviceMin, serviceMax, err := openstackutil.GetSupportedMicroversions(*compute)
+	if err != nil {
+		return nil, fmt.Errorf("unable to verify compatible server version: %w", err)
+	}
+
+	supported, err := openstackutil.MicroversionSupported(MinimumNovaMicroversion, serviceMin, serviceMax)
+	if err != nil {
+		return nil, fmt.Errorf("unable to verify compatible server version: %w", err)
+	}
+	if !supported {
+		return nil, fmt.Errorf("no compatible server version. CAPO requires %s, but min=%s and max=%s",
+			MinimumNovaMicroversion, serviceMin, serviceMax)
+	}
+
+	compute.Microversion = MinimumNovaMicroversion
+
+	return &computeClient{client: compute, minVersion: serviceMin, maxVersion: serviceMax}, nil
 }
 
 func (c computeClient) ListAvailabilityZones() ([]availabilityzones.AvailabilityZone, error) {
@@ -154,6 +186,21 @@ func (c computeClient) ListServerGroups() ([]servergroups.ServerGroup, error) {
 	return servergroups.ExtractServerGroups(allPages)
 }
 
+// WithMicroversion checks that the required Nova microversion is supported and sets it for
+// the ComputeClient.
+func (c computeClient) WithMicroversion(required string) (ComputeClient, error) {
+	supported, err := openstackutil.MicroversionSupported(required, c.minVersion, c.maxVersion)
+	if err != nil {
+		return nil, err
+	}
+	if !supported {
+		return nil, fmt.Errorf("microversion %s not supported. Min=%s, max=%s", required, c.minVersion, c.maxVersion)
+	}
+	versionedClient := c
+	versionedClient.client.Microversion = required
+	return versionedClient, nil
+}
+
 type computeErrorClient struct{ error }
 
 // NewComputeErrorClient returns a ComputeClient in which every method returns the given error.
@@ -194,5 +241,9 @@ func (e computeErrorClient) DeleteAttachedInterface(_, _ string) error {
 }
 
 func (e computeErrorClient) ListServerGroups() ([]servergroups.ServerGroup, error) {
+	return nil, e.error
+}
+
+func (e computeErrorClient) WithMicroversion(_ string) (ComputeClient, error) {
 	return nil, e.error
 }
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -36,6 +36,7 @@ import (
 	orcv1alpha1 "github.com/k-orc/openstack-resource-controller/api/v1alpha1"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/clients"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/record"
 	capoerrors "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/errors"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/filterconvert"
@@ -104,7 +105,29 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, instanceSpec *I
 		}
 	}
 
-	server, err := s.getComputeClient().CreateServer(
+	compute := s.getComputeClient()
+	if requiresTagging(instanceSpec) {
+		s.scope.Logger().V(4).Info("Tagging support is required for creating this Openstack instance")
+		computeWithTags, err := compute.WithMicroversion(clients.NovaTagging)
+		if err != nil {
+			return nil, fmt.Errorf("tagging is not supported by the server: %w", err)
+		}
+		compute = computeWithTags
+	}
+
+	multiattachRequired, err := s.requiresMultiattach(blockDevices)
+	if err != nil {
+		return nil, fmt.Errorf("error checking if multiattach is required: %w", err)
+	}
+	if multiattachRequired {
+		s.scope.Logger().V(4).Info("Multiattach support is required for creating this Openstack instance")
+		computeWithMultiattach, err := compute.WithMicroversion(clients.NovaMultiAttachVolume)
+		if err != nil {
+			return nil, fmt.Errorf("multiattach is not supported by the server: %w", err)
+		}
+		compute = computeWithMultiattach
+	}
+	server, err := compute.CreateServer(
 		keypairs.CreateOptsExt{
 			CreateOptsBuilder: serverCreateOpts,
 			KeyName:           instanceSpec.SSHKeyName,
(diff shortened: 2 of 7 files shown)


pkg/clients/compute.go
```diff
@@ -25,23 +25,29 @@ import (
...
-CAPO uses server tags, which were added in microversion 2.52.
+CAPO uses server tags, which were first added in microversion 2.26 and then refined
+in 2.52 so it is possible to apply them when creating a server (which is what CAPO does).
 CAPO supports multiattach volume types, which were added in microversion 2.60.
 */
-const NovaMinimumMicroversion = "2.60"
+var NovaSupportedVersions = []*utils.Version{
+	{ID: "2.1", Priority: 10, Suffix: "/v2.1/"},   // Initial microversion release, same as specifying no version
+	{ID: "2.53", Priority: 20, Suffix: "/v2.53/"}, // Maximum in Pike
+	{ID: "2.60", Priority: 30, Suffix: "/v2.60/"}, // Maximum in Queens
```
jichenjc, line 49: we forced to 2.60 at commit 2df3778e989a227a542201ee7873b8d51d7f9c8d
so I think at some specific call we should use that version ,but I didn't see we have such negotiation 
in the following code, can you point me where's the logic that choose 2.60 after the change?Thanks

pkg/cloud/services/compute/instance.go
```diff
@@ -345,7 +346,19 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
...
 
-	server, err = s.getComputeClient().CreateServer(keypairs.CreateOptsExt{
+	// Filter supported versions based on features used
+	recognizedMicroversions := clients.NovaSupportedVersions
+	supportedMicroversions := []*utils.Version{}
+	if len(instanceSpec.Tags) > 0 {
+		for i := range recognizedMicroversions {
+			if recognizedMicroversions[i].ID == "2.1" {
+				continue
+			}
```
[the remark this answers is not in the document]
lentzi90: Good question! I'm not sure if this is how we should do it, but I can explain what I'm trying to do here.
Before we [set the minimum version](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/af5069cd715e401a5cbc261500286d4217267b1a) we didn't set any version at all. If I understand correctly that would mean the same thing as setting 2.1.

> If no version is specified then the API will behave as if a version request of v2.1 was requested.

From https://docs.openstack.org/nova/latest/reference/api-microversion-history.html

So what I'm trying to do here is to allow fallback to the default if possible. If tags are used, then we require 2.53 minimum so then I remove 2.1 from the versions that it can pick from. Probably a better way to do it would be to convert the versions to floats and then remove anything that is < 2.53, but for now I'm just playing with it.

pkg/clients/compute.go
lentzi90: The version is chosen [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c38aa7f44e28ab6316033555ba25d46a3b1455f0/pkg/clients/compute.go#L81). It will check what versions are supported by the server and then pick a version from this list that is supported. I also made sure to put highest priority for 2.60 so this is what should be preferred if it is available.

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
mdbooth: Incidentally however we do it, I'd probably go with ints rather than floats and drop the `2.`.

I wonder if this could be something like:
```golang
compute := s.getComputeClient()
if len(instanceSpec.Tags) > 0 {
  err := compute.RequireMicroVersion(NovaTagging)
  if err != nil {
    ...tagging not supported...
  }
}
```

We could have similar for neutron extensions.

Thoughts:
* how can we inform the user up front, if their configuration requires more than their cloud supports?
* If we can't do it up front, how do we make the best UX we can in the failure case?
* How do we do the 'we need the multi-attach microversion' thing, because we can't detect that from k8s objects?

On the latter point: we could pre-fetch the specified volume-type, or infer the default volume type (don't know if this is possible) and check if multi-attach is set. Or we could react to a failure by trying again with a higher microversion, but the granularity of OpenStack error code likely makes this messy.

pkg/clients/compute.go
```diff
@@ -25,23 +25,29 @@ import (
...
 https://docs.openstack.org/nova/latest/reference/api-microversion-history.html
 
-CAPO uses server tags, which were added in microversion 2.52.
+CAPO uses server tags, which were first added in microversion 2.26 and then refined
+in 2.52 so it is possible to apply them when creating a server (which is what CAPO does).
 CAPO supports multiattach volume types, which were added in microversion 2.60.
 */
-const NovaMinimumMicroversion = "2.60"
+var NovaSupportedVersions = []*utils.Version{
+	{ID: "2.1", Priority: 10, Suffix: "/v2.1/"},   // Initial microversion release, same as specifying no version
```
dulek, line 47: I'd put the bar higher here, it's hard to correlate but we might be implicitly depending on some features only available later. Starting from Newton [^1] would make sense to me.

[^1]: https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#maximum-in-newton

pkg/cloud/services/compute/instance.go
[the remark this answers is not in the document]
lentzi90: > how can we inform the user up front, if their configuration requires more than their cloud supports?

I think upfront is impossible here. We don't do any API calls in the webhook and for good reason, so I don't see how we would be able to do it unfortunately.

> If we can't do it up front, how do we make the best UX we can in the failure case?

Probably the best we can do here is to bubble up the error to the CAPI level where it will be visible on the Machine or the Cluster.

>   How do we do the 'we need the multi-attach microversion' thing, because we can't detect that from k8s objects?
> On the latter point: we could pre-fetch the specified volume-type, or infer the default volume type (don't know if this is possible) and check if multi-attach is set. Or we could react to a failure by trying again with a higher microversion, but the granularity of OpenStack error code likely makes this messy.

Now we are getting to the core of the issue! I *think* the reasonable thing to do is to keep a list of supported versions with higher priority for higher versions (as seen in this PR). That way the highest supported version would be picked (I hope), which should support as many features as possible. My thinking is that this would solve the issue without complicating the code with extra checks, errors and retries.

However, it builds on the assumption that picking the highest supported version is always best and I'm not completely sure if this is true. Could there be a case similar to the multi-attach volume situation, where OpenStack is configured in a way that a microversion is in theory supported, but in practice a *lower* version is needed because of `specific-volume-type-or-similar` set up by the admins? I want to think the answer is no and that it would be safe to always pick the highest available versions.

pkg/clients/compute.go
lentzi90: Fair point! I picked 2.1 because that seems to be what is used if no version is specified, so this would have been the situation before CAPO specified a version. Basically this list then represents the history of CAPO:
- No version (=2.1)
- 2.53 (added as requirement for tags)
- 2.60 (added as requirement for multi-attach volumes)

That said, I agree that it could be a good idea to set the bar higher and 2.38 sounds like a good starting place.

lentzi90: > but I'd like to make it configuration variable so someone might change it if they really want

This was also my original suggestion. I wanted to make CAPO/Gophercloud pick up the microversion from `clouds.yaml`. This way the user could set the version in there if needed and it would work the same way as for the OpenStack CLI. However, there were objections since it would be very easy to break things by setting versions that CAPO cannot work with. So now I'm trying to propose a way where CAPO still has control but it is a bit more flexible.

pkg/clients/compute.go
```diff
@@ -72,7 +86,18 @@ func NewComputeClient(providerClient *gophercloud.ProviderClient, providerClient
...
 	}
-	compute.Microversion = NovaMinimumMicroversion
+
+	// Make an initial version negotiation.
+	// The caller can redo this later with more restrictions via RequireMicroversion.
+	supportedVersions := []*utils.Version{}
+	for k := range NovaSupportedVersions {
+		supportedVersions = append(supportedVersions, NovaSupportedVersions[k])
+	}
+	version, _, err := utils.ChooseVersion(providerClient, supportedVersions)
```
mdbooth, line 96: I think it would have to be `compute` rather than `providerClient` here? Not sure what `providerClient` would test here. Keystone? Nothing?

Except that it seems this interface needs a provider client. It makes me think ChooseVersion may not be the right API for microversions, which are an artifact of the Nova API specifically.

mdbooth (commented):
I think I can see a way forward with this!

pkg/clients/compute.go
```diff
@@ -72,7 +86,18 @@ func NewComputeClient(providerClient *gophercloud.ProviderClient, providerClient
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to negotiation compute service version: %w", err)
+	}
+	compute.Microversion = version.ID
```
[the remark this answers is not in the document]
lentzi90: Hmm I'm not sure I understand. If we have deterministic microversion and no negotiation, then how would this return failure? Then we end up with the same as we already have where CAPO says the microversion is X and then we will see when making the first API call if that works or not. With negotiation we get to know this already here.
I can accept negotiating with a single microversion (so it is that microversion or nothing), but then I still think it is better to do the negotiation than just setting it without checking if it is supported.

pkg/clients/compute.go
lentzi90: Interesting! I will try to investigate this.

pkg/clients/compute.go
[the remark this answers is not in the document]
mdbooth: 2 things:

I anticipate that we'd fetch the maximum supported microversion from Nova when initially creating the service client and store it in the service client. We would then use this every time we needed to check that the required microversion was supported. i.e.

On instantiating the service client we check that microversion 38 is supported because that's the minimum. This is the microversion that all code without a higher requirement will run against.

When adding tags to a server before creation we'd do something like:
```go
  createClient, err := compute.WithMicroversion(TAGS)
  if err != nil {
    createClient = compute
  } else {
    ... add tags to server create opts ...
  }
```

pkg/clients/compute.go
lentzi90: Thank you for pointing this out! This function was just for setting the Keystone microversion (even though the docs indicated something different).

pkg/clients/compute.go
```diff
@@ -29,19 +29,28 @@ import (
...
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
+	openstackutil "sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/openstack"
 )
 
 /*
-NovaMinimumMicroversion is the minimum Nova microversion supported by CAPO
-2.60 corresponds to OpenStack Queens
+Constants for specific microversion requirements.
+2.60 corresponds to OpenStack Queens and 2.53 to OpenStack Pike,
+2.38 is the maximum in OpenStack Newton.
```
MaysaMacedo, line 38: Would it be possible to expand the comment with the reason on why 2.38 is the `MinimumNovaMicroversion`?

lentzi90: Good point! I added a note about it now. It is not set in stone at all, for more details see the [discussion here above](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1567#discussion_r1210568162)

lentzi90:
```
E1011 07:53:34.531883       1 controller.go:324] "Reconciler error" err="create OpenStack instance: error creating Openstack instance: Bad request with: [POST http://10.0.3.15/compute/v2.1/servers], error message: {\"badRequest\": {\"code\": 400, \"message\": \"Invalid input for field/attribute 1. Value: {'boot_index': -1, 'delete_on_termination': True, 'destination_type': 'volume', 'source_type': 'volume', 'tag': 'extravol', 'uuid': '270c1689-b622-4f38-8635-2bc9992e73c9'}. Additional properties are not allowed ('tag' was unexpected)\"}}" controller="openstackmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackMachine" OpenStackMachine="e2e-y1cygi/cluster-e2e-y1cygi-control-plane-svfm7" namespace="e2e-y1cygi" name="cluster-e2e-y1cygi-control-plane-svfm7" reconcileID=eebc8754-c59b-4e56-9f1c-e35f0df73a90
```

The test is failing on the [multi-AZ](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ce90495b0d975772f3b16d17c96536447bae6232/test/e2e/suites/e2e/e2e_test.go#L456) test because it does *not* see any tag in the instance spec, but a tag is still in the request. I'm a bit lost. Where is the tag coming from?
Any ideas @mdbooth ?

lentzi90:
I think it is because of the tag in the additional block volume!
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ce90495b0d975772f3b16d17c96536447bae6232/pkg/cloud/services/compute/instance.go#L503

lentzi90:
Alright, it *does* work! The failed tests before confirm that it is actually doing something also. It failed because there were tags in the `AdditionalBlockDevices` and I was only checking the instance spec. This is encouraging! I will now try to shuffle things around so that I use the gophercloud code from CAPO and see if I can get something useful from it.

jichenjc:
this is great PR, seems ok now ,we can follow up on other parts later on just work on tags in this one
still WIP now?

lentzi90:
I'm marking this ready for review now. It has been tested with and without tags and the correct version has been picked for each case, and it is passing e2e tests.

There is one part missing for multiattachable volumes though. They are trickier because we need to check the volume type in order to determine if we need it or not. Additionally, both nova and cinder are involved. For Cinder, we are not even setting any microversion at the moment.

I'm not sure if I should attempt to add multiattach here or do it separately. I will at least start working on it in a separate commit.

lentzi90:
It is failing to get devstack up...

EmilienM (approved):
Really good. I wonder if we want to document it?

lentzi90:
> Really good. I wonder if we want to document it?

Thanks! Yes some documentation would probably be good. Release notes definitely, maybe also something [here](https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration#openstack-version)?

EmilienM:
@lentzi90 feel free to add doc in this PR or do it later.
I'm going to LGTM because I think it's good as is.

Do we want to backport it? I'll let you decide and call the bot if yes.

/lgtm
/hold
please hold cancel when you want it merged.

lentzi90:
Thanks! I will do docs separately (first thing on Monday!). This has been such a long process :face_exhaling: 
/hold cancel

pkg/clients/compute.go
```diff
@@ -154,6 +186,21 @@ func (c computeClient) ListServerGroups() ([]servergroups.ServerGroup, error) {
...
+func (c computeClient) WithMicroversion(required string) (ComputeClient, error) {
+	supported, err := openstackutil.MicroversionSupported(required, c.minVersion, c.maxVersion)
+	if err != nil {
+		return nil, err
+	}
+	if !supported {
+		return nil, fmt.Errorf("microversion %s not supported. Min=%s, max=%s", required, c.minVersion, c.maxVersion)
+	}
+	versionedClient := c
+	versionedClient.client.Microversion = required
```
mdbooth, line 200: Because client is a pointer this is actually changing the Microversion globally, meaning we need to be careful to put it back again after. I wonder, is there any way to copy the ServiceClient? I haven't checked, but I have a feeling that creating a ServiceClient may not involve any API calls? If so, perhaps we could just create a new one?

lentzi90: Oh I see! We should probably do a copy. Creating a new requires one API call to check supported microversions:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7c38ac82f70f8c78e0b0244db80775c4881be8ef/pkg/clients/compute.go#L91

PR #1571, opened 2023-05-31
🌱 Bump Kubernetes version used in e2e tests

lentzi90:

We are still testing v1.25.0 in the e2e tests. I think it is time to bump to v1.26. Probably we should even go to v1.27, but let's check that this is working first.

- Remove workaround for cri-tools.
- Switch the e2e v1alpha5 and v1alpha6 cluster templates to use external
cloud provider (required for k8s v1.26).
- Move the CCM ClusterResourceSet to the common patches component to
  avoid repetition

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90:
@tormath1 could you provide a flatcar image with v1.26.5? :pray: :sweat_smile:

mdbooth:
> @lentzi90 the image is uploaded on a personal GCS bucket for testing purposes. If someone with access can upload the image on the CAPO GCS bucket once the CI is green (@mdbooth ?) that would be awesome.

@lentzi90 should also be able to do this now! If he can't that's a bug.

jichenjc:
>>>could you provide a flatcar image with v1.26.5

I didn't do flatcar before and googled this a little bit , but can someone point me the flatcar background support or it's actually another name of coreOS image?  Thanks

lentzi90:
Thank you for the image @tormath1 ! I managed to upload it myself :tada: Let's see how this works :slightly_smiling_face:

test/e2e/shared/cluster.go
```diff
@@ -91,6 +91,7 @@ func setupBootstrapCluster(config *clusterctl.E2EConfig, scheme *runtime.Scheme,
 			Name:               config.ManagementClusterName,
 			RequiresDockerSock: config.HasDockerProvider(),
 			Images:             config.Images,
+			KubernetesVersion:  config.GetVariable(KubernetesVersion),
```
lentzi90, line 94: I noticed that the kind cluster was using a different version from the workload clusters. Probably it makes more sense to use the same version, so I put it here.

lentzi90:
Oh there is no Kind node image for v1.26.5... I'll revert that part and let kind take the default version like it used to do.

test/e2e/shared/cluster.go
lentzi90: Reverted. There is no kind node image with this tag.

test/e2e/data/e2e_conf.yaml
```diff
@@ -128,9 +128,9 @@ providers:
 variables:
   # used to ensure we deploy to the correct management cluster
   KUBE_CONTEXT: "kind-capo-e2e"
-  KUBERNETES_VERSION: "v1.25.0"
-  KUBERNETES_VERSION_UPGRADE_FROM: "v1.24.9"
-  KUBERNETES_VERSION_UPGRADE_TO: "v1.25.0"
+  KUBERNETES_VERSION: "v1.26.5"
+  KUBERNETES_VERSION_UPGRADE_FROM: "v1.25.10"
+  KUBERNETES_VERSION_UPGRADE_TO: "v1.26.5"
   ETCD_VERSION_UPGRADE_TO: "3.5.4-0"
```
seanschneeweiss, line 134: Kubernetes v1.26.5 uses etcd v3.5.5, at least when checking the `go.mod`. Not sure if that is a great source of truth :)

https://github.com/kubernetes/kubernetes/blob/890a139214b4de1f01543d15003b5bda71aae9c7/go.mod#L69

seanschneeweiss: Cluster API tests upgrading to v1.26 with etcd v3.5.6
https://github.com/kubernetes/test-infra/blob/8d3c72bc9953d935ce31f109e004e25ec03c4f47/config/jobs/kubernetes-sigs/cluster-api/cluster-api-periodics-release-1-4-upgrades.yaml#L239-L240

seanschneeweiss: By the way, could we use  `stable-v1.25` and  `stable-v1.26` here as CAPI does it?
https://github.com/kubernetes/test-infra/blob/8d3c72bc9953d935ce31f109e004e25ec03c4f47/config/jobs/kubernetes-sigs/cluster-api/cluster-api-periodics-release-1-4-upgrades.yaml#L235-L238

lentzi90: Cool! I was not aware that this was possible. For now we actually have the upgrade test disabled because we use a plain cloud image. Since we use `preKubeadmCommands` to install kubelet, kubeadm etc. we cannot "switch" version in the middle of the test. :disappointed: 

I have it on my TODO to build images that we could use for these tests so we can enable them again.
I'm curious how CAPI makes sure the necessary images are always available? :thinking:

lentzi90: Ok now I see what they do. They have additional logic in their scripts that resolves these versions and builds the relevant images before starting the test.
https://github.com/kubernetes-sigs/cluster-api/blob/75c29acbb784ce26daf05d670d949eb44f76b439/scripts/ci-e2e-lib.sh#L81

lentzi90: ETCD version bumped, but I left the other versions as they were since we don't have the fancy script that CAPI use to translate `stable-v1.25` and the like. :slightly_smiling_face:

seanschneeweiss (commented):
/lgtm

PR #1572, opened 2023-05-31
🐛 Always filter cluster subnets by cluster network ID

### LINKED ISSUE #1570: Default subnet filter broken in 0.7.2
/kind bug

I have a pre-existing network with exactly one subnet, and I create an `OpenStackCluster` as follows:

```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha6
kind: OpenStackCluster
spec:
  network:
    id: <network-id>
```

i.e. specify a network filter but not a subnet filter.

Up to 0.7.1, this would deploy nodes on the single subnet for the network. As of 0.7.2, I get the following error:

```
failed to find subnet: filter matched more than one resource
```

To use the single subnet for the specified network.

Prior to 0.7.2, the network ID from the network discovered from the network filter was added to the subnet filter. In 0.7.2 this appears to have changed so that all subnets are searched, regardless of network.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.7.2
- Cluster-API version: 1.3.3
- OpenStack version: devstack HEAD
- Minikube/KIND version: N/A
- Kubernetes version (use `kubectl version`): 1.26
- OS (e.g. from `/etc/os-release`): Ubuntu

mdbooth:
Fixes a bug introduced in #1519: previously when fetching cluster subnets we filtered subnet by the cluster network id. #1519 broke that, which caused a failure due to matching too many subnets.

We additionally add an envtest which asserts the fix.

Fixes #1570

/hold

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -351,16 +351,31 @@ func (s *Service) GetSubnetsByFilter(opts subnets.ListOptsBuilder) ([]subnets.Su
 // GetSubnetByFilter gets a single subnet specified by the given SubnetFilter.
 // It returns an ErrFilterMatch if no or multiple subnets are found.
 func (s *Service) GetSubnetByFilter(filter *infrav1.SubnetFilter) (*subnets.Subnet, error) {
+	return s.getSubnetByFilter(filter.ToListOpt())
+}
+
+// GetNetworkSubnetByFilter gets a single subnet of the given network, specified by the given SubnetFilter.
+// It returns an ErrFilterMatch if no or multiple subnets are found.
+func (s *Service) GetNetworkSubnetByFilter(networkID string, filter *infrav1.SubnetFilter) (*subnets.Subnet, error) {
+	listOpt := filter.ToListOpt()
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/networking/network.go
```diff
@@ -351,16 +351,31 @@ func (s *Service) GetSubnetsByFilter(opts subnets.ListOptsBuilder) ([]subnets.Su
 // GetSubnetByFilter gets a single subnet specified by the given SubnetFilter.
 // It returns an ErrFilterMatch if no or multiple subnets are found.
 func (s *Service) GetSubnetByFilter(filter *infrav1.SubnetFilter) (*subnets.Subnet, error) {
+	return s.getSubnetByFilter(filter.ToListOpt())
+}
+
+// GetSubnetByFilter gets a single subnet of the given network, specified by the given SubnetFilter.
```
lentzi90, line 357: nit: this comment should match the function

mdbooth: Updated.

lentzi90 (approved):

PR #1577, opened 2023-06-01
⚠️ Update NetworkStatus ready for dual stack

mdbooth:
This is a breaking API change to OpenStackClusterStatus. It makes 2 changes to `status.Network`. These are done in a single PR as they're somewhat related and can't be merged separately without conflicts. I have intentionally kept them in multiple commits.

The first change moves `apiServerLoadBalancer` and `router` out of `status.network`. These are not properties of the cluster network and may evolve separately in the future. Separating them now before v1beta1 reduces the scope of future churn.

The second change transforms `network.subnet` into a slice of subnets. This is in anticipation of dual stack support, which we hope to add to a future release based on v1beta1.

The multiple commits are deliberate. I don't intend to squash them. I recommend reviewing this PR by commit, as the changes are logically separate.

Where it was simpler to do so I have updated code to handle multiple subnets, but not in all places. This change doesn't intend to support multiple subnets. It simply allows the API to represent multiple subnets.

This does not represent all API changes required for dual stack support, only one of them.

/hold


api/v1alpha7/openstackcluster_types.go
```diff
@@ -162,6 +162,12 @@ type OpenStackClusterStatus struct {
 	// externalNetwork contains information about the external network used for default ingress and egress traffic.
 	ExternalNetwork *NetworkStatus `json:"externalNetwork,omitempty"`
 
+	// router describes the default cluster router
+	Router *Router `json:"router,omitempty"`
+
+	// apiServerLoadBalancer describes the api server load balancer if one exists
```
seanschneeweiss, line 168: Comment case should match variable case `Router`, `ExternalNetwork`, `APIServerLoadBalancer` and end with a dot.

The dot linter would be great but is unfortunately not reliable enough.

api/v1alpha6/conversion.go
```diff
@@ -580,3 +591,37 @@ func Convert_v1alpha7_BindingProfile_To_Map_string_To_Interface(in *infrav1.Bind
...
+	}
+
+	// Router and APIServerLoadBalancer have been moved out of Network in v1alpha7
+	if in.Router != nil || in.APIServerLoadBalancer != nil {
+		if out.Network == nil {
+			out.Network = &Network{}
+		}
+
+		out.Network.Router = (*Router)(in.Router)
+		out.Network.APIServerLoadBalancer = (*LoadBalancer)(in.APIServerLoadBalancer)
```
seanschneeweiss, line 610: `Router` or `APIServerLoadBalancer` still could be nil, right?
If so, we could have to separate `if` blocks. One for `in.Router != nil`, one for `in.APIServerLoadBalancer != nil`.

pkg/cloud/services/compute/instance.go
```diff
@@ -182,17 +182,20 @@ func (s *Service) constructPorts(openStackCluster *infrav1.OpenStackCluster, ins
...
-		ports = []infrav1.PortOpts{{
+		port := infrav1.PortOpts{
 			Network: &infrav1.NetworkFilter{
 				ID: openStackCluster.Status.Network.ID,
 			},
-			FixedIPs: []infrav1.FixedIP{{
+			Trunk: &instanceSpec.Trunk,
+		}
+		for _, subnet := range openStackCluster.Status.Network.Subnets {
+			port.FixedIPs = append(port.FixedIPs, infrav1.FixedIP{
```
seanschneeweiss, line 192: Can help me to understand, why we need all subnets in `FixedIPs`.

api/v1alpha7/openstackcluster_types.go
```diff
@@ -157,7 +157,7 @@ type OpenStackClusterStatus struct {
 
 	// Network contains all information about the created OpenStack Network.
 	// It includes Subnets and Router.
```
seanschneeweiss, line 159: Does it still include `Router`?

seanschneeweiss:
Consider setting label `tide/merge-method-squash` for keeping the commits.

tobiasgiese:
> Consider setting label `tide/merge-method-squash` for keeping the commits.

@seanschneeweiss it‘s the other way around. Setting the squash label would squash the commits. The default merge behavior of Prow is a merge commit :)

api/v1alpha7/openstackcluster_types.go
mdbooth: Interesting. I have no strong personal preference, but as our API astronauts in OpenShift always insist on using the json tag in API godocs I was assuming it was a documented convention. And it is, but the only reference I can find is for OpenShift specifically: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-json-field-names-in-godoc

The reasoning is that users see this output in `kubectl explain`, and only the json tags make sense in that context.

But... this isn't documented in kubernetes 🤔 The [API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) don't have anything to say on the matter, and the [coding conventions](https://www.kubernetes.dev/docs/guide/coding-convention/) just point to the Go conventions, which obviously say to use the Go field. I'm genuinely surprised by that. I was kinda hoping to do a big API cleanup and change it everywhere so it's consistent, but now I'm wondering.

Thoughts?

api/v1alpha6/conversion.go
mdbooth: I recall the fuzzer tests beat me up a bit on this one. The issue is that in either case you need to check if Network is nil, so I combined them. They're no-ops if the fields are nil anyway.

pkg/cloud/services/compute/instance.go
mdbooth: This is only if we have explicitly defined multiple subnets in the cluster network. In a dual stack cluster we will have separate ipv4 and ipv6 subnets for the same network, and the port needs an address from both. In a single stack cluster this just does exactly the same as it always did.

api/v1alpha7/openstackcluster_types.go
mdbooth: Nope! I need to update that.

mdbooth (commented):
Thanks, Sean!

api/v1alpha7/openstackcluster_types.go
tobiasgiese: I think we don't even have to start with the object name itself. If we take a look at `kubectl explain` we can see that the name will be printed anyway.

Example

```
❯ k explain osc.spec | grep -A1 "^  network"
  network	<Object>
    If NodeCIDR cannot be set this can be used to detect an existing network.
```

If we want to keep the name in the comment I'd suggest to use the go notation and not the json tag.
But tbh, I don't really have a hard opinion on this.

config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
```diff
@@ -4306,6 +4306,32 @@ spec:
           status:
             description: OpenStackClusterStatus defines the observed state of OpenStackCluster.
             properties:
+              apiServerLoadBalancer:
+                description: apiServerLoadBalancer describes the api server load balancer
+                  if one exists
+                properties:
+                  allowedCIDRs:
```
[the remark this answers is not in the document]
mdbooth: For consistency I'm going to stick with Go type case for now. I thought we were obviously flouting a definite convention, which it seems we're not. As the cleanup is uncertain I'll stick with how most of them are currently defined.

api/v1alpha7/openstackcluster_types.go
mdbooth: On that there is a convention: https://www.kubernetes.dev/docs/guide/coding-convention/ points to https://go.dev/blog/godoc, which says:

```
Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.
```

mdbooth:
This update rebases on to the latest main and updates some godocs as discussed above.

seanschneeweiss (commented):
/lgtm

api/v1alpha7/openstackcluster_types.go
```diff
@@ -155,13 +155,18 @@ type OpenStackClusterSpec struct {
 type OpenStackClusterStatus struct {
 	Ready bool `json:"ready"`
 
-	// Network contains all information about the created OpenStack Network.
-	// It includes Subnets and Router.
-	Network *Network `json:"network,omitempty"`
+	// Network contains information about the created OpenStack Network.
+	Network *NetworkStatusWithSubnets `json:"network,omitempty"`
 
 	// externalNetwork contains information about the external network used for default ingress and egress traffic.
```
seanschneeweiss, line 161: Not really part of your PR, would you mind changing that too?
```suggestion
	// ExternalNetwork contains information about the external network used for default ingress and egress traffic.
```

mdbooth: I think we need a general godoc cleanup, tbh.

PR #1579, opened 2023-06-06
🐛 Fix deadlock

### LINKED ISSUE #1578: 🐛  Deletion Deadlock
/kind bug

@mdbooth your input is treasured as per usual...

When I delete a cluster (properly this time, you can prevent Argo from deleting anything and let CAPI manage the whole process), what I'm seeing is control plane machines hanging.

From what I can tell there is a race where:

* CAPI adds a owner reference on the infrastructure
* I delete the Cluster
* CAPI should delete... the MDs, the KCP, then the infrastructure
* However the owner reference trigger the infrastructure delete early...

What happens is CAPO keeps trying to delete the infra and the CP machines at the same time.
As soon as the ports are detached from the network, that completes and the infrastructure gets deleted.
However the CP machines haven't been fully deleted yet, and cannot be deleted because they need to see the infrastructure resource in order to determine whether there's a loadbalancer that needs reconciling.

I expect either:

* CAPI needs to stop messing with owner references and let its internal ordering take precedence
* CAPO needs to cache infrastructure configuration in the machines so it doesn't need to refer to the OSC resource

You can actually delete things without a hang.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.7.1
- Cluster-API version: 1.3.2
- OpenStack version: Zed
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): 1.27

spjmurray:

There's a race where the infrastructure can be deleted before the machines, but the deletion of machines is dependent on the infrastruture, and we get stuck deleting forever (unless you manually delete the machines from Nova and remove the finalizer).  Simple fix is to defer deletion of infrastructure until the machines have been purged.

Fixes #1578 

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -30,6 +31,7 @@ import (
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/collections"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -119,15 +121,29 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// Handle deleted clusters
 	if !openStackCluster.DeletionTimestamp.IsZero() {
-		return reconcileDelete(scope, cluster, openStackCluster)
+		return r.reconcileDelete(ctx, scope, cluster, openStackCluster)
 	}
 
 	// Handle non-deleted clusters
 	return reconcileNormal(scope, cluster, openStackCluster)
 }
 
-func reconcileDelete(scope scope.Scope, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster) (ctrl.Result, error) {
+func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope scope.Scope, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster) (ctrl.Result, error) {
 	scope.Logger().Info("Reconciling Cluster delete")
+
+	// Wait for machines to be deleted before removing the finalizer as they
+	// depend on this resource to deprovision.  Additionally it appears that
+	// allowing the Kubernetes API to vanish too quickly will upset the capi
+	// kubeadm control plane controller.
+	machines, err := collections.GetFilteredMachinesForCluster(ctx, r.Client, cluster)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if len(machines) != 0 {
+		scope.Logger().Info("Waiting for machines to be deleted", "remaining", len(machines))
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
 
 	if err := deleteBastion(scope, cluster, openStackCluster); err != nil {
 		return reconcile.Result{}, err


spjmurray:
Reviewers please note this is based on 0.7, so I'm expecting a 0.7.4 (fingers crossed)... If you want it against main, please speak up!

lentzi90:
I was just about to ask why is this PR for `release-0.7`. The normal process would be that you make the PR for `main`, then if it makes sense to backport, we make a cherry pick for the release branch and merge that once the original PR has been merged

spjmurray:
Noted, I'll rebase...

lentzi90:
Otherwise the next minor release (v0.8.0) would still have the bug :sweat_smile:

lentzi90:
Thanks for the effort so far!

I have a feeling this may be solved in a nicer way with finalizers. It seems the machine controller really needs the OpenStackCluster to do its job, so perhaps it should mark it with a finalizer then to make things official? We would then need to detect when the last machine is deleted and remove the finalizer at that time.

Not sure if it is a good idea or not but I would love to hear opinions :smile:

mdbooth:
My initial thoughts are:
* I agree with @lentzi90 that this looks a bit complicated
* I agree with @spjmurray that this is way better than doing it manually

I would personally be ok landing this as long as we've determined there's no simpler existing solution.

We could use a finalizer, but as you say I think that gets complicated. You'd presumably need a separate finalizer for each individual machine, which doesn't feel right either.

Can we add a second ownerReference with blockOwnerDeletion? Would that work?

spjmurray:
So, if I understand this right, every OSM would be "owned" by the OSC, that would prevent deletion of the OSC until all dependent OSMs are terminated 🤖 .   The only side effect that is that deletion of the OSC would propagate down to the OSMs, outside of the control of the KCP, dunno how that'd react!  That said, it has legs, so I'll give it a spin on the flipside, gotta enjoy that sun before the UK resorts back to torrential rain.

spjmurray:
Meh, I found time @mdbooth 😸 It appears to work from a quick test, but I'll let you know more tomorrow.  Not quite sure how it's going to play out with conversion webhooks and GVKs changing all the time.  I'll let you muse on that...

lentzi90:
I have done some digging! We could probably take inspiration from CAPI for this. They use a finalizer, but not in the way I suggested. The control plane controller just checks that all machines are gone before removing its own finalizer.
What do you think about this approach?

Ref:
https://github.com/kubernetes-sigs/cluster-api/blob/6c6543448bee351430e48f39883265f409c44662/controlplane/kubeadm/internal/controllers/controller.go#L434-L445

jichenjc:
THere are some comments in that function
```
// The implementation does not take non-control plane workloads into consideration. This may or may not change in the future.
// Please see https://github.com/kubernetes-sigs/cluster-api/issues/2064.
```

and I agree seems check the number of machines is the right way to go

spjmurray:
I've updated to that methodology, and updated the commit.  Gotta say it's ostensibly the same as my initial commit 😄  I dunno if adding the context back in, and making the delete function a receiver is a good thing or not... think that's why I initially had the check at the top level 🤷🏻 well, you have 3 commits to consider which is the best.  Tests just fine again.

lentzi90 (commented):
Looks good! Just squash the commits and I think this is good to go in :slightly_smiling_face:

spjmurray:
You know, I'm not so sure, I've seen a couple hangs now with the CAPI kubeadm control plane controller.  Perhaps there's a race still where the load balancer/API vanish too early...
```
{"ts":1686232434559.7458,"caller":"cache/reflector.go:424","msg":"k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: failed to list *v1.ClusterRole: Get \"https://185.47.227.202:6443/apis/rbac.authorization.k8s.io/v1/clusterroles?resourceVersion=5519&timeout=10s\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)\n","v":0}
{"ts":1686232434559.7974,"caller":"trace/trace.go:205","msg":"Trace[1708403150]: \"Reflector ListAndWatch\" name:k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169 (08-Jun-2023 13:53:44.559) (total time: 10000ms):\nTrace[1708403150]: ---\"Objects listed\" error:Get \"https://185.47.227.202:6443/apis/rbac.authorization.k8s.io/v1/clusterroles?resourceVersion=5519&timeout=10s\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers) 10000ms (13:53:54.559)\nTrace[1708403150]: [10.000743907s] [10.000743907s] END\n","v":0}
{"ts":1686232434559.815,"caller":"cache/reflector.go:140","msg":"k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: Failed to watch *v1.ClusterRole: failed to list *v1.ClusterRole: Get \"https://185.47.227.202:6443/apis/rbac.authorization.k8s.io/v1/clusterroles?resourceVersion=5519&timeout=10s\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)\n"}
```

Restarting the service kicks it back into life.  Which ain't perfect.  Perhaps blocking deletion was the right approach, rather than blocking removal of the finalizer.  I shall be back once I've tested some more.

spjmurray:
Right, I am definitely happy that this is hang free, for now!

controllers/openstackcluster_controller.go
```diff
@@ -119,16 +120,29 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	// depend on this resource to deprovision.  Additionally it appears that
+	// allowing the Kubernetes API to vanish too quickly will upset the capi
+	// kubeadm control plane controller.
+	machines, err := collections.GetFilteredMachinesForCluster(ctx, r.Client, cluster)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if len(machines) != 0 {
+		return ctrl.Result{}, fmt.Errorf("waiting for %d machines to be deleted", len(machines))
```
tobiasgiese, line 143: Do we want to requeue this immediately? As the machines needs some time to delete it's mayube fair to requeue  after x seconds? I don't have a hard opinion on this.

lentzi90: That's a good point! I'm not sure how long it is expected to take to delete the machines but waiting a few seconds (5?) is probably a good idea. So something like this?
```suggestion
		return ctrl.Result{RequeueAfter: 5 * time.Second}, fmt.Errorf("waiting for %d machines to be deleted", len(machines))
```

tobiasgiese: 5 seconds is a good value (tbh that‘s exactly the value I wanted to suggest 😄)

lentzi90 (approved):
/approve

lentzi90:
I think so
/cherry-pick release-0.7

spjmurray:
I knew this would happen 😄  10 minutes...

PR #1581, opened 2023-06-07
🌱 E2e: Implement LogCollector interface

lentzi90:

CAPI provides an interface LogCollector that the providers can implement to get provider specific logs from the CAPI test specs. This gives us logs from the clusterctl upgrade, self-hosted and remediation tests.

Moves the `capo-e2e-image.tar` out of the `_artifacts` since we do not need to save this with the logs. It just takes up space for no reason.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/shared/common.go
+++ b/test/e2e/shared/common.go
@@ -31,7 +31,10 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	expv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/cluster-api/util"
@@ -199,15 +202,15 @@ func dumpMachine(ctx context.Context, e2eCtx *E2EContext, machine infrav1.OpenSt
 			},
 			{
 				title: "containerd-info",
-				cmd:   "crictl info",
+				cmd:   "crictl --runtime-endpoint unix:///run/containerd/containerd.sock info",
 			},
 			{
 				title: "containerd-containers",
-				cmd:   "crictl ps",
+				cmd:   "crictl --runtime-endpoint unix:///run/containerd/containerd.sock ps",
 			},
 			{
 				title: "containerd-pods",
-				cmd:   "crictl pods",
+				cmd:   "crictl --runtime-endpoint unix:///run/containerd/containerd.sock pods",
 			},
 			{
 				title: "cloud-final",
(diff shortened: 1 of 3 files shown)


test/e2e/shared/suite.go
```diff
@@ -182,7 +182,8 @@ func AllNodesBeforeSuite(e2eCtx *E2EContext, data []byte) {
 	e2eCtx.Settings.ArtifactFolder = conf.ArtifactFolder
 	e2eCtx.Settings.ConfigPath = conf.ConfigPath
 	e2eCtx.Environment.ClusterctlConfigPath = conf.ClusterctlConfigPath
-	e2eCtx.Environment.BootstrapClusterProxy = framework.NewClusterProxy("bootstrap", conf.KubeconfigPath, e2eCtx.Environment.Scheme)
+	withLogCollector := framework.WithMachineLogCollector(OpenStackLogCollector{E2EContext: *e2eCtx})
+	e2eCtx.Environment.BootstrapClusterProxy = framework.NewClusterProxy("bootstrap", conf.KubeconfigPath, e2eCtx.Environment.Scheme, withLogCollector)
```
lentzi90, line 186: This is where the magic happens. We pass the log collector to the cluster proxy so the CAPI test specs can make use of it. The only exception I found is the clusterctl upgrade test. There is doesn't help that the bootstrap cluster proxy has the log collector, because it is the workload cluster that is of interest. I'm planning to work on a way to pass the log collector also to the workload cluster in CAPI to improve this in the future.

lentzi90:
An obvious follow up would be to refactor the rest of the e2e tests to also make use of the LogCollector. It should allow us to reduce/simplify the code a bit.

test/e2e/shared/common.go
```diff
@@ -262,3 +265,123 @@ func SetEnvVar(key, value string, private bool) {
...
+	srv := srvs[m.Name]
+
+	serverJSON, err := json.MarshalIndent(srv, "", "    ")
+	if err != nil {
+		return fmt.Errorf("error marshalling server %v: %s", srv, err)
+	}
+	if err := os.WriteFile(path.Join(machineLogBase, "server.txt"), serverJSON, 0o600); err != nil {
+		return fmt.Errorf("error writing server JSON %s: %s", serverJSON, err)
+	}
+	_, _ = fmt.Fprintf(f, "instance found: %q\n", srv.ID)
```
lentzi90, line 334: This part confuses me. It should be the same as for the normal e2e tests but for some reason it just gives "empty" server instances with all fields set to their default value.

lentzi90:
Checking if job migration worked (ref https://github.com/kubernetes/test-infra/pull/29739 )
/test pull-cluster-api-provider-openstack-build

test/e2e/shared/common.go
seanschneeweiss: For some reason `m.Spec.InfrastructureRef.Name` doesn't match any server name or is an empty string maybe?
But there is one (empty) server returned here, but with an empty name, empty id, etc.?

lentzi90: I figured it out! I was using the Machine name in `srv := srvs[m.Name]` :facepalm: it should of course be the OpenStackMachine name.

lentzi90:
All the logs are there now :slightly_smiling_face: 
I think this is ready to go in

lentzi90:
Ping @mdbooth @tobiasgiese @seanschneeweiss :slightly_smiling_face:

mdbooth (commented):
Nice! This is writing [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1581/pull-cluster-api-provider-openstack-e2e-test/1668516451239071744/artifacts/clusters/bootstrap/instances/e2e-89o9gs/cluster-e2e-89o9gs-control-plane-dl7hx/), right?

I think we need to fix the use of status.Addresses because that is going to panic on some test failures. I'd also like to understand hard-coding the containerd socket location. I'd prefer not to do that if at all possible because it feels like the sort of thing that's going to silently break at some point.

Other things are for your consideration.

test/e2e/shared/common.go
```diff
@@ -199,15 +202,15 @@ func dumpMachine(ctx context.Context, e2eCtx *E2EContext, machine infrav1.OpenSt
...
 			},
 			{
 				title: "containerd-containers",
-				cmd:   "crictl ps",
+				cmd:   "crictl --runtime-endpoint unix:///run/containerd/containerd.sock ps",
 			},
 			{
 				title: "containerd-pods",
-				cmd:   "crictl pods",
+				cmd:   "crictl --runtime-endpoint unix:///run/containerd/containerd.sock pods",
```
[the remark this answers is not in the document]
lentzi90: We are using containerd in all our tests. Previously we were relying on `crictl` checking all default locations and finding containerd that way, but if you check the logs from any recent runs you will see that it no longer works. It is now erroring out because we do not specify the endpoint. I'm not sure when exactly this changed.

test/e2e/shared/common.go
```diff
@@ -262,3 +265,122 @@ func SetEnvVar(key, value string, private bool) {
...
+
+	ip := m.Status.Addresses[0].Address
+
+	srvs, err := GetOpenStackServers(&o.E2EContext, openStackCluster, sets.New(m.Spec.InfrastructureRef.Name))
+	if err != nil {
+		return fmt.Errorf("cannot dump machines, could not get servers from OpenStack: %v", err)
+	}
+	if len(srvs) != 1 {
+		return fmt.Errorf("expected exactly 1 server but got %d", len(srvs))
+	}
```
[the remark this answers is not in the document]
lentzi90: Good point! May I suggest changing this in a follow up where I unify the log collection so that all our e2e tests use the same method? (Currently we have our own method for "our" tests and nothing for the CAPI test specs until this PR.)

test/e2e/shared/common.go
```diff
@@ -262,3 +265,122 @@ func SetEnvVar(key, value string, private bool) {
...
+		return fmt.Errorf("error writing server JSON %s: %s", serverJSON, err)
+	}
+	_, _ = fmt.Fprintf(f, "instance found: %q\n", srv.ID)
+
+	srvUser := o.E2EContext.E2EConfig.GetVariable(SSHUserMachine)
+	executeCommands(
+		ctx,
+		o.E2EContext.Settings.ArtifactFolder,
+		o.E2EContext.Settings.Debug,
+		filepath.Dir(f.Name()),
```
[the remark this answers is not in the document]
lentzi90: I must confess I just copied this from [dumpMachine](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/bc57b7f750977aebf17ec820cfa5030dc72281f2/test/e2e/shared/common.go#L186). We do write something to that file, although I do not see how it is useful. It is this one: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1581/pull-cluster-api-provider-openstack-e2e-test/1668516451239071744/artifacts/clusters/bootstrap/instances/e2e-89o9gs/cluster-e2e-89o9gs-control-plane-dl7hx/instance.log

> instance found: "fe61d653-7997-483c-ac73-839cae7fe3ce"

The content comes from this line in `dumpMachine` or line 333 here above: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/bc57b7f750977aebf17ec820cfa5030dc72281f2/test/e2e/shared/common.go#L181

`executeCommands` will not write anything to the file. It will instead write a new file in the folder for each of the commands we send.

test/e2e/shared/common.go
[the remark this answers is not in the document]
lentzi90: Here is an example where you an see how it currently fails: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1590/pull-cluster-api-provider-openstack-e2e-test/1668820465856024576#1:build-log.txt%3A2358
and as a result we do not get the corresponding logs in the artifacts: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1590/pull-cluster-api-provider-openstack-e2e-test/1668820465856024576/artifacts/clusters/bootstrap/instances/e2e-1ydzzp/cluster-e2e-1ydzzp-control-plane-4bx9j/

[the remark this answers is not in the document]
lentzi90: Crictl has this note about the endpoint:

> Note: The default endpoints are now deprecated and the runtime endpoint should always be set instead. The performance maybe affected as each default connection attempt takes n-seconds to complete before timing out and going to the next in sequence.

From https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md#usage

lentzi90:
> Nice! This is writing [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1581/pull-cluster-api-provider-openstack-e2e-test/1668516451239071744/artifacts/clusters/bootstrap/instances/e2e-89o9gs/cluster-e2e-89o9gs-control-plane-dl7hx/), right?

Exactly! :slightly_smiling_face: 

> I think we need to fix the use of status.Addresses because that is going to panic on some test failures. I'd also like to understand hard-coding the containerd socket location. I'd prefer not to do that if at all possible because it feels like the sort of thing that's going to silently break at some point.
> 
> Other things are for your consideration.

Added a guard for the address and commented on the containerd socket

test/e2e/shared/common.go
```diff
@@ -262,3 +265,125 @@ func SetEnvVar(key, value string, private bool) {
...
+		return nil
+	}
+	defer f.Close()
+
+	openStackCluster, err := getOpenStackClusterFromMachine(ctx, managementClusterClient, m)
+	if err != nil {
+		return fmt.Errorf("error getting OpenStackCluster for Machine: %s", err)
+	}
+
+	if (m.Status.Addresses == nil) || (len(m.Status.Addresses) < 1) {
```
mdbooth, line 315: Go nit: idiomatically this would be
```suggestion
	if len(m.Status.Addresses) < 1 {
```
because the length of a nil slice is well defined (as zero). Incidentally it's also safe and idiomatic to iterate over a nil slice in a for loop.

mdbooth (commented):
/lgtm

Fix the nit only if you don't want me to have a tic 😉

test/e2e/shared/common.go
lentzi90: Thanks! Fixing

PR #1585, opened 2023-06-12
🌱 Replace github.com/pkg/errors

pierreprinetti:

The package [pkg/errors](https://github.com/pkg/errors) has served its purpose and its proposed changes have been incorporated into Go's standard library in Go v1.13.

This change removes it as a direct dependency, replacing its functionality with the corresponding features of `fmt` and `errors`.

The old `pkg/errors` defined the method `Cause() error`, whereas the stdlib incorporated its functionality with the method `Unwrap() error`. As such, if a consumer of CAPO directly called `errors.Cause(err)` to match the causing error, then their code would silently break because the underlying sentinel error wouldn't be identified.

This problem is susceptible to affect any code that uses CAPO as a library, that uses  a version of `pkg/errors` below [`v0.9.0`](https://github.com/pkg/errors/releases/tag/v0.9.0).

The last versions of `pkg/errors` (v0.9.0+) [incorporated](https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L248) a compatibility layer with the new `Unwrap()` syntax.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

_EDIT: Added a note pointing out that there is a risk of compatibility issues_

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -18,11 +18,11 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"time"
 
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
@@ -109,7 +109,7 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	defer func() {
 		if err := patchHelper.Patch(ctx, openStackCluster); err != nil {
 			if reterr == nil {
-				reterr = errors.Wrapf(err, "error patching OpenStackCluster %s/%s", openStackCluster.Namespace, openStackCluster.Name)
+				reterr = fmt.Errorf("error patching OpenStackCluster %s/%s: %w", openStackCluster.Namespace, openStackCluster.Name, err)
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


pierreprinetti:
Incidentally, with this change the CI is now checking the style of the errors... which is why I had to remove the capital letter from an error message :)

controllers/openstackcluster_controller.go
```diff
@@ -232,24 +232,24 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
 		for _, address := range addresses {
 			if address.Type == corev1.NodeExternalIP {
 				if err = networkingService.DeleteFloatingIP(openStackCluster, address.Address); err != nil {
-					handleUpdateOSCError(openStackCluster, errors.Errorf("failed to delete floating IP: %v", err))
-					return errors.Errorf("failed to delete floating IP: %v", err)
+					handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
+					return fmt.Errorf("failed to delete floating IP: %v", err)
```
lentzi90, line 236:
```suggestion
					return fmt.Errorf("failed to delete floating IP: %w", err)
```

lentzi90 (commented):
Looks like a quite small change for us so I'm in favor of taking this in. From what I understand CAPI and kubeadm would have much larger changes so they are waiting for now. But I see plenty of PRs like this in k/k so I think it makes sense for us to follow.

controllers/openstackcluster_controller.go
```diff
@@ -443,16 +443,16 @@ func reconcileNetworkComponents(scope scope.Scope, cluster *clusterv1.Cluster, o
 		netOpts := openStackCluster.Spec.Network.ToListOpt()
 		networkList, err := networkingService.GetNetworksByFilter(&netOpts)
 		if err != nil {
-			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to find network: %v", err))
-			return errors.Errorf("failed to find network: %v", err)
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find network: %w", err))
+			return fmt.Errorf("failed to find network: %w", err)
 		}
 		if len(networkList) == 0 {
-			handleUpdateOSCError(openStackCluster, errors.Errorf("failed to find any network: %v", err))
```
pierreprinetti, line 450: wait. `err` is `nil` here, isn't it? Why did we wrap it, here and below?

controllers/openstackcluster_controller.go
pierreprinetti: hawk eye!

controllers/openstackcluster_controller.go
pierreprinetti: I have changed this error (and its message) in a separate commit. PTAL!

jichenjc:
/approve

Agree it's small change and we can follow other k/k behavior..
leave the unhold to other folks in case they want to take another look

lentzi90 (commented):
/hold cancel

PR #1595, opened 2023-06-21
🌱 E2e: Unify machine log collection

lentzi90:
This is a follow up to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1581.

This unifies the e2e log collection for machines so that it uses the LogCollector interface. It removes the instance.log which was just showing the ID of the openstack server and simplifies the code a bit.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/shared/openstack.go
+++ b/test/e2e/shared/openstack.go
@@ -485,6 +485,41 @@ func GetOpenStackServers(e2eCtx *E2EContext, openStackCluster *infrav1.OpenStack
 	return srvs, nil
 }
 
+// GetOpenStackServer returns the server with the given ID along with the first
+// IP address it has in the OpenStackCluster network.
+func GetOpenStackServerWithIP(e2eCtx *E2EContext, id string, openStackCluster *infrav1.OpenStackCluster) (ServerExtWithIP, error) {
+	srvExtWithIP := ServerExtWithIP{}
+	providerClient, clientOpts, _, err := GetTenantProviderClient(e2eCtx)
+	if err != nil {
+		_, _ = fmt.Fprintf(GinkgoWriter, "error creating provider client: %s\n", err)
+		return srvExtWithIP, nil
+	}
+
+	computeClient, err := clients.NewComputeClient(providerClient, clientOpts)
+	if err != nil {
+		return srvExtWithIP, fmt.Errorf("unable to create compute client: %w", err)
+	}
+	srvExt, err := computeClient.GetServer(id)
+	if err != nil {
+		return srvExtWithIP, fmt.Errorf("unable to get server: %w", err)
+	}
+	srvExtWithIP.ServerExt = *srvExt
+
+	instanceStatus := compute.NewInstanceStatusFromServer(srvExt, logr.Discard())
+	instanceNS, err := instanceStatus.NetworkStatus()
+	if err != nil {
+		return srvExtWithIP, fmt.Errorf("error getting network status for server %s: %v", srvExt.Name, err)
+	}
+
+	ip := instanceNS.IP(openStackCluster.Status.Network.Name)
+	if ip == "" {
+		return srvExtWithIP, fmt.Errorf("error getting internal ip for server %s: internal ip doesn't exist (yet)", srvExt.Name)
+	}
+	srvExtWithIP.ip = ip
+
+	return srvExtWithIP, nil
+}
+
 func GetTenantProviderClient(e2eCtx *E2EContext) (*gophercloud.ProviderClient, *clientconfig.ClientOpts, *string, error) {
 	openstackCloud := e2eCtx.E2EConfig.GetVariable(OpenStackCloud)
 	return getProviderClient(e2eCtx, openstackCloud)
--- a/test/e2e/shared/suite.go
+++ b/test/e2e/shared/suite.go
@@ -182,7 +182,7 @@ func AllNodesBeforeSuite(e2eCtx *E2EContext, data []byte) {
 	e2eCtx.Settings.ArtifactFolder = conf.ArtifactFolder
 	e2eCtx.Settings.ConfigPath = conf.ConfigPath
 	e2eCtx.Environment.ClusterctlConfigPath = conf.ClusterctlConfigPath
-	withLogCollector := framework.WithMachineLogCollector(OpenStackLogCollector{E2EContext: *e2eCtx})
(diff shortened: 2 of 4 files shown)


mdbooth:
Looks like a flake:
```
Error response from daemon: Get "https://gcr.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

lentzi90:
:open_mouth: 
That is the correct place. I must have made some mistake. Maybe I'm missing to create the folder :thinking: 
I did a test run locally where I saw the logs, but then it already had the folder structure in place... Will investigate!

lentzi90:
Alright I think I found the issue. I was not properly passing the name of the cluster which made it not find any machines.

I have one more doubt: the way I get the IP [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/facb1f0625a4136ab4709e555e8a4f58a2806ea0/test/e2e/shared/common.go#L318) does not seem to be as good as what we [do here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/facb1f0625a4136ab4709e555e8a4f58a2806ea0/test/e2e/shared/openstack.go#L474). I can see a difference for the `(without lb)` test where it cannot get the IP with the new method for one of the machines. Probably I should change it to the old method of getting the IP.

lentzi90:
Putting in draft mode until I have figured this out. Something is obviously different between my local runs and the CI

lentzi90:
Ok now it works. The folder structure is a bit strange though...
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1595/pull-cluster-api-provider-openstack-e2e-test/1673259762076094464/artifacts/clusters/bootstrap/instances/machines/cluster-e2e-11nvmz-control-plane-fkt9j/instances/e2e-11nvmz/cluster-e2e-11nvmz-control-plane-fkt9j/

test/e2e/shared/suite.go
```diff
@@ -182,7 +182,7 @@ func AllNodesBeforeSuite(e2eCtx *E2EContext, data []byte) {
 	e2eCtx.Settings.ArtifactFolder = conf.ArtifactFolder
 	e2eCtx.Settings.ConfigPath = conf.ConfigPath
 	e2eCtx.Environment.ClusterctlConfigPath = conf.ClusterctlConfigPath
-	withLogCollector := framework.WithMachineLogCollector(OpenStackLogCollector{E2EContext: *e2eCtx})
+	withLogCollector := framework.WithMachineLogCollector(OpenStackLogCollector{E2EContext: e2eCtx})
```
lentzi90, line 185: This obviously needed a pointer to the E2EContext since we are making changes to the context below.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -605,17 +614,14 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 	})
 })
 
-func createCluster(ctx context.Context, configCluster clusterctl.ConfigClusterInput) []*clusterv1.MachineDeployment {
-	result := &clusterctl.ApplyClusterTemplateAndWaitResult{}
+func createCluster(ctx context.Context, configCluster clusterctl.ConfigClusterInput, result *clusterctl.ApplyClusterTemplateAndWaitResult) {
```
lentzi90, line 617: The name of the cluster is needed when collecting logs. I had to refactor this a bit to make the name known in the `AfterEach`. The cluster object is part of the result here so we can get the name that way. This is also mimicking what CAPI is doing (although they use `ApplyClusterTemplateAndWaitResult` directly).

lentzi90:
I think this is ready to take in now. The folder structure changed a little bit from what we have before, but nothing major.
Please check the logs: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1595/pull-cluster-api-provider-openstack-e2e-test/1673285952287543296/artifacts/clusters/bootstrap/

test/e2e/shared/openstack.go
```diff
@@ -485,6 +485,42 @@ func GetOpenStackServers(e2eCtx *E2EContext, openStackCluster *infrav1.OpenStack
 	return srvs, nil
 }
 
+// GetOpenStackServer returns the server with the given ID along with the first
+// IP address it has in the OpenStackCluster network.
+func GetOpenStackServerWithIP(e2eCtx *E2EContext, id string, openStackCluster *infrav1.OpenStackCluster) (*ServerExtWithIP, error) {
```
seanschneeweiss, line 490: This is provider OpenStack, so maybe we can name this simply `GetServerWithIP`.

test/e2e/shared/openstack.go
```diff
@@ -485,6 +485,42 @@ func GetOpenStackServers(e2eCtx *E2EContext, openStackCluster *infrav1.OpenStack
...
+	ip := instanceNS.IP(openStackCluster.Status.Network.Name)
+	if ip == "" {
+		return nil, fmt.Errorf("error getting internal ip for server %s: internal ip doesn't exist (yet)", srvExt.Name)
+	}
+	srvExtWithIP := ServerExtWithIP{
+		ServerExt: *srvExt,
+		ip:        ip,
+	}
+
+	return &srvExtWithIP, nil
```
seanschneeweiss, line 521: What is the reason for returning a pointer here?

test/e2e/shared/openstack.go
lentzi90: Sure, but then we should also rename `GetOpenStackServers` above. I'm fine with either but would prefer to keep the same naming convention.

seanschneeweiss: Okay for me :)

test/e2e/shared/openstack.go
lentzi90: Good question! I think I just mimicked the function above. It would make sense to remove the pointer here. :+1:

lentzi90: Fixed!

seanschneeweiss (approved):

lentzi90:
All good from my side :slightly_smiling_face: 
I just tried explaining some of the changes that may not be obvious at first
/hold cancel

PR #1491, opened 2023-02-23
🐛  no more reconcileExternalnetwork when not use external network

ruifaling:
NodeCIDR is empty string

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1490 

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #1597, opened 2023-06-29
no more reconcileExternalnetwork when not use external network

### LINKED ISSUE #1490: No need to ReconcileExternalNetwork when disableApiserverFloatingIP and NodeCIDR is empty string
/kind bug

In openstackcluster_controller，when disableApiserverFloatingIP, will not use external network to create a floatingIP;when NodeCIDR is empty string, there is no need to reconcile network and create router between external network and subnet; 

when disableApiserverFloatingIP and the NodeCIDR is empty string, it won't excute the  ReconcileExternalNetwork

**Environment:**

- Cluster-API version:1.1.4
- OpenStack version:0.6.3
- Minikube/KIND version:

ruifaling:

Fixes #1490 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


ruifaling:
@mdbooth  @jichenjc @tobiasgiese Because the previous pr code has fallen too far behind, I resubmit the new pr,please have a re-review

The previous one is https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1491

controllers/openstackcluster_controller.go
```diff
@@ -431,10 +431,12 @@ func reconcileNetworkComponents(scope scope.Scope, cluster *clusterv1.Cluster, o
 
 	scope.Logger().Info("Reconciling network components")
 
-	err = networkingService.ReconcileExternalNetwork(openStackCluster)
-	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile external network: %w", err))
-		return fmt.Errorf("failed to reconcile external network: %w", err)
+	if !openStackCluster.Spec.DisableAPIServerFloatingIP || openStackCluster.Spec.NodeCIDR != "" {
```
mdbooth, line 434: We will reconcile external network either if APIServerFloatingIP is enabled (*), meaning we need to create a floating IP somewhere, or if NodeCIDR is set, meaning we're going to create the cluster default network.

We also need to do it if there's a Bastion defined.

* I un-doubled the negative!

mdbooth (changes requested):
This fix is messy and hard to reason about, but I think that's due to architectural tech debt elsewhere. Specifically:
* I'm more and more convinced we need a separate ControlPlane endpoint controller.
* I think we should re-implement creation of the bastion using the machine controller.

Lets add in the bastion case as I've noticed it. If other folks could think carefully if there might be more I'd appreciate it, but I think we should get this PR merged anyway and we can simplify our controllers in due course.

controllers/openstackcluster_controller_test.go
```diff
@@ -233,14 +233,14 @@ var _ = Describe("OpenStackCluster controller", func() {
 		networkClientRecorder := mockScopeFactory.NetworkClient.EXPECT()
 
 		// Fetch external network
-		networkClientRecorder.ListNetwork(networks.ListOpts{
+		/*networkClientRecorder.ListNetwork(networks.ListOpts{
```
mdbooth, line 236: ?

ruifaling: According to the Settings in the testCluster.Spec(line 218: DisableAPIServerFloatingIP is true and there is no NodeCIDR ), there is no need to check the externalNetwork, it will not execute ReconcileExternalNetwork(openStackCluster)

mdbooth: Got it. Please can you remove this code instead of commenting it out?

mdbooth (commented):
You still need to handle the case in openstackcluster_controller.go where there is a Bastion defined as, for now at least, we still need the external network in this case. We will get a panic otherwise.

mdbooth:
@EmilienM Does your external network change resolve this?

controllers/openstackcluster_controller.go
EmilienM: We now have:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/a44323cc3cb23b021a4eade9c29111a7bff0691e/pkg/cloud/services/networking/network.go#L83-L87

If this is getting too complicated, I suggest we abandon this PR, because we also have the bastion to handle and then all networking related parameters :-/.

Now we have a simple API to not reconcile external networks.

EmilienM:
> @EmilienM Does your external network change resolve this?

We now provide a way to disable the external network bits: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/622defd0bd6fa3f28f4672a1c309a71c55f4529c/docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md?plain=1#L90-L93

EmilienM:
@mdbooth I suggest we close this one. We now have a way to disable ExternalNetwork. The PR here makes it for the user but like you said this might not work fine for the bastion.

PR #1610, opened 2023-07-14
:sparkles: Allow Use of Public CAs

spjmurray:
As a public cloud operator it strikes me as odd that I have to supply a CA certificate for every cluster that's created.  This leads to a bunch of code that has to Dial the endpoint and extract the root CA (rather than hard coding it - ugh!).  The whole point of the distroless base container image is to distribute TLS root CAs, so allow the option to use what's already present.

As stated in the commit message.

None

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/scope/provider.go
+++ b/pkg/scope/provider.go
@@ -153,13 +153,13 @@ func NewProviderClient(cloud clientconfig.Cloud, caCert []byte, logger logr.Logg
 	}
 
 	config := &tls.Config{
-		RootCAs:    x509.NewCertPool(),
 		MinVersion: tls.VersionTLS12,
 	}
 	if cloud.Verify != nil {
 		config.InsecureSkipVerify = !*cloud.Verify
 	}
 	if caCert != nil {
+		config.RootCAs = x509.NewCertPool()
 		config.RootCAs.AppendCertsFromPEM(caCert)
 	}
 


pkg/scope/provider.go
```diff
@@ -153,14 +163,25 @@ func NewProviderClient(cloud clientconfig.Cloud, caCert []byte, logger logr.Logg
...
 		MinVersion: tls.VersionTLS12,
 	}
 	if cloud.Verify != nil {
 		config.InsecureSkipVerify = !*cloud.Verify
 	}
-	if caCert != nil {
-		config.RootCAs.AppendCertsFromPEM(caCert)
+
+	if useSystemRootCAs {
+		roots, err := x509.SystemCertPool()
```
spjmurray, line 173: Just a note about this block.  I did initially try without, and I thought it would just pick them up implicitly but it didn't, that is unless I made a mistake at some point, which is highly likely considering how complex this entire ecosystem is.  Besides this provides additional error checking, so never a bad thing.

spjmurray:
Implementation note: I left the default behaviour as-is, so it doesn't work unless you provide a CA certificate... odd security posture that, if you want to force the use of a CA, have admission control do it.

Anyway, I wouldn't be averse to backing out 90% of the changes, making public CAs the default, and have the end user explicitly override for self-signed infra.  Seems like the infinitely more useful way of doing it imo 🤷🏻

spjmurray:
Ah, I see what I did wrong!  Helm defaults injecting an invalid CA.  Yeah, I'm going to take the KISS approach; use the system CAs if a cert if not explicitly specified. It's literally 2 lines then.  If you want to enforce the specification of an exact CA then I can add some logic into the validating webhook.

mdbooth (approved):
https://pkg.go.dev/crypto/tls#Config says:
> // RootCAs defines the set of root certificate authorities
	// that clients use when verifying server certificates.
	// If RootCAs is nil, TLS uses the host's root CA set.
	RootCAs *[x509](https://pkg.go.dev/crypto/x509).[CertPool](https://pkg.go.dev/crypto/x509#CertPool)

With this change RootCAs is nil unless cacerts is non-nil. Following cacerts up the chain of callers, it is also nil unless --ca-certs was given on the command line. IOW, I think this behaviour is probably what most folks would have expected it already does.

PR #1614, opened 2023-07-20
🌱 chore: use capo cluster agent for API requests

### LINKED ISSUE #1611: Use reasonable User-Agent for requests
/kind bug

No details for the version of CAPO in the User Agent.

It would be very nice to be able to have the CAPO version in the User Agent to simplify troubleshooting.

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

mnaser:
At the moment, we do not modify the user agent for all API requests we make which make it hard to identify if certain requests were done by CAPO or not.

This change prepends the CAPO version to the Gophercloud user agent which will give plenty of benefits.

Fixes #1611 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/scope/provider.go
+++ b/pkg/scope/provider.go
@@ -37,6 +37,7 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha7"
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/clients"
+	"sigs.k8s.io/cluster-api-provider-openstack/version"
 )
 
 const (
@@ -151,6 +152,10 @@ func NewProviderClient(cloud clientconfig.Cloud, caCert []byte, logger logr.Logg
 	if err != nil {
 		return nil, nil, "", fmt.Errorf("create providerClient err: %v", err)
 	}
+
+	ua := gophercloud.UserAgent{}
+	ua.Prepend(fmt.Sprintf("cluster-api-provider-openstack/%s", version.Get().String()))
+	provider.UserAgent = ua
 
 	config := &tls.Config{
 		MinVersion: tls.VersionTLS12,


jichenjc:
/lgtm

make paste the updated output will be informative

mdbooth (commented):
This looks great. Unfortunately we don't currently set the version in our builds, so we're going to have to fix that too.

For an example of how this is done see: https://github.com/kubernetes/cloud-provider-openstack/blob/b229f849774e7e1976256faf2df4931841d242fa/Makefile#L39

mnaser:
@mdbooth ... i think we do..?

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/Makefile#L114-L115

though i haven't really validated here.

mdbooth:
> @mdbooth ... i think we do..?
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/Makefile#L114-L115
> 
> though i haven't really validated here.

I didn't know that was there!

/approve

pkg/scope/provider.go
```diff
@@ -152,6 +153,10 @@ func NewProviderClient(cloud clientconfig.Cloud, caCert []byte, logger logr.Logg
 		return nil, nil, "", fmt.Errorf("create providerClient err: %v", err)
 	}
 
+	ua := gophercloud.UserAgent{}
+	ua.Prepend(fmt.Sprintf("cluster-api-provider-openstack/%s", version.Get().String()))
```
stephenfin, line 157: Is there benefit in using the default user agent and simply prepending to it? That way we'll also get the gophercloud version in use. Perhaps that's overkill...?

mnaser: Personally I mimicked the patter that was used by the CPO so that's how I got where I got.

PR #1618, opened 2023-07-24
 🌱 Migrate pkg/cloud/services/networking/network.go to structured logging

jddaggett:

Networking logging structure changes to adhere to guidelines in Kubernetes documentation here --> https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

/hold

### DIFF
--- a/pkg/cloud/services/networking/router.go
+++ b/pkg/cloud/services/networking/router.go
@@ -40,7 +40,7 @@ func (s *Service) ReconcileRouter(openStackCluster *infrav1.OpenStackCluster, cl
 		return nil
 	}
 	if openStackCluster.Status.ExternalNetwork == nil || openStackCluster.Status.ExternalNetwork.ID == "" {
-		s.scope.Logger().V(3).Info("No need to create router, due to missing ExternalNetworkID")
+		s.scope.Logger().V(3).Info("No need to reconcile router since no external network exists")
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/networking/router.go
```diff
@@ -32,15 +32,15 @@ import (
...
 		return nil
 	}
 	if len(openStackCluster.Status.Network.Subnets) == 0 {
-		s.scope.Logger().V(4).Info("No need to reconcile router since no subnet exists.")
+		s.scope.Logger().V(4).Info("No need to reconcile router since no subnet exists")
 		return nil
 	}
 	if openStackCluster.Status.ExternalNetwork == nil || openStackCluster.Status.ExternalNetwork.ID == "" {
-		s.scope.Logger().V(3).Info("No need to create router, due to missing ExternalNetworkID.")
+		s.scope.Logger().V(3).Info("No need to create router, due to missing ExternalNetworkID")
```
stephenfin, line 43: We probably should rework the grammar of this while we're here. How about:

```
No need to reconcile router since no external network exists
```

?

pkg/cloud/services/networking/router.go
```diff
@@ -70,7 +70,7 @@ func (s *Service) ReconcileRouter(openStackCluster *infrav1.OpenStackCluster, cl
 		}
 		router = *createdRouter
 	} else {
-		s.scope.Logger().V(6).Info(fmt.Sprintf("Reuse existing Router %s with id %s", router.Name, router.ID))
+		s.scope.Logger().V(6).Info("Reuse existing Router", "name", router.Name, "id", router.ID)
```
stephenfin, line 73: nit: `Reusing`? Also, `Router` is title-case here but lower-case above and elsewhere. We probably want the latter unless [@mdbooth decides otherwise](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1617#discussion_r1271990511).

stephenfin (changes requested):

Issue #1314, opened 2022-07-30
Improve logging with following CAPI guideline


tobiasgiese:
There is a new logging guideline in CAPI and following this guideline and making logs consistent with CAPI will improve UX.

Guideline: https://github.com/kubernetes-sigs/cluster-api/pull/6969 / https://main.cluster-api.sigs.k8s.io/developer/logging.html

and this is the CAPI implementation: https://github.com/kubernetes-sigs/cluster-api/pull/6150

/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

mdbooth:
@seanschneeweiss Definitely agree we should keep this open. I've added it to future topics for the office hours, but my current thoughts are:

* Lets encourage all our reviewers to read the linked guidelines (I just did, and much it it was new to me)
* Where practical, lets try to ensure that all newly added or significantly changed logs conform to the guidelines.
* Where time permits, lets gradually update the existing codebase file by file and add incremental linting as described in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#prevent-re-adding-banned-functions-after-migration

seanschneeweiss:
@mdbooth 
Sounds like a great plan, thank you.

AryanSharma9917:
Hi @tobiasgiese, Just wanted to check if this issue still needs to be worked on. If yes, I’d be interested in contributing.

lentzi90:
I think this is still needed. It may not be a good first issue though. But if you want to dive in and try to find some parts where the logs can be improved, it would be very welcome! 🙂

PR #1621, opened 2023-07-24
:seedling: Migrating /pkg/cloud/services/loadbalancer/loadbalancer.go and /controllers/openstackmachine_controller.go to structured logging

anastaruno:
It fixes the format of the log messages and log parameters. Following the guidelines [here](https://main.cluster-api.sigs.k8s.io/developer/logging.html) and [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md). Part of the process to fix issue #1314.

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -169,7 +169,7 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
 		return lb, nil
 	}
 
-	s.scope.Logger().Info(fmt.Sprintf("Creating load balancer in subnet: %q", subnetID), "name", loadBalancerName)
+	s.scope.Logger().Info("Creating load balancer in subnet", "subnetID", subnetID, "name", loadBalancerName)
 
 	lbCreateOpts := loadbalancers.CreateOpts{
 		Name:        loadBalancerName,
@@ -199,7 +199,7 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
 		return listener, nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer listener", "name", listenerName, "lb-id", lbID)
+	s.scope.Logger().Info("Creating load balancer listener", "name", listenerName, "loadBalancerID", lbID)
 
 	listenerCreateOpts := listeners.CreateOpts{
 		Name:           listenerName,
@@ -267,7 +267,7 @@ func (s *Service) getOrUpdateAllowedCIDRS(openStackCluster *infrav1.OpenStackClu
 	listener.AllowedCIDRs = capostrings.Unique(listener.AllowedCIDRs)
 
 	if !reflect.DeepEqual(allowedCIDRs, listener.AllowedCIDRs) {
-		s.scope.Logger().Info("CIDRs do not match, start to update listener", "expected CIDRs", allowedCIDRs, "load balancer existing CIDR", listener.AllowedCIDRs)
+		s.scope.Logger().Info("CIDRs do not match, updating listener", "expectedCIDRs", allowedCIDRs, "currentCIDRs", listener.AllowedCIDRs)
 		listenerUpdateOpts := listeners.UpdateOpts{
 			AllowedCIDRs: &allowedCIDRs,
 		}
@@ -316,7 +316,7 @@ func (s *Service) getOrCreatePool(openStackCluster *infrav1.OpenStackCluster, po
 		return pool, nil
 	}
 
-	s.scope.Logger().Info(fmt.Sprintf("Creating load balancer pool for listener %q", listenerID), "name", poolName, "lb-id", lbID)
+	s.scope.Logger().Info("Creating load balancer pool for listener", "loadBalancerID", lbID, "listenerID", listenerID, "name", poolName)
 
 	method := pools.LBMethodRoundRobin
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -48,7 +48,7 @@ const loadBalancerProvisioningStatusActive = "ACTIVE"
 
 func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string, apiServerPort int) (bool, error) {
 	loadBalancerName := getLoadBalancerName(clusterName)
-	s.scope.Logger().Info("Reconciling load balancer", "name", loadBalancerName)
+	s.scope.Logger().Info("Reconciling load balancer", "loadBalancerName", loadBalancerName)
```
mdbooth, line 51: Lets hold off on this change until we have agreement in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1617#discussion_r1271990511

I suspect that we may decide we're going to stick with `name` here.

mdbooth:
This is fun: you have a test failure!
image

If you click on details you'll see it's a linter failure:
```
pkg/cloud/services/loadbalancer/loadbalancer.go:657: File is not `gci`-ed with --skip-generated -s standard,default (gci)
```

Honestly I still haven't worked out exactly what 'gci'-ed means, but in this case I'm guessing it's referring to the lack of a newline at the end of the file. Normally it's complaining about the ordering of imports, but not today!

If you want to run the linter tests locally you can do `make lint`, which should give you the same error. You can fix it manually, but frequently you can fix it automatically by running the linter in update mode. In this repo you can do that with `make lint-update`.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -267,7 +267,7 @@ func (s *Service) getOrUpdateAllowedCIDRS(openStackCluster *infrav1.OpenStackClu
 	listener.AllowedCIDRs = capostrings.Unique(listener.AllowedCIDRs)
 
 	if !reflect.DeepEqual(allowedCIDRs, listener.AllowedCIDRs) {
-		s.scope.Logger().Info("CIDRs do not match, start to update listener", "expected CIDRs", allowedCIDRs, "load balancer existing CIDR", listener.AllowedCIDRs)
+		s.scope.Logger().Info("CIDRs do not match, start to update listener", "expected CIDRs", allowedCIDRs, "existing CIDR", listener.AllowedCIDRs)
```
mdbooth, line 270: This is basically a trace level log message. When we get around to it, I'm pretty sure we'll put this at `V(5)`. However...

I'm not 100% sure, but I don't think we want spaces in parameter names. Perhaps `expected` and `current`?

We also probably want `..., updating listener`.

And lastly: are you able to find out how `allowedCIDRs` renders in this log message in practise? I wonder if we should explicitly format it somehow. However, don't spend much time on this given that it's most likely a trace level log.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -316,7 +316,7 @@ func (s *Service) getOrCreatePool(openStackCluster *infrav1.OpenStackCluster, po
 		return pool, nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer pool for listener", "listenerID", listenerID, "poolName", poolName, "loadBalancerID", lbID)
+	s.scope.Logger().Info("Creating load balancer pool for listener", "id", lbID, "listenerID", listenerID, "name", poolName)
```
mdbooth, line 319: I think 'pool' is the subject here, so `loadBalancerID`, `listenerID`, `name`?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -356,7 +356,7 @@ func (s *Service) getOrCreateMonitor(openStackCluster *infrav1.OpenStackCluster,
 		return nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer monitor for pool", "poolID", poolID, "monitorName", monitorName, "loadBalancerID", lbID)
+	s.scope.Logger().Info("Creating load balancer monitor for pool", "id", lbID, "name", monitorName, "poolID", poolID)
```
mdbooth, line 359: Ditto, but for 'monitor' instead of 'pool'.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -400,7 +400,7 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 	}
 
 	loadBalancerName := getLoadBalancerName(clusterName)
-	s.scope.Logger().Info("Reconciling load balancer member", "loadBalancerName", loadBalancerName)
+	s.scope.Logger().Info("Reconciling load balancer member", "name", loadBalancerName)
```
mdbooth: Tricky. The only parameter doesn't refer to the subject of the log message. i.e. We're creating a 'member', but `name` refers to the load balancer, not the member. I'd probably use `loadBalancerName` here.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -169,7 +169,7 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
 		return lb, nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer in subnet", "subnetID", subnetID, "loadBalancerName", loadBalancerName)
+	s.scope.Logger().Info("Creating load balancer in subnet", "id", subnetID, "name", loadBalancerName)
```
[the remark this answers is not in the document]
dulek: Agree! Subnet is a secondary resource here.

PR #1623, opened 2023-07-26
🌱Structured logging migration of instance.go, service.go, floatingip.go, and securitygroups.go

skylerxhu:
Fix up the string formatting log messages so that they use structured logging as documented here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

/hold

### DIFF
--- a/api/v1alpha7/openstackcluster_types.go
+++ b/api/v1alpha7/openstackcluster_types.go
@@ -229,7 +229,6 @@ type OpenStackClusterStatus struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this OpenStackCluster belongs"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="Cluster infrastructure is ready for OpenStack instances"
 // +kubebuilder:printcolumn:name="Network",type="string",JSONPath=".status.network.id",description="Network the cluster is using"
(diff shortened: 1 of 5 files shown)


pkg/cloud/services/networking/securitygroups.go
```diff
@@ -37,9 +37,9 @@ const (
 
 // ReconcileSecurityGroups reconcile the security groups.
 func (s *Service) ReconcileSecurityGroups(openStackCluster *infrav1.OpenStackCluster, clusterName string) error {
-	s.scope.Logger().Info("Reconciling security groups", "cluster", clusterName)
+	s.scope.Logger().Info("Reconciling security groups", "name", clusterName)
```
mdbooth, line 40: Couple of things here: `name` refers to the cluster not the security groups, so would need to be qualified. However...

We actually already added this parameter way back when we created the logger: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/434cfe26c624685cc574cab23e3805add6766dde/controllers/openstackcluster_controller.go#L96-L97

I think we should remove this parameter entirely.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -37,9 +37,9 @@ const (
...
 	if !openStackCluster.Spec.ManagedSecurityGroups {
-		s.scope.Logger().V(4).Info("No need to reconcile security groups", "cluster", clusterName)
+		s.scope.Logger().V(4).Info("No need to reconcile security groups", "name", clusterName)
```
mdbooth, line 42: Ditto

mdbooth (commented):
/approve

mdbooth:
Please squash commits before merging

PR #1625, opened 2023-07-31
🌱 Simplify types of Instance.SecurityGroup and Instance.Networks

mdbooth:
These both have a type of pointer to slice, which is:
* Redundant in Go
* Is marshalled identically in the CRD to a slice with no pointer

To verify the latter, note that there are no changes to the CRD despite the type change.

Also note that the values are unused.

The reason for fixing these types is that Semantic.DeepEqual can't compare them correctly so they break the fuzzer tests.

/hold


jichenjc:
um.. why the PR hold for CLA ? guess something wrong with github pipeline check now?
I can force merge if problem still exist

PR #1631, opened 2023-08-02
🌱Structured logging migration of instance.go, service.go, floatingip.go, and securitygroups.go

skylerxhu:
Fix up the string formatting log messages so that they use structured logging as documented here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -73,7 +73,7 @@ func (s *Service) normalizePortTarget(port *infrav1.PortOpts, openStackCluster *
 	// No network, but fixed IPs are defined(we handled the no fixed
 	// IPs case above): try to infer network from a subnet
 	if noNetwork {
-		s.scope.Logger().V(4).Info("No network defined for port %d, attempting to infer from subnet", portIdx)
+		s.scope.Logger().V(4).Info("No network defined for port, attempting to infer from subnet", "port", portIdx)
 
 		// Look for a unique subnet defined in FixedIPs.  If we find one
 		// we can use it to infer the network ID. We don't need to worry
@@ -95,7 +95,7 @@ func (s *Service) normalizePortTarget(port *infrav1.PortOpts, openStackCluster *
 				if err != nil {
 					// Multiple matches might be ok later when we restrict matches to a single network
 					if errors.Is(err, networking.ErrMultipleMatches) {
-						s.scope.Logger().V(4).Info("Can't infer network from subnet %d: %s", i, err)
+						s.scope.Logger().V(4).Info("Couldn't infer network from subnet", "subnetIndex", i, "err", err)
 						continue
 					}
 
@@ -433,7 +433,7 @@ func (s *Service) getOrCreateRootVolume(eventObject runtime.Object, instanceSpec
 			return nil, fmt.Errorf("exected to find volume %s with size %d; found size %d", name, size, volume.Size)
 		}
 
-		s.scope.Logger().Info("using existing root volume %s", name)
+		s.scope.Logger().V(3).Info("Using existing root volume", "name", name)
 		return volume, nil
 	}
 
@@ -580,7 +580,7 @@ func (s *Service) DeleteInstance(eventObject runtime.Object, instanceStatus *Ins
 				return nil
 			}
 
-			s.scope.Logger().Info("deleting dangling root volume %s(%s)", volume.Name, volume.ID)
+			s.scope.Logger().V(2).Info("Deleting dangling root volume", "name", volume.Name, "ID", volume.ID)
 			return s.getVolumeClient().DeleteVolume(volume.ID, volumes.DeleteOpts{})
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/compute/instance.go
```diff
@@ -95,7 +95,7 @@ func (s *Service) normalizePortTarget(port *infrav1.PortOpts, openStackCluster *
 				if err != nil {
 					// Multiple matches might be ok later when we restrict matches to a single network
 					if errors.Is(err, networking.ErrMultipleMatches) {
-						s.scope.Logger().V(4).Info("Can't infer network from subnet %d: %s", i, err)
+						s.scope.Logger().V(4).Info("Couldn't infer network from subnet", "subnet", i, "err", err)
```
mdbooth, line 98: Do we have any other instances where we're passing an error in a structured log? Do we have a standard name for them already?

mdbooth (commented):
/approve

I think it's good with or without a change of "err" to "error" (maybe), but if you could have a quick look for precedent I'd appreciate it. Thanks!

pkg/cloud/services/compute/instance.go
dulek: Also it doesn't seem like `i` is the subnet here. It's the index of the `Port`'s `FixedIPs`, we should rather use `fixedIP.Subnet` here.

pkg/cloud/services/networking/floatingip.go
```diff
@@ -179,7 +179,7 @@ func (s *Service) DisassociateFloatingIP(eventObject runtime.Object, ip string)
 }
 
 func (s *Service) waitForFloatingIP(id, target string) error {
-	s.scope.Logger().Info("Waiting for floating IP", "id", id, "targetStatus", target)
+	s.scope.Logger().Info("Waiting for floating IP", "ID", id, "status", target)
```
dulek, line 182: `status` is vague here, it doesn't say if it's the current or expected status. How about `expectedStatus`?

dulek (changes requested):

pkg/cloud/services/compute/instance.go
skylerxhu: I think we should use "err" instead of "error" according to "With the exception of acronyms like "IP" and the standard "err", don't shorten names. " from this structured logging migration guideline(https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md).

skylerxhu: There is another instance like this in this file. On line 261: `s.scope.Logger().V(4).Error(err, "Failed to clean up ports after failure")`. I'm not sure if we already have a standard name for them.

skylerxhu: For the naming issue, I had a discussion with @mdbooth last week. We think `subnetIndex` is a proper choice. Sorry for not updating it in this PR. FixedIPs is a weird naming thing from OpenStack which really means subnets.

pkg/cloud/services/networking/floatingip.go
skylerxhu: The guideline(https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md) mentioned "Do not provide additional context for how value is used. Don't use podIP, do use IP." And "Context in name is acceptable to distinguish between values that normally go under same key. For example using both status and oldStatus in log that needs to show the change between statuses."

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -321,7 +321,7 @@ func (s *Service) createSecurityGroupIfNotExists(openStackCluster *infrav1.OpenS
 		return err
 	}
 	if secGroup == nil || secGroup.ID == "" {
-		s.scope.Logger().V(6).Info("Group doesn't exist, creating it.", "name", groupName)
+		s.scope.Logger().V(6).Info("Group doesn't exist, creating it", "name", groupName)
```
stephenfin, line 324: Somewhat unrelated, but I assume there are no issues with using contractions (`doesn't` rather than `does not`) in logs?

stephenfin:
One question inline but it's more out of curiosity and not really related to this change.

/lgtm

pkg/cloud/services/networking/securitygroups.go
skylerxhu: Yeah, the guideline didn't mention the contraction issue.

pkg/cloud/services/networking/floatingip.go
dulek: I still disagree, `status` here can be understood both ways.

mdbooth: This could go in the log message: "Waiting for floating IP status", and then just "status" in the parameters. I think that would be more explicit.

stephenfin:
/lgtm

now that I have the appropriate powers

mdbooth:
Bah, needs a rebase.

mdbooth (commented):
/approve
/lgtm

PR #1633, opened 2023-08-04
✨ Add support to IPv6 in Machine's status

MaysaMacedo:
Dual-stack clusters require that the Machine's status contains the
IPv6 address. This commit removed the check that would skip the addition
of IPv6 to the Machines. Also, it ensures the IPv4 addresses of each
Network remains being the first ones listed in the Machines given there are
operations that rely on the first address of each network.

### DIFF
--- a/pkg/cloud/services/compute/instance_types.go
+++ b/pkg/cloud/services/compute/instance_types.go
@@ -150,15 +150,9 @@ func (is *InstanceStatus) NetworkStatus() (*InstanceNetworkStatus, error) {
 			return nil, fmt.Errorf("error unmarshalling addresses for instance %s: %w", is.ID(), err)
 		}
 
-		var addresses []corev1.NodeAddress
+		var IPv4addresses, IPv6addresses []corev1.NodeAddress
 		for i := range interfaceList {
 			address := &interfaceList[i]
-
-			// Only consider IPv4
-			if address.Version != 4 {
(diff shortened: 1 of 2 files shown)


MaysaMacedo:
Here is how Machine's address field looks like once this change is included:
```
status:
  addresses:
  - address: fd2e:6f44:5dd8:c956::165
    type: InternalIP
  - address: 192.168.25.158
    type: InternalIP
  - address: ostest-twv6v-master-2
    type: Hostname
  - address: ostest-twv6v-master-2
    type: InternalDNS
```

mnaser:
@MaysaMacedo maybe this is really an edge case, but would it make sense for us to try and find a way to order IPv4 addreses first then IPv6 next?

I'm just wondering in the edge case scenario that someone is relying on the first item of the `addresses` section as the IP address and they're surprised to see it become IPv6.

It's totally a small thing though, I think it's probably fine as is, but thought i'd call it out.

MaysaMacedo:
Here is how Machine's status looks like with the latest changes:
```
status:
  addresses:
  - address: 192.168.25.50
    type: InternalIP
  - address: fd2e:6f44:5dd8:c956::23
    type: InternalIP
  - address: 10.196.3.139
    type: InternalIP
  - address: ostest-gqvh8-worker-0-j9trj
    type: Hostname
  - address: ostest-gqvh8-worker-0-j9trj
    type: InternalDNS
```

mnaser (approved):

mdbooth (commented):
/approve

Looks good, thanks!

PR #1636, opened 2023-08-09
:seedling: Adding logging verbosity levels for pkg/cloud/services/loadbalancer/loadbalancer.go

anastaruno:
This PR is part of the process to improve logging in CAPO (issue #1314). It adds the proper verbosity levels to log messages in loadbalancer.go, as described in these [guidelines](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/compare/main...anastaruno:cluster-api-provider-openstack:log_levels).

In this file, I used mainly verbosity levels 2, and 3. According to the guidelines, 
- level 2: Useful steady state information about the service and important log messages that may correlate to significant changes in the system
- level 3:  Extended information about changes 

Thus, I used level 2 for the creation/deletion of important objects (load balancer, listener, pool, monitor). 
I used level 3 for 'reconciling' and 'waiting for' messages. I think those are slightly less important for understanding the process flow or debugging, but they do signify changes and deserve the attention of the developers.

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -48,7 +48,7 @@ const loadBalancerProvisioningStatusActive = "ACTIVE"
 
 func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string, apiServerPort int) (bool, error) {
 	loadBalancerName := getLoadBalancerName(clusterName)
-	s.scope.Logger().Info("Reconciling load balancer", "name", loadBalancerName)
+	s.scope.Logger().V(3).Info("Reconciling load balancer", "name", loadBalancerName)
 
 	var fixedIPAddress string
 	switch {
@@ -169,7 +169,7 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
 		return lb, nil
 	}
 
-	s.scope.Logger().Info(fmt.Sprintf("Creating load balancer in subnet: %q", subnetID), "name", loadBalancerName)
+	s.scope.Logger().V(2).Info("Creating load balancer in subnet", "subnetID", subnetID, "name", loadBalancerName)
 
 	lbCreateOpts := loadbalancers.CreateOpts{
 		Name:        loadBalancerName,
@@ -199,7 +199,7 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
 		return listener, nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer listener", "name", listenerName, "lb-id", lbID)
+	s.scope.Logger().V(2).Info("Creating load balancer listener", "name", listenerName, "loadBalancerID", lbID)
 
 	listenerCreateOpts := listeners.CreateOpts{
 		Name:           listenerName,
@@ -267,7 +267,7 @@ func (s *Service) getOrUpdateAllowedCIDRS(openStackCluster *infrav1.OpenStackClu
 	listener.AllowedCIDRs = capostrings.Unique(listener.AllowedCIDRs)
 
 	if !reflect.DeepEqual(allowedCIDRs, listener.AllowedCIDRs) {
-		s.scope.Logger().Info("CIDRs do not match, start to update listener", "expected CIDRs", allowedCIDRs, "load balancer existing CIDR", listener.AllowedCIDRs)
+		s.scope.Logger().V(3).Info("CIDRs do not match, updating listener", "expectedCIDRs", allowedCIDRs, "currentCIDRs", listener.AllowedCIDRs)
 		listenerUpdateOpts := listeners.UpdateOpts{
 			AllowedCIDRs: &allowedCIDRs,
 		}
@@ -316,7 +316,7 @@ func (s *Service) getOrCreatePool(openStackCluster *infrav1.OpenStackCluster, po
 		return pool, nil
 	}
 
-	s.scope.Logger().Info(fmt.Sprintf("Creating load balancer pool for listener %q", listenerID), "name", poolName, "lb-id", lbID)
+	s.scope.Logger().V(2).Info("Creating load balancer pool for listener", "loadBalancerID", lbID, "listenerID", listenerID, "name", poolName)
 
 	method := pools.LBMethodRoundRobin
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -169,7 +169,7 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
 		return lb, nil
 	}
 
-	s.scope.Logger().Info(fmt.Sprintf("Creating load balancer in subnet: %q", subnetID), "name", loadBalancerName)
+	s.scope.Logger().V(2).Info("Creating load balancer in subnet", "subnetID", subnetID, "name", loadBalancerName)
```
dulek, line 172: This can be just "id" here.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -199,7 +199,7 @@ func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster
 		return listener, nil
 	}
 
-	s.scope.Logger().Info("Creating load balancer listener", "name", listenerName, "lb-id", lbID)
+	s.scope.Logger().V(2).Info("Creating load balancer listener", "name", listenerName, "loadBalancerID", lbID)
```
dulek, line 202: Just "id" here too.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -267,7 +267,7 @@ func (s *Service) getOrUpdateAllowedCIDRS(openStackCluster *infrav1.OpenStackClu
 	listener.AllowedCIDRs = capostrings.Unique(listener.AllowedCIDRs)
 
 	if !reflect.DeepEqual(allowedCIDRs, listener.AllowedCIDRs) {
-		s.scope.Logger().Info("CIDRs do not match, start to update listener", "expected CIDRs", allowedCIDRs, "load balancer existing CIDR", listener.AllowedCIDRs)
+		s.scope.Logger().V(3).Info("CIDRs do not match, updating listener", "expectedCIDRs", allowedCIDRs, "currentCIDRs", listener.AllowedCIDRs)
```
dulek, line 270: If `V(2)` informs about changes in the system, it's reasonable this should be `V(2)` too.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -539,7 +539,7 @@ func (s *Service) DeleteLoadBalancerMember(openStackCluster *infrav1.OpenStackCl
 			return err
 		}
 		if pool == nil {
-			s.scope.Logger().Info("Load balancer pool does not exist", "name", lbPortObjectsName)
+			s.scope.Logger().V(2).Info("Load balancer pool does not exist", "name", lbPortObjectsName)
```
dulek, line 542: Seems like `V(3)` or even `V(4)`.

dulek (changes requested):

pkg/cloud/services/loadbalancer/loadbalancer.go
anastaruno: But subnet is not the main subject of the message, load balancer is. Hence, no prefixes for load balancer, and "subnetID" for subnet. That was my logic

pkg/cloud/services/loadbalancer/loadbalancer.go
anastaruno: Same logic here. Load balancer is not the main subject

pkg/cloud/services/loadbalancer/loadbalancer.go
anastaruno: Agreed, missed that one

pkg/cloud/services/loadbalancer/loadbalancer.go
anastaruno: Is it important or useful enough to be V(2)? Seems like something that the system does on its own without impacting the user much

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: Yep. We should also update the documentation with the upstream links we used, plus any supplemental conventions we've decided on such as this.

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: I agree with @dulek here. This branch is making an OpenStack API call and reconfiguring the listener (see line 275). This should probably log at V(2).

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -539,7 +539,7 @@ func (s *Service) DeleteLoadBalancerMember(openStackCluster *infrav1.OpenStackCl
 			return err
 		}
```
mdbooth, line 540: @dulek Not related this this PR, but this looks pretty janky. Does this need an issue to track it?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -539,7 +539,7 @@ func (s *Service) DeleteLoadBalancerMember(openStackCluster *infrav1.OpenStackCl
...
 		if pool == nil {
-			s.scope.Logger().Info("Load balancer pool does not exist", "name", lbPortObjectsName)
+			s.scope.Logger().V(4).Info("Load balancer pool does not exist", "name", lbPortObjectsName)
```
mdbooth, line 542: V(4) looks correct to me, but I'm interested to know what @dulek thinks. I don't know this code super well: would it be 'normal' for the lb pool to not exist, i.e. an expected state of affairs, even if only briefly? If so, V(4) is appropriate. If it would be unusual then perhaps V(3) would be more appropriate.

mdbooth (commented):
Nice, this is really thoughtful. I have some queries for @dulek inline as he knows OpenStack load balancers much better than I do.

pkg/cloud/services/loadbalancer/loadbalancer.go
dulek: We should probably just make a call and ignore 404's? Is this what you mean here?

mdbooth:
@anastaruno This needs a rebase.

PR #1637, opened 2023-08-09
Set failure only on instance error when no nodeRef

### LINKED ISSUE #1582: Reconciliation stops working as soon as machine is in error state
/kind bug

We had a little incidence, which broke the network for a while, the instance went in error state, causing the openstackmachine to change to Failed state.
Once the network was back, the instance went back to running state, the node (on k8s point of view) returned in Ready state, but the openstackmachine didn't recover.

This is due to the fact that capo code doesn't even try to reconcile the machine if the FailureReason (UpdateError) or FailureMessage are not nil https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/152c874c666eaa87e52bbe5591b2e7bfe4452130/controllers/openstackmachine_controller.go#L298

And this stops reconciliation even before attempting to verify instance status here https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/152c874c666eaa87e52bbe5591b2e7bfe4452130/controllers/openstackmachine_controller.go#L370

So my understanding is that this machine will never be able to recover, once it enters in error state. Even if the instance comes back.

Reconciliation is not active, so I guess it will stay stuck forever?

Reconciliation should have picked up the new state

**Environment:**

- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

mquhuy:

Setting a failure message is permanent and requires user action to get
rid of. We cannot know if the error is permanent though. The idea with
this commit is that if the machine has a nodeRef, we know that the
instance was working at some point, so the error could well be
temporary. Therefore we do not set failure in this case.
On the other hand, if the machine does not have a nodeRef, we set
failure since it is more likely that there is some configuration error
that will not go away.

Fixes #1582 . For more details please read my comment on that issue.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -334,14 +334,6 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 		return ctrl.Result{}, err
 	}
 
-	// Set an error message if we couldn't find the instance.
-	if instanceStatus == nil {
-		err = errors.New("OpenStack instance not found")
-		openStackMachine.SetFailure(capierrors.UpdateMachineError, err)
-		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "")
-		return ctrl.Result{}, nil
-	}
-
 	// TODO(sbueringer) From CAPA: TODO(ncdc): move this validation logic into a validating webhook (for us: create validation logic in webhook)
 
 	openStackMachine.Spec.ProviderID = pointer.String(fmt.Sprintf("openstack:///%s", instanceStatus.ID()))
@@ -364,10 +356,14 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
-		// Error is unexpected, thus we report error and never retry
+		// If the machine has a NodeRef then it must have been working at some point,
+		// so the error could be something temporary.
+		// If not, it is more likely a configuration error so we set failure and never retry.
 		scope.Logger().Info("Machine instance state is ERROR", "id", instanceStatus.ID())
-		err = fmt.Errorf("instance state %q is unexpected", instanceStatus.State())
-		openStackMachine.SetFailure(capierrors.UpdateMachineError, err)
+		if machine.Status.NodeRef == nil {
+			err = fmt.Errorf("instance state %q is unexpected", instanceStatus.State())
+			openStackMachine.SetFailure(capierrors.UpdateMachineError, err)
+		}
 		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceStateErrorReason, clusterv1.ConditionSeverityError, "")
 		return ctrl.Result{}, nil
 	case infrav1.InstanceStateDeleted:
@@ -429,6 +425,8 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, computeService *compute.Service, userData string) (*compute.InstanceStatus, error) {
 	instanceStatus, err := computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
 	if err != nil {
+		logger.Info("Unable to get OpenStack instance", "name", openStackMachine.Name)
+		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.OpenStackErrorReason, clusterv1.ConditionSeverityError, err.Error())
 		return nil, err
 	}
 
--- a/api/v1alpha7/conditions_consts.go
+++ b/api/v1alpha7/conditions_consts.go
@@ -40,6 +40,8 @@ const (
 	InstanceNotReadyReason = "InstanceNotReady"
 	// InstanceDeleteFailedReason used when deleting the instance failed.
 	InstanceDeleteFailedReason = "InstanceDeleteFailed"
+	// OpenstackErrorReason used when there is an error communicating with OpenStack.
+	OpenStackErrorReason = "OpenStackError"
 )
 
 const (


wwentland:
Thank you, @mquhuy! Could you be so kind as to look into the status of your CLA again?

lentzi90:
He is in vacation this week. I will check it with him when he is back :slightly_smiling_face:

mdbooth:
We've long known we've got a bunch of cleanup work to do here, but until now I've thought the goal would be:
* Failure before machine is Ready -> Failed
* Failure after machine is Ready -> Retry forever

My thinking is that if the machine has not become Ready then it has no workload to worry about, so you might as well delete it. I'm prepared to be challenged on this thinking, though.

A concern I have is that this would mask some errors which probably shouldn't be retried, e.g. specifying an invalid resource. Some of these things we (sometimes) validate prior to server creation, but ultimately we rely on Nova to return us an error if, e.g. the user specified an invalid image ID. Not putting the machine into a failed state in this situation makes a configuration error much harder to find.

If we were going to ignore creation failure like this, at a minimum I'd like to work out precisely which errors are ignorable. We should also look at Nova's return value for server create to work out if it distinguishes between an invalid request and a transient error. It probably does, and if so we should distinguish between those two. Ideally we should ignore transient errors and retry, but we should not ignore invalid requests. If we can't do the work to distinguish between the two yet, my preference would probably be to leave it as-is and make *all* errors prior to machine Ready fatal. You can still just delete the machine.

Again, this is just my thinking, though. I'm very much prepared to have it challenged by real-world experience.

Interested to know if @mkjpryor or @mnaser have an opinion.

wwentland:
> We've long known we've got a bunch of cleanup work to do here, but until now I've thought the goal would be:
> 
>     * Failure before machine is Ready -> Failed
>     * Failure after machine is Ready -> Retry forever

To me this sounds like exactly the desired  behaviour. The issue detailed in #1582 does, AFAIUI, happen when the machine has been Ready before, but does not recover from an error state once the network has been unavailable for extended periods of time.

I think this PR is motivated by the suspicion that the controller runs into the following code path when there is network instability as `instanceStatus` might be nil, even if the instance actually exists:

```
// Set an error message if we couldn't find the instance.
	if instanceStatus == nil {
		err = errors.New("OpenStack instance not found")
		openStackMachine.SetFailure(capierrors.UpdateMachineError, err)
		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "")
		return ctrl.Result{}, nil
	}
```

lentzi90:
Exactly on point @wwentland ! :dart: 
The issue we saw was a machine that was ready, then connection to nova was lost, which put it into error state where it then stayed forever.

lentzi90:
I'm taking this over since @mquhuy is gone for a bit longer (we are colleagues).

lentzi90:
@mdbooth @mkjpryor @mnaser any more comments?

mdbooth:
> I think this PR is motivated by the suspicion that the controller runs into the following code path when there is network instability as `instanceStatus` might be nil, even if the instance actually exists:
> 
> ```
> // Set an error message if we couldn't find the instance.
> 	if instanceStatus == nil {
> 		err = errors.New("OpenStack instance not found")
> 		openStackMachine.SetFailure(capierrors.UpdateMachineError, err)
> 		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "")
> 		return ctrl.Result{}, nil
> 	}
> ```

I would prefer to fix `instanceStatus` might be nil, even if the instance actually exists. That is certainly not the intention. If `instanceStatus` exists but can't be fetched for some reason (I think cell down would be the only cause of this other than a regular bug @stephenfin?) it should be an error, so we should not continue. If we are returning a nil instanceStatus with no error in any other circumstance I would prefer to address that instead.

mkjpryor:
@mdbooth 

This all sounds sensible. My one thought would be whether we have a timeout that results in a transition to Failed rather than retrying infinitely. This would help those of us building UIs on top of CAPI, where an infinitely retrying entity is quite hard for users to grok, especially non-technical ones…

lentzi90:
The whole point is that if we call `SetFailure`, that is a *permanent* condition that the Machine at CAPI level will pick up and never update. We should not do that for a simple network issue that could be resolved on the next reconciliation. I'm saying we *should* retry, because that is the only way to resolve this kind of situation automatically. The alternative is to have the user delete/patch the Machine and this is definitely madness for non-technical users.

Regarding the `instanceStatus` being nil I'm a bit lost. I don't think there is a situation where the instance exists and the status is still nil. What can happen is that we manage to get the instances fine, but find that the one we are looking for is not yet created so we try to create it. At that point we can see network issues and end up with an error because we were unable to create it.

My suggestion is to only set conditions (not failure) for these two cases:

1. We are unable to get the OpenStack instance
2. We are unable to create the OpenStack instance

If we have *specific* cases that we know are permanent errors, then I'm fine with adding checks for them, but without checking the error I say we should treat it as temporary.

controllers/openstackmachine_controller.go
```diff
@@ -334,8 +334,6 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 
 	// Set an error message if we couldn't find the instance.
 	if instanceStatus == nil {
-		err = errors.New("OpenStack instance not found")
```
mdbooth, line 337: I wonder if the `if instanceStatus == nil` check here is bogus. Looking at https://github.com/Nordix/cluster-api-provider-openstack/blob/3c742d2cd136ca2c64b1245a97230ae038eafe84/controllers/openstackmachine_controller.go#L427-L444 I don't think this should ever happen. One of the following things should have happened:
* The instance existed and we returned it (instanceStatus != nil)
* The instance did not exist and we created it successfully (instanceStatus != nil)
* The there was an error either determining if the instance exists, or creating the instance (err != nil, handled above)

I think this might actually be dead code. Probably best to delete it entirely. I think this code is my fault, actually. I think I was intending to make additional changes in which CreateInstance could return nil but no error. Looking at it now I don't think there are any. Lets not leave in half-implemented code.

Are you sure this is the code which is causing your issue?

jichenjc:
>>>  We are unable to get the OpenStack instance
    We are unable to create the OpenStack instance

can we define this more clearly ? e.g many cases we can say we can't create, like no network connection temply, or it's param error, or something wrong server side (openstack side return 500) but they can be fixed soon etc
so need define this and doc clearly on the desired behavior

controllers/openstackmachine_controller.go
lentzi90: Oh I see! I was somehow mixing this up with the same check in `getOrCreate` below :facepalm: 
I will remove it

mdbooth:
There are currently only 3 places where we call machine.SetFailure:

* [reconcileNormal if getOrCreate returns nil InstanceStatus](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1384d61f356e53010a1a1aca56752c4c3348f149/controllers/openstackmachine_controller.go#L340-L341)
* [reconcileNormal if the machine is in ERROR state](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1384d61f356e53010a1a1aca56752c4c3348f149/controllers/openstackmachine_controller.go#L370-L371)
* [reconcileDelete if we can't get a NetworkStatus](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1384d61f356e53010a1a1aca56752c4c3348f149/controllers/openstackmachine_controller.go#L257-L261)

Presumably your issue is the result of one of these 3.

The first is this one. As noted above I think that's dead code and we should just delete it, but (assuming it's dead code) it's not your issue.

The last isn't your issue either because you're not deleting. AFAICT the only way that could happen is if OpenStack was returning us data we can't parse, which should never happen. It's probably not worth spending a lot of time thinking about it.

So I'm guessing your issue is the second one. At the end of a reconcile the machine is in the ERROR state. My take on this one is that we should only SetFailure() here if the machine has ever been ready.

lentzi90:
Thanks @mdbooth that makes sense! I will try to reproduce the issue to double check this is indeed the case.

lentzi90:
Alright I now have some more information on this. It is not about network (at least not between CAPO and Openstack). The issue was that some part of Openstack was down (I guess Nova?), which caused all servers to report error status. So to summarize:

- The VMs had been ready
- There was a temporary issue with the Openstack environment that caused all servers to report error status
- When the situation was resolved, all OpenStackMachines were stuck permanently in failed state

Sounds to me like it was working as intended. From a user point of view I can also understand the frustration that the Machines and OpenStackMachines stayed in failed state, but I guess this is such a rare/unique situation that it may still be for the best.

lentzi90:
> I still think we can do better here. I don't think we should be setting the failed condition on a Machine which has had the opportunity to run a workload. How about we look to see if the Machine has a nodeRef? If it doesn't we set failed. If it does we retry forever and... do something else to make it really obvious there's a problem which needs looking at.

Sounds good, I will give it a go!

controllers/openstackmachine_controller.go
```diff
@@ -364,10 +356,13 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
-		// Error is unexpected, thus we report error and never retry
+		// If the machine has a NodeRef the error could be something temporary.
```
mdbooth, line 359: Can we expand on this comment a bit? The comment is correct, but it might not be immediately obvious why. Could we explain that if the Machine has a NodeRef then we know it was working at some point, and therefore the error (presumably) isn't in the spec.

controllers/openstackmachine_controller.go
```diff
@@ -429,6 +424,8 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, computeService *compute.Service, userData string) (*compute.InstanceStatus, error) {
 	instanceStatus, err := computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
 	if err != nil {
+		logger.Info("Unable to get OpenStack instance", "Machine", openStackMachine.Name)
```
mdbooth, line 427: nit
```suggestion
		logger.Info("Unable to get OpenStack instance", "name", openStackMachine.Name)
```

controllers/openstackmachine_controller.go
lentzi90: Good point, fixed!

controllers/openstackmachine_controller.go
```diff
@@ -429,6 +425,8 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *clusterv1.Cluster, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, computeService *compute.Service, userData string) (*compute.InstanceStatus, error) {
 	instanceStatus, err := computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
 	if err != nil {
+		logger.Info("Unable to get OpenStack instance", "name", openStackMachine.Name)
+		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, err.Error())
```
mdbooth, line 429: I wonder if we need a new reason here. This isn't really 'InstanceNotFound', it's ... 'OpenStackError'?

lentzi90: Right! I did a bit too much copy paste on that...

mdbooth (commented):
/lgtm

PR #1639, opened 2023-08-09
🌱 Remove embedding in NetworkStatusWithSubnets

mdbooth:
This reverts commit d961508e4ef7e6786ced7cf778de8a241f15e119

The apiequality.Semantic.DeepEqual can't correctly compare all possible zero values of a struct with an embedded struct, making this more trouble than it's worth.

/hold


mdbooth:
Turns out this wasn't the problem at all. The problem was the difference between `&{}` and `nil`, which, unlike for map and slice, are genuinely different.

PR #1640, opened 2023-08-09
✨ Add EphemeralDisk to InstanceSpec

EmilienM:

An additional block device can be attached to the instance. It won't be bootable and only optional.
It can be added aside of a `RootVolume` or not.

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
metadata:
  name: <cluster-name>-controlplane
  namespace: <cluster-name>
spec:
...
  ephemeralDisk:
     format: "ext4"
     size: 10
...
```

The local disk will be deleted when the instance is removed.

A use case is when having etcd running on machines that boot from volumes, you can now have a dedicated local disk with higher I/O for it.


EmilienM:
This is just a prototype for now. Also for some reasons I can't make the conversion working yet.

EmilienM:
struggling a bit with the conversion & their tests...

ching-kuo:
You might need to restore the value during up conversion like this PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1517

jichenjc:
for new spec I suggest create an issue to track the proposed change and create release notes for it... @EmilienM

mdbooth:
Couple of things to think about:

There's prior art here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e7773fe017f3493be65c3368808d726a6343911e/api/v1beta1/azuremachine_types.go#L77-L82

It might be worth considering using the same terminology unless there's a good reason to do otherwise.

Also consider: how do I specify that I want a secondary disk and it's the nova local disk, vs it's a cinder disk? I suppose it might be neater to give the local disk a separate field in the API because there can only be one of them, vs any number of cinder disks.

I believe the CAPZ folks have done the whole thing end-to-end, i.e. they're also able to mount the disk. Might be worth looking into how they did that.

EmilienM:
@mdbooth thanks, I'm looking into this
@ching-kuo I pushed a new patchset, I managed to reduce the failures, but I still have a couple. I wonder if you have an idea of what I've missed.
@jichenjc I'll do it, once we settle on the API needs, based on Matt's comments. I'll come up with something and create an issue.

PR #1644, opened 2023-08-14
WIP: ✨ Add DisableHealthMonitor option on API Server LoadBalancer

### LINKED ISSUE #1643: Add the option to disable Octavia healthmonitor creation
/kind feature

Currently, Openstack Cloud Controller Manager have the option on its plugin to disable the creation of health-monitors on newly provisioned Load Balancers from the workload cluster. 

I would like to discuss the possibility to also add this option on CAPO for the creation of the API Server Loadbalancer. I'm facing some issues where the health-monitor is breaking the LoadBalancer status during instance reconciliation, changing it to Operating Status Failed. Due to this feature is also available on OCCM, I don't see any problems implementing it on CAPO side, but I would like to discuss this before moving forward. 

Even if the description of the issue I'm facing isn't really deep, my team is dealing with it for several months, as we're waiting for a deep debug to really understand this infra problem, I'm able to commit the code for this feature, as I've already started to work on it.

jfcavalcante:

Fixes #1643 

1. This is my first contribution, so please, if there's any guideline not being followed, point me to it and I'll fix.

**TODOs**:

- [ ] finish fuzzy tests implementation

/hold


dulek:
This makes perfect sense, some Octavia backends still doesn't support health monitors. What is the expected default value of the option? Shouldn't we indicate it on the option's description?

dulek:
I've just got some more context. Let's hash out my concerns on https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1645 first. I do not understand why the current solution of skipping creation in case of 501 doesn't work with your env.

jfcavalcante:
> I've just got some more context. Let's hash out my concerns on https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1645 first. I do not understand why the current solution of skipping creation in case of 501 doesn't work with your env.

This PR isn't related to that issue. This is another problem that I've hit with other env.

dulek:
> > I've just got some more context. Let's hash out my concerns on #1645 first. I do not understand why the current solution of skipping creation in case of 501 doesn't work with your env.
> 
> This PR isn't related to that issue. This is another problem that I've hit with other env.

Alright, can you explain why disabling health monitors explicitly is important for your use case then?

jfcavalcante:
My team is facing some issues with health monitors crashing loadbalancers status and preventing cluster booststrap. Our infra team couldn't find exactly what is the problem with health monitors yet, but they suspect of some problem with the messaging system. Currently, we can disable health monitor creation on new provisioned LBs from the workload cluster, but not the api server one created during bootstrap. As openstack cloud controller manager already enables this feature, I think we can bring it to CAPO as well.

api/v1alpha7/types.go
```diff
@@ -292,6 +292,8 @@ type APIServerLoadBalancer struct {
 	AllowedCIDRs []string `json:"allowedCidrs,omitempty"`
 	// Octavia Provider Used to create load balancer
 	Provider string `json:"provider,omitempty"`
+	// Enabled defines whether health monitors should be created for the load balancer.
```
mdbooth, line 295: `DisabledHealthMonitor` defines...

api/v1alpha7/types.go
```diff
@@ -292,6 +292,8 @@ type APIServerLoadBalancer struct {
...
+	DisabledHealthMonitor bool `json:"disabledHealthMonitor,omitempty"`
```
mdbooth, line 296: Please could we call the option `DisableHealthMonitors`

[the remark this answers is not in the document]
mdbooth: I think this is ok: the default is the zero value, which is false in this case.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -133,8 +133,10 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 			return false, err
 		}
 
-		if err := s.getOrCreateMonitor(openStackCluster, lbPortObjectsName, pool.ID, lb.ID); err != nil {
-			return false, err
+		if !openStackCluster.Spec.APIServerLoadBalancer.DisabledHealthMonitor {
+			if err := s.getOrCreateMonitor(openStackCluster, lbPortObjectsName, pool.ID, lb.ID); err != nil {
+				return false, err
+			}
```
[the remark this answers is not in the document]
mdbooth: I think it's ok to add that separately, as without allowing it explicitly in the webhook the value will be immutable.

api/v1alpha7/types.go
jfcavalcante: I think it makes sense

mdbooth:
@jfcavalcante are you still interested in this? Seems promising.

jfcavalcante:
Hey, @mdbooth! Yes, I was in a shortage of time, but I'll be able to work on this in the next days.

PR #1646, opened 2023-08-17
🐛  Fix application credential support in env.rc

wwentland:

This allows users to utilise authentication via application credentials again rather than expecting them to use passwords.

Specifically, users that use a cloud.conf with application credentials would run into an error such as:

```
$ source ./templates/env.rc ~/.config/openstack/clouds.yaml example
Error: Value for env variable 'PASSWORD' not provided in env()
```

with the resulting `CAPO_OPENSTACK_CLOUD_YAML_SELECTED_CLOUD_B64` environment variable containing content similar to:

```
clouds:
  example:
```

This commit only includes the password if it does indeed have a value, to ensure that the cloud config file can be rendered correctly.

The erroneous behaviour was introduced in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1381

/hold


lentzi90 (commented):
/approve

PR #1655, opened 2023-09-07
avoid duplicate error event

jichenjc:

this is from code read when doing other PR 

seems we have event https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/trunk.go#L73 so no need duplicate event to confuse admin

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/networking/trunk.go
+++ b/pkg/cloud/services/networking/trunk.go
@@ -70,7 +70,6 @@ func (s *Service) getOrCreateTrunk(eventObject runtime.Object, clusterName, trun
 
 	trunk, err := s.client.CreateTrunk(trunkCreateOpts)
 	if err != nil {
-		record.Warnf(eventObject, "FailedCreateTrunk", "Failed to create trunk %s: %v", trunkName, err)
 		return nil, err
 	}
 


jichenjc:
there's only one usage, so update directly in the function...

pkg/cloud/services/networking/port.go
```diff
@@ -179,7 +179,6 @@ func (s *Service) GetOrCreatePort(eventObject runtime.Object, clusterName string
 	if portOpts.Trunk != nil && *portOpts.Trunk {
 		trunk, err := s.getOrCreateTrunk(eventObject, clusterName, port.Name, port.ID)
 		if err != nil {
-			record.Warnf(eventObject, "FailedCreateTrunk", "Failed to create trunk for port %s: %v", portName, err)
```
mdbooth, line 182: I was thinking keep this one and remove the events in getOrCreateTrunk()? By emitting the event at both error points we're no longer avoiding duplicating the error event.

dulek: Yeah, I'd probably go with what Matt says here, but no strong preference.

PR #1662, opened 2023-09-08
⚠️ Add OpenStackMachineSpec.RootVolume.UseMachineAZ

### LINKED ISSUE #1649: Compute AZs should not be used as Volume AZs
/kind feature

In #1030, we added support for defining root volume AZs, closing #1021. When this was implemented, a decision was made to use the MachineSpec's `failureDomain` attribute to populate `rootVolume.availabilityZone` if the latter was not configured. This feels like an anti-pattern. Compute AZs are not Volume AZs, and while they may be identical in smaller, simpler deployments with a single AZ (Cinder's default AZ is `nova` to allow this) or hyper-converged deployments, where `nova-compute` and `cinder-volume` run side-by-side on hosts, it's unlikely to be the case elsewhere. I've seen this pop up already in different deployments.

Is there any reason why we can't default to no implicit AZ or at least validate the implicit AZ exists?

I'm not sure if this change in behavior would need to be placed behind an API version or not. Would welcome input in any case.

mdbooth:
TODO: write me

Fixes #1649 

cc @stephenfin 

/hold


stephenfin (commented):
If I understand correctly, this means that by default, we will not explicitly request an AZ for Cinder and will instead get whatever is provided by Cinder. If you set `UseMachineAZ=true`, you'll get the Compute AZ, and if you set `AvailabilityZone=foo`, you'll get the `foo` AZ. If so, this makes sense to me.

The _one_ suggestion for improvement I have is to name the option `allow_cross_az_attach` and default it to `True`. This is the name of an option in Nova which, when set to `False`, causes Nova to use the AZ of the compute host when creating the volume. Assuming people understand this option, we could make their lives a bit easier through aligning the two? Again though, I'm happy with this also.

mdbooth:
Fixed by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2008

mdbooth:
We'll redesign this when it comes up again. I don't think there's value in keeping this PR around.

PR #1663, opened 2023-09-08
✨Allow changing allowAllInClusterTraffic in a deployed cluster

### LINKED ISSUE #1653: Changes on OpenstackCluster spec is forbidden
/kind bug

I installed clusters with `spec.allowAllInClusterTraffic` disabled. Then I attempted to set `spec.allowAllInClusterTraffic` as true to let kube-prom-stack to communicate w/ k8s components. But admission webhook declined this change.

There are usecases apparently to be required to change not only `allowAllInClusterTraffic` but also other fields in OpenstckCluster's spec. I expect for CAPO to allow to change OpenstackCluster spec.

**Environment:**
Environment(version infor) doesn't matter. Btw, here is my version infor

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built):  v0.7.1
- Cluster-API version: v1.4.4
- OpenStack version: zed
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): 1.24
- OS (e.g. from `/etc/os-release`): ubuntu jammy

okozachenko1203:
…affic

It allows the change on `spec.allowAllInClusterTraffic` of `OpenstackCluster` CRs.

Fixes #1653

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha7/openstackcluster_webhook.go
+++ b/api/v1alpha7/openstackcluster_webhook.go
@@ -130,6 +130,10 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) (admission.Warn
 	old.Spec.ControlPlaneAvailabilityZones = []string{}
 	r.Spec.ControlPlaneAvailabilityZones = []string{}
 
+	// Allow change to the allowAllInClusterTraffic.
+	old.Spec.allowAllInClusterTraffic = false
+	r.Spec.allowAllInClusterTraffic = false
+
 	if !reflect.DeepEqual(old.Spec, r.Spec) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
 	}
--- a/api/v1alpha7/openstackcluster_webhook.go
+++ b/api/v1alpha7/openstackcluster_webhook.go
@@ -131,8 +131,8 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) (admission.Warn
 	r.Spec.ControlPlaneAvailabilityZones = []string{}
 
 	// Allow change to the allowAllInClusterTraffic.
-	old.Spec.allowAllInClusterTraffic = false
-	r.Spec.allowAllInClusterTraffic = false
+	old.Spec.AllowAllInClusterTraffic = false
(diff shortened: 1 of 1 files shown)


mdbooth:
Just to be clear, the reason we currently don't allow users to change this field is that we don't support changing it. i.e. It's good UX to prevent it from being changed, because if the user did change it it wouldn't work.

We can't just allow the user to change the field. It won't make it work, and now their cluster might be broken because the controller is making assumptions about state which we've broken.

The difficult work here is implementing the ability to change the configuration of an existing cluster. We need to do that in the same change (i.e. this PR) which allows the API field to be updated.

jichenjc:
>>>The difficult work here is implementing the ability to change the configuration of an existing cluster. We need to do that in the same change (i.e. this PR) which allows the API field to be updated.

right, this is the key request, we should reconcile to the desire state after the API is accepted...

dulek:
/lgtm

From my point of view it looks good.

@mdbooth: Yeah, though the comparisons done in CAPO code are much simpler than in CPO, where[ we actually compare more](https://github.com/kubernetes/cloud-provider-openstack/blob/aeb7a5ef4c8290eec720509bdda68ebf753083ea/pkg/openstack/loadbalancer.go#L2180) than just `remote_group_id`.

jichenjc:
sorry missed one comment in the original issue :( , thanks for the updated comments
one comment is this change worthy a release NOTES , can you help update with that? Thanks
@okozachenko1203 

/approve

okozachenko1203:
> sorry missed one comment in the original issue :( , thanks for the updated comments one comment is this change worthy a release NOTES , can you help update with that? Thanks @okozachenko1203
> 
> /approve

@jichenjc sorry, could you guide me how i can make release note? this is the first contribution to this project for me :)

mdbooth:
Unfortunately we don't do separate release notes in CAPO (yet).

@jichenjc I've been meaning to ask you how CPO release notes work, because I suspect we should be doing something similar. This PR will be mentioned in the generated release notes of the release it's included in, but just the title. Is there some GH action or prow module we should be using to do better?

PR #1668, opened 2023-09-13
:sparkles: Additional data volumes for machines

### LINKED ISSUE #1286: Ability to mount additional volumes
/kind feature

openstack template  only has rootvolume, but i want to my own volume

mkjpryor:

This PR adds the ability for a machine to have additional data volumes attached, e.g. for etcd.

Fixes #1286 


api/v1alpha6/conversion.go
```diff
@@ -60,14 +60,11 @@ func restorev1alpha7MachineSpec(previous *infrav1.OpenStackMachineSpec, dst *inf
...
+	dst.DataVolumes = previous.DataVolumes
 }
 
 func restorev1alpha7Bastion(previous **infrav1.Bastion, dst **infrav1.Bastion) {
-	// PropagateUplinkStatus has been added in v1alpha7.
-	// We restore the whole Ports since they are anyway immutable.
-	if *previous != nil && (*previous).Instance.Ports != nil && *dst != nil && (*dst).Instance.Ports != nil {
-		(*dst).Instance.Ports = (*previous).Instance.Ports
-	}
+	restorev1alpha7MachineSpec(&(*previous).Instance, &(*dst).Instance)
```
mdbooth, line 67: Doesn't this panic if (*previous) is null? There maybe some context-dependent reason why that never ends up happening, but to make me happy could we keep the previous null checks before dereferencing (*previous) and (*dst) to call restorev1alpha7MachineSpec?

api/v1alpha7/types.go
```diff
@@ -163,6 +163,15 @@ type RootVolume struct {
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+type DataVolume struct {
+	NameSuffix       string `json:"nameSuffix"`
+	Size             int    `json:"diskSize,omitempty"`
+	VolumeType       string `json:"volumeType,omitempty"`
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	//+optional
+	Tags []string `json:"tags,omitempty"`
```
mdbooth, line 172: These are just going to be the attachment tags, right?

api/v1alpha7/types.go
```diff
@@ -163,6 +163,15 @@ type RootVolume struct {
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+type DataVolume struct {
+	NameSuffix       string `json:"nameSuffix"`
+	Size             int    `json:"diskSize,omitempty"`
+	VolumeType       string `json:"volumeType,omitempty"`
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
```
mdbooth, line 170: @stephenfin Very much interested in your input here.

I've proposed adding `useMachineAZ` (default false) to a *future* api version. What do you think the semantics of this one should be?

Options:
* Empty AZ means no AZ
* Empty AZ means something machine related
* Anything else?

mdbooth (commented):
Looking good! Lets start the discussion.

api/v1alpha7/types.go
```diff
@@ -163,6 +163,15 @@ type RootVolume struct {
...
+	//+optional
```
mdbooth, line 171: I believe `+optional` is a no-op. @JoelSpeed can you confirm?

I think you can test this by removing it and re-running `make generate`. If it didn't make any change in the CRD it was a no-op.

api/v1alpha7/types.go
```diff
@@ -163,6 +163,15 @@ type RootVolume struct {
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+type DataVolume struct {
+	NameSuffix       string `json:"nameSuffix"`
```
mdbooth, line 167: `omitempty`? If not, it's probably also `+required`.

api/v1alpha7/types.go
```diff
@@ -163,6 +163,15 @@ type RootVolume struct {
...
+	Size             int    `json:"diskSize,omitempty"`
```
mdbooth, line 168: `diskSize` for consistency with `rootVolume`, right? Do we want to change it?

mdbooth (commented):
Please could you also add field comments of the form
```go
// FieldName does X. It defaults to Y. It is incompatible with Z.
```

api/v1alpha7/types.go
mkjpryor: I was planning to make the semantics the same as the root volume for now. I think that is less confusing.

Then when we change the root volume we can change this as well. My plan is to have one "volume creating" function so hopefully the logic doesn't have to be duplicated.

api/v1alpha7/types.go
mkjpryor: As it is now, it comes out in the `required` keys in the CRD schema. Not sure it can be `omitempty` - if you want a data volume you have to name it I think

api/v1alpha7/types.go
mkjpryor: Similar to AZ, I'd rather match rootVolume and change them both at once in a future patch.

api/v1alpha6/conversion.go
mkjpryor: 👍

api/v1alpha7/types.go
mdbooth: The problem is we know the semantics of the rootVolume are wrong, and we're going to have a minor upgrade problem there. If we know what we're going to do with rootVolume I'd prefer to get this right in the first instance.

See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1662 for WIP.

mkjpryor: What is left to do on the root volume WIP? Should we get that merged first?

api/v1alpha7/openstackmachine_types.go
```diff
@@ -84,6 +84,9 @@ type OpenStackMachineSpec struct {
 	// The volume metadata to boot from
 	RootVolume *RootVolume `json:"rootVolume,omitempty"`
 
+	// Data volumes to attach to the server instance
```
mdbooth, line 87: I know we don't even remotely follow this in the existing APIs, but for new APIs we should follow the convention that the first word of the field description is the exact name of the field.

So something like
```golang
// DataVolumes defines a list of volumes to attach to the server instance
```

api/v1alpha7/types.go
```diff
@@ -163,6 +163,14 @@ type RootVolume struct {
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+type DataVolume struct {
```
mdbooth, line 166: Already mentioned: docstrings for all these fields, please.

api/v1alpha7/types.go
JohnGarbutt: I guess the issue here is you don't want to fall foul of the cross_az_attach checks, for clouds where that matters. In theory the Cinder AZ is independent from the Nova AZ, but most people don't do that. Network AZ on the ports is different again, of course.

FWIW, I think the main three cases are:
- 80% case: don't specify AZ at all, for anything, nova, neutron and cinder
- Volume and server AZ matches, to pass the cross_az_attach check (mostly for when you have multiple fibrechannel storage appliances, or similar, I think)
- Allow people to specify a custom AZ for storage that can be unrelated to the Nova and network AZ (because people bought multiple storage appliances, and want to expose that to their users for some reason, shrug)

(I am curious if there are other cases people care about.)

api/v1alpha7/openstackmachine_types.go
```diff
@@ -84,6 +84,9 @@ type OpenStackMachineSpec struct {
 	// The volume metadata to boot from
 	RootVolume *RootVolume `json:"rootVolume,omitempty"`
 
+	// DataVolumes is a list of specifications for volumes to attach to the server instance
+	DataVolumes []DataVolume `json:"dataVolumes,omitempty"`
```
EmilienM, line 88: Not a big deal, but just to mention the CAPZ used `DataDisks`.
I wonder if we want to follow that pattern, since our data might end-up being on volumes and/or local disks at some point.

api/v1alpha7/types.go
```diff
@@ -163,6 +163,24 @@ type RootVolume struct {
...
+	// If omitted, the default type will be used.
+	VolumeType string `json:"volumeType,omitempty"`
+	// AvailabilityZone is the volume availability zone to create the volume in.
+	// If omitted, the availability zone of the server will be used.
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	// Tag is a tag to add to the volume attachment.
+	// This is then available in the instance metadata, allowing devices to be
+	// identified by tag inside the server.
+	Tag string `json:"tag,omitempty"`
+}
```
EmilienM, line 182: This can be addressed in a separate PR but for supporting local disks, we'll need to add more fields.

EmilienM:
:+1: excellent start to me!

Beside a few comments and missing doc (and tests?) this is good to me.

Note: I'm very interested by local disks instead volumes, but with your PR I can easily see it extended via a separate PR. However I have no objection (beside making the PR too big) if you include it in here.

api/v1alpha7/openstackmachine_types.go
mkjpryor: I used `DataVolumes` to match `RootVolume`. I don't really have a strong opinion.

api/v1alpha7/types.go
mkjpryor: Just one I think, to indicate whether it should be a local or Cinder disk? The rest can be reused in both cases.

mkjpryor:
> Note: I'm very interested by local disks instead volumes, but with your PR I can easily see it extended via a separate PR. However I have no objection (beside making the PR too big) if you include it in here.

I feel like it should be a separate PR, just in case there are unintended consequences that need addressing.

api/v1alpha7/types.go
mdbooth: *If* we accept my argument that there will be a single local disk then I would prefer it to have a separate field because it's more obvious to the user and doesn't require validation. I would only put a local flag here if we decide it's useful to have multiple local disks.

But as you say we don't need to consider this yet.

api/v1alpha7/types.go
mdbooth: @mkjpryor We need a firm decision on this, btw.

api/v1alpha7/openstackmachine_types.go
mkjpryor: The things you have to set in kubeadmcontrolplane and/or kubeadmbootstrapconfig correspond to cloud-init fields, so `DiskSetup` and `Mounts`.

mkjpryor: But the OpenStack terminology for these things is volume, so my feeling is that we stick with that 🤷‍♂️

api/v1alpha7/types.go
mkjpryor: I'd rather we did the same thing as the root volume in this PR, then change them all at once in another PR, personally, so that it stays consistent.

My preference would be that we add a flag that says `useServerAZ` and defaults to `false`, despite that being a breaking change, and keep the existing `availabilityZone` override. The logic would be:

1. If `availabilityZone` is set use it
2. If `useServerAZ = true`, set the AZ to the server AZ
3. Don't set an AZ at all

mkjpryor: I guess if we want a non-breaking change, then the `useServerAZ` flag defaults to `true`.

EmilienM: > I guess if we want a non-breaking change, then the useServerAZ flag defaults to true.

That's what I believe is already CAPO's behavior on rootVolumes:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance.go#L439-L442

api/v1alpha7/openstackmachine_types.go
EmilienM: Could we settle on the API, considering that we'll support multiple ephemeral disks and volumes.

_Option A_. `DataVolumes` and `LocalDataVolumes` where each of them have the options needed for their use-case.

_Option B_. One field (maybe name it `BlockDevices` to avoid confusion with volumes`) with all fields that we need for both use cases.


I have a preference for _Option B_. (and the rename) for simplicity's sake.

mkjpryor: My preference is also for option B. I like using a name that mirrors the OpenStack API attribute that we are ultimately setting - I think it is probably less confusing for users.

api/v1alpha7/types.go
mkjpryor: @mdbooth as I raised in Slack, I think I have a use case for multiple ephemeral devices carved out of the ephemeral volume, so I would like to see an option that permits that.

mkjpryor: But it is something for a future PR anyway

api/v1alpha7/types.go
mkjpryor: Yes - I mirrored the current root behaviour for the additional volumes on purpose.

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,143 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
+			}
+			blockDevices = append(blockDevices, bootfromvolume.BlockDevice{
+				SourceType:          bootfromvolume.SourceVolume,
+				DestinationType:     bootfromvolume.DestinationVolume,
+				UUID:                blockDevice.ID,
+				BootIndex:           -1,
+				DeleteOnTermination: true,
+				// TODO(mkjpryor) uncomment once gophercloud patch is merged
+				// https://github.com/gophercloud/gophercloud/pull/2777
+				// Tag:                 dataVolumeSpec.Tag,
```
EmilienM, line 509: Once https://github.com/gophercloud/gophercloud/pull/2789 is merged and a release was produced, I'll update vendoring here.

EmilienM:
TODO:
- [] Update Gophercloud vendoring once https://github.com/gophercloud/gophercloud/pull/2789 has merged and 1.7.0 was tagged.
- [] Uncomment `Tag`
- Documentation
- Squash commits with Co-Authoring.

api/v1alpha7/types.go
EmilienM: @mkjpryor Please look at my recent refactor & patches, I think the way we're proposing now is the same behaviour as with rootVolumes. And indeed this can be bulked changed later.

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,129 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
-	rootVolume := instanceSpec.RootVolume
-	if !hasRootVolume(rootVolume) {
-		return nil, nil
+func (s *Service) getOrCreateVolume(eventObject runtime.Object, opts volumes.CreateOpts) (*volumes.Volume, error) {
+	existingVolume, err := s.getVolumeByName(opts.Name)
+	if err != nil {
+		return nil, err
 	}
+	if existingVolume != nil {
+		if existingVolume.Size != opts.Size {
```
mandre, line 395: Don't we need to also check the availability zone and the type to ensure the existing volume conforms to the spec?
Alternatively, it might be simpler to error out when there is an existing volume with the name we want to use.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -569,33 +570,61 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
-				bootVolume, err := shared.GetOpenStackVolume(e2eCtx, volumes[0].ID)
+				firstVolume, err := shared.GetOpenStackVolume(e2eCtx, volumes[0].ID)
 				Expect(err).NotTo(HaveOccurred(), "failed to get OpenStack volume %s for machine %s", volumes[0].ID, machine.Name)
+				secondVolume, err := shared.GetOpenStackVolume(e2eCtx, volumes[1].ID)
+				Expect(err).NotTo(HaveOccurred(), "failed to get OpenStack volume %s for machine %s", volumes[0].ID, machine.Name)
+
+				bootVolume := firstVolume
+				additionalVolume := secondVolume
+				// The boot volume is the one with the "bootable" flag set.
+				if firstVolume.Bootable != "true" { // This is genuinely a string, not a bool
```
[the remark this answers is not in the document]
EmilienM: I need to do more testing to confirm that listing volumes provide an output ordered by creation time. I'll look at that today because I agree that this is ugly :smile: .

pkg/cloud/services/compute/instance_test.go
```diff
@@ -298,6 +299,15 @@ func TestService_ReconcileInstance(t *testing.T) {
 					"test-metadata": "test-value",
 				},
 				"user_data": &userData,
+				"block_device_mapping_v2": []map[string]interface{}{
+					{
+						"delete_on_termination": true,
+						"destination_type":      "local",
```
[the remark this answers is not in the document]
EmilienM: it's the default block device mapping when no root volume is being used.

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,129 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
+			return []bootfromvolume.BlockDevice{}, err
+		}
+		blockDevices = append(blockDevices, bootfromvolume.BlockDevice{
+			SourceType:          bootfromvolume.SourceVolume,
+			DestinationType:     bootfromvolume.DestinationVolume,
+			UUID:                rootVolume.ID,
+			BootIndex:           0,
+			DeleteOnTermination: true,
+		})
+	} else {
```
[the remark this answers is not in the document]
EmilienM: yes exactly. We need an ephemeral storage to be created using the imageID and be bootable (bootindex 0).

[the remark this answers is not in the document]
mkjpryor: Basically, yeah. OpenStack won't let you set non-bootable block devices without setting a bootable one

pkg/cloud/services/compute/instance.go
```diff
@@ -545,38 +585,24 @@ func (s *Service) GetManagementPort(openStackCluster *infrav1.OpenStackCluster,
...
-			We create a boot-from-volume instance in 2 steps:
-			1. Create the volume
-			2. Create the instance with the created root volume and set DeleteOnTermination
+			Attaching volumes to an instance is a two-step process:
 
-			This introduces a new failure mode which has implications for safely deleting instances: we
-			might create the volume, but the instance create fails. This would leave us with a dangling
-			volume with no instance.
+			  1. Create the volume
+			  2. Create the instance with the created volume in iAdditionalBlockDevices with DeleteOnTermination=true
```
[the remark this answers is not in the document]
EmilienM: Good point. I'm updating the doc.

test/e2e/suites/e2e/e2e_test.go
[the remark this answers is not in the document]
EmilienM: I tested it and indeed it seems that the volume used for boot is always first.

pkg/cloud/services/compute/instance.go
EmilienM: What if we do this:
* When we create a volume, we put the Kubernetes UUIDs of the OpenStackCluster and the OpenStackMachine in the volume metadata
  * openstack.infrastructure.cluster.x-k8s.io/cluster-uuid
  * openstack.infrastructure.cluster.x-k8s.io/machine-uuid
* When we search for a volume by name and find one, we then check if the UUIDs in the metadata match the cluster/machine being reconciled
* If they do, we own the volume and that is the volume we return
* If they don't, we don't own the volume and we error out

If that's a pattern people agree on, I'll be happy to do this via a separated PR if possible. For the sake of not adding too many design changes into one PR.

mkjpryor: @mdbooth and I also discussed doing this for Neutron resources using tags

mandre: That looks reasonable. I would just change the logic to search for volume by name and by metadata in a single query. If it returns a result, that's the one we own and we can use it, otherwise, we know we need to create a new one.

mkjpryor: Not sure the single query is possible unfortunately, although I might have missed something. I would be surprised if the metadata items were indexed for searching.

EmilienM: I think it's possible:

https://github.com/gophercloud/gophercloud/blob/1ef3b0843934066d5c65163e64683657608e6975/openstack/blockstorage/v3/volumes/requests.go#L125-L126

EmilienM:
/hold cancel

This is ready for review.
There are a list of things we want to do next, but we believe that this should be done via a separated PR:
* [ ] Address this [comment](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1668#discussion_r1337423351) to better filter volumes by metadata
* [ ] Add ephemeral storage support to the `AdditionalBlockDevices` API; with doc/tests.

pkg/cloud/services/compute/instance.go
EmilienM: We came to the agreement that we (I) will take care of it in a separated PR.

mdbooth (commented):
This is really close, thanks!

Can you please also consider how soon we can add the volume metadata support? Ideally this will be within the next couple of months to avoid an upgrade issue.

EmilienM:
> This is really close, thanks!
> 
> Can you please also consider how soon we can add the volume metadata support? Ideally this will be within the next couple of months to avoid an upgrade issue.

I'm working on it now, I think I can be done before Tuesday next week.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -569,33 +570,51 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
-				// Check that all machines have the expected boot volume
+				// Check that all machines have the expected volumes:
+				// - 1 root volume
+				// - 1 additional volume
 				volumes := server.AttachedVolumes
-				Expect(volumes).To(HaveLen(1))
+				Expect(volumes).To(HaveLen(2))
 
-				bootVolume, err := shared.GetOpenStackVolume(e2eCtx, volumes[0].ID)
+				rootVolume, err := shared.GetOpenStackVolume(e2eCtx, volumes[0].ID)
```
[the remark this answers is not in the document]
EmilienM: it's apparently stable. The volume attachments listed for a server seem to be indexed by their creation, and the rootVolume is always created first. I was surprised myself too but I've tested it a few times and it was always the case :shrug: .

I'll try to find the code in Nova to confirm what I saw.

pkg/cloud/services/compute/instance.go
EmilienM: Like I said on another comment, I'm working on it and this will be done before the next release is cut.

api/v1alpha7/types.go
EmilienM: We came to an agreement to bulk change it via another PR.

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,131 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
+		})
 	}
 
-	block := bootfromvolume.BlockDevice{
-		SourceType:          bootfromvolume.SourceVolume,
-		BootIndex:           0,
-		UUID:                volume.ID,
-		DeleteOnTermination: true,
-		DestinationType:     bootfromvolume.DestinationVolume,
+	if instanceSpec.AdditionalBlockDevices != nil && len(instanceSpec.AdditionalBlockDevices) > 0 {
```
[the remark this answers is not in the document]
EmilienM: done

pkg/cloud/services/compute/instance.go
```diff
@@ -625,6 +653,36 @@ func (s *Service) DeleteInstance(openStackCluster *infrav1.OpenStackCluster, eve
 	return s.deleteInstance(eventObject, instanceStatus.InstanceIdentifier())
 }
 
+func (s *Service) deleteVolumes(instanceSpec *InstanceSpec) error {
+	if hasRootVolume(instanceSpec) {
+		if err := s.deleteVolume(instanceSpec.Name, "root"); err != nil {
+			return err
+		}
+	}
+	if instanceSpec.AdditionalBlockDevices != nil {
```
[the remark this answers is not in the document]
EmilienM: done

test/e2e/suites/e2e/e2e_test.go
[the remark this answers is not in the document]
stephenfin: The API output is generated [here](https://github.com/openstack/nova/blob/27.1.0/nova/api/openstack/compute/views/servers.py#L730-L749). The `bdms` parameter is populated [here](https://github.com/openstack/nova/blob/27.1.0/nova/api/openstack/compute/views/servers.py#L365-L370). I followed that call to `nova.objects.BlockDeviceMappingList.bdms_by_instance_uuid` all the way down to the DB layer and didn't spot any sorting being done. This leads me to believe we're probably relying on the order of the rows being returned by the DB. As you likely know, MySQL doesn't guarantee sorting without a specific `ORDER BY` so that doesn't seem wise since it won't be consistent.

You should double check this (!) but assuming this is correct, you probably want to explore device tagging. In general, nova doesn't guarantee anything wrt device ordering and asks that you use [device tagging](https://specs.openstack.org/openstack/nova-specs/specs/mitaka/approved/virt-device-role-tagging.html) if you need this. I don't know if gophercloud support this but you simply want to add the `tags` field to the BDM for the root volume, e.g.

```json
{
  "block_device_mapping_v2": [
    {
      "uuid": "<image_id>",
      "boot_index": 0,
      "source_type": "image",
      "destination_type": "volume",
      "volume_size": "<volume_size>",
      "tags": ["foo", "bar"]
    }
  ]
}
```

(note: this requires API microversion 2.42).

You can then use the volume attachments API (`GET /servers/{serverID}/os-volume-attachments`) to list the attached volumes with tags, since this info is not returned in the `os-extended-volumes:volumes_attached` field in the `GET /servers/{serverID}` response.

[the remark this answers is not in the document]
EmilienM: right, that was my thought as well. I'll investigate it.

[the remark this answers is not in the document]
EmilienM: For now, I've addressed this comment with:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6855f55964b2a57eddddafc74af6502fb389c5ee/test/e2e/suites/e2e/e2e_test.go#L579-L592

Until we find a better solution. With device tagging that I plan to work on next, we'll probably make this better.

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,129 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
 	return &volumeList[0], nil
 }
 
-func (s *Service) getOrCreateRootVolume(eventObject runtime.Object, instanceSpec *InstanceSpec, imageID string) (*volumes.Volume, error) {
-	rootVolume := instanceSpec.RootVolume
-	if !hasRootVolume(rootVolume) {
-		return nil, nil
+func (s *Service) getOrCreateVolume(eventObject runtime.Object, opts volumes.CreateOpts) (*volumes.Volume, error) {
```
lentzi90, line 389: Please add a doc string for this function

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,129 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
+			return true, nil
+		case "error":
+			return false, fmt.Errorf("volume %s is in error state", volumeID)
+		default:
+			return false, nil
+		}
+	})
+}
 
+func (s *Service) getOrCreateVolumeBuilder(eventObject runtime.Object, instanceSpec *InstanceSpec, blockDevice infrav1.AdditionalBlockDevice, imageID string, description string) (*volumes.Volume, error) {
```
lentzi90, line 435: Please add a doc string for this function

pkg/cloud/services/compute/instance.go
```diff
@@ -407,68 +386,129 @@ func (s *Service) getVolumeByName(name string) (*volumes.Volume, error) {
...
-	return volume, err
+
+	return s.getOrCreateVolume(eventObject, createOpts)
 }
 
-// applyRootVolume sets a root volume if the root volume Size is not 0.
-func applyRootVolume(opts servers.CreateOptsBuilder, volume *volumes.Volume) servers.CreateOptsBuilder {
-	if volume == nil {
-		return opts
+func (s *Service) getBlockDevices(eventObject runtime.Object, instanceSpec *InstanceSpec, imageID string, timeout time.Duration, retryInterval time.Duration) ([]bootfromvolume.BlockDevice, error) {
```
lentzi90, line 454: Please add a doc string for this function

lentzi90 (commented):
Good job with this!
I agree with the decision to implement the AZ in the same way as for the root volume. It will be easier to understand it this way and then break both at the same time.

Could you please add some doc strings for at least the more complicated functions? I commented below on those I think are most important.

pkg/cloud/services/compute/instance.go
EmilienM: done

EmilienM:
@lentzi90 comments addressed. Thanks for the review!

pkg/cloud/services/compute/instance.go
lentzi90: @EmilienM what about this one?

api/v1alpha7/types.go
```diff
@@ -163,6 +163,24 @@ type RootVolume struct {
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+type AdditionalBlockDevice struct {
+	// NameSuffix is the suffix to be added to volume names.
+	// It will be combined with the machine name to make the volume name.
+	NameSuffix string `json:"nameSuffix"`
```
mdbooth, line 169: Could we call this `name` instead? The docstring can explain that the volume name will be the machine name with this appended.

api/v1alpha7/types.go
```diff
@@ -163,6 +163,24 @@ type RootVolume struct {
...
+	// VolumeType is the volume type of the volume.
+	// If omitted, the default type will be used.
+	VolumeType string `json:"volumeType,omitempty"`
+	// AvailabilityZone is the volume availability zone to create the volume in.
+	// If omitted, the availability zone of the server will be used.
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	// Tag is a tag to add to the volume attachment.
+	// This is then available in the instance metadata, allowing devices to be
+	// identified by tag inside the server.
+	Tag string `json:"tag,omitempty"`
```
mdbooth, line 181: Is there any reason to have a separate tag element instead of just using the name?

api/v1alpha7/openstackmachine_types.go
```diff
@@ -84,6 +84,9 @@ type OpenStackMachineSpec struct {
 	// The volume metadata to boot from
 	RootVolume *RootVolume `json:"rootVolume,omitempty"`
 
+	// AdditionalBlockDevices is a list of specifications for additional block devices to attach to the server instance
+	AdditionalBlockDevices []AdditionalBlockDevice `json:"additionalBlockDevices,omitempty"`
```
mdbooth, line 88: Please could you add
```go
// +listType=map
// +listMapKey=name
```

That last assuming you change `nameSuffix` to `name` (see below). Don't forget to re-run `make generate`, as this affects the CRDs.

This allows tools (e.g. kustomize/PATCH) to apply strategic merge patches to this list by name instead of having to replace the whole list.

mdbooth (changes requested):
Please could you respond to the 3 API requests:
* Rename `nameSuffix` to `name`
* Remove tag in favour of just using `name`
* Addition of strategic merge patch annotation to `additionalBlockDevices`

There's also an outstanding request from @lentzi90 for a missing docstring.

Apart from that, this looks good and has good test coverage. I don't think there's nothing complicated in the above requests.

api/v1alpha7/types.go
EmilienM: I don't understand why you want to name it `Name`. This is not a name, this is a name suffix.
This would be a lie to the API user that their block device would be named like this, because this wouldn't be the case.

pkg/cloud/services/compute/instance.go
EmilienM: done

api/v1alpha7/types.go
mdbooth: See below. It would be the name of the volume *in the context of this machine*. It would also be the name of the volume in the `additionalBlockDevices` list.

mdbooth:
To explain the above API requests, consider the case where a user wants to specify an additional volume to hold etcd.

Before:
```yaml
additionalBlockDevices:
- nameSuffix: etcd
  size: 10
  tag: etcd
```

After:
```yaml
additionalBlockDevices:
- name: etcd
  size: 10
```

I prefer the latter.

api/v1alpha7/types.go
EmilienM: fair enough, I'll do it.

EmilienM:
@mdbooth @lentzi90 I've applied all of your suggestions, thanks for the feedback. I think this time it's ready to go!

mdbooth (approved):

lentzi90 (commented):
/lgtm

PR #1669, opened 2023-09-13
🌱Fix various issues with CRD generation

mdbooth:
Kustomize was emitting warnings due to use of deprecated methods. The first commit addresses all of these except use of `vars`.

We were also hitting https://github.com/kubernetes-sigs/kustomize/issues/5031. Although this is still broken in kustomize, there is a workaround in https://github.com/kubernetes-sigs/controller-tools/pull/800, so we bump controller-tools to get the fix.

/hold


config/default/kustomization.yaml
```diff
@@ -7,22 +7,31 @@ commonLabels:
...
+    version: v1
+    kind: MutatingWebhookConfiguration
+    name: mutating-webhook-configuration
+  path: cainjection_patch.yaml
+- target:
+    group: admissionregistration.k8s.io
+    version: v1
+    kind: ValidatingWebhookConfiguration
+    name: validating-webhook-configuration
+  path: cainjection_patch.yaml
```
lentzi90, line 34: Nice solution!

lentzi90 (commented):
/lgtm

PR #1678, opened 2023-09-24
:book: Clarify release notes content for releases and pre-releases

wwentland:

This clarifies the content of release notes for releases and pre-releases.

The change is motivated by the understanding that a significant number of users will only ever upgrade from one stable release to the next, without necessarily being aware of the changes in the interim pre-releases.

These users would have previously had to review not only the release notes of the release they are upgrading to, but also those of all interim pre-releases. A process that is at best arduous, and at worst would have led to them missing important information.

Another nice side-effect of this change is that community contributions are highlighted in a prominent place even if they had been included in a pre-release already.

See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/discussions/1676 for additional information.

/hold

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -27,6 +27,15 @@ to version 1.2.3 would be called `v1.2.3-rc.0`.
 
 It is recommended to create at least one release candidate when bumping `X` or `Y`.
 
+## Release notes
+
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
/ok-to-test
/approve
Thank you for this! :pray:

PR #1690, opened 2023-09-27
🌱 gitignore: ignore vendor/ directory

EmilienM:
CAPO doesn't use vendoring, so let's ignore that directory from git.
That will prevent someone from committing it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,6 @@ docs/book/book/
 
 # Development container files (https://containers.dev/)
 .devcontainer
+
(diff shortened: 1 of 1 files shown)


mdbooth:
/approve

Please can you update the PR title with the prefix :seedling: before releasing the hold? It's used by the release notes generation.

/hold

dulek:
Just out of curiosity…

dulek:
Okay, so it looks good now!

/hold cancel

PR #1691, opened 2023-09-28
✨ Add InternalDNS to Machine Addresses

### LINKED ISSUE #1689: CAPO should set InternalDNS in status.Addresses
/kind feature

CAPO currently writes `InternalIP` and `ExternalIP` records to `OpenStackMachine.Status.Addresses`. However, it does not write `InternalDNS`. It should write an `InternalDNS` record containing the hostname of the machine.

This is useful information for other controllers consuming CAPI. Specifically, it is required by OpenShift's machine approver to automatically approve CSRs generated by a node for the target machine.

InternalDNS should match the hostname as reported to the server by the metadata service. This is available in the `OS-EXT-SRV-ATTR:hostname` server attribute, but unfortunately this was admin-only until microversion 2.90. However, the way nova generates the hostname is relatively simple: it mangles the server name with `sanitize_hostname`, defined here: https://github.com/openstack/nova/blob/87d4807848bb9546c1fca972da2eb2eda13eb08d/nova/utils.py#L356-L389. It is probably a more pragmatic approach to re-implement this utility function in CAPO than bump to such a relatively recent microversion.

cc @stephenfin

mdbooth:
Fixes #1689

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -22,6 +22,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -348,6 +350,19 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 	}
 
 	addresses := instanceNS.Addresses()
+
+	// Append the server's hostname to the list of addresses
+	// This must be the same value returned to the server by the metadata
+	// service. From Nova microversion 2.90 onwards this is available in the
+	// `OS-EXT-SRV-ATTR:hostname` attribute. However, the value is generated
+	// from the server name by a relatively simple mangling funcion. To
+	// avoid the microversion bump we instead reproduce the mangling
+	// function in sanitizeHostname() below.
+	addresses = append(addresses, corev1.NodeAddress{
+		Type:    corev1.NodeInternalDNS,
+		Address: sanitizeHostname(instanceStatus.Name()),
+	})
+
 	openStackMachine.Status.Addresses = addresses
 
 	switch instanceStatus.State() {
@@ -504,6 +519,36 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
 	instanceSpec.Ports = openStackMachine.Spec.Ports
 
 	return &instanceSpec
+}
+
+var (
+	spaceDotUnderscore = regexp.MustCompile(`[ _\.]`)
+	notWordDotDash     = regexp.MustCompile(`[^\w.-]+`)
+)
+
+// sanitizeHostname mangles a server's name into a hostname. It is required to be functionally identical to the version in nova:
+//
+//	https://github.com/openstack/nova/blob/87d4807848bb9546c1fca972da2eb2eda13eb08d/nova/utils.py#L356-L389
+//
+// Do not change the behaviour of this function except to match the behaviour in nova.
+func sanitizeHostname(name string) string {
+	// Remove all non-latin-1 characters
+	hostname := strings.Map(func(r rune) rune {
+		if r > 0xFF {
+			return -1
+		}
+		return r
+	}, name)
+
+	if len(hostname) > 63 {
+		hostname = hostname[:63]
+	}
+	hostname = spaceDotUnderscore.ReplaceAllString(hostname, "-")
+	hostname = notWordDotDash.ReplaceAllString(hostname, "")
+	hostname = strings.ToLower(hostname)
+	hostname = strings.Trim(hostname, ".-")
+
+	return hostname
 }
 
 func (r *OpenStackMachineReconciler) reconcileLoadBalancerMember(scope scope.Scope, openStackCluster *infrav1.OpenStackCluster, openStackMachine *infrav1.OpenStackMachine, instanceNS *compute.InstanceNetworkStatus, clusterName string) error {
--- a/controllers/openstackmachine_controller_test.go
+++ b/controllers/openstackmachine_controller_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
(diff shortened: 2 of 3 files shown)


mdbooth:
Tests look good. Artifacts look good. I'm going to test this against an OpenShift cluster in the morning.

@stephenfin  would you be able to take a look at the sanitizeHostname function and its tests to validate that it's doing what Nova is doing?

mdbooth:
Confirmed through manual testing that this PR allows OpenShift to approve CSRs issued by CAPO-provisioned workers.

controllers/openstackmachine_controller.go
```diff
@@ -506,6 +521,36 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
...
+var (
+	spaceDotUnderscore = regexp.MustCompile(`[ _\.]`)
+	notWordDotDash     = regexp.MustCompile(`[^\w.-]+`)
+)
+
+// sanitizeHostname mangles a server's name into a hostname. It is required to be functionally identical to the version in nova:
+//
+//	https://github.com/openstack/nova/blob/87d4807848bb9546c1fca972da2eb2eda13eb08d/nova/utils.py#L356-L389
+//
+// Do not change the behaviour of this function except to match the behaviour in nova.
```
lentzi90, line 533: Is there any risk that we end up in a situation where nova has different behaviors for different versions?

mdbooth: I would hope not. If it did that would be a potential upgrade issue for Nova itself: afaict it doesn't store the mangled value but calculates it when it's used, so changing the function would change the hostname for a running server, which probably isn't a good idea.

Perhaps @stephenfin has more insight.

lentzi90 (commented):
/lgtm
Leaving the hold in case there was more things to double check

mdbooth:
> /lgtm Leaving the hold in case there was more things to double check

Thanks. I'm going to wait for @stephenfin to weigh in first just in case as he knows this code.

controllers/openstackmachine_controller.go
stephenfin: Unfortunately you already do have different behaviours. The code we're reimplemented here was changed relatively recently ([in Wallaby](https://github.com/openstack/nova/commit/9046f0fff4be424eda25401a3f9b8752964de775), to be precise). Prior to this, we did not replace the periods as part of the sanitisation. As such, you are going to see slightly different behaviour if this is deployed on an older cloud in situations where a user requests an instance name with periods in it e.g. `my-instance.example.com`.

Adding to the confusion, RHOSP includes a downstream-only workaround config option to opt-out of the new behaviour and disable the additional sanitization. IIRC this requires a support exception to use but it does exist. When this workaround config option is enabled, the cloud will behave like an older cloud. I don't recall what version of RHOSP this is present in.

Finally, Nova has the ability to request hostnames via the API [starting in microversion 2.90 (Xena)](https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#maximum-in-xena-and-yoga). We presumably don't use this in CAPO right now, but if it ever is used then we should be careful to start using the `OS-EXT-SRV-ATTR:hostname` field returned in the API instead of this.

mdbooth: Eurgh 😢 

I'm going to try to process this to see what it would mean for us.

lentzi90: Ouch! So there would potentially be issues for consumers of this in cases where CAPO and Nova does not agree on the hostname then. :slightly_frowning_face:

mdbooth:
In order for a CSR to be approved there must be a matching InternalDNS entry on the Machine for the nodename requested in the CSR generated by kubelet.

We had an upgrade problem in OpenShift already moving from in-tree to external CCM. The in-tree provider always used server name for the Node name taken directly from Nova, but when we switched to external CCM kubelet used hostname. We discovered that this hostname could be an FQDN in some circumstances (what circumstances?), which meant that on upgrade kubelet tried to create a new Node for the FQDN rather than using the existing one. We worked round this by passing an explicit hostname to kubelet which we fetch on-node from the metadata service:

https://github.com/openshift/machine-config-operator/blob/master/templates/common/openstack/files/usr-local-bin-openstack-kubelet-nodename.yaml

We're also careful to persist the value to disk and not re-generate it if it exist already, so even if OpenStack did change what it was reporting, kubelet would continue to use the same value.

Where can kubelet get nodename from:
* hostname, mangled without dot removal, FQDN (pre-Wallaby)
* hostname, mangled with dot removal, FQDN (Wallaby+)
* server name, not mangled (legacy provider compatibility)
* Theoretically, any arbitrary string

Additionally, hostname will change and therefore break kubelet and probably other things if OpenStack is upgraded across Wallaby and the server name contains dots.

Lets ignore the arbitrary option for now.

It looks like the FQDN part comes from static configuration in Nova: https://github.com/openstack/nova/blob/87d4807848bb9546c1fca972da2eb2eda13eb08d/nova/api/metadata/base.py#L528-L534. I don't think we have any way to know what that is before it's exposed in 2.90.

Also worth noting that from 2.94 it's possible to specify servers with FQDNs as hostnames: https://github.com/openstack/nova/blob/master/releasenotes/notes/microversion-2-94-59649401d5763286.yaml

We need an InternalDNS entry matching whatever is in use. We know what hostname is if we have microversion 2.90, so we can use that. I think we *don't* know what hostname is if we have microversion <2.90 because we don't know `dhcp_domain` is. I propose that for now we just don't emit InternalDNS if we can't get it.

We also need to support the case where, for legacy compatibility, we're using the unmangled server name. I think we have 2 options here:
1. A field in the spec, `hostnameUsesServerName`. If this is set on the machine then we emit server name instead of hostname.
1. We unconditionally emit both. I suspect this would work in practise, but is it nasty? @JoelSpeed

lentzi90:
> I propose that for now we just don't emit InternalDNS if we can't get it.

This sounds sensible to me! :+1:

PR #1695, opened 2023-09-29
WIP - Volumes metadata

### LINKED ISSUE #1693: Identify Volumes (rootVolumes and additional block devices) with Metadata
/kind feature

* When we create a volume, we put the Kubernetes UUIDs of the OpenStackCluster and the OpenStackMachine in the volume metadata
  * openstack.infrastructure.cluster.x-k8s.io/cluster-uuid
  * openstack.infrastructure.cluster.x-k8s.io/machine-uuid
* When we search for a volume by name and find one, we then check if the UUIDs in the metadata match the cluster/machine being reconciled
* If they do, we own the volume and that is the volume we return
* If they don't, we don't own the volume and we error out

Context: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1668#discussion_r1337423351

EmilienM:
- Additional data volumes for machines
- WIP - volumes metadata

<TO DO>

Fixes #1693

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


pkg/cloud/services/compute/instance.go
```diff
@@ -377,15 +356,29 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, openStackCluste
 	}
 
 	record.Eventf(eventObject, "SuccessfulCreateServer", "Created server %s with id %s", createdInstance.Name(), createdInstance.ID())
+
+	// iterate over blockDevices and update the metadata
+	opts := volumes.UpdateOpts{
+		Metadata: map[string]string{
+			// add openstack.infrastructure.cluster.x-k8s.io/cluster-uuid
+			"openstack.infrastructure.cluster.x-k8s.io/machine-uuid": createdInstance.ID(),
```
EmilienM, line 364: this is not the instance ID that we want but the OpenStackMachine ID, a mistake. I'll fix this.

PR #1696, opened 2023-09-29
✨ Add ephemeral storage support to the AdditionalBlockDevices

### LINKED ISSUE #1692: Add ephemeral storage support to the `AdditionalBlockDevices`
/kind feature

This is a follow-up of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1286

Now that we support additional block devices as volumes, we want to extend the API to support ephemeral storage.
We wanted to initially support additional volumes and separate the PRs.

We will extend what was done for AdditionalBlockDevices, in term of API, testing, doc, etc.
Also, we want to document how to actually use the volumes by tag using diskSetup and mounts - see https://github.com/stackhpc/azimuth-images/pull/64/files for examples.

EmilienM:
Add ephemeral storage support to the `AdditionalBlockDevices`

Fixes #1692

1. We only support unformated disks for now (we don't expose `guestFormat`) so the user can handle the filesystem themselves. Note that we initially added it and tested swap but it didn't work in CAPO CI. The field can be added later.
2. While we support multiple ephemeral disks in this PR, we documented that this is to use at your own risks, this hasn't been well tested in Nova.
3. API union discriminators aren't used but are here just in case at some point kube api will support it.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests


api/v1alpha7/types.go
```diff
@@ -163,6 +163,39 @@ type RootVolume struct {
...
 }
 
+type AdditionalBlockDevice struct {
+	// BlockDeviceType is the type of block device to create.
+	// This can be either "volume" or "local".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=volume;local
+	BlockDeviceType string `json:"blockDeviceType"`
+
+	// Here I'm hesitant to add GuestFormat. Not to let users choose the format of the block device
```
EmilienM, line 173: question for reviewers.

Same question for `BlockDeviceType`, do we want to support things like Snapshots, Images, etc?
In other words, do we want to support any block device type that nova/cinder can do?



Task | openstack server create parameter(s) | Information
-- | -- | --
Boot an instance from an image and attach a non-bootable volume. | --block-device | Boot_instance_from_image_and_attach_non-bootable_volume
Create a volume from an image and boot an instance from that volume. | --boot-from-volume and --image; --block-device | Create_volume_from_image_and_boot_instance
Boot from an existing source image, volume, or snapshot. | --volume or --snapshot; --block-device | Create_volume_from_image_and_boot_instance
Attach a swap disk to an instance. | --swap | Attach_swap_or_ephemeral_disk_to_an_instance
Attach an ephemeral disk to an instance. | --ephemeral | Attach_swap_or_ephemeral_disk_to_an_instance


Source: https://docs.openstack.org/nova/latest/user/launch-instance-from-volume.html

EmilienM: cc @mdbooth @mkjpryor

EmilienM:
TODO:

- [ ] Figure out API
- [ ] Add doc string for `getBlockDevices`.

mandre:
I'll help on this PR as we'd like to get that merge relatively soon. For now, this is just a simple rebase to resolve the conflicts, it obviously needs more work, and I suspect I broke the tests in the process :sweat_smile:

api/v1alpha7/types.go
mandre: It depends how far we want to go with the `BlockDeviceType` abstraction vs exposing `source_type`, `destination_type`, and `guest_format` directly.

Looking at [what is currently possible](https://docs.openstack.org/nova/2023.2/user/block-device-mapping.html#valid-source-destination-combinations), we could have the following device types:
- `volume`:
  - source: volume
  - destination: volume
  - uuid: required
- `snapshot`:
  - source_type: snapshot
  - destination_type: volume
  - uuid: required
- `image`:
  - source_type: image
  - destination_type: volume
  - uuid: required
- `local`:
  - source_type: blank
  - destination_type: local
  - guest_format: whatever the user wants
  - size: required
- `swap`:
  - source_type: blank
  - destination_type: local
  - guest_format: swap
  - size: required

Swap is just a local block device, with `guest_format=swap`, so we may not want to create an additional type.

mandre: And of course, adding support for the `snapshot` and `image` types belongs to separate PRs, if that is something we want.

mkjpryor: I asked @mdbooth if we could just replace the root volume stuff with `source_type: image` and he seemed quite against it. I think I understand why, but I feel that we probably won't support the image source type for additional devices (i.e. non-boot)? Not sure what the use case for that would be.

mkjpryor: `swap` might be useful to support in the future, now that Kubernetes has a bit more support for running with swap on.

mkjpryor: It's still not encouraged though, so I'm not sure we should support it... 🤷‍♂️

EmilienM: > replace the root volume stuff

Do you mean, removing `rootVolume` and make `additionalBlockDevice` supporting `source_type: image` like you said?
I wouldn't be against but this seems like yet another API change (although conversion would be possible).

Also in that case we would maybe remove the "Additional" from the field name, and have `BlockDevices` ?

> swap

I agree it's useful for users who know what they're doing. I wonder if we can support that later though or if it needs to be part of that API for now.

mdbooth: I don't want to proxy the OpenStack API. Every time we've done that so far it has created an almighty mess. We need to understand our own use cases and present them to the user in a way which makes sense in our context. Specifically I would not be in favour of directly exposing `source_type`, `destination_type`, and `guest_format`.

I agree that we don't have a current use case for swap or snapshot, and I don't see any reason to support them until we do.

My primary concern here is actually the supportability of the 'multiple ephemeral disks' feature in Nova, which this API implies. When I maintained this code it was my impression that it was not well tested or widely used. Most users have a single ephemeral disk. Being possibly the only user of this feature makes me nervous.

I could be persuaded that it's ok if we test it ourselves on both current OpenStack and the oldest OpenStack we want to support. If it currently works in our window, I think we can trust OpenStack to fix regressions if they arise.

api/v1alpha7/types.go
```diff
@@ -164,10 +164,25 @@ type RootVolume struct {
 }
 
 type AdditionalBlockDevice struct {
-	// Name of the Cinder volume in the context of a machine.
-	// It will be combined with the machine name to make the actual volume name.
+	// BlockDeviceType is the type of block device to create.
+	// This can be either "volume" or "local".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=volume;local
+	BlockDeviceType string `json:"blockDeviceType"`
```
mdbooth, line 171: What we've got here is a discriminated union.

BlockDeviceType = `volume`:
* `VolumeType`
* `AvailabilityZone`

BlockDeviceType = `local`:
* `GuestFormat`

Please can you find a few examples of prior art in the k8s api so we can consider the pros and cons of existing designs?

mdbooth: This looks like it might be interesting reading: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md

api/v1alpha7/types.go
EmilienM: > we don't have a current use case for swap or snapshot, and I don't see any reason to support them until we do.

:+1:

> the supportability of the 'multiple ephemeral disks' feature in Nova

I think this was asked by @mkjpryor who has a valid use case.
I would say that we add it and maybe document that this is tested as best effort in our CI (for the `etcd` use case ultimately) but that there are some risks due to OpenStack upstream testing coverage.

pkg/cloud/services/compute/instance_test.go
```diff
@@ -765,8 +787,79 @@ func TestService_ReconcileInstance(t *testing.T) {
...
+					},
+				}
+				expectCreateServer(r.compute, createMap, false)
+				expectServerPollSuccess(r.compute)
+
+				// Don't delete ports because the server is created: DeleteInstance will do it
+			},
+			wantErr: false,
+		},
+		// This tests is currently passing but it shouldn't, we need to find out why.
```
EmilienM, line 810: A mystery to solve.

api/v1alpha7/types.go
```diff
@@ -163,18 +174,42 @@ type RootVolume struct {
...
 
+// AdditionalBlockDevice is a block device to attach to the server.
 type AdditionalBlockDevice struct {
-	// Name of the Cinder volume in the context of a machine.
-	// It will be combined with the machine name to make the actual volume name.
+	// Type is the type of block device to create.
+	// This can be either "volume" or "local".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=volume;local
+	// +unionDiscriminator
```
EmilienM, line 183: @mdbooth please review the API now that we have unions.

EmilienM:
e2e fail here:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1696/pull-cluster-api-provider-openstack-e2e-test/1712049907029774336/artifacts/clusters/bootstrap/resources/e2e-nlyee7/OpenStackMachine/cluster-e2e-nlyee7-control-plane-b2ffp.yaml

api/v1alpha7/types.go
EmilienM: re: Union, I checked CAPZ (quite a reference to me) and they have one `unionDiscriminator` in their API and the `enum` is defined at the tpe level so for us that would be in `BlockDeviceType`. In OpenShift API we have set the enum differently to allow more granular values but I think in CAPO I'll follow what CAPZ did because it looks cleaner here.

Other than that I think the API is "final" for this PR and ready to be reviewed.

api/v1alpha7/types.go
EmilienM: Just had a chat with @mdbooth about `swap`, he seems against it.
If we want users to add swap, we'll have to add a check for `swap` string in `guestFormat`. Is that what we want?

cc @mandre

EmilienM: A chat with @mdbooth today:
* We don't want to prevent users to deploy `swap`, but we don't want to add explicit support for it.
* Nova formatted ephemeral disks have always been problematic so we want to warn our users that using `guestFormat` is at their own risk for both swap or any filesystem.

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -218,17 +218,40 @@ profile:
...
+Here is an example on how to attach two ephemeral disks to the instance, where the second disk will be formated to `ext4`:
+
+```yaml
+- name: disk1
+  type: local
+  diskSize: 1
+- name: disk2
+  type: local
+  diskSize: 1
+  guestFormat: ext4
```
EmilienM, line 242: We probably want to document that `guestFormat` should be used at user's own risks. Nova formatted ephemeral disks have always been problematic.

EmilienM:
Somehow swap doesn't work in e2e. The instance must have failed to be created, I'll have to investigate unless we are fine not supporting swap for now.

mdbooth:
I just discussed this briefly with @EmilienM. Because we (Red Hat) don't yet use CAPO directly (it's coming!) we will have to merge this separately into the OpenShift API repo before we can use it. An advantage and a disadvantage of that is that we tend to get a much higher level of API scrutiny there than we get here.

In this case I'm proposing that we lean on it to effectively get API review on this PR from the OpenShift API reviewers. @EmilienM is going to submit an OpenShift PR to add this support to OpenShift. He's going to propose the exact same API we have here, and we'll see how it fares. We can then take that into account when deciding what to do here.

/hold
/cc @mandre

EmilienM:
> I just discussed this briefly with @EmilienM. Because we (Red Hat) don't yet use CAPO directly (it's coming!) we will have to merge this separately into the OpenShift API repo before we can use it. An advantage and a disadvantage of that is that we tend to get a much higher level of API scrutiny there than we get here.
> 
> In this case I'm proposing that we lean on it to effectively get API review on this PR from the OpenShift API reviewers. @EmilienM is going to submit an OpenShift PR to add this support to OpenShift. He's going to propose the exact same API we have here, and we'll see how it fares. We can then take that into account when deciding what to do here.

for reference: https://github.com/openshift/api/pull/1623

lentzi90 (commented):
Looks good so far! Let's see what the OpenShift folks have to say about it :slightly_smiling_face:

docs/book/src/topics/crd-changes/v1alpha6-to-v1alpha7.md
```diff
@@ -218,17 +218,29 @@ profile:
...
-  size: 50
+  type: volume
+  diskSize: 50
   volumeType: my-volume-type
   availabilityZone: az0
 ```
 
+Here is an example on how to attach a ephemeral disk to the instance:
+
+```yaml
```
EmilienM, line 235: I missed `additionalBlockDevices`.

EmilienM:

sounds like a flake with a LB stuck in PENDING

EmilienM:
@lentzi90 @mdbooth any ideas why e2e fails?
I haven't seen anything in logs. This happens since my new API proposal. The logic should remain the same. I wonder if you know where to look. Thanks

lentzi90:
This is just a wild guess: Could it be that you run out of quota? I think the default quota for volumes (and many other things) is 10. With this PR you add more volumes to the multi-az test case which could mean that we just hit that limit.

EmilienM:
> This is just a wild guess: Could it be that you run out of quota? I think the default quota for volumes (and many other things) is 10. With this PR you add more volumes to the multi-az test case which could mean that we just hit that limit.

Good idea. I'll try and also check in the OpenStack logs if we see a quota error. What's weird is that it worked last week with the same numbers.

lentzi90:
> Good idea. I'll try and also check in the OpenStack logs if we see a quota error. What's weird is that it worked last week with the same numbers.

It could be flaky since we run 2 tests in parallel. If two tests with volumes happen to run in parallel it could trigger and otherwise not. Or it could be that my guess is completely wrong and we should be looking elsewhere :grimacing:

EmilienM:

The only OpenStack errors we can see in the logs are Neutron/OVS related.

EmilienM:
```
I1023 13:50:57.848720       1 controller.go:114] "Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference" controller="openstackmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackMachine" OpenStackMachine="e2e-b2ybbv/cluster-e2e-b2ybbv-control-plane-fctpw" namespace="e2e-b2ybbv" name="cluster-e2e-b2ybbv-control-plane-fctpw" reconcileID=712cdfe2-2bf9-4fcc-b4f0-18b05aa3e293
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x16984a1]

goroutine 192 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:115 +0x1fa
panic({0x1843380, 0x2a97550})
	/usr/local/go/src/runtime/panic.go:884 +0x212
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*Service).getOrCreateVolumeBuilder(0x44?, {0x1d26010, 0xc0017dd680}, 0x0?, {{0xc0013a20f0, 0x8}, 0x1, {{0xc0013a20f8, 0x6}, 0x0}}, ...)
	/workspace/pkg/cloud/services/compute/instance.go:442 +0x81
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*Service).getBlockDevices(0x1a09d00?, {0x1d26010, 0xc0017dd680}, 0xc000934800, {0xc000bb82a0?, 0x24?}, 0xc000121440?, 0x1d26010?)
	/workspace/pkg/cloud/services/compute/instance.go:502 +0x766
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*Service).createInstanceImpl(0xc000c43730, {0x1d26010, 0xc0017dd680}, 0x1aaff06?, 0xc000934800, {0xc001bb2a68, 0x12}, 0x2?, 0x20?)
	/workspace/pkg/cloud/services/compute/instance.go:318 +0xa5a
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*Service).CreateInstance(...)
	/workspace/pkg/cloud/services/compute/instance.go:225
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackMachineReconciler).getOrCreate(0xc00075d6e0?, {{0x1d397e0?, 0xc000a8f480?}, 0x0?}, 0xc0011224e0, 0x0?, 0x0?, 0xc0017dd680, 0x0?, {0xc00115c000, ...})
	/workspace/controllers/openstackmachine_controller.go:444 +0x1bc
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackMachineReconciler).reconcileNormal(0x1d397e0?, {0x1d35ef8, 0xc000fa4f30}, {0x1d3a640, 0xc001288e40?}, 0xc0011224e0, 0xc00144ac00, 0xc0017ddb80, 0xc0017dd680)
	/workspace/controllers/openstackmachine_controller.go:331 +0x325
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackMachineReconciler).Reconcile(0xc0000d02a0, {0x1d35ef8, 0xc000fa4f30}, {{{0xc0016dd6e0?, 0x10?}, {0xc0016f6300?, 0x40dc07?}}})
	/workspace/controllers/openstackmachine_controller.go:154 +0x9fd
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x1d35ef8?, {0x1d35ef8?, 0xc000fa4f30?}, {{{0xc0016dd6e0?, 0x17b68e0?}, {0xc0016f6300?, 0x10?}}})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:118 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0007320a0, {0x1d35e50, 0xc000610f80}, {0x18d4c00?, 0xc0013c97a0?})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:314 +0x3a5
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0007320a0, {0x1d35e50, 0xc000610f80})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:265 +0x1d9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:226 +0x85
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/internal/controller/controller.go:222 +0x333
```

Something's wrong in the PR. I'll fix it.

EmilienM:
@mdbooth @lentzi90 @mkjpryor This PR is now ready for review again.

@JoelSpeed (OpenShift API lead) has provided great insights on https://github.com/openshift/api/pull/1623 and I've copy-pasted this API in this PR; which I think would be the final proposal for this feature.

Feedback is welcome, thanks.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,81 @@ type RootVolume struct {
...
-	// VolumeType is the volume type of the volume.
-	// If omitted, the default type will be used.
-	VolumeType string `json:"volumeType,omitempty"`
-	// AvailabilityZone is the volume availability zone to create the volume in.
+// blockDeviceStorage is the storage type of a block device to create and
+// contains additional storage options.
+// +union
+//
+//nolint:godot
+type BlockDeviceStorage struct {
```
JoelSpeed, line 171: You should add CEL to validate that the `Volume` field is present iff `type: Volume`. There's examples of this in the openshift API examples API group

api/v1alpha7/types.go
```diff
@@ -163,20 +163,81 @@ type RootVolume struct {
...
+
+	// availabilityZone is the volume availability zone to create the volume in.
 	// If omitted, the availability zone of the server will be used.
+	// The availability zone must NOT contain spaces otherwise it will lead to volume that belongs
+	// to this availability zone register failure, see kubernetes/cloud-provider-openstack#1379 for
+	// further information.
+	// The maximum length of availability zone name is 63 as per labels limits.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[^ ]*$`
```
JoelSpeed, line 203: If you switched this to CEL, you'd be able to provide a custom error message. Have you tried
```suggestion
	// +kubebuilder:validation:XValidation:rule="!self.contains(' ')",message="availabilityZone may not contain spaces"
```

api/v1alpha7/types.go
EmilienM: @JoelSpeed can you confirm that I in fact don't need it because the 2 fields in `BlockDeviceVolume` are optional?

api/v1alpha7/types.go
EmilienM: ack, I'll do that.

api/v1alpha7/types.go
JoelSpeed: `BlockDeviceStorage` is a discriminated union, you need to apply the validations for it right now because the discriminated union code within the API is not yet implemented.

See https://github.com/openshift/api/blob/79b9cd5e6020c2e430b0d90ad7413a0ab85f55c8/example/v1/types_stable.go#L89 for an example.

To be clear, when `type: Local`, we do not want `volume` (the member) to be set, and when `type: Volume` we do want `volume` to be set right?

EmilienM:
Ouch, this API test error:

```=== RUN   TestAPIs
Running Suite: Controller Suite - /home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers
==========================================================================================================
Random Seed: [1m1698160668[0m

Will run [1m7[0m of [1m7[0m specs
[38;5;243m------------------------------[0m
[38;5;9m[BeforeSuite] [FAILED] [9.203 seconds][0m
[38;5;9m[1m[BeforeSuite] [0m
[38;5;243m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:62[0m

  [38;5;243mTimeline >>[0m
  [1mSTEP:[0m bootstrapping test environment [38;5;243m@ 10/24/23 15:17:48.016[0m
  [38;5;9m[FAILED][0m in [BeforeSuite] - /home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77 [38;5;243m@ 10/24/23 15:17:57.219[0m
  [38;5;243m<< Timeline[0m

  [38;5;9m[FAILED] Unexpected error:
      <*fmt.wrapError | 0xc000578000>: 
      unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "openstackclusters.infrastructure.cluster.x-k8s.io": CustomResourceDefinition.apiextensions.k8s.io "openstackclusters.infrastructure.cluster.x-k8s.io" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
      {
          msg: "unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
          err: <*fmt.wrapError | 0xc000791fe0>{
              msg: "unable to create CRD instances: unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
              err: <*fmt.wrapError | 0xc000791fc0>{
                  msg: "unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
                  err: <*errors.StatusError | 0xc000000a00>{
                      ErrStatus: {
                          TypeMeta: {Kind: "", APIVersion: ""},
                          ListMeta: {
                              SelfLink: "",
                              ResourceVersion: "",
                              Continue: "",
                              RemainingItemCount: nil,
                          },
                          Status: "Failure",
                          Message: "CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
                          Reason: "Invalid",
                          Details: {
                              Name: "openstackclusters.infrastructure.cluster.x-k8s.io",
                              Group: "apiextensions.k8s.io",
                              Kind: "CustomResourceDefinition",
                              UID: "",
                              Causes: [
                                  {
                                      Type: "FieldValueForbidden",
                                      Message: "Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
                                      Field: "spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices...

  Gomega truncated this representation as it exceeds 'format.MaxLength'.
  Consider having the object provide a custom 'GomegaStringer' representation
  or adjust the parameters in Gomega's 'format' package.

  Learn more here: https://onsi.github.io/gomega/#adjusting-output

  occurred[0m
  [38;5;9mIn [1m[BeforeSuite][0m[38;5;9m at: [1m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77[0m [38;5;243m@ 10/24/23 15:17:57.219[0m
[38;5;243m------------------------------[0m

[38;5;9m[1mSummarizing 1 Failure:[0m
  [38;5;9m[FAIL][0m [38;5;9m[1m[BeforeSuite] [0m
  [38;5;243m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77[0m

[38;5;9m[1mRan 0 of 7 Specs in 10.687 seconds[0m
[38;5;9m[1mFAIL![0m -- [38;5;14m[1mA BeforeSuite node failed so all tests were skipped.[0m
--- FAIL: TestAPIs (10.69s)
FAIL
```

JoelSpeed:
> Ouch, this API test error:
> 
> ```
> Running Suite: Controller Suite - /home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers
> ==========================================================================================================
> Random Seed: �[1m1698160668�[0m
> 
> Will run �[1m7�[0m of �[1m7�[0m specs
> �[38;5;243m------------------------------�[0m
> �[38;5;9m[BeforeSuite] [FAILED] [9.203 seconds]�[0m
> �[38;5;9m�[1m[BeforeSuite] �[0m
> �[38;5;243m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:62�[0m
> 
>   �[38;5;243mTimeline >>�[0m
>   �[1mSTEP:�[0m bootstrapping test environment �[38;5;243m@ 10/24/23 15:17:48.016�[0m
>   �[38;5;9m[FAILED]�[0m in [BeforeSuite] - /home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77 �[38;5;243m@ 10/24/23 15:17:57.219�[0m
>   �[38;5;243m<< Timeline�[0m
> 
>   �[38;5;9m[FAILED] Unexpected error:
>       <*fmt.wrapError | 0xc000578000>: 
>       unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "openstackclusters.infrastructure.cluster.x-k8s.io": CustomResourceDefinition.apiextensions.k8s.io "openstackclusters.infrastructure.cluster.x-k8s.io" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
>       {
>           msg: "unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
>           err: <*fmt.wrapError | 0xc000791fe0>{
>               msg: "unable to create CRD instances: unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
>               err: <*fmt.wrapError | 0xc000791fc0>{
>                   msg: "unable to create CRD \"openstackclusters.infrastructure.cluster.x-k8s.io\": CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
>                   err: <*errors.StatusError | 0xc000000a00>{
>                       ErrStatus: {
>                           TypeMeta: {Kind: "", APIVersion: ""},
>                           ListMeta: {
>                               SelfLink: "",
>                               ResourceVersion: "",
>                               Continue: "",
>                               RemainingItemCount: nil,
>                           },
>                           Status: "Failure",
>                           Message: "CustomResourceDefinition.apiextensions.k8s.io \"openstackclusters.infrastructure.cluster.x-k8s.io\" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices].items.properties[storage].properties[volume].properties[availabilityZone].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
>                           Reason: "Invalid",
>                           Details: {
>                               Name: "openstackclusters.infrastructure.cluster.x-k8s.io",
>                               Group: "apiextensions.k8s.io",
>                               Kind: "CustomResourceDefinition",
>                               UID: "",
>                               Causes: [
>                                   {
>                                       Type: "FieldValueForbidden",
>                                       Message: "Forbidden: estimated rule cost exceeds budget by factor of 2.9x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)",
>                                       Field: "spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[bastion].properties[instance].properties[additionalBlockDevices...
> 
>   Gomega truncated this representation as it exceeds 'format.MaxLength'.
>   Consider having the object provide a custom 'GomegaStringer' representation
>   or adjust the parameters in Gomega's 'format' package.
> 
>   Learn more here: https://onsi.github.io/gomega/#adjusting-output
> 
>   occurred�[0m
>   �[38;5;9mIn �[1m[BeforeSuite]�[0m�[38;5;9m at: �[1m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77�[0m �[38;5;243m@ 10/24/23 15:17:57.219�[0m
> �[38;5;243m------------------------------�[0m
> 
> �[38;5;9m�[1mSummarizing 1 Failure:�[0m
>   �[38;5;9m[FAIL]�[0m �[38;5;9m�[1m[BeforeSuite] �[0m
>   �[38;5;243m/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/controllers/suite_test.go:77�[0m
> 
> �[38;5;9m�[1mRan 0 of 7 Specs in 10.687 seconds�[0m
> �[38;5;9m�[1mFAIL!�[0m -- �[38;5;14m�[1mA BeforeSuite node failed so all tests were skipped.�[0m
> --- FAIL: TestAPIs (10.69s)
> FAIL
> ```

Can you place a reasonable length limit on the number of `additionalBlockDevices` that would be allowed to be added? eg is it reasonable to say no one would add more than 256 additional block devices? Or would you go even lower than that?

EmilienM:
> Can you place a reasonable length limit on the number of `additionalBlockDevices` that would be allowed to be added? eg is it reasonable to say no one would add more than 256 additional block devices? Or would you go even lower than that?

This is technically unlimited: https://opendev.org/openstack/nova/src/commit/d0b459423dd81644e8d9382b6c87fabaa4f03ad4/nova/compute/manager.py#L1619-L1623
But I'll limit to 255 and explain why in a comment for developers.

api/v1alpha7/types.go
EmilienM: > To be clear, when type: Local, we do not want volume (the member) to be set

yes

> when type: Volume we do want volume to be set right?

Not necessarily, both options in `BlockDeviceVolume` are optional and have no default; they are up to the OpenStack cloud and how it was deployed.

EmilienM: IIUC I need one CEL validation for `type: local` that verifies that no `volume` is given.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,83 @@ type RootVolume struct {
...
+	// type is the type of block device to create.
+	// This can be either "Volume" or "Local".
+	// +kubebuilder:validation:Enum="Volume";"Local"
+	// +kubebuilder:validation:Required
+	// +unionDiscriminator
+	Type BlockDeviceType `json:"type"`
+
+	// volume contains additional storage options for a volume block device.
+	// +optional
+	// +unionMember,optional
```
EmilienM, line 182: @JoelSpeed I added that.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,83 @@ type RootVolume struct {
...
-	Name string `json:"name"`
-	// Size is the size in GB of the volume.
-	Size int `json:"diskSize"`
-	// VolumeType is the volume type of the volume.
-	// If omitted, the default type will be used.
-	VolumeType string `json:"volumeType,omitempty"`
-	// AvailabilityZone is the volume availability zone to create the volume in.
+// blockDeviceStorage is the storage type of a block device to create and
+// contains additional storage options.
+// +kubebuilder:validation:XValidation:rule="self.type == 'Local' ?  !has(self.volume) : has(self.volume)",message="volume is forbidden for type 'Local'"
```
EmilienM, line 168: @JoelSpeed the only rule that I think we need. Since `volume` is optional, I don't think we should have any other CEL here.

EmilienM:
The CEL broke the e2e job:

```
  The OpenStackMachineTemplate "cluster-e2e-3e4n6m-control-plane" is invalid: spec.template.spec.additionalBlockDevices[0].storage: Invalid value: "object": volume is forbidden for type 'Local'
```

I probably did something backward; I'll look into it asap.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,83 @@ type RootVolume struct {
...
-	Name string `json:"name"`
-	// Size is the size in GB of the volume.
-	Size int `json:"diskSize"`
-	// VolumeType is the volume type of the volume.
-	// If omitted, the default type will be used.
-	VolumeType string `json:"volumeType,omitempty"`
-	// AvailabilityZone is the volume availability zone to create the volume in.
+// blockDeviceStorage is the storage type of a block device to create and
+// contains additional storage options.
+// +kubebuilder:validation:XValidation:rule="self.type == 'Local' && has(self.volume)",message="volume is forbidden for type 'Local'"
```
EmilienM, line 168: @JoelSpeed the only rule that I think we need. Since volume is optional, I don't think we should have any other CEL here.

EmilienM:
Tests still failing with CEL, will check tomorrow

api/v1alpha7/types.go
JoelSpeed: The validation should pass if you want the object to be accepted, you need `!has(self.volume)` on the second clause

JoelSpeed (commented):
API changes LGTM

api/v1alpha7/types.go
```diff
@@ -163,20 +163,83 @@ type RootVolume struct {
...
+	// Also, this name will be used for tagging the block device.
+	// Information about the block device tag can be obtained from the OpenStack
+	// metadata API or the config drive.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// sizeGiB is the size of the block device in gibibytes (GiB).
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Required
+	SizeGiB int `json:"sizeGiB"`
```
mkjpryor, line 224: I'm not a massive fan of field names that contain the unit. Is it not sufficient for the description and docs to say the unit?

mkjpryor (commented):
Just one small comment on the API from me. Other than that it looks sensible. I'm happy to be overruled if others feel differently.

api/v1alpha7/types.go
JoelSpeed: If you prefer not to have the units in the name, then you should use a `resource.Quantity`, which would mean the field would look like `size: 128Gi` to the user.

This is as per current [Kube API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units).

I'm happy with either but @EmilienM preferred this over the `resource.Quantity` IIUC

lentzi90: Not a strong opinion, but for me it seems preferable to have it part of the name.

lentzi90 (commented):
/lgtm

api/v1alpha7/types.go
EmilienM: having `size: 128Gi` would be error prone IMO (we would certainly need some validation). I prefer to have the unit in the name. Unless someone has a strong opinion.

EmilienM:
I've been asked (offline) to add tests for CEL. I agree we should have them and I plan to work on it, however this is going to be some work that I think needs to happen outside this PR.
I know it's not ideal but for now I'll suggest we move forward with it without CEL tests, as it has proven to actually report failures when the CEL was wrong, which means in its current state it's not breaking anything and in fact the CEL was tested already in the clusters that were deployed.

I'll work on adding testsuites for API and etc; but I really hope we can get it outside this PR. Otherwise, I'll remove all CELs from this PR so I can get this merged asap.

EmilienM:
I've removed CEL validations from the PR into a separate commit, see commit message for details, please.

api/v1alpha7/openstackmachine_types.go
```diff
@@ -89,7 +89,6 @@ type OpenStackMachineSpec struct {
 	// + ---
 	// + While it's possible to have unlimited number of block devices attached to a server instance, the number is
 	// + limited to 255 to avoid rule cost exceeded error in CRD validation.
-	// +kubebuilder:validation:MaxItems=255
```
EmilienM: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1736 is an issue that I created so at some point we can re-add CEL.

mdbooth:
The offline CEL discussion was with me, btw. For the record I'm very much in favour of CEL and I'd like to replace all webhook validations that can be implemented with it. My concern is that we should not add CEL without tests because it will inevitably be broken, and this will break folks using a cluster where it's enabled. If we're going to add CEL we need to test it. If we can't test it, we shouldn't add it yet. But we will add it.

pkg/cloud/services/compute/instance_test.go
```diff
@@ -765,8 +798,17 @@ func TestService_ReconcileInstance(t *testing.T) {
 						"boot_index":            float64(-1),
 						"delete_on_termination": true,
 						"destination_type":      "volume",
+						"volume_size":           float64(50),
```
mdbooth, line 801: We shouldn't be setting volume_size here because the volume already exists.

pkg/cloud/services/compute/instance.go
```diff
@@ -490,16 +502,31 @@ func (s *Service) getBlockDevices(eventObject runtime.Object, instanceSpec *Inst
...
 		blockDevices = append(blockDevices, bootfromvolume.BlockDevice{
-			SourceType:          bootfromvolume.SourceVolume,
-			DestinationType:     bootfromvolume.DestinationVolume,
-			UUID:                blockDevice.ID,
+			SourceType:          sourceType,
+			DestinationType:     destinationType,
+			UUID:                bdUUID,
 			BootIndex:           -1,
 			DeleteOnTermination: true,
+			VolumeSize:          blockDeviceSpec.SizeGiB,
```
mdbooth, line 529: We shouldn't set this unconditionally here because it doesn't apply to volumes. Hurray for unit tests!

mdbooth (changes requested):
Nothing major from me. This is nearly ready to go.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
-	Size int `json:"diskSize"`
-	// VolumeType is the volume type of the volume.
-	// If omitted, the default type will be used.
-	VolumeType string `json:"volumeType,omitempty"`
-	// AvailabilityZone is the volume availability zone to create the volume in.
+// blockDeviceStorage is the storage type of a block device to create and
+// contains additional storage options.
+// +union
+//
+//nolint:godot
```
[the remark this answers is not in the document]
lentzi90: I can help with this. It is complaining about missing dot in the comment above. And this is also the directive that `no-unused` was complaining about.

.golangci.yml
```diff
@@ -126,6 +126,9 @@ linters-settings:
       - pkg: sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1
         alias: controlplanev1
 
+  nolintlint:
+    # https://github.com/golangci/golangci-lint/issues/3228
+    allow-unused: true
```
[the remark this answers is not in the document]
EmilienM: In this PR, it was problematic for the `nolint:godot` and I found the issue that explains this behaviour can be a known issue and the workaround for now is to allow unused nolints to avoid the error.

api/v1alpha7/openstackmachine_types.go
```diff
@@ -84,9 +84,14 @@ type OpenStackMachineSpec struct {
...
 
-	// AdditionalBlockDevices is a list of specifications for additional block devices to attach to the server instance
+	// additionalBlockDevices is a list of specifications for additional block devices to attach to the server instance
+	//
+	// + ---
+	// + While it's possible to have unlimited number of block devices attached to a server instance, the number is
+	// + limited to 255 to avoid rule cost exceeded error in CRD validation.
 	// +listType=map
 	// +listMapKey=name
+	// +optional
```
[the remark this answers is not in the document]
EmilienM: no, I was asked by Joel to put it actually and if you look at this file it's already in some places.

pkg/cloud/services/compute/instance.go
```diff
@@ -463,10 +470,15 @@ func (s *Service) getBlockDevices(eventObject runtime.Object, instanceSpec *Inst
 
 	if hasRootVolume(instanceSpec) {
 		rootVolumeToBlockDevice := infrav1.AdditionalBlockDevice{
-			Name:             "root",
-			AvailabilityZone: instanceSpec.RootVolume.AvailabilityZone,
-			Size:             instanceSpec.RootVolume.Size,
-			VolumeType:       instanceSpec.RootVolume.VolumeType,
+			Name:    "root",
```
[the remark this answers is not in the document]
EmilienM: I added a guard.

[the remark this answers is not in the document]
mdbooth: Where?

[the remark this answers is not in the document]
EmilienM: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1696/files#diff-5e7dbfd52305a21a11fb94fab201062ae2c7b851db93c2a8443a1d82d7aa7936R507-R509

pkg/cloud/services/compute/instance.go
```diff
@@ -490,16 +500,36 @@ func (s *Service) getBlockDevices(eventObject runtime.Object, instanceSpec *Inst
...
-		blockDevice, err := s.getOrCreateVolumeBuilder(eventObject, instanceSpec, blockDeviceSpec, "", fmt.Sprintf("Additional block device for %s", instanceSpec.Name))
-		if err != nil {
-			return []bootfromvolume.BlockDevice{}, err
+		var bdUUID string
+		var localDiskSizeGiB int
+		var sourceType bootfromvolume.SourceType
+		var destinationType bootfromvolume.DestinationType
+		if blockDeviceSpec.Name == "root" {
+			return []bootfromvolume.BlockDevice{}, fmt.Errorf("block device name 'root' is reserved")
+		}
```
mdbooth, line 509: Can you put this in the validating webhook instead?

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/f28da1834f1ea79c0e8e5617aa14607b045ca53b/api/v1alpha7/openstackmachine_webhook.go#L58

EmilienM: done

api/v1alpha7/openstackmachine_webhook.go
```diff
@@ -62,6 +62,14 @@ func (r *OpenStackMachine) ValidateCreate() (admission.Warnings, error) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	if r.Spec.RootVolume != nil && r.Spec.AdditionalBlockDevices != nil {
```
EmilienM, line 65: this file has no unit tests, but I started to write it and i'll publish via another PR if reviewers don't mind.

EmilienM: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1737

mdbooth: Ideally these would be envtests, btw. We could then re-use them to validate CEL.

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+type BlockDeviceStorage struct {
+	// type is the type of block device to create.
+	// This can be either "Volume" or "Local".
+	// +unionDiscriminator
```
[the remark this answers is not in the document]
mdbooth: We discussed this. These stay at least for now.

mdbooth (approved):

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
-	// Name of the Cinder volume in the context of a machine.
-	// It will be combined with the machine name to make the actual volume name.
-	Name string `json:"name"`
-	// Size is the size in GB of the volume.
-	Size int `json:"diskSize"`
-	// VolumeType is the volume type of the volume.
-	// If omitted, the default type will be used.
-	VolumeType string `json:"volumeType,omitempty"`
-	// AvailabilityZone is the volume availability zone to create the volume in.
+// blockDeviceStorage is the storage type of a block device to create and
```
lentzi90, line 166:
```suggestion
// BlockDeviceStorage is the storage type of a block device to create and
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+// +union
+//
+//nolint:godot
+type BlockDeviceStorage struct {
+	// type is the type of block device to create.
+	// This can be either "Volume" or "Local".
+	// +unionDiscriminator
+	Type BlockDeviceType `json:"type"`
+
+	// volume contains additional storage options for a volume block device.
```
lentzi90, line 177:
```suggestion
	// Volume contains additional storage options for a volume block device.
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+	// +optional
+	// +unionMember,optional
+	Volume *BlockDeviceVolume `json:"volume,omitempty"`
+}
+
+// blockDeviceVolume contains additional storage options for a volume block device.
+type BlockDeviceVolume struct {
+	// type is the Cinder volume type of the volume.
```
lentzi90, line 185:
```suggestion
	// Type is the Cinder volume type of the volume.
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+	// If omitted, the default Cinder volume type that is configured in the OpenStack cloud
+	// will be used.
+	// +optional
+	Type string `json:"type,omitempty"`
+
+	// availabilityZone is the volume availability zone to create the volume in.
```
lentzi90, line 191:
```suggestion
	// AvailabilityZone is the volume availability zone to create the volume in.
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
 	// If omitted, the availability zone of the server will be used.
+	// The availability zone must NOT contain spaces otherwise it will lead to volume that belongs
+	// to this availability zone register failure, see kubernetes/cloud-provider-openstack#1379 for
+	// further information.
+	// +optional
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
+// additionalBlockDevice is a block device to attach to the server.
```
lentzi90, line 200:
```suggestion
// AdditionalBlockDevice is a block device to attach to the server.
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+type AdditionalBlockDevice struct {
+	// name of the block device in the context of a machine.
+	// If the block device is a volume, the Cinder volume will be named
+	// as a combination of the machine name and this name.
+	// Also, this name will be used for tagging the block device.
+	// Information about the block device tag can be obtained from the OpenStack
+	// metadata API or the config drive.
+	Name string `json:"name"`
+
+	// sizeGiB is the size of the block device in gibibytes (GiB).
```
lentzi90, line 210:
```suggestion
	// SizeGiB is the size of the block device in gibibytes (GiB).
```

api/v1alpha7/types.go
```diff
@@ -163,20 +163,69 @@ type RootVolume struct {
...
+	SizeGiB int `json:"sizeGiB"`
+
+	// storage specifies the storage type of the block device and
```
lentzi90, line 213:
```suggestion
	// Storage specifies the storage type of the block device and
```

lentzi90 (commented):
Not sure if I missed it before or if it was changed, but the doc strings should be matching the code also for capitalization. I get that this will then make the yaml description not match the yaml fields but this is the system we currently follow.

api/v1alpha7/openstackmachine_types.go
```diff
@@ -84,9 +84,10 @@ type OpenStackMachineSpec struct {
 	// The volume metadata to boot from
 	RootVolume *RootVolume `json:"rootVolume,omitempty"`
 
-	// AdditionalBlockDevices is a list of specifications for additional block devices to attach to the server instance
+	// additionalBlockDevices is a list of specifications for additional block devices to attach to the server instance
```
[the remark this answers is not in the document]
EmilienM: no, I've been told by Joel to use the pascal name per k8s convention.

lentzi90 (commented):
/lgtm

api/v1alpha7/openstackmachine_types.go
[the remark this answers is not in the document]
EmilienM: for now we agreed to follow existing patterns in CAPO...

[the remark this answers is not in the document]
lentzi90: As agreed offline, currently the files are formatted in this way (go doc) so we stick to it.

Issue #1458, opened 2023-01-25
E2e tests fail because of cri-tools version bump


lentzi90:
/kind bug

**What steps did you take and what happened:**

E2e tests currently fail with due to kubeadm init pre-flight issues:

```
Jan 19 11:43:05 cluster-e2e-e2n2w5-control-plane-mm7kc cloud-init[1402]: [init] Using Kubernetes version: v1.25.0
Jan 19 11:43:05 cluster-e2e-e2n2w5-control-plane-mm7kc cloud-init[1402]: [preflight] Running pre-flight checks
Jan 19 11:43:05 cluster-e2e-e2n2w5-control-plane-mm7kc cloud-init[1402]: error execution phase preflight: [preflight] Some fatal errors occurred:
Jan 19 11:43:05 cluster-e2e-e2n2w5-control-plane-mm7kc cloud-init[1402]:         [ERROR CRI]: container runtime is not running: output: time="2023-01-19T11:42:58Z" level=fatal msg="validate service connection: CRI v1 runtime API is not implemented for endpoint \"unix:///var/run/containerd/containerd.sock\": rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService"
Jan 19 11:43:05 cluster-e2e-e2n2w5-control-plane-mm7kc cloud-init[1402]: , error: exit status 1
```

The cri-tools package was updated to v1.26 even tough we install kubeadm v1.25!
Working:
```
Get:1 https://packages.cloud.google.com/apt kubernetes-xenial/main amd64 cri-tools amd64 1.25.0-00 [17.9 MB]
```
Not working:
```
Get:1 https://packages.cloud.google.com/apt kubernetes-xenial/main amd64 cri-tools amd64 1.26.0-00 [18.9 MB]
```

Now the issue is that cri-tools is installed by [this script](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/framework/kubernetesversions/data/debian_injection_script.envsubst.sh.tpl) as a dependency to kubeadm. And even though kubeadm is explicitly installed with version v1.25.0, we still get cri-tools v1.26.

**What did you expect to happen:**

E2e tests should pass.

Issue #1429, opened 2022-12-21
Wrong kubeadm, kubelet versions in e2e tests


lentzi90:
/kind bug

**What steps did you take and what happened:**

Checking the [cloud-final-log](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-openstack-e2e-test-main/1605411522220134400/artifacts/clusters/bootstrap/instances/e2e-xlwutp/cluster-e2e-xlwutp-control-plane-5q5mb/) for the periodic e2e tests shows this:

```
Dec 21 04:47:03 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:34] kubeadm version:  v1.23.10
Dec 21 04:47:03 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:34] kubectl version:  Client Version: v1.23.10
Dec 21 04:47:03 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:34] kubelet version:  Kubernetes v1.23.10
Dec 21 04:47:03 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:34] *************************************************
Dec 21 04:47:03 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:34] [init] Using Kubernetes version: v1.25.0
```

Note that v1.23.10 is the installed version of kubeadm, kubelet and kubectl even though the test is supposed to run with v1.25.0.
The reason for this can be seen earlier in the log:

```
Dec 21 04:46:42 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:14] The following packages will be upgraded:
Dec 21 04:46:42 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:14]   kubeadm
Dec 21 04:46:42 cluster-e2e-xlwutp-control-plane-5q5mb cloud-init[648]: [2022-12-21 04:45:14] 1 upgraded, 0 newly installed, 0 to remove and 20 not upgraded.
[...]

PR #1699, opened 2023-10-02
🌱 E2e: Use pre-build node images

lentzi90:

We are currently using a plain ubuntu cloud image for the nodes in most e2e tests. Cloud-init scripts are used to install the Kubernetes components on startup. This is convenient since we do not have to build specific images. However, the scripts are not well maintained, they increase the complexity and running time of the tests.

This changes the tests to use a pre-built image. The image was build using image-builder, which is a standard for many CAPI providers.

As a consequence, the containerd patches are no longer needed. (In fact they probably are not needed anyway since the only reason we needed them was to get a newer version of containerd that was not available in the ubuntu repos at the time.)

A bit more on why I think we need this:
The [script for injecting CI artifacts](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/framework/kubernetesversions/data/debian_injection_script.envsubst.sh.tpl) has not proven to be very reliable. We have had a number of issues caused by us relying on them. See for example #1429 and #1458. We have [disabled the k8s upgrade test](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c73d55dd47006207a15c93bbae69811c97274827/test/e2e/suites/e2e/self_hosted_test.go#L35-L40) because the use of CI artifacts makes it hard/impossible to run it. I also expect that we will run into issues the next time we try to bump the Kubernetes version because of the [change to package repositories](https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/). The CI injection script has not been updated to handle the new repositories even though the old repositories are already frozen.
It seems like we are quite alone using this script and need to decide if we take ownership of it or just start building images.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


test/e2e/shared/suite.go
```diff
@@ -58,6 +58,8 @@ func Node1BeforeSuite(e2eCtx *E2EContext) []byte {
 	flag.Parse()
 	Expect(e2eCtx.Settings.ConfigPath).To(BeAnExistingFile(), "Invalid test suite argument. configPath should be an existing file.")
 	Expect(os.MkdirAll(e2eCtx.Settings.ArtifactFolder, 0o750)).To(Succeed(), "Invalid test suite argument. Can't create artifacts-folder %q", e2eCtx.Settings.ArtifactFolder)
+	templatesDir := path.Join(e2eCtx.Settings.ArtifactFolder, "templates")
+	Expect(os.MkdirAll(templatesDir, 0o750)).To(Succeed(), "Can't create templates folder %q", templatesDir)
```
lentzi90, line 62: This folder is created by the `GenerateCIArtifactsInjectedTemplateForDebian` function that is called for templates that should get CI artifacts injected. If there are no such templates, we still need to ensure that the folder exists since the templates without CI artifacts will also be written to it.

lentzi90:
Unsure if flake or actual issue... 5 tests passed, 1 failed and 4 skipped.

lentzi90:
Looks like this shaves off about 7-8 minutes from the tests. Not much but always something :slightly_smiling_face:

Makefile
```diff
@@ -151,19 +151,20 @@ E2E_NO_ARTIFACT_TEMPLATES_DIR=test/e2e/data/infrastructure-openstack-no-artifact
...
+		 cluster-template.yaml \
 		 cluster-template-flatcar.yaml)
+# Currently no templates that require CI artifacts
+# $(addprefix $(E2E_TEMPLATES_DIR)/, add-templates-here.yaml) \
 
-$(E2E_TEMPLATES_DIR)/cluster-template.yaml: $(E2E_KUSTOMIZE_DIR)/with-tags $(KUSTOMIZE) FORCE
+$(E2E_NO_ARTIFACT_TEMPLATES_DIR)/cluster-template.yaml: $(E2E_KUSTOMIZE_DIR)/with-tags $(KUSTOMIZE) FORCE
 	$(KUSTOMIZE) build "$<" > "$@"
 
 $(E2E_TEMPLATES_DIR)/cluster-template-%.yaml: $(E2E_KUSTOMIZE_DIR)/% $(KUSTOMIZE) FORCE
```
mdbooth, line 170: Can this rule be removed? Looks like nothing will use it now.

mdbooth (commented):
Thanks! Is anything still using the templates with artifacts? If not, could we remove what we're not using?

/lgtm

Makefile
lentzi90: Yes it is unused for now. I was thinking we may want to use it in some specific job in the future, but we can of course just add it back at that point. Removing... :slightly_smiling_face:

mdbooth:
@lentzi90 I completely agree that this approach is the best way to go. However, remember that we originally did this because we weren't doing a good job of maintaining the images. So the maintenance problem persists, we'd just be moving it from the e2e templates to an image builder pipeline that we don't have yet.

Which is the lesser evil? If we want to merge this, what do you think the chances are of getting the image builder pipeline effort going again?

lentzi90:
> @lentzi90 I completely agree that this approach is the best way to go. However, remember that we originally did this because we weren't doing a good job of maintaining the images. So the maintenance problem persists, we'd just be moving it from the e2e templates to an image builder pipeline that we don't have yet.
> 
> Which is the lesser evil? If we want to merge this, what do you think the chances are of getting the image builder pipeline effort going again?

That is a good point.
I would honestly rather work on the pipeline than trying to patch together the on-the-fly building. It has been painful trying to debug the failures that happen in cloud-init of the workload cluster nodes. :slightly_frowning_face: 
The image-builder works perfectly if I just run it locally, and that is what I have done for this PR. I understand that it is not ideal to rely on me or other maintainers to do this manually though. However, I find it slightly better to be able to build those images manually and then it "just works", instead of using plain cloud images but having to fight the pre-kubeadm scripts.

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -53,8 +53,17 @@
...
     # We upload the Amphora image so it doesn't have to be build
-    IMAGE_URLS="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/amphora/2022-12-05/amphora-x64-haproxy.qcow2"
+    # Upload the images so we don't have to upload them from Prow
+    # NOTE: If you get issues when changing/adding images, check if the limits
+    # are sufficient and change the variable above if needed.
+    # https://docs.openstack.org/glance/latest/admin/quotas.html
+    IMAGE_URLS="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/amphora/2022-12-05/amphora-x64-haproxy.qcow2,"
+    IMAGE_URLS+="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/cirros/2022-12-05/cirros-0.6.1-x86_64-disk.img,"
+    IMAGE_URLS+="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2023-09-29/ubuntu-2204-kube-v1.27.2.img,"
+    IMAGE_URLS+="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/flatcar/flatcar-stable-3510.2.4-kube-v1.27.2.img"
```
wwentland, line 66: @mdbooth kindly[ uploaded a new version](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1713#issuecomment-1752768084), so we might want to probably use it here:

- `https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/flatcar/flatcar-stable-3602.2.0-kube-v1.27.2.img`

This would get us all the security fixes in that (and interim) release(s):

- https://www.flatcar.org/releases#release-3602.2.0

wwentland (commented):
I like it, with the downside of having to manually maintain/build/upload newer versions of the images in use here.

hack/ci/cloud-init/controller.yaml.tpl
lentzi90: Thanks for pointing this out. I have updated it :slightly_smiling_face:

PR #1701, opened 2023-10-02
✨ Add option to customize private network MTU

### LINKED ISSUE #1670: Add option to customize private network MTU
/kind feature

For now, when we create the network using CAPO, we have no option to set the network's maximum transmission unit (MTU). 
We can add a new `OpenStackCluster` api optional field to provide this customization option.

For instance it could be:
```golang
  // [...]

  // NetworkMTU sets the maximum transmission unit (MTU) value to address fragmentation for the private network ID.
  // This value will be used only if the Cluster actuator creates the network.
  // If leaved empty, the network will have the default MTU defined in Openstack network service.
  // +optional
  NetworkMTU *int `json:"networkMtu,omitempty"`

  // [...]
}
```
 
I would be happy to add this option if you approve it.

alexandrevilain:

Fixes #1670

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -63,6 +63,7 @@ type createOpts struct {
 	AdminStateUp        *bool  `json:"admin_state_up,omitempty"`
 	Name                string `json:"name,omitempty"`
 	PortSecurityEnabled *bool  `json:"port_security_enabled,omitempty"`
+	MTU                 *int   `json:"mtu,omitempty"`
 }
 
 func (c createOpts) ToNetworkCreateMap() (map[string]interface{}, error) {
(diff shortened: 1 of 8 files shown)


pkg/cloud/services/networking/network.go
```diff
@@ -135,18 +136,17 @@ func (s *Service) ReconcileNetwork(openStackCluster *infrav1.OpenStackCluster, c
...
-	} else {
-		opts = createOpts{
-			AdminStateUp: gophercloud.Enabled,
-			Name:         networkName,
-		}
+		opts.PortSecurityEnabled = gophercloud.Disabled
+	}
+
+	if openStackCluster.Spec.NetworkMTU > 0 {
+		opts.MTU = &openStackCluster.Spec.NetworkMTU
```
seanschneeweiss, line 149: Do we actually have to check if the OpenStack installation supports the `net-mtu` extension? Or can we assume a more or less up to date neutron version that supports setting the MTU?

alexandrevilain: So you want to check for this extension by doing a equivalent of `openstack extension show net-mtu`.
If it exits, add this field, otherwise return an error to ask the user remove this field as it's not supported by the underlying provider ?

seanschneeweiss: Either way, this feels like a non-transient error.

alexandrevilain: Yes, you're right, adding `requires net-mtu extension` in the comment should be sufficient!

alexandrevilain: Field description is now updated :)

dulek:
/ok-to-test
/lgtm

This does seem to be okay, I'm not 100% sure on the placement of the API option though.

dulek:
Looks like conversion code needs to be added.

alexandrevilain:
> Looks like conversion code needs to be added.

Thanks! This is now fixed :)

seanschneeweiss (approved):

PR #1708, opened 2023-10-04
WIP - rework ports

EmilienM:

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


pkg/cloud/services/compute/ports.go
```diff
@@ -0,0 +1,226 @@
...
+		// List ports by name and tags
+		ports, err := networkingClient.ListPort(ports.ListOpts{
+			Name:      portName,
+			NetworkID: port.Network.ID,
+		})
+		if err != nil {
+			return err
+		}
+		if len(ports) == 0 {
+			return nil
```
EmilienM, line 149: @mdbooth I don't get why we returned `nil` here, it'll stop the whole function if a port doesn't exist. So what? I'm trying to figure out how you wanted to find out what ports actually need creation/deletion after adoption.

mdbooth: When we write this again we should include some comments!

It will have been because we always create ports in order. This means that if port `N` doesn't exist we know that ports `>N` don't exist.

PR #1709, opened 2023-10-04
🌱 Ignore vendor directories in boilerplate check

mdbooth:
This allows `make verify-boilerplate` to pass when using vendoring.


lentzi90 (commented):
/lgtm

PR #1710, opened 2023-10-05
🌱 Allow custom KUBEBUILDER_ASSETS_DIR in tests

mdbooth:
The primary purpose of this change is to enable running the envtests in environments where neither $XDG_DATA_HOME nor $HOME are set.

Additionally, by executing setup-envtest explicitly in the `test` rule rather than in a global variable definition, we:
* Only run it when it's required
* Log its execution and output

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,15 +129,25 @@ endif
 $(ARTIFACTS):
 	mkdir -p $@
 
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

mdbooth:
I've also tested this in our CI and it *looks* like it fixes it 🥳

PR #1712, opened 2023-10-05
🌱 Add explicit dependency on github.com/golang/mock/mockgen/model

mdbooth:
This allows mockgen to run when using vendoring.


mandre (commented):
Even though CAPO doesn't do vendoring, this allows to minimize the number of patches to maintain in our downstream fork, and could be of use for anyone else doing vendoring.
/lgtm

lentzi90 (commented):
/lgtm

Issue #1502, opened 2023-03-10
build CAPO supported images


tormath1:
/kind feature

**Describe the solution you'd like**:

During the CAPO office hours, @mdbooth suggested to build CAPO supported images in the CI. These images could be directly consumed by the tests (and consumed by the users?).

These images could be built for multiple Kubernetes version and multiple OS version (ubuntu-2004, flatcar, etc.)

**Anything else you would like to add:**

On the implementation side, we could use Github Actions and image-builder Docker image. Example with Flatcar:
```
sudo docker run --privileged --net host -v "${PWD}/docker-output:/output" -v /dev:/dev --rm \
    -e PACKER_FLAGS="--var 'kubernetes_semver=v1.26.2' --var 'kubernetes_series=v1.26'" \
    -e OEM_ID=openstack \
    tormath1/cluster-node-image-builder-amd64 \
    build-qemu-flatcar
```

(Note: it needs https://github.com/kubernetes-sigs/image-builder/pull/1092)

For Flatcar: https://github.com/flatcar/Flatcar/issues/928 (https://github.com/kubernetes-sigs/image-builder/issues/924)  can be interesting

Produced images could be then stored on a GCS bucket.

mdbooth:
@tormath1 We would probably use cloudbuild rather than GH actions, and I think that *might* support KVM.

mdbooth:
That's awesome!

I think we're going to need this in cloudbuild for 2 reasons:
* We need the artifacts in a GCS bucket for CI or we'll exceed our ingress quota.
* The only automated way (caveat: that I'm aware of) to use the project's gcloud credentials is from the prow secure cluster.

That said, I don't currently know when I or anybody else will find time to implement this and you seem to have something looks close to useful for at least some people already. I wonder if we should merge a GH actions solution now and aim to replace it with a cloudbuild solution in prow when we get to it.

Thoughts?
/cc @lentzi90 @tobiasgiese @jichenjc

mdbooth:
> Moving it to cloudbuild is not an issue as it mainly relies on the Makefile and Docker to run. It's just less community friendly as one can easily browse the logs on the Github actions (or maybe we can have the logs from cloubuild ? I don't know how it works :D)

Yeah, we have the logs: they're in the prow job that runs the cloudbuild, same as regular tests. Admittedly they're not as easy to find, though. This feels like an opportunity for improved docs 🤔

mdbooth:
[...]

PR #1713, opened 2023-10-06
:seedling: ci: bump Flatcar tested version

tormath1:
**What this PR does / why we need it**: Test a [new](https://www.flatcar.org/releases#release-3602.2.0) Flatcar major Stable

**Special notes for your reviewer**: As discussed previously, let's bump Flatcar tested version only for new major. I followed these instructions: https://gist.github.com/tormath1/acbae5c6cd12420bb8ea137e25655c99 (if the tests are green one maintainer could upload the Flatcar image to the CAPO GCS bucket :pray: , thanks!)  

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/hack/ci/cloud-init/controller.yaml.tpl
+++ b/hack/ci/cloud-init/controller.yaml.tpl
@@ -138,7 +138,7 @@
     # https://docs.openstack.org/glance/latest/admin/quotas.html
     /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/cirros/2022-12-05/cirros-0.6.1-x86_64-disk.img
     /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2023-01-14/focal-server-cloudimg-amd64.img
-    /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/flatcar/flatcar-stable-3510.2.4-kube-v1.27.2.img
+    /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/flatcar/flatcar-stable-3602.2.0-kube-v1.27.2.img
 
     # Add the controller to its own host aggregate and availability zone
     aggregateid=$(openstack aggregate create --zone "${PRIMARY_AZ}" "${PRIMARY_AZ}" -f value -c id)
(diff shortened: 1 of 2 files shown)


hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -138,7 +138,7 @@
     # https://docs.openstack.org/glance/latest/admin/quotas.html
     /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/cirros/2022-12-05/cirros-0.6.1-x86_64-disk.img
     /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2023-01-14/focal-server-cloudimg-amd64.img
-    /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/flatcar/flatcar-stable-3510.2.4-kube-v1.27.2.img
+    /opt/stack/devstack/tools/upload_image.sh https://storage.googleapis.com/tormath1-flatcar/flatcar-stable-3602.2.0-kube-v1.27.2.img
```
wwentland, line 141: Is there any way in which we wouldn't have to switch between `artifacts.k8s-staging-capi-openstack.appspot.com/test` and `tormath1-flatcar` whenever a new flatcar image is being made available?

I appreciate that an automatic process would be best (cf. https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1502), but maybe we can find a different solution in the interim?

wwentland (commented):
Thank you, looking forward to testing the current stable version. 

FWIW, I've used that version in my own clusters already and did not run into issues.

hack/ci/cloud-init/controller.yaml.tpl
wwentland: > Yes, it's not ideal and it creates bottleneck: one needs to generate the image with the image-builder and one with CAPO GCS Bucket write access needs to upload it.

Indeed, that's exactly what I was wondering we could achieve here: Find a person with the permissions above or grant them.

> One different approach would be to remove the image-builder dependency and to use plain Flatcar image: [...] In the Flatcar template, using Ignition, we could pull Kubernetes dependencies using [systemd-sysext](https://www.flatcar.org/docs/latest/container-runtimes/getting-started-with-kubernetes/#setup-the-control-plane):

I am *very* much looking forward to utilising that functionality in all my clusters and, quite likely, in the context of tests here. How far has that feature progressed by now? My impression was that it is still quite experimental. But, if we can make it work, I'd very much like that. Until then, my preference would be for tests to mirror the setup that most users would use.

On that note: Please give me a shout if you need any help in getting the templates ready.

mdbooth: I still think the ideal situation here is to generate them directly in the CI. Incidentally, the 'OpenStack' packer builder is vastly quicker and more reliable than the qemu builder. The only problem with it is you need an OpenStack to run it against. That's hurting us.

mdbooth:
@tormath1 I've copied the new image to our storage bucket. I'm happy to merge this once you've bumped the location. Thanks!

/approve

wwentland (commented):
That's great, thanks again!

/lgtm

wwentland (commented):
Cheers!

/lgtm

PR #1715, opened 2023-10-06
✨ Add server name for the Machine InternalDNS

### LINKED ISSUE #1689: CAPO should set InternalDNS in status.Addresses

MaysaMacedo:
For a node to join the cluster, a CSR generated by kubelet with the node name must be approved. The approval happens if the value of the NodeInternalDNS entry on the Machine matches the node name. When using legacy cloud provider the value of that node name is the Server name. This commit ensures the nodes handled with legacy cloud provides get approved by adding the server name to the list of Machine addresses.

Fixes #1689 

/hold


controllers/openstackmachine_controller.go
```diff
@@ -348,6 +348,12 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 	}
 
 	addresses := instanceNS.Addresses()
+
+	// Append the server's name as the NodeInternalDNS to match the node name of a CSR created for a node that used in-tree provider.
```
mdbooth, line 352: Can we expand on this comment a bit? This makes it sound like we might be able to remove this now. However, we can probably never remove this because it's required for nodes which were upgraded from in-tree.

MaysaMacedo: Done

controllers/openstackmachine_controller.go
```diff
@@ -348,6 +348,14 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 	}
 
 	addresses := instanceNS.Addresses()
+
+	// A node joins the cluster if the CSR generated by kubelet with the node name is approved.
```
mdbooth, line 352: Can you make a note here that this is specific to OpenShift but likely also relevant to other systems.

mdbooth (commented):
@MaysaMacedo Expanded comment looks good, thanks! Can I ask you to just clarify that the specific behaviour described is OpenShift's?

@lentzi90 I've worked on this with @MaysaMacedo so I'd appreciate an independent look if you get a moment.

/approve

lentzi90 (commented):
Do I understand correctly that the idea is that the server name will in most cases be the same as the node name, especially if migrating from the in-tree provider?
Any plans to support other cases, or is this enough?

mdbooth:
> Do I understand correctly that the idea is that the server name will in most cases be the same as the node name, especially if migrating from the in-tree provider? Any plans to support other cases, or is this enough?

I'd expect node name to be the same as server name *only* in cases where it's been overridden in order to be compatible with the in-tree provider. Honestly, outside of OpenShift I'd expect this to be an unusual configuration now that the in-tree provider has been removed.

I expect the normal case to be where nodename is the same as hostname, which this doesn't support. The plan is to add a second InternalDNS entry containing hostname, but only when the server supports microversion 2.90. This may end up being one of the first consumers of your microversion patch.

So:
* Everybody gets an InternalDNS entry containing server name. This is a niche case, but it's also the easiest to implement. It's important to Red Hat because it's currently all OpenShift nodes.
* If you have microversion >= 2.90 you *also* get an InternalDNS entry containing hostname (tbd). This is the usual case, but it requires variable microversion support. We'd also like to implement this asap.

mdbooth (commented):
Thanks. As this just updates the comment I'm going to add back @lentzi90 's lgtm

/lgtm

MaysaMacedo:
/hold cancel

We tested it and it worked.
@mdbooth Thank you!

PR #1716, opened 2023-10-09
🌱 Bump devstack image to ubuntu 22.04

lentzi90:

We have been seeing issues with cloud-final failing while setting up the devstack. One issue could be that the image we use is just too old. Let's try something newer.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


mdbooth:
The tests will decide, so

/lgtm

lentzi90:
Amazing. The test passed just like that. I was sure we would need to fiddle with some config :sweat_smile: 
/hold cancel

Issue #1717, opened 2023-10-10
Old flags used in ensure-kubectl.sh


lentzi90:
/kind bug

**What steps did you take and what happened:**
A periodic job failed due to

```console
$ kubectl version --client --short
error: unknown flag: --short 
```

in [ensure-kubectl.sh](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ensure-kubectl.sh).

See: <https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-openstack-e2e-full-test-main/1711561098999631872>

**What did you expect to happen:**

The job should not fail.

**Anything else you would like to add:**

We just need to update the way we extract the version from `kubectl version`.

mdbooth:
Possible fix(es) in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1713

PR #1718, opened 2023-10-10
🌱 ci: Several CI fixes

mdbooth:
This PR contains several critical fixes, all of which are required for passing CI.

Firstly, fixes several issues which seem to stem from a bump in the base prow image.

Secondly, reverts the effect of #1716 by putting the devstack image back to Ubuntu 20.04. Devstack for OpenStack Yoga does not support Ubuntu 22.04.

Lastly, fixes a GCE provisioning issue which prevented #1716 from being tested properly. It looks like we could non-deterministically reuse an old disk image depending on which zone we landed in, meaning we were not necessarily testing the change in the PR under test.

Fixes: #1717 

/hold

### DIFF
--- a/hack/ci/create_devstack.sh
+++ b/hack/ci/create_devstack.sh
@@ -77,32 +77,26 @@ function retry {
 function ensure_openstack_client {
     if ! command -v openstack;
     then
-        # We are running in a Debian Buster image with python 3.7. Python 3.7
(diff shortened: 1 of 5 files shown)


lentzi90 (commented):
/lgtm

lentzi90 (commented):
/lgtm

PR #1723, opened 2023-10-12
🌱 Use k8s v1.28 in tests

lentzi90:
Bump Kubernetes to v1.28.2 for e2e tests and bump envtest Kubernetes
version to v1.28.0. Also removes the now unused USE_CI_ARTIFACTS
environment variable that was set in some places.


lentzi90:
I cannot find the issue. Rerunning to rule out flake...

lentzi90:
:facepalm: triggered wrong test...

lentzi90:
@tormath1 could there be something wrong with the flatcar image? I created it with image-builder in the same way as I did the ubuntu image. Any special settings needed?
I'm a bit lost on why it fails, but it has failed 2 times in the flatcar test. I will debug more myself also a bit later

lentzi90:
Thanks @tormath1 ! I have built a new image and uploaded. Let's see how it goes! :crossed_fingers: 

lentzi90:
Fantastic! It worked! :slightly_smiling_face: Thanks again for the help!

lentzi90:
Updated the table in the readme. I just added v1.28 to it and did not remove any of the older versions or check marks. My reasoning is that we have, for example, tested CAPO v1alpha6 with k8s v1.23 at some point so the check mark is still valid, even though we do not currently test it.

jichenjc:
>>> did not remove any of the older versions or check marks. My reasoning is that we have, for example, tested CAPO v1alpha6 with k8s v1.23 at some point so the check mark is still valid, even though we do not currently test it.

jichenjc:
>>> did not remove any of the older versions or check marks. My reasoning is that we have, for example, tested CAPO v1alpha6 with k8s v1.23 at some point so the check mark is still valid, even though we do not currently test it.

we can achieve those `older` version at some point later and keep newer enough info

Issue #1674, opened 2023-09-15
Remove FloatingIP from OpenStackMachineSpec


mdbooth:
/kind feature

FloatingIP is not used by the machine controller, only the Bastion. We should move the floating IP configuration to the Bastion spec (OpenStackCluster.Spec.Bastion) instead.

EmilienM:
I agree with Matt, it's only used in the Bastion, see https://github.com/shiftstack/cluster-api-provider-openstack/blob/070714eaa24a25ec6774cbae1c6e140eb8babebe/controllers/openstackcluster_controller.go#L362-L363

EmilienM:
/reopen

We are reverting. I'll re-add it to main once we have v1alpha8.

mdbooth:
@EmilienM We put this back, right?

PR #1725, opened 2023-10-12
✨ Enable Worker Nodes to Associate with Floating IPs

mikaelgron:

This PR aims to enable worker nodes to associate with pre-allocated Floating IPs in OpenStack. The use-case is primarily for clients who want to whitelist their worker nodes in external services. By specifying a set of Floating IPs during the cluster setup, worker nodes can be configured to reuse these IPs, thus avoiding the need to track new floating IPs.

> **Note:** This feature is currently in an experimental phase, and we are looking for feedback. It's not targeted to join alpha8 yet.

**Future Steps**

We're already contemplating further enhancements to add dynamicity to this feature. One idea is to allow the cluster CRD to dynamically create Floating IPs based on the number of machine deployments + X or something like that. While this isn't part of the current PR, it is a direction we're interested in exploring.

**Feedback Requested**
We are actively looking for feedback on this feature to understand its viability and possible improvements. Feel free to comment with your suggestions, questions, or concerns.

**TODOs**:
- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


controllers/openstackmachine_controller.go
```diff
@@ -426,6 +434,36 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+			}
+
+			port, err := computeService.GetManagementPort(openStackCluster, instanceStatus)
+			if err != nil {
+				conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.FloatingIPErrorReason, clusterv1.ConditionSeverityError, "Obtaining management port for worker machine failed: %v", err)
+				return ctrl.Result{}, fmt.Errorf("get management port for worker machine: %w", err)
+			}
+
+			if fp.PortID != "" {
+				scope.Logger().Info("Floating IP already associated to a port", "id", fp.ID, "fixedIP", fp.FixedIP, "portID", port.ID)
```
dulek, line 454: This feels like something happening all the time, it should have lower log level.

controllers/openstackmachine_controller.go
```diff
@@ -426,6 +434,36 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+			} else {
+				err = networkingService.AssociateFloatingIP(openStackMachine, fp, port.ID)
+				if err != nil {
+					conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.FloatingIPErrorReason, clusterv1.ConditionSeverityError, "Associating floating IP failed: %v", err)
+					return ctrl.Result{}, fmt.Errorf("associate floating IP %q to worker node with port %q: %w", fp.FloatingIP, port.ID, err)
+				}
+			}
+		}
```
dulek, line 462: We should log a situation when node haven't got any FIP because pool got exhausted.

controllers/openstackmachine_controller.go
```diff
@@ -426,6 +434,36 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 	}
 	conditions.MarkTrue(openStackMachine, infrav1.APIServerIngressReadyCondition)
 
+	// Apply floating ip on worker if enabled in openStackCluster spec
+	if !util.IsControlPlaneMachine(machine) && openStackCluster.Spec.WorkerFloatingIPConfig.Enabled {
+		scope.Logger().Info("Processing worker floating IPs")
+		for _, floatingIP := range openStackCluster.Spec.WorkerFloatingIPConfig.IPAddresses {
+			fp, err := networkingService.GetOrCreateFloatingIP(openStackMachine, openStackCluster, clusterName, floatingIP)
```
dulek, line 441: If FIP actually got created, where do we remove it on cluster deinstallation?

api/v1alpha7/types.go
```diff
@@ -29,6 +29,11 @@ type ExternalRouterIPParam struct {
 	Subnet SubnetFilter `json:"subnet"`
 }
 
+type WorkerFloatingIPConfig struct {
+	Enabled     bool     `json:"enabled,omitempty"`
```
dulek, line 33: Do we even need this flag? Could a `nil` or empty list work as a disabled state?

dulek (changes requested):

mdbooth (commented):
Let me start with a negative: I'm not a fan of the design here, particularly that this would apply to all workers without any obvious mechanism for restricting it. It also touches floating IPs in the machine controller, which is already an almighty mess. I actually opened https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1674 to track removing floatingIP from the machine spec because it can't be used correctly.

However, the use case you describe sounds interesting. I'd like to help you implement it.

This sounds like it could be an IPAM controller? I believe this was first implemented by metal3, so @lentzi90 may know more about it. How about we created an OpenStackFloatingIPPool. It could contain your pre-populated list of FIPs, or create them on demand. We could add floatingIP to PortOpts, except that instead of being an IP it would be a reference to an IPAM provider (in this case our OpenStackFloatingIPPool). The machine controller would create an IPClaim for the port and wait for an IPAddress to be allocated, at which point it could create the port.

The metal3 docs are here: https://book.metal3.io/ipam/introduction.html. There was talk of CAPI adopting it, and I understand CAPV are also using it.

controllers/openstackmachine_controller.go
mikaelgron: Agreed. It should perhaps be reduced on the APIServerLoadBalancer flow on line 426 as well.

controllers/openstackmachine_controller.go
mikaelgron: For the use case where you have to create IPs first to get the actual IP address strings to put in the spec, they would need to be removed manually anyway. It would be better to use a getter instead of a get-or-create in this case.

controllers/openstackmachine_controller.go
```diff
@@ -273,6 +273,14 @@ func (r *OpenStackMachineReconciler) reconcileDelete(scope scope.Scope, cluster
 		}
 	}
 
+	// If this is a worker node and it has a floating IP attached, disassociate it.
+	if !util.IsControlPlaneMachine(machine) && openStackMachine.Spec.FloatingIP != "" {
```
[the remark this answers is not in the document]
mikaelgron: True, and I believe it's unnecessary to disassociate FIP in code. Guessing that happens automatically when the worker disappears.

dulek:
Looks like something being implemented by #1763?

mdbooth:
Closed in favour of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1762

PR #1727, opened 2023-10-13
✨  Add env var to allow set different prefix/suffix for LB

### LINKED ISSUE #1719: Add option to customize resource prefixes
/kind feature

I want to be able to specify in the `OpenStackCluster` resource prefixes for the following resources which are now constants:
* network - [`networkPrefix string = "k8s-clusterapi"` ](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/cc304b8af1d5f637419273b0702a0311b3c17b01/pkg/cloud/services/networking/service.go#L32)
* router - uses network prefix
* subnet - uses network prefix
* ApiServerLoadBalancer - [`networkPrefix   string = "k8s-clusterapi"`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/cc304b8af1d5f637419273b0702a0311b3c17b01/pkg/cloud/services/loadbalancer/loadbalancer.go#L42C1-L45)
* SecurityGroup - [`secGroupPrefix     string = "k8s"`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/securitygroups.go#L30-L36)

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
kind: OpenStackCluster
metadata:
  name: <cluster-name>
  namespace: <cluster-namespace>
spec:
  ...
  network:
    prefix: myPrefix
  apiServerLoadBalancer:
    prefix: myPrefix
  securityGroups:
    prefix: myPrefix
```

* I want this to improve operator visibility into resource association with individual clusters when multiple clusters run inside a single OpenStack project.
* This feature request is a result of the following issue, which may give a more complete context of our use case: https://github.com/SovereignCloudStack/k8s-cluster-api-provider/issues/495

jichenjc:

Fixes #1719 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -19,6 +19,7 @@ package loadbalancer
 import (
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"time"
 
@@ -40,8 +41,10 @@ import (
 )
 
 const (
-	networkPrefix   string = "k8s-clusterapi"
-	kubeapiLBSuffix string = "kubeapi"
+	networkPrefixEnvVar   string = "LB_NETWORK_PREFIX"
+	networkPrefix         string = "k8s-clusterapi"
+	kubeapiLBSuffixEnvVar string = "LB_KUBEAPI_SUFFIX"
+	kubeapiLBSuffix       string = "kubeapi"
 )
 
 const loadBalancerProvisioningStatusActive = "ACTIVE"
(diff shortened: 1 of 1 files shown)


dulek:
Personally I don't understand why setting cluster name would be enough as it's already part of the LB name.

mdbooth:
> Personally I don't understand why setting cluster name would be enough as it's already part of the LB name.

I'm guessing when you've got independent management clusters (e.g. CI/per user) using the same OpenStack tenant? @jichenjc could you expand on the use case?

It doesn't seem unreasonable, though, and environment variables are more kustomize friendly than command line arguments.

Could you also please add some documentation? Definitely needs docs.

mdbooth:
After reading the issue, I agree with @dulek. Lets go back to the use case. I'm not sure I'm following it yet.

fdobrovolny:
Just a side note to the #1719:
it was meant not only for LB but also for networks, routers, subnets, and security groups. But the same solution can be used for others as well.

jichenjc:
>>>it was meant not only for LB but also for networks, routers, subnets, and security groups. But the same solution can be used for others as well.

 I think it's valid use case from the LB level after I read original issue 

>>>It doesn't seem unreasonable, though, and environment variables are more kustomize friendly than command line arguments.

um... this seems more reasonable

PR #1728, opened 2023-10-13
🐛 Don't remove BYO api server loadbalancer floating IP

### LINKED ISSUE #1615: Disable deletion of floating IP on cluster deletion
/kind feature

I would like to have the option of not deleting the floating IP when deleting the cluster.

I have a scenario in which the floating ip is generated a priori and provided as input to Cluster API cluster
creation mechanisms (and also to a cilium helm chart). I am still experimenting with Cluster API so I'm going
through workflows of creating clusters and removing them - things generally work fine and when the cluster
is created, the floating ip is attached to the load balancer and the cluster is accessible.

One specific issue that I have is that when the cluster is deleted, the floating IP is removed from my Openstack project - this means that every time I want to recreate the cluster, I need to generate a new floating ip and modify my config to use this new floating ip. It would be preferable if I could have some flag which could control whether the floating ip is deleted or not in the case that it is provided as an input.

We are a user of an Openstack service; we don't manage the Openstack cluster ourselves, so we prob
don't have authorizations to demand a specific floating ip all the time; in any case, removing the floating ip
from the project gives someone else the opportunity to use it and even if we did have higher level privileges,
we may not be easily able to take it from a user in the case that a user had taken this IP and attached it
to some resource.

alexandrevilain:

Fixes #1615

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1alpha7/openstackcluster_webhook.go
+++ b/api/v1alpha7/openstackcluster_webhook.go
@@ -134,6 +134,12 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) (admission.Warn
 	old.Spec.AllowAllInClusterTraffic = false
 	r.Spec.AllowAllInClusterTraffic = false
 
+	// Allow change on the spec.APIServerFloatingIP only if it matches the current api server loadbalancer IP.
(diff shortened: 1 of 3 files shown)


mdbooth (commented):
/approve

lentzi90 (commented):
/lgtm

PR #1731, opened 2023-10-20
🌱 Bump `controller-runtime` to `v0.16.2` and deps

EmilienM:
- [vendoring] bump `controller-runtime` to `v0.16.2`
- main: adjustments for new `controller-runtime`

bump `controller-runtime` to `v0.16.2` and apply adjustments to `main.go`.

**TODOs**:

- [X] squashed commits - we don't squash since it involves vendoring.
- if necessary:
  - [x] includes documentation - no doc update is needed.
  - [x] adds unit tests - no test updates are needed.


main.go
```diff
@@ -181,21 +185,31 @@ func main() {
...
-			&corev1.Secret{},
+		Scheme:           scheme,
+		Metrics:          metricsserver.Options{BindAddress: metricsBindAddr},
+		LeaderElection:   enableLeaderElection,
+		LeaderElectionID: "controller-leader-election-capo",
+		LeaseDuration:    &leaderElectionLeaseDuration,
+		RenewDeadline:    &leaderElectionRenewDeadline,
+		RetryPeriod:      &leaderElectionRetryPeriod,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{watchNamespace: {}},
```
seanschneeweiss, line 196: Does this work if `watchNamespace == ""`.
Might be good to adapt what CAPI does: https://github.com/kubernetes-sigs/cluster-api/blob/dc64556b0a19e11e14ea3ba28095d64547f1aa12/bootstrap/kubeadm/main.go#L197-L202

EmilienM: Very good suggestion, thanks!

go.mod
```diff
@@ -12,23 +12,23 @@ require (
...
+	k8s.io/klog/v2 v2.100.1
 	k8s.io/kubernetes v1.27.2
-	k8s.io/utils v0.0.0-20230313181309-38a27ef9d749
-	sigs.k8s.io/cluster-api v1.5.1
-	sigs.k8s.io/cluster-api/test v1.5.1
-	sigs.k8s.io/controller-runtime v0.15.1
+	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2
+	sigs.k8s.io/cluster-api v1.5.2
+	sigs.k8s.io/cluster-api/test v1.5.2
+	sigs.k8s.io/controller-runtime v0.16.2
```
EmilienM, line 31: https://github.com/kubernetes-sigs/cluster-api/blob/v1.5.2/go.mod#L46

CAPI 1.5.2 is stillon `sigs.k8s.io/controller-runtime v0.15.1` so let's not bump it.

EmilienM:
/hold
I'll come back to it once we have a new CAPI release.

go.mod
pierreprinetti, line 31: Please make sure you bump controller-runtime to `v0.16.3`+ to get the RapidReset mitigation. This is what CAPI v1.6 will very likely get anyway: https://github.com/kubernetes-sigs/cluster-api/blob/dd553a2e82b4d76f6f9bf1f0f66b37fd1d640416/go.mod#L49

EmilienM: ack, I'll keep that in mind when we come back to this but first we need a new CAPI release.

EmilienM: I'll need to update this PR once CAPI V1.6.0, end of November.
See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1734

EmilienM:
I'm trying to bump Golang into a separate PR if possible, so we separate problems and fix them one by one: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1740

EmilienM:
This is working, cool. We are now waiting for the final CAPI 1.6.0 release (end of November), then I'll update the vendoring but the other commit should be an easy update.

lentzi90 (commented):
Looks great!
Waiting with LGTM until the release is finalized

main.go
```diff
@@ -143,6 +145,10 @@ func InitFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&showVersion, "version", false, "Show current version and exit.")
 }
 
+// Add RBAC for the authorized diagnostics endpoint.
+// +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
```
mdbooth, line 150: I don't see this new RBAC showing up in the generated config. Have I missed it?

Do you have a link to any documentation describing how this is used, btw?

EmilienM: I stole it from https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4182/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R244-R245

And I think it comes from: https://github.com/kubernetes-sigs/cluster-api/pull/9264/files#diff-5f809e136425aaf67e69338ebe143aa06634a81c0632ddaed85894129029d85bR21-R32

EmilienM: @mboersma maybe you have more context than I do to answer Matt's question. Thanks!

mboersma: It's probably not showing up in `rbac.yaml` because the `make generate-manifests` target isn't looking at `main.go`. 

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ca2516072b453fa4cefbb60452e3343e69317929/Makefile#L274-L281

Adding `paths=./ \` like [CAPI has](https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L281) should fix that. (I had made a similar change to CAPZ but must have lost it in refactoring; I'm glad you pointed this out.)

The docs on this feature in CAPI are here as part of PR 9264: https://github.com/kubernetes-sigs/cluster-api/pull/9264/files#diff-c60ebbe9702722c1666539c3ca1c30aaeafa37e5d420a62130752da313cf1156, not in the CAPI Book yet I assume because this isn't yet in a released version.

EmilienM: Thanks a lot @mboersma ! :-)

config/manager/manager.yaml
```diff
@@ -22,14 +22,18 @@ spec:
...
           image: controller:latest
           imagePullPolicy: Always
           name: manager
           ports:
             - containerPort: 9440
               name: healthz
               protocol: TCP
+            - containerPort: 8443
+              name: metrics
+              protocol: TCP
```
mdbooth, line 36: Hah, we were never exposing port 8080 anyway!

config/manager/manager.yaml
```diff
@@ -22,14 +22,18 @@ spec:
           args:
             - "--leader-elect"
             - "--v=2"
-            - "--metrics-bind-addr=127.0.0.1:8080"
+            - "--diagnostics-address=${CAPO_DIAGNOSTICS_ADDRESS:=:8443}"
+            - "--insecure-diagnostics=${CAPO_INSECURE_DIAGNOSTICS:=false}"
```
mdbooth, line 26: Until we've sorted out how we want to expose secure metrics, please can we put this back to the old, insecure. Also I'd prefer not to add new cluster api variables for substitution in this change. Lets do that separately and deliberately when we understand this feature.

For now we could either:
* Leave it as it is. `--metrics-bind-addr` is deprecated but still works and has the same behaviour as before.
* Change it to `--diagnostics-address=localhost:8080 --insecure-diagnostics=true`, which keeps the same behaviour but without using the deprecated flag.

My feeling is that the prudent option is to entire revert the change to this file, leaving it exactly as it was before, and open an issue to track adding support for secure metrics.

mdbooth:
Given that `--metrics-bind-addr` is still supported I think this doesn't have to be a breaking change. However, we're changing the default behaviour in the templates which *is* a breaking change. We should revert that part of the change.

I think if we revert the change to `config/manager/manager.yaml` this is no longer breaking change.

I pre-emptively opened an issue to track adding it properly: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1777

PR #1732, opened 2023-10-23
🌱 Enable k8s upgrade in self hosted test

lentzi90:

This enables the k8s upgrade that is part of the self hosted test. It was previously disabled since the dynamic installation of kubernetes components on the nodes made it hard to change the version in the middle of a test.

Also updated the README tables for what versions are tested.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


README.md
```diff
@@ -48,25 +48,28 @@ This provider's versions are compatible with the following versions of Cluster A
...
-| OpenStack Provider v1alpha7        |        | +     | +     | +     | +      | ✓        | ✓       | ✓    | ✓    |
+| OpenStack Provider v1alpha5 (v0.6) | +      | +     | +     | +     | +      | ✓        | ✓       | ✓    | ★    |
+| OpenStack Provider v1alpha6 (v0.7) | +      | +     | +     | +     | +      | ✓        | ✓       | ✓    | ★    |
+| OpenStack Provider v1alpha7        |        | +     | +     | +     | +      | ✓        | ✓       | ✓    | ★    |
 
 Test status:
 
-- `✓` tested
+- `★` currently testing
+- `✓` previously tested
```
jichenjc, line 68: this is very great update

README.md
```diff
@@ -48,25 +48,28 @@ This provider's versions are compatible with the following versions of Cluster A
...
-|                                    | v1.22 | v1.23 | v1.24 | v1.25 | v1.26 | v1.27 | v1.28 |
-|------------------------------------| ----- | ----- | ----- | ----- | ----- | ----- | ----- |
-| OpenStack Provider v1alpha5 (v0.6) | +     | ✓     | ✓     | ✓     | +     | +     | +     |
-| OpenStack Provider v1alpha6 (v0.7) | +     | ✓     | ✓     | ✓     | ✓     | ✓     | +     |
-| OpenStack Provider v1alpha7        | +     | +     | +     | +     | ✓     | ✓     | ✓     |
+|                                    | v1.25 | v1.26 | v1.27 | v1.28 |
+|------------------------------------| ----- | ----- | ----- | ----- |
+| OpenStack Provider v1alpha5 (v0.6) | ✓     | +     | +     | +     |
+| OpenStack Provider v1alpha6 (v0.7) | ✓     | ✓     | ✓     | +     |
+| OpenStack Provider v1alpha7        | +     | ✓     | ★     | ★     |
```
jichenjc, line 55: maybe a little picky, but 1.27 seems not testing 
we only test whether 1.28 be upgraded from 1.27?

lentzi90: Yes I was wondering how should I handle this. 1.28 is much more tested but we do create one cluster with 1.27 and then upgrade to 1.28. Maybe I should put ✓ instead?

README.md
lentzi90: All credit goes to @wwentland :blush:

README.md
jichenjc: yes, I prefer this way..

test/e2e/data/kustomize/k8s-upgrade/kustomization.yaml
lentzi90, line 1: This is a bit silly, but allow me to explain.
We need a separate OpenStackMachineTemplate with a different k8s version in order to do the upgrade. The test spec itself is making the switch from the initial template to the "upgrade to" template, that is specified in the e2e config.
Our normal template is already using k8s v1.28.2 so here I add a template witch v1.27.2 to get a different initial version. 

The silly part is that we cannot upgrade to the "normal" template because it gets a dynamic name when applied in the test. All clusters are created with a random suffix and this propagates also to the template, so it is impossible to specify this template in the e2e config. We could change the naming scheme of the template to work around this. Let me know if you think this would be better. What I did instead was to just add another statically named template, so we end up with 2 extra templates for this test.

test/e2e/data/kustomize/k8s-upgrade/upgrade-to-template.yaml
```diff
@@ -0,0 +1,43 @@
...
+# The normal templates has the cluster name as part of the OSMT name.
+# See CONTROL_PLANE_MACHINE_TEMPLATE_UPGRADE_TO and WORKERS_MACHINE_TEMPLATE_UPGRADE_TO
+# in the e2e config.
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
+kind: OpenStackMachineTemplate
+metadata:
+  name: upgrade-to-control-plane
+  labels:
+    clusterctl.cluster.x-k8s.io/move: ""
```
lentzi90, line 12: This label is important to get the (not yet in use) template moved to the self hosted cluster. Otherwise it would be left behind since it is not part of the owner chain at that point.

jichenjc: yes, I think it's something I tried to add into CAPO before, IIRC it's moving stuffs from boostrap to deployed cluster?

lentzi90: Exactly. To be clear, it is only needed on this template because it is not part of the cluster owner chain. All resources that are part of the cluster are moved automatically. For example, the OpenStackMachineTemplates that are used when initializing the workload cluster are automatically moved.

jichenjc:
/lgtm

leave unhold to someone who might want to take another look

PR #1735, opened 2023-10-26
✨ Use kustomize new syntax for patches

### LINKED ISSUE #1730: templates: 'patchesStrategicMerge' is deprecated
/kind feature

While running `make templates`, we can see the following output:
```
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

Can be observed in the CI output too. 

This comes from the new Kustomize version being used: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.0
> We are deprecating the following kustomization fields. These fields will never be removed from the kustomize.config.k8s.io/v1beta1 Kustomization API, but they will not be included in the kustomize.config.k8s.io/v1 Kustomization API. When Kustomization v1 is available, we will announce the deprecation of the v1beta1 version. There will be at least two releases between deprecation and removal of Kustomization v1beta1 support from the kustomize CLI, and removal itself will happen in a future major version bump:
> * patchesStrategicMerge. Please use patches instead

mquhuy:
Fixes #1730 

`Kustomize` version needs to be updated to 5.2.1, as the current version (5.1.1) gets confused with multiple objects in the same patch file (it only understands if the patch file is split so that each object is in one file).

I also updated the kustomize version in `ensure_kustomize.sh`. Not sure if it's in use though, as I can see that the version is already very old.


kustomize/v1alpha7/flatcar/kustomization.yaml
```diff
@@ -1,5 +1,7 @@
 resources:
 - ../default
 
-patchesStrategicMerge:
-- patch-flatcar.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
lentzi90, line 5: It is not technically anything wrong with having the `apiVersion` and `kind` in the middle of the file, but I would prefer that we keep it at the top (if specified at all). Now it looks a bit like it has something to do only with the `patches` :smile:

kustomize/v1alpha7/default/kustomization.yaml
```diff
@@ -1,2 +1,4 @@
 resources:
 - cluster-template.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
lentzi90, line 4:
```suggestion
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- cluster-template.yaml
```

kustomize/v1alpha7/without-lb/kustomization.yaml
```diff
@@ -1,5 +1,7 @@
 resources:
 - ../default
 
-patchesStrategicMerge:
-- patch-without-lb.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
lentzi90, line 5: Same here

lentzi90 (commented):
Looks good! Could you move the `apiVersion` and `kind` to the top of the files?

kustomize/v1alpha7/without-lb/kustomization.yaml
mquhuy: Sure. I'll do that.

kustomize/v1alpha7/default/kustomization.yaml
mquhuy: Sure

lentzi90 (commented):
Great!
I think we can also update the same in [v1alpha6](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/main/kustomize/v1alpha6), [v1alpha5](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/main/kustomize/v1alpha5) and [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/data/kustomize/md-remediation/kustomization.yaml). Then I think we got rid of all the old syntax :slightly_smiling_face:

lentzi90 (commented):
/approve

lentzi90:
Triggering full e2e since this touches the remediation tests

lentzi90:
It is unclear to me if the test failure is related to the PR

PR #1738, opened 2023-10-30
🐛 loadbalancer: resolve ControlPlaneEndpoint.Host when needed

### LINKED ISSUE #1714: We assume ControlPlaneEndpoint.Host is IP address
/kind bug

From code inspection, we set fixedIPAddress to ControlPlaneEndpoint.Host if not using APIServer FloatingIP: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/62d9d76feaf0cf876e52138cde8c52d1ceb8e00b/pkg/cloud/services/loadbalancer/loadbalancer.go#L54-L59

We then pass that to getOrCreateLoadBalancer, which expects it to be a VipAddress: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/62d9d76feaf0cf876e52138cde8c52d1ceb8e00b/pkg/cloud/services/loadbalancer/loadbalancer.go#L174-L182

However, ControlPlaneEndpoint.Host is not required to be an IP address. It may be a DNS name.

Same when creating the load balancer FIP: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/62d9d76feaf0cf876e52138cde8c52d1ceb8e00b/pkg/cloud/services/loadbalancer/loadbalancer.go#L92-L97

EmilienM:

`ControlPlaneEndpoint.Host` is not guaranteed to be an IP address, it
can also be an hostname.

Now we'll try to lookup the hostname if it's not an IP and set that for
the LB VipAddress.

Fixes #1714

Unit tests are tricky. We stub the call to net.LookupHost to avoid actual DNS lookups.

- [X] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -17,8 +17,10 @@ limitations under the License.
 package loadbalancer
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"net"
 	"slices"
 	"time"
 
@@ -27,7 +29,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/utils/net"
+	utilsnet "k8s.io/utils/net"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 
@@ -42,10 +44,28 @@ import (
 const (
 	networkPrefix   string = "k8s-clusterapi"
 	kubeapiLBSuffix string = "kubeapi"
+	resolvedMsg     string = "ControlPlaneEndpoint.Host is not an IP address, using the first resolved IP address"
 )
 
 const loadBalancerProvisioningStatusActive = "ACTIVE"
 
+// We wrap the LookupHost function in a variable to allow overriding it in unit tests.
+//
+//nolint:gocritic
+var lookupHost = func(host string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
+	defer cancel()
+	ips, err := net.DefaultResolver.LookupHost(ctx, host)
+	if err != nil {
+		return "", err
+	}
+	if ip := net.ParseIP(ips[0]); ip == nil {
+		return "", fmt.Errorf("failed to resolve IP address for host %s", host)
+	}
+	return ips[0], nil
+}
+
+// ReconcileLoadBalancer reconciles the load balancer for the given cluster.
 func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string, apiServerPort int) (bool, error) {
 	loadBalancerName := getLoadBalancerName(clusterName)
 	s.scope.Logger().Info("Reconciling load balancer", "name", loadBalancerName)
@@ -57,13 +77,18 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	}
 
 	var fixedIPAddress string
+	var err error
+
 	switch {
 	case lbStatus.InternalIP != "":
 		fixedIPAddress = lbStatus.InternalIP
 	case openStackCluster.Spec.APIServerFixedIP != "":
 		fixedIPAddress = openStackCluster.Spec.APIServerFixedIP
 	case openStackCluster.Spec.DisableAPIServerFloatingIP && openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-		fixedIPAddress = openStackCluster.Spec.ControlPlaneEndpoint.Host
+		fixedIPAddress, err = lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
+		if err != nil {
+			return false, fmt.Errorf("lookup host: %w", err)
+		}
 	}
 
 	providers, err := s.loadbalancerClient.ListLoadBalancerProviders()
@@ -108,7 +133,10 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 		case openStackCluster.Spec.APIServerFloatingIP != "":
 			floatingIPAddress = openStackCluster.Spec.APIServerFloatingIP
 		case openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-			floatingIPAddress = openStackCluster.Spec.ControlPlaneEndpoint.Host
+			floatingIPAddress, err = lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
+			if err != nil {
+				return false, fmt.Errorf("lookup host: %w", err)
+			}
 		}
 		fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterName, floatingIPAddress)
 		if err != nil {
@@ -308,9 +336,9 @@ func validateIPs(openStackCluster *infrav1.OpenStackCluster, definedCIDRs []stri
 
 	for _, v := range definedCIDRs {
 		switch {
-		case net.IsIPv4String(v):
+		case utilsnet.IsIPv4String(v):
 			marshaledCIDRs = append(marshaledCIDRs, v+"/32")
-		case net.IsIPv4CIDRString(v):
+		case utilsnet.IsIPv4CIDRString(v):
 			marshaledCIDRs = append(marshaledCIDRs, v)
 		default:
 			record.Warnf(openStackCluster, "FailedIPAddressValidation", "%s is not a valid IPv4 nor CIDR address and will not get applied to allowed_cidrs", v)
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -46,16 +47,29 @@ const (
 
 const loadBalancerProvisioningStatusActive = "ACTIVE"
 
+var lookupHost = func(host string) (addrs []string, err error) {
+	if net.ParseIP(host) != nil {
+		return []string{host}, nil
+	}
```
jichenjc, line 53: is it worthy log about the change we are implicitly change from DNS name to IP here?

EmilienM: good point, I'm not sure I added into the right place, see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1738/files#diff-29cb3ac9349ca04685d8b48d0b1bc13a90e76242d0300fea7ed14f4f71ef2eecR74

EmilienM:
```
 pkg/cloud/services/loadbalancer/loadbalancer_test.go:32: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
	. "github.com/onsi/gomega"
pkg/cloud/services/loadbalancer/loadbalancer_test.go:37: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1" 
```

I'm not sure why I have this error.

EmilienM:
Thanks for the review @jichenjc - I'll also tag @mdbooth since he reported the issue, I want to make sure he's fine too with the proposal.

pkg/cloud/services/loadbalancer/loadbalancer.go
seanschneeweiss: We could directly call `net.LookupHost` as it either returns the IP if it is an IP already or resolves the host otherwise. https://pkg.go.dev/net#LookupHost 

https://go.dev/play/p/06Zss3v3dnI

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -46,16 +47,32 @@ const (
...
+	var err error
+
 	switch {
 	case openStackCluster.Spec.APIServerFixedIP != "":
 		fixedIPAddress = openStackCluster.Spec.APIServerFixedIP
 	case openStackCluster.Spec.DisableAPIServerFloatingIP && openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-		fixedIPAddress = openStackCluster.Spec.ControlPlaneEndpoint.Host
+		ips, err := lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
+		if err != nil {
+			return false, err
```
seanschneeweiss, line 70: Might want to wrap the error here
```suggestion
			return false, fmt.Errorf("lookup host: %w", err)
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -46,16 +47,32 @@ const (
...
 	case openStackCluster.Spec.DisableAPIServerFloatingIP && openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-		fixedIPAddress = openStackCluster.Spec.ControlPlaneEndpoint.Host
+		ips, err := lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
+		if err != nil {
+			return false, fmt.Errorf("lookup host: %w", err)
+		}
+		fixedIPAddress = ips[0]
+		if net.ParseIP(fixedIPAddress) == nil {
+			s.scope.Logger().Info("ControlPlaneEndpoint.Host is not an IP address, using the first resolved IP address", "host", openStackCluster.Spec.ControlPlaneEndpoint.Host, "ip", fixedIPAddress)
+		}
```
mdbooth, line 75: This code block feels slightly too long to be duplicated below. Could we DRY this somehow?

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -46,16 +47,32 @@ const (
 
 const loadBalancerProvisioningStatusActive = "ACTIVE"
 
+// We wrap the net.LookupHost function in a variable to allow overriding it in unit tests.
```
mdbooth, line 50: Eww, but I don't personally know of a cleaner way to do this. I'll let it pass, but if anybody else knows better... @pierreprinetti maybe?

pkg/cloud/services/loadbalancer/loadbalancer.go
EmilienM: I've put it into a constant. Let me know if that works.

EmilienM:
@mdbooth are we good now or do you expect me to change something else?
Thanks

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -42,20 +43,37 @@ import (
...
+	resolvedMsg     string = "ControlPlaneEndpoint.Host is not an IP address, using the first resolved IP address"
 )
 
 const loadBalancerProvisioningStatusActive = "ACTIVE"
 
+// We wrap the net.LookupHost function in a variable to allow overriding it in unit tests.
+//
+//nolint:gocritic
+var lookupHost = func(host string) ([]string, error) {
+	return net.LookupHost(host)
```
mdbooth, line 55: A few things here:

This should probably use Resolver.LookupHost() as documented in net.LookupHost(), but we can't because we don't currently pass the context into this ReconcileLoadBalancer. We should most likely also do that.

I seem to recall we have a forcing function coming soon which will require us to use Context everywhere. So we don't miss this one, which isn't obvious without reading the godocs, could we update it to use Resolver.LookupHost anyway and create a context explicitly to pass in?

Secondly, does this have the correct behaviour when given an IP address? Either way, I wonder if we should check if it's an IP address before calling the resolver and just return it.

pkg/cloud/services/loadbalancer/loadbalancer_test.go
```diff
@@ -38,9 +41,24 @@ func Test_ReconcileLoadBalancer(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
+	// Stub the call to net.LookupHost
+	lookupHost = func(host string) (addrs []string, err error) {
+		if net.ParseIP(host) != nil {
```
mdbooth, line 46: I wonder if we should do this specifically in the non-test function.

pkg/cloud/services/loadbalancer/loadbalancer.go
EmilienM: 1. Using `net.DefaultResolver.LookupHost`: I agree, consider it done in the next patchset.


2. Does this have the correct behaviour when given an IP address? yes:

```golang
// LookupHost looks up the given host using the local resolver.
// It returns a slice of that host's addresses.
func (r *Resolver) LookupHost(ctx context.Context, host string) (addrs []string, err error) {
	// Make sure that no matter what we do later, host=="" is rejected.
	if host == "" {
		return nil, &DNSError{Err: errNoSuchHost.Error(), Name: host, IsNotFound: true}
	}
	if _, err := netip.ParseAddr(host); err == nil {             <----------- HERE
		return []string{host}, nil
	}
	return r.lookupHost(ctx, host)
}
```

pkg/cloud/services/loadbalancer/loadbalancer_test.go
EmilienM: I don't think we need as I pointed out above.

pkg/cloud/services/loadbalancer/loadbalancer.go
pierreprinetti: The alternative that pops to mind is dependency injection: If `ReconcileLoadBalancer` can accept a resolver as an additional argument, I'd inject the resolver from the controller. This `lookupHost` woud then be defined inside the controller, rather than being a global var (which can be dangerous). However, my solution doesn't work if you want to unit-test the controller this way.

If that's what you're looking for:
```Go
type Resolver func(string) ([]string, error)

func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string, apiServerPort int, lookupHost Resolver) (bool, error) {
```
and in `openstackcluster_controller.go`:
```Go
		resolver := func(host string) ([]string, error) {
			ctx, cancel := context.WithTimeout(context.TODO(), time.Second)
			defer cancel()
			return net.DefaultResolver.LookupHost(ctx, host)
		}
		terminalFailure, err := loadBalancerService.ReconcileLoadBalancer(openStackCluster, clusterName, apiServerPort, resolver)

```

However, my suggestion is not really a game changer.

EmilienM: Thanks Pierre for your analysis. I think the PR is then ready.

EmilienM:
can someone review it again please?

mdbooth (commented):
/approve

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -42,10 +44,23 @@ import (
...
+//nolint:gocritic
+var lookupHost = func(host string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
+	defer cancel()
+	return net.DefaultResolver.LookupHost(ctx, host)
+}
+
+// ReconcileLoadBalancer reconciles the load balancer for the given cluster.
+//
+//nolint:gocyclo // Function requires a lot of conditions to be checked
```
EmilienM, line 63: I'll remove that nolint and do some painful refactoring.

Issue #1612, opened 2023-07-18
Randomly deleted port for VMs


mnaser:
/kind bug

**What steps did you take and what happened:**
At some point, a VM will simply get it's interface deleted, nothing gets logged anywhere.  Access logs confirm that it was CAPO responsbile behind it.

**What did you expect to happen:**
CAPO to not delete logs.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built):
- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

mnaser:
I believe I am into the root cause, I am seeing an OpenStack API call which is listing ports with the name of the instance and the network ID.  

The only reference to this is here:

https://github.com/mnaser/cluster-api-provider-openstack/blob/95ef1e3f511a09dc693b79aaf9aec5840e14a1c5/pkg/cloud/services/networking/port.go#L57-L60

Now, the only place this function gets called from is here:

https://github.com/mnaser/cluster-api-provider-openstack/blob/95ef1e3f511a09dc693b79aaf9aec5840e14a1c5/pkg/cloud/services/compute/instance.go#L196-L197

and within the same exact function.. we have this:

https://github.com/mnaser/cluster-api-provider-openstack/blob/95ef1e3f511a09dc693b79aaf9aec5840e14a1c5/pkg/cloud/services/compute/instance.go#L160-L169

Now the only way we are in that function is if `CreateInstance` is called, which is called here:

https://github.com/mnaser/cluster-api-provider-openstack/blob/95ef1e3f511a09dc693b79aaf9aec5840e14a1c5/controllers/openstackmachine_controller.go#L437-L438

It seems that it is for some reason failing to get the instance status.. trying to create a new server with the same name, failing somewhere after `GetOrCreatePort` successfully returns the list of ports and getting appended then being deleted because of the `defer`'d function...

I _believe_ that's how we've ended up here.. now why did the instance status not be correct.. looking

mnaser:
Another proof for this, I see a request to `/flavors/detail` at the same timeline which is part of `getAndValidateFlavor`.

mnaser:
bingo.

```
[...]

Issue #1632, opened 2023-08-02
Floating IPs can leak if they cannot be attached


mnaser:
/kind bug

**What steps did you take and what happened:**
If you have a scenario where you create a cluster in a network that doesn't have a router (or a router that has no path to the external network) .. then it will keep looping on every reconcile and allocate a new floating ip but fail to attach it without cleaning up..

**What did you expect to happen:**
It shouldn't keep leaking them non-stop

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built):
- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

mdbooth:
I think this was fixed by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1829

PR #1753, opened 2023-11-22
🐛 Fix random instance port deletion

zioc:

Under certain circumstances, when they are non-responsive cells, nova may return a partial result instead of an error.

(this is controlled by list_records_by_skipping_down_cells parameter that is true by default - no error will be thrown)

When it faces this issue, capo controller will try to create a new server for the machine, resulting in deleting ports of existing one.

In order avoid this issue, use GET instead of LIST to retrieve server using its uuid. For that purpose we must ensure that uuid is stored in machine spec as soon as the server is created, even if server is in error state.

Fixes: #1612 

Yes

**TODOs**:

- [ X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1alpha5/types.go
+++ b/api/v1alpha5/types.go
@@ -273,8 +273,8 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
+	// InstanceStateBuild is the string representing an instance in a build state.
+	InstanceStateBuild = InstanceState("BUILD")
 
 	// InstanceStateActive is the string representing an instance in an active state.
 	InstanceStateActive = InstanceState("ACTIVE")
@@ -290,6 +290,9 @@ var (
 
 	// InstanceStateDeleted is the string representing an instance in a deleted state.
 	InstanceStateDeleted = InstanceState("DELETED")
+
+	// InstanceStateUndefined is the string representing an undefined instance state.
+	InstanceStateUndefined = InstanceState("")
 )
 
 // Bastion represents basic information about the bastion node.
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -215,10 +215,24 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
 		return err
 	}
 
-	instanceName := fmt.Sprintf("%s-bastion", cluster.Name)
-	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, instanceName)
-	if err != nil {
-		return err
+	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		if err = networkingService.DeleteFloatingIP(openStackCluster, openStackCluster.Status.Bastion.FloatingIP); err != nil {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
+			return fmt.Errorf("failed to delete floating IP: %w", err)
+		}
+	}
+
+	var instanceStatus *compute.InstanceStatus
+	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.ID != "" {
+		instanceStatus, err = computeService.GetInstanceStatus(openStackCluster.Status.Bastion.ID)
+		if err != nil {
+			return err
+		}
+	} else {
+		instanceStatus, err = computeService.GetInstanceStatusByName(openStackCluster, bastionName(cluster.Name))
+		if err != nil {
+			return err
+		}
 	}
 
 	if instanceStatus != nil {
@@ -230,6 +244,7 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
 
 		for _, address := range addresses {
 			if address.Type == corev1.NodeExternalIP {
+				// Floating IP may not have properly saved in bastion status (thus not deleted above), delete any remaining floating IP
 				if err = networkingService.DeleteFloatingIP(openStackCluster, address.Address); err != nil {
 					handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
 					return fmt.Errorf("failed to delete floating IP: %w", err)
@@ -277,8 +292,9 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
 		return reconcile.Result{}, err
 	}
 
-	if err = reconcileBastion(scope, cluster, openStackCluster); err != nil {
-		return reconcile.Result{}, err
+	result, err := reconcileBastion(scope, cluster, openStackCluster)
+	if err != nil || !reflect.DeepEqual(result, reconcile.Result{}) {
+		return result, err
 	}
 
 	availabilityZones, err := computeService.GetAvailabilityZones()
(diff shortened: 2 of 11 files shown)


jichenjc:
@mdbooth maybe need your help on this to see fit for your original idea

mdbooth (commented):
This looks great: I'm very much in favour of this!

The e2e test failure looks real, though. Something odd is going on with Bastion creation. It looks like we end up creating bastions continuously until devstack runs out of quota.

To take a look, have a look at the e2e job failure in the PR. At the top you'll see 'Artifacts' which link to everything we dump out of the test run. Some interesting logs:

* [CAPO logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1753/pull-cluster-api-provider-openstack-e2e-test/1728006436866232320/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-m2pff/manager.log)
* [Events for cluster e2e-j9dl6w](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1753/pull-cluster-api-provider-openstack-e2e-test/1728006436866232320/artifacts/clusters/bootstrap/resources/e2e-j9dl6w/events.log)

There's one other factor to consider here from a design POV: we need to handle the case where we fail to write the machine/bastion ID to the status. It turns out it is surprisingly easy for this to happen due to the way we patch objects, but we should also consider that brief outages of the k8s control plane are expected from time to time.

My solution to this to date has been that if the ID is not stored we *always* check by name, and adopt any resource we find. This solution has its own edge cases, but I think it's a lesser evil.

One more thing, and I mention this only because it looks like you know what you're doing and appreciate clean controllers. Feel free to ignore this entirely if the scope creep is too much:

I would *really* like to move machine creation out of the cluster controller, but we haven't gotten around to it yet. Instead of calling directly into CreateInstance, I would like to see the cluster controller create a Machine/OpenStackMachine for the Bastion and watch it. That would eliminate a bunch of cruft and duplication, and enable more robustification of the machine cleanup flow.

Again, only if you're up for it. It would make this cleanup simpler, but I'm pretty sure the total work involved would be greater.

zioc:
Thanks a lot for the feedback and suggestions @mdbooth , as well as tips to understand e2e failures, they were very helpful.
                                                                                
It is indeed much safer to try to get instance by name when there is no ID saved in resource, I've added that fallback in various places.
                                                                                
While reworking the reconcileBastion method to fix the observed issue, I figured out that current implementation is not really reconcilling bastion floating IP as it exits [as soon as instance hash has been saved in annotations](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/141013b0e89b1110d422e3f6af01fa562c936ebc/controllers/openstackcluster_controller.go#L334C7-L334C28).

With current implementation, if floating IP fails to be attached to the instance (it can be the case if we provide a cluster network without router), it will be creating floating IPs in loop (similarly to what has been described in [this issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1632).
                                                                                
I tried to revisit that part to make it more robust and save allocated floatingIP in cluster status. The proposed patch is functional in most cases, but there is a race condition that sometimes leads to the creation of two floating IPs. It is very likely to be caused by "the way we patch objects" as you said: when bastion is created, its [object is patched before the status subresource](https://github.com/kubernetes-sigs/cluster-api/blob/0970c323d8ebf247bbe1c124e7e94dc4ad1ac35f/util/patch/patch.go#L143-L144) as we are adding the bastionHash annotation. Consequently cluster may reconcile without updated status and create a new floatingIP.
Unfortunately, as floatingIPs objects don't have names, we can't check by name as a fallback as we do for Instances (we could use description for that purpose, but it is not very clean...).
Other option would be to patch resources status as soon as we want to store some identifier (like instance IDs or floating IPs), this way we would have less chances to face this kind of race conditions.
Last option (and probably the most reasonable one) will be to avoid revisiting the floating IP attachment part (it probably deserves its own PR to keep this one simple), but I felt it was worth sharing [this patch](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/d2070463c2d17d81b44b0f3fe43eba4cd1910553) anyway for the shake of discussion.

Regarding your suggestion to use Machine/OpenstackMachine for bastion, it seems very interesting at first glance, but there are several points that remains unclear to me:
                                                                                
- I'm not sure if it's very clean to have a machine that does not belong to any controlplane or machine deployment from a capi controller perspective, as it expects to find a corresponding node in kubernetes at some point.
- As a large part of the code in reconcile/deleteBastion is handling floatingIP attachment and status update, we would still have to handle that part even if we were using an openstackMachine to control the openstack instance. I'm wondering if we would be sharing that much code at the end (but I may be missing something)

pkg/cloud/services/compute/instance_types.go
```diff
@@ -116,7 +116,6 @@ func (is *InstanceStatus) BastionStatus(openStackCluster *infrav1.OpenStackClust
 
 	clusterNetwork := openStackCluster.Status.Network.Name
 	i.IP = ns.IP(clusterNetwork)
-	i.FloatingIP = ns.FloatingIP(clusterNetwork)
```
zioc, line 119: Here we were persisting the floating IP in status only if it was attached to the bastion VM. In order to properly track the created floating IP, it is now [saved independently in reconcileBastion](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/398e4439f4a82649e1c6e4232d08c275e4b2c189/controllers/openstackcluster_controller.go#L421)

api/v1alpha5/types.go
```diff
@@ -273,8 +273,8 @@ func (r SecurityGroupRule) Equal(x SecurityGroupRule) bool {
 type InstanceState string
 
 var (
-	// InstanceStateBuilding is the string representing an instance in a building state.
-	InstanceStateBuilding = InstanceState("BUILDING")
+	// InstanceStateBuild is the string representing an instance in a build state.
+	InstanceStateBuild = InstanceState("BUILD")
```
zioc, line 277: While it is being created Nova server status [is BUILD and not BUILDING](https://docs.openstack.org/api-guide/compute/server_concepts.html#server-status). As this variable was not used elsewhere I felt it was ok to change it, please let me know it you foresee any issue to do that.

zioc (commented):
Finally I couldn't manage to avoid an important refactoring of reconcileBastion part. The race condition I was referring to in my previous message is fixed by checking if bastion already has a floating IP attached. (The race may still occur if floating IP failed to be attached, so we may create 2 floating IP in some cases, but it's already much better than current code that will create as many floating IP as it can if it fails to attach them for some reason).

api/v1alpha5/types.go
zioc: VM state seems to be [derived from status](https://github.com/openstack/nova/blob/d4f34ab8d8ba255e2514b88f22d49f9968d22e16/nova/api/openstack/common.py#L115), and the correspondance is not staightforward, for BUILDING state, we have a [BUILD status](https://github.com/openstack/nova/blob/d4f34ab8d8ba255e2514b88f22d49f9968d22e16/nova/api/openstack/common.py#L65)

controllers/openstackcluster_controller.go
```diff
@@ -215,10 +215,26 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
 		return err
 	}
 
-	instanceName := fmt.Sprintf("%s-bastion", cluster.Name)
-	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, instanceName)
-	if err != nil {
-		return err
+	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		// Floating IP could have been created but not associated to instance, attempt to delete it from saved status first
```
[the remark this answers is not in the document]
zioc: Actually this is the regular case (floating IP is stored in status, and deleted here), I haven't removed [the other block](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d2070463c2d17d81b44b0f3fe43eba4cd1910553/controllers/openstackcluster_controller.go#L241-L254) that removes floating IP attached to bastion, but it will we be useful only if floating IP is was not properly saved in status for some reason. Maybe I should reword the comment to make it clearer.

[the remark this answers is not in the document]
zioc: As the removal of floating IP will already be logged [when it will be removed](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8a49b9f23c3a8bd2f41c4e071213edc8bc8ede7b/pkg/cloud/services/networking/floatingip.go#L111) at this point, it does not seem very meaningfull to add another log. I [changed my comment](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/19a6cb5fff75a67b792cf274f37d55b1738c93e1/controllers/openstackcluster_controller.go#L248), hoping that it will be more explicit. Let me know if you want me to revisit that.

zioc:
rebased on top of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1783/commits/0c65d147838209ccc6c5bae1c9d45c8c583f242a 

/retest (CI issue seemed unrelated)

jichenjc:
/lgtm

I don't have further questions on this, but I guess let more people to review might be helpful for such fundmental logic change..

zioc:
rebased on top of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1782/commits/f549ede91eb79f5774a9cefa1383f3e0ef6aea8f

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
+		return ctrl.Result{}, err
 	}
-	err = networkingService.AssociateFloatingIP(openStackCluster, fp, port.ID)
+	fp, err := networkingService.GetFloatingIPByPortID(port.ID)
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to associate floating IP with bastion: %w", err))
-		return fmt.Errorf("failed to associate floating IP with bastion: %w", err)
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create floating IP for bastion: %w", err))
+		return ctrl.Result{}, fmt.Errorf("failed to get floating IP for bastion port: %w", err)
+	} else if fp != nil {
```
mdbooth, line 400: redundant else.

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
 	}
 	port, err := computeService.GetManagementPort(openStackCluster, instanceStatus)
 	if err != nil {
 		err = fmt.Errorf("getting management port for bastion: %w", err)
 		handleUpdateOSCError(openStackCluster, err)
-		return err
+		return ctrl.Result{}, err
 	}
-	err = networkingService.AssociateFloatingIP(openStackCluster, fp, port.ID)
+	fp, err := networkingService.GetFloatingIPByPortID(port.ID)
```
mdbooth, line 396: How would this have a FIP already? Are you looking forward to floating IP IPAM support in the machine spec?

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to associate floating IP with bastion: %w", err))
-		return fmt.Errorf("failed to associate floating IP with bastion: %w", err)
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create floating IP for bastion: %w", err))
```
mdbooth, line 398: Shouldn't use handleUpdateOSCError() here.

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
-	fp, err := networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterName, openStackCluster.Spec.Bastion.Instance.FloatingIP)
-	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create floating IP for bastion: %w", err))
-		return fmt.Errorf("failed to get or create floating IP for bastion: %w", err)
+		return ctrl.Result{}, err
 	}
 	port, err := computeService.GetManagementPort(openStackCluster, instanceStatus)
 	if err != nil {
 		err = fmt.Errorf("getting management port for bastion: %w", err)
 		handleUpdateOSCError(openStackCluster, err)
```
mdbooth, line 393: We shouldn't use handleUpdateOSCError() here either, but this is existing code 🤦  Guessing you cribbed the other uses from here. Would you mind deleting this one, too?

There are very few circumstances when we would ever set the cluster to failed. It would only really be before it has become initially ready, and even then only if the cause was due to an invalid spec.

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
 	}
-	bastion.FloatingIP = fp.FloatingIP
-	openStackCluster.Status.Bastion = bastion
-	annotations.AddAnnotations(openStackCluster, map[string]string{BastionInstanceHashAnnotation: bastionHash})
-	return nil
+	openStackCluster.Status.Bastion.FloatingIP = fp.FloatingIP
+
+	err = networkingService.AssociateFloatingIP(openStackCluster, fp, port.ID)
+	if err != nil {
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to associate floating IP with bastion: %w", err))
```
mdbooth, line 421: As above: it's just a regular error.

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
+	clusterName := fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)
+	floatingIP := openStackCluster.Spec.Bastion.Instance.FloatingIP
+	if openStackCluster.Status.Bastion.FloatingIP != "" {
+		// Some floating IP has already been created for this bastion, make sure we re-use it
+		floatingIP = openStackCluster.Status.Bastion.FloatingIP
+	}
+	fp, err = networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterName, floatingIP)
 	if err != nil {
-		return err
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create floating IP for bastion: %w", err))
```
mdbooth, line 414: As above: regular error.

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
+		return ctrl.Result{}, fmt.Errorf("failed to get floating IP for bastion port: %w", err)
+	} else if fp != nil {
+		// Floating IP is already attached to bastion, no need to proceed
+		openStackCluster.Status.Bastion.FloatingIP = fp.FloatingIP
+		return ctrl.Result{}, nil
 	}
 
-	bastion, err := instanceStatus.BastionStatus(openStackCluster)
+	clusterName := fmt.Sprintf("%s-%s", cluster.Namespace, cluster.Name)
```
mdbooth, line 406: I see this is just code motion, but could you remove this and use `cluster.Name` below, instead?

controllers/openstackcluster_controller_test.go
mdbooth, line 1: Please can you add new test cases covering the new behaviours. These should hopefully be mostly cut/paste with minor changes. That I can think of:
* Create bastion, server already exists
* Delete bastion, ID is not in status but server exists
* Create bastion with existing FIP
* Create bastion re-queues

Let me know if any of these turn out to be a rabbithole, though. There may be other ways.

controllers/openstackmachine_controller.go
```diff
@@ -431,14 +440,24 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+			logger.Info("Unable to get OpenStack instance", "name", openStackMachine.Name)
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.OpenStackErrorReason, clusterv1.ConditionSeverityError, err.Error())
+			return nil, err
+		}
 	}
 
 	if instanceStatus == nil {
+		// First check if there is an existing instance with machine name, in case where instance ID would not have been stored in machine status
+		if instanceStatus, err = computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name); err == nil {
+			if instanceStatus != nil {
```
mdbooth, line 457: We should set `openStackMachine.Spec.InstanceID` here.

mdbooth (commented):
A few things, but nothing major except some new tests.

@lentzi90 @jichenjc This is a fairly major change and I'm close to approving it. Do either of you have time to look closely, too?

jichenjc:
>>>@lentzi90 @jichenjc This is a fairly major change and I'm close to approving it. Do either of you have time to look closely, too?

I did review this deeply before and yes, will take another look this week, thanks for reminder~

controllers/openstackcluster_controller.go
```diff
@@ -215,10 +215,25 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
 		return err
 	}
 
-	instanceName := fmt.Sprintf("%s-bastion", cluster.Name)
-	instanceStatus, err := computeService.GetInstanceStatusByName(openStackCluster, instanceName)
-	if err != nil {
-		return err
+	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		if err = networkingService.DeleteFloatingIP(openStackCluster, openStackCluster.Status.Bastion.FloatingIP); err != nil {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
```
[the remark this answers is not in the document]
zioc: I was not aware of that consequence, thanks for pointing that out. 

But on the other hand, if a failed status blocks any further reconciliation in openstackmachine_controller, it doesn't seems to be the case in openstackcluster_controller, I wonder if if is really relevant to remove all these calls as FailureReason & FailureMessage in status is very useful to help users figuring out what is going wrong without having to look at controller logs...

controllers/openstackcluster_controller.go
```diff
@@ -215,10 +215,25 @@ func deleteBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackClust
...
+		instanceStatus, err = computeService.GetInstanceStatus(openStackCluster.Status.Bastion.ID)
+		if err != nil {
+			return err
+		}
+	} else {
+		instanceName := fmt.Sprintf("%s-bastion", cluster.Name)
+		instanceStatus, err = computeService.GetInstanceStatusByName(openStackCluster, instanceName)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
zioc: done

controllers/openstackcluster_controller.go
```diff
@@ -308,82 +325,104 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
...
-		return err
+		return reconcile.Result{}, err
 	}
 
 	instanceSpec := bastionToInstanceSpec(openStackCluster, cluster.Name)
 	bastionHash, err := compute.HashInstanceSpec(instanceSpec)
 	if err != nil {
-		return fmt.Errorf("failed computing bastion hash from instance spec: %w", err)
+		return reconcile.Result{}, fmt.Errorf("failed computing bastion hash from instance spec: %w", err)
+	} else if bastionHashHasChanged(bastionHash, openStackCluster.ObjectMeta.Annotations) {
```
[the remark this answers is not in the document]
zioc: fixed

controllers/openstackcluster_controller.go
zioc: This is useful to handle the race condition that I was discussing in my [previous comment](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753#issuecomment-1832018501): As we are patching [object before the status subresource](https://github.com/kubernetes-sigs/cluster-api/blob/0970c323d8ebf247bbe1c124e7e94dc4ad1ac35f/util/patch/patch.go#L143-L144) once first reconcile succeeds, it is quite frequent to have a subsequent reconciliation occurring before the status update (from client cache perspective). This check helps to prevent the creation of an additional floatingIP in that case. I've added a comment to explain that point.

controllers/openstackcluster_controller.go
zioc: I'm not sure to properly get your point here: clusterName is like a namespaced name, do you want me to remove the leading namespace and only use cluster.Name? (for now it is only used in floatingIP description so it would not make much difference...)

controllers/openstackcluster_controller_test.go
zioc: done, I've added suggested test cases.

controllers/openstackmachine_controller.go
zioc: For now `openStackMachine.Spec.InstanceID` was set right after `getOrCreate` [function call](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/f271006c0dd0cc60df682f60ed969503e07eb0c2/controllers/openstackmachine_controller.go#L346C11-L346C11). I'll do it here instead, but it also needs to be done before final return of `getOrCreate` in that case.

controllers/openstackcluster_controller.go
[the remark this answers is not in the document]
mdbooth: You're absolutely right: that's only for the Machine controller. Please ignore these comments for now. We can move a cleanup of this stuff into another PR.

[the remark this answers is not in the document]
zioc: Ok, I'll re-add the handleUpdateOSCError calls that I had removed in the meantime following your comments.

controllers/openstackcluster_controller.go
zioc: ok, done!

controllers/openstackcluster_controller.go
zioc: ok, done!

controllers/openstackcluster_controller_test.go
```diff
@@ -195,15 +200,179 @@ var _ = Describe("OpenStackCluster controller", func() {
...
+
+		computeClientRecorder := mockScopeFactory.ComputeClient.EXPECT()
+		computeClientRecorder.GetServer("adopted-fip-bastion-uuid").Return(&server, nil)
+
+		bastionFloatingips := make([]floatingips.FloatingIP, 1)
+		bastionFloatingips[0].FloatingIP = "1.2.3.4"
+
+		networkClientRecorder := mockScopeFactory.NetworkClient.EXPECT()
+		networkClientRecorder.ListPort(gomock.Any()).Return(make([]ports.Port, 1), nil)
+		networkClientRecorder.ListFloatingIP(gomock.Any()).Return(bastionFloatingips, nil)
```
mdbooth, line 304: As above.

mdbooth (commented):
/approve

This looks great, and especially thanks for writing good tests with good descriptions.

I've approved, but I'd appreciated it if you could clear up the portID thing in the tests. It works, but it raised a flag for me that detracted from the clarity of the tests.

controllers/openstackcluster_controller_test.go
```diff
@@ -195,15 +200,179 @@ var _ = Describe("OpenStackCluster controller", func() {
...
+		server.Status = "ACTIVE"
+
+		computeClientRecorder := mockScopeFactory.ComputeClient.EXPECT()
+		computeClientRecorder.ListServers(servers.ListOpts{
+			Name: "^capi-cluster-bastion$",
+		}).Return([]clients.ServerExt{server}, nil)
+
+		networkClientRecorder := mockScopeFactory.NetworkClient.EXPECT()
+		networkClientRecorder.ListPort(gomock.Any()).Return(make([]ports.Port, 1), nil)
+		networkClientRecorder.ListFloatingIP(gomock.Any()).Return(make([]floatingips.FloatingIP, 1), nil)
```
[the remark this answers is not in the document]
zioc: make sense, done

controllers/openstackcluster_controller_test.go
zioc: done here too

controllers/openstackmachine_controller.go
zioc: Yes, on the other hand, setting InstanceID here could also help ensuring that it will always be done as soon as we retrieve it, whereas if we do it in `reconcileNormal` fiunction, it may be delayed by some future refactoring that would add code between the call to `getOrCreate` and the setting of the ID....

I've switched back to previous version as it's slighlty cleaner as you said.

jichenjc:
@zioc can you help address above comments and solve teh conflict ? Thanks

mdbooth:
FWIW, all of my comments are addressed 👍

jichenjc:
/lgtm
/hold cancel

looks good to me ,thanks~

zioc:
thanks!

/retest

PR #1754, opened 2023-11-22
🐛  Update e2e testing image of CAPI version from 1.5.0 to 1.5.1 

### LINKED ISSUE #1698: E2e misconfigured logger
/kind bug

All our e2e tests currently show a log like this:

```
 [controller-runtime] log.SetLogger(...) was never called, logs will not be displayed:
goroutine 82 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot() 
...
```

I thought this was due to https://github.com/kubernetes-sigs/cluster-api/issues/9101, but it has already been fixed in CAPI v1.5.1 (which we are using) so I suspect there is something more to it.
We need to find out where this is coming from and fix it.

Edit: It could still be from the CAPI version. We are using CAPI v1.5.0 for e2e tests it seems:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c73d55dd47006207a15c93bbae69811c97274827/test/e2e/data/e2e_conf.yaml#L35

We should not have misconfigured loggers that emit stack traces on successful e2e jobs.

I think this would be a good issue to get started with the project.
/help

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): main
- Cluster-API version:
- OpenStack version:
- Minikube/KIND version:

maxrantil:
Fix misconfigured logger

Fixes #1698

**TODOs**:
- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/suites/e2e/e2e_suite_test.go
+++ b/test/e2e/suites/e2e/e2e_suite_test.go
@@ -25,6 +25,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"sigs.k8s.io/cluster-api-provider-openstack/test/e2e/shared"
 )
@@ -44,6 +46,7 @@ func init() {
 
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
+	ctrl.SetLogger(klog.Background())
 	RunSpecs(t, "capo-e2e")
 }
 
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -12,11 +12,11 @@
 managementClusterName: capo-e2e
 
 images:
-- name: gcr.io/k8s-staging-cluster-api/cluster-api-controller:v1.5.0
+- name: gcr.io/k8s-staging-cluster-api/cluster-api-controller:v1.5.1
   loadBehavior: tryLoad
-- name: gcr.io/k8s-staging-cluster-api/kubeadm-bootstrap-controller:v1.5.0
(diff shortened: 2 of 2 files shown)


lentzi90:
Please check all occurrences in the file! There are many places with the same version. One for each of bootstrap, control plane and core providers, and also for the images for them.

jichenjc:
I think we are only updating the testing image of CAPI? not the CAPI itself in the code from the change.. so the commit message might need update?

maxrantil:
Yes that is right. Thanks you for the feedback @jichenjc . It is now changed.

lentzi90:
Looks like it was not enough :disappointed: 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1754/pull-cluster-api-provider-openstack-e2e-test/1727570398410706944#1:build-log.txt%3A1930
```
[controller-runtime] log.SetLogger(...) was never called, logs will not be displayed:
```

I had hoped that it would be enough to just bump the CAPI version but it could be that we have to do something on our side also.

test/e2e/suites/e2e/e2e_suite_test.go
```diff
@@ -45,6 +47,7 @@ func init() {
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "capo-e2e")
+	ctrl.SetLogger(klog.Background())
```
lentzi90, line 50: Umm now you add the logger *after* running the specs :sweat_smile:

maxrantil: whoops

lentzi90 (commented):
Looking good now from what I can see in the logs!
/approve

Issue #1749, opened 2023-11-15
TCP for Loadbalancer Monitor causes API server downtime during upgrades


MPV:
/kind bug

**What steps did you take and what happened:**

1. Upgrade k8s version in a CAPO target cluster.
2. When an old control plane node leaves, the API server is removed from the load balancer pool after it has stopped.
   - TCP health monitor fails at time `0s`, node is removed at time `90s`.
   - ...as the current implementation means:
       >After `90 (=3*30)` seconds of downtime, API server pool members will be marked as down.
    - ...as per:
    https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/821a1a2ef25ac615db5fb26379eb0c4b947ad284/pkg/cloud/services/loadbalancer/loadbalancer.go#L361-L369
3. This means any/all API server clients will get intermittent failures (as they round-robin will reach an API server node which isn't there anymore).

**What did you expect to happen:**

1. ...
2. No downtime/errors on the API server, and thus none of the sub-issues seen above.

**Anything else you would like to add:**

Could we start supporting either/both of:
- https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1221
- https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1748

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `v0.7.0`
- Cluster-API version: `v1.3.3`
- OpenStack version: ...
- Minikube/KIND version: N/A
- Kubernetes version (use `kubectl version`): 1.25 -> 1.26
- OS (e.g. from `/etc/os-release`): N/A

...but we have manually changed our LB health monitor's to include the changes from #1360 (to alleviate issues seen in #1221 and #1375).

MPV:
An example of a (semi-CAPO-related) symptom we've seen caused by this:

1. Nodes can't connect to the API server during this time, marking themselves as unready.
2. Then, pods on these nodes thus get evicted.
3. ...and workloads starts shifting around in the cluster, potentially triggering cluster-autoscaler, which will have a hard time sizing the cluster correctly as the amount of ready nodes keep changing based on the intermittent (OK/error) answers from the API servers.

mnaser:
@MPV is there any conceivable way we can do this while kepeing TCP load balancer monitors?

mdbooth:
[...]

Issue #1256, opened 2022-06-06
Use a server group to ensure anti-affinity for control plane nodes


mkjpryor:
/kind feature

**Describe the solution you'd like**
We should put control plane nodes in a server group with the soft anti-affinity policy to decrease the risk that a hypervisor failure takes out the whole control plane.

This becomes more important once #1252 is merged, but is something that I think we should do even when using explicit AZs.

**Anything else you would like to add:**

jichenjc:
I knew OCP seems has such by default, are you suggesting we should default it in our template or in the code directly?

https://github.com/openshift/installer/blob/master/docs/user/openstack/README.md#using-a-server-group

mkjpryor:
@jichenjc 

IMHO, it would be nice if we could manage a server group for the control plane nodes. It would be better for the "Infrastructure-as-Code" approach to have the whole thing defined in Kubernetes resources, rather than having to make the server group ahead of time.

jichenjc:
yes, we have such definition in api/v1alpha5/openstackmachine_types.go
```
// The server group to assign the machine to
        ServerGroupID string `json:"serverGroupID,omitempty"`
```

I think maybe we can refer to other impl, like sec group we did, a bool var to control whether we create for customer
if the bool is true, we create server group and honor the ID to each control plane node
if it's false, let customer decide whether they need such ..
and during deletion phase, delete the server group all together at last

mdbooth:
For the control plane we certainly have the opportunity to create and manage this server group: the server group will have the same lifecycle as the cluster, so it can be managed with the cluster.

For machines in a machineset it is not currently possible to do this in CAPI as far as I'm aware. I have previously mooted the possibility of some CAPI object whose lifecycle spans multiple machines. The best suggestion so far is something along the lines of MachinePools, however this would be an abuse of this interface IMHO as OpenStack has no such concept. An object with the same lifecycle as a MachineSet would probably be sufficient: you'd need a new machineset for every server group, but this doesn't sound unreasonable to me.

dalees:
I'd like to re-open this issue and help implement this.

We are a cloud provider working on implementing OpenStack Magnum managed CAPO clusters and these must have `anti-affinity` applied to control plane nodes. (Use case: Region with a single AZ).

The preferred defaults would be `anti-affinity` for control plane to ensure separation and `soft-anti-affinity` for machines to allow scaling beyond available compute hypervisors.

Proposal:
[...]

PR #1756, opened 2023-11-22
✨ proposal: new API for managed Security Groups

EmilienM:

KEP to enhance Security Group API.

Related #1752

### DIFF
--- a/docs/proposals/20231122-security-groups.md
+++ b/docs/proposals/20231122-security-groups.md
@@ -1,0 +1,216 @@
+---
+title: Flexible managed security sroups
+authors:
+  - "@emilienm"
+reviewers:
+  - "@jichenjc"
+  - "@lentzi90"
+  - "@mdbooth"
+creation-date: 2023-11-22
+last-updated: 2024-02-21
+
+status: provisional
+---
+
+## Title
+
+Flexible managed security groups
+
+## Table of Contents
+
+- [Flexible managed security groups](#title)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Alternatives](#alternatives)
+  - [Upgrade Strategy](#upgrade-strategy)
+  - [Additional Details](#additional-details)
+    - [Test Plan [optional]](#test-plan-optional)
+    - [Graduation Criteria [optional]](#graduation-criteria-optional)
+    - [Version Skew Strategy [optional]](#version-skew-strategy-optional)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+
+- Egress - Egress traffic is traffic that is leaving the cluster.
+- Ingress - Ingress traffic is traffic that is coming into the cluster.
+- Security Group - A security group is a named set of rules that control traffic to and from virtual machine instances.
+  The rules can specify which ports and protocols to allow, and which IP address ranges to allow traffic to or from.
+  Security groups can be associated with multiple instances, and multiple security groups can be associated with a single instance.
+  Security groups act as a firewall for associated instances, controlling both inbound and outbound traffic at the instance level.
+
+## Summary
+
+This proposal introduces a new API to manage more security groups and rules.
+
+## Motivation
+
+We want users to be able to not think about the security groups that are needed for their Control Plane and Worker Nodes. However, we want to be able to provide a way for users to have more control over the managed security groups and not have to worry about rules that are not needed by their cluster. They will be able to add additional security groups and rules to the cluster.
+
+### Goals
+
+1. Create a new API to manage more security groups and rules:
+   1. Default rules provided by `OpenStackCluster.spec.managedSecurityGroups` will have a new API used to manage additional security groups and rules.
+   1. Successfully be able to new manage security groups rules for the Bastion, the Control Plane and the Worker Nodes.
+   1. Successfully be able to manage the lifecycle of additional security groups that later can be used in the Machine Template.
+
+### Non-Goals/Future Work
+
+1. Removing pre-existing security groups support in `OpenStackMachineTemplate`, via `OpenStackMachineSpec` and for the Bastion in `OpenStackCluster`.
+1. Deprecate `OpenStackCluster.spec.managedSecurityGroups.llowAllInClusterTraffic`, as we'll still need it.
+
+## Proposal
+
+### User Stories
+
+- When creating a cluster as an operator, I don't want to worry about the security groups that are needed for my Control Plane and Worker Nodes to be functional and
+  I want these security groups to be managed automatically.
+
+- As an operator, I can already pre-create security groups and provide them to the Machine Templates. However, to remove that burden from the user,
+  I now want to let the controller manage the lifecycle of these security groups.
+  and have a new API to provide the additional security groups and their rules.
+
+- As an operator, I want to be able to manage additional security group rules for the Bastion, the Control Plane and the Worker Nodes in order to provide more
+  rules than the defaults, so I have more flexibility in my cluster.
+
+- As an operator, I want to be able to manage additional security group rules for the CNI so I have more options than Calico. I need to provide my CNI's specific security group rules
+  that I need applied to all the machines, so I have more flexibility in my cluster.
+
+### Implementation Details/Notes/Constraints
+
+The plan is to create a new controller and migrate the management of security groups and their rules in there when the `OpenStackCluster.Spec.ManagedSecurityGroups` is set to `True`.
+
+A cluster operator will be able to add additional security groups and rules to the cluster by adding them to the `OpenStackCluster` spec.
+The controller will then reconcile the security groups and their rules to the cluster between what's in the spec and what's in the status.
+Special care will be taken to not remove security groups that are still in use by the machines.
+Also, during an upgrade, the controller will make sure to add the new security groups and their rules to the machines before reducing the default rules
+which won't have the Calico rules anymore.
+
+#### Data model changes
+
+The type for `OpenStackCluster.Spec.ManagedSecurityGroups` was a boolean but will become a structure with new fields.
+
+Example `ManagedSecurityGroups` yaml:
+
+```yaml
+managedSecurityGroups:
+  # Enable the management of security groups with the default rules (kubelet, etcd, etc.)
+  # The default stays false
+  enabled: true
+  # Allow to extend the default security group rules for the Bastion,
+  # the Control Plane and the Worker security groups
+  additionalBastionSecurityGroupRules:
+    - direction: ingress
+      ethertype: IPv4
+      portRangeMin: 1234
+      portRangeMax: 1234
+      protocol: tcp
+  additionalControlPlaneSecurityGroupRules:
+    - direction: ingress
+      ethertype: IPv4
+      portRangeMin: 1234
+      portRangeMax: 1234
+      protocol: tcp
+  additionalWorkerSecurityGroupRules:
+    - direction: ingress
+      ethertype: IPv4
+      portRangeMin: 1234
+      portRangeMax:1234
+      protocol: tcp
+  # Allow to provide rules that will be applied to all nodes
+  allNodesSecurityGroupRules:
+    - direction: ingress
+      ethertype: IPv4
+      portRangeMin: 1234
+      portRangeMax: 1234
+      protocol: tcp
+  # Allow to provide additional security groups and rules that can be used in the Machine Template
+  additionalSecurityGroups:
+    - name: my-security-group
+      description: My security group
+      rules:
+        - direction: ingress
+          ethertype: IPv4
+          portRangeMin: 1234
+          portRangeMax: 1234
+          protocol: tcp
+  # When set to `True`, the controller will add the Calico rules to the All nodes security group and update
+  # the machines before the default rules are removed from the managed security groups
+  useLegacyCalicoRules: false
+```
+
+#### Enable or disable the management of security groups
+
+`OpenStackCluster.Spec.ManagedSecurityGroups.enabled` will be used to enable or disable the management of security groups in general.
+It was a boolean and will now be a structure with new fields.
+
+Conversion will be done automatically for the user.
+
+#### Additional security rules
+
+`OpenStackCluster.Spec.ManagedSecurityGroups.additionalBastionSecurityGroupRules` will be used to add additional security rules to the Bastion security group.
+`OpenStackCluster.Spec.ManagedSecurityGroups.additionalControlPlaneSecurityGroupRules` will be used to add additional security rules to the Control Plane security group.
+`OpenStackCluster.Spec.ManagedSecurityGroups.additionalWorkerSecurityGroupRules` will be used to add additional security rules to the Worker security group.
+
+Theses rules will be managed by the new controller and added to respectively the Bastion, Control Plane and Worker security groups status.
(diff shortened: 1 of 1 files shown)


docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,169 @@
...
+        - direction: ingress
+          ethertype: IPv4
+          portRangeMin: 443
+          portRangeMax: 443
+          protocol: tcp
+```
+
+We will replace the `managedSecurityGroups` field with `managedLegacySecurityGroups` which will be a boolean and will be used to determine if the legacy rules should be managed or not. This will be used to provide a migration path for users who were using the legacy rules provided by `OpenStackCluster.spec.managedSecurityGroups`. We'll add conversion logic to convert the old `managedSecurityGroups` field to the new `managedLegacySecurityGroups` field.
+
+Legacy rules and new rules will be managed in parallel, until we decide to remove the legacy rules. Which means that they can be used together, as they'll live in their own Security Groups.
```
EmilienM, line 139: Do we really want/need that? I have mixed feelings

jichenjc: >>>Legacy rules and new rules will be managed in parallel, until we decide to remove the legacy rules. Which means that they can be used together, as they'll live in their own Security Groups.

seems we don't , you provide a migration path and we should stick to newer way and drop older not good enough method to save effort and avoid maintain complexity

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,169 @@
...
+This proposal introduces a new API to manage OpenStack Security Groups in a flexible way that will replace legacy rules that are can optionaly be managed by CAPO.
+
+## Motivation
+
+We want users to be able to manage their security groups in a flexible way for both the Control Plane and the Worker Nodes without enforcing unnecessary rules that are not needed by the user.
+
+### Goals
+
+1. Make the feature configurable via a new API
+2. Avoid causing and breaking changes to previous users, they'll still be able to use the legacy rules
```
jichenjc, line 54: will this still be true based on our previous discussion?

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,169 @@
...
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: OpenStackCluster
+metadata:
+  name: cluster-sample
+  namespace: default
+  spec:
+    securityGroupsSpec:
+      controlPlaneSecurityGroupRules:
```
jichenjc, line 93: previously we talked about whether we should create a CRD and reusable in other clusters 
e.g you create multiple clusters and if modify one json file or yaml file then it can be used in multiple clusters?

I know it's complicated so I think we can defer to next stage to add such extension?

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,169 @@
...
+          portRangeMax: 80
+          protocol: tcp
+        - direction: ingress
+          ethertype: IPv4
+          portRangeMin: 443
+          portRangeMax: 443
+          protocol: tcp
+```
+
+Legacy rules and new rules will be managed exclusively, until we decide to remove the legacy rules.
```
lentzi90, line 137: If we want to remove the legacy rules, I think we *must* then provide working/minimal defaults for the new managed rules. The user should not need to specify "common sense" rules that would be needed in standard clusters IMO

lentzi90 (commented):
I like the overall design of this

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,169 @@
...
+
+We want users to be able to manage their security groups in a flexible way for both the Control Plane and the Worker Nodes without enforcing unnecessary rules that are not needed by the user.
+
+### Goals
+
+1. Make the feature configurable via a new API
+2. Avoid causing and breaking changes to previous users, they'll still be able to use the legacy rules
+3. Successfully be able to create and manage security groups for both the Control Plane and the Worker Nodes
+4. Provide a migration path (documentation or code) for users who are currently using the legacy rules provided by `OpenStackCluster.spec.managedSecurityGroups`
+5. Deprecate `OpenStackCluster.spec.apiServerLoadBalancer.additionalPorts` as we'll have an API to create Control Plane Security Group rules (with more flexibility than we have now). Conversion will happen between this parameter and the one for Control Plane Security Group rules, in order to maintain backward compatibility
```
mdbooth, line 57: We can't do this. This parameters has more uses than just security group rules and it would break existing use cases. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1749#issuecomment-1826054055

However, I would like to separate responsibility for these rules into:
* A separate security group
* Managed by a separate Octavia control plane endpoint provider

i.e. Create a new controller for managing an Octavia API loadbalancer. That controller is responsible for creating a security group suitable for its needs and attaching that security group to control plane machines.

In this case, it would simply be the responsibility of the cluster controller to remove the legacy rules from the default security group.

docs/proposals/20231122-security-groups.md
EmilienM: After some discussion with Matt, i've reworked the whole KEP. Please have a look again.

docs/proposals/20231122-security-groups.md
EmilienM: After some discussion with Matt, i've reworked the whole KEP. Please have a look again.

docs/proposals/20231122-security-groups.md
EmilienM: For now, let's try without a new CRD and see how it goes. I'll keep this comment open as it's still relevant in the last KEP's iteration.

docs/proposals/20231122-security-groups.md
EmilienM: After some discussion with Matt, i've reworked the whole KEP. Please have a look again.

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,225 @@
...
+
+This proposal moves the management of security groups into a dedicated controller and introduces a new API to manage more security groups and rules.
+
+## Motivation
+
+We want users to be able to not think about the security groups that are needed for their Control Plane and Worker Nodes. However, we want to be able to provide a way for users to have more control over their security groups and not have to worry about rules that are not needed by their cluster.
+
+### Goals
+
+1. Move the management of security groups into a dedicated controller since we will have to reconcile the assignement of security groups to the machines.
```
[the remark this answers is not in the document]
mdbooth: The question is really whether it makes sense for us. CAPA has a multitude of security groups already: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3373251e75afd25e11b256a645a95469b0139328/api/v1beta2/network_types.go#L399-L418

If they don't need to alter their security group attachments they don't need a separate controller.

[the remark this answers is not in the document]
EmilienM: Something interesting [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1256#issuecomment-1829999292) which could justify a dedicated CRD & controller, eventually.
I'm waiting for @mkjpryor's feedback as well.

docs/proposals/20231122-security-groups.md
```diff
@@ -0,0 +1,225 @@
...
+      portRangeMin: 1234
+      portRangeMax:1234
+      protocol: tcp
+  # Allow to provide rules for the CNI that will be applied to all nodes
+  cniSecurityGroupRules:
+    - direction: ingress
+      ethertype: IPv4
+      portRangeMin: 1234
+      portRangeMax: 1234
+      protocol: tcp
```
dulek, line 142: Why do we call this one CNI, when it's basically a thing gathering rules added to each of the nodes? I'd abstain from designing this against a CNI. From the technical standpoint I think it's possible to use different CNIs on different nodes. It's totally not supported at the moment, but might be one day.

dulek (changes requested):
There is an important design decision to be made here - how do we deal with users messing with SGs managed by CAPO through OpenStack API.

My idea would be to adopt K8s model here. The controller should periodically monitor SG rules of the SGs it's managing and SG associations and reconcile them if user made any changes. I think this is the only sane way to do it, otherwise we allow users to modify SGs on their own and we never know if the SG that's attached is to be kept or removed during e.g. upgrade.

mkjpryor:
@EmilienM 

Do you think that this should be consistent with whatever we do for server groups? Personally, I favour new CRDs for both, i.e. `OpenStackSecurityGroup` and `OpenStackServerGroup` CRDs.

I worry about us cluttering up the spec of the `OpenStackCluster` with too much complicated stuff like this. IMHO we should start to establish a pattern of having additional CRDs for resources that don't have a direct 1-2-1 correspondence with a cluster/machine.

EmilienM:
> @EmilienM
> 
> Do you think that this should be consistent with whatever we do for server groups? Personally, I favour new CRDs for both, i.e. `OpenStackSecurityGroup` and `OpenStackServerGroup` CRDs.
> 
> I worry about us cluttering up the spec of the `OpenStackCluster` with too much complicated stuff like this. IMHO we should start to establish a pattern of having additional CRDs for resources that don't have a direct correspondence with a cluster/machine.
> 
> For example, a port is something that I don't mind having in `OpenStackMachine.spec`, as it is clearly related to the machine.

To me there is a relation between Security Groups and Machines, as we want to update a Machine if a security group is added to the control plane machines.
Example given: if we move the Calico rules in their own Security Group, we'll need a migration path where we remove the rules from the control plane security group and then create a new one with Calico rules and add that group to the machines. This require the controller to update the machine with the new SG.
Do we want to do that in the OpenStackMachine (or Cluster) controller or via a new controller dedicated to security groups but that can also update a Machine?

EmilienM:
> There is an important design decision to be made here - how do we deal with users messing with SGs managed by CAPO through OpenStack API.
> 
> My idea would be to adopt K8s model here. The controller should periodically monitor SG rules of the SGs it's managing and SG associations and reconcile them if user made any changes. I think this is the only sane way to do it, otherwise we allow users to modify SGs on their own and we never know if the SG that's attached is to be kept or removed during e.g. upgrade.

I think we all agree; the question remains whether we have a new CRD (+ controller) dedicated to Security Groups lifecycle or continue to use OpenStackCluster.

mkjpryor:
It just feels odd to me that all the security groups are defined in the `OpenStackCluster` object and the `OpenStackMachine` object says "I want to use security group X from the cluster object", vs having a new `OpenStackSecurityGroup` CRD 🤷‍♂️ Same for server groups when we do that thing.

Happy to be overruled though.

jichenjc:
>>>It just feels odd to me that all the security groups are defined in the OpenStackCluster object and the OpenStackMachine object says "I want to use security group X from the cluster object", vs having a new OpenStackSecurityGroup CRD 🤷‍♂️ Same for server groups when we do that thing.

that's exactly what I asked above, and I Think maybe we can check comments from @EmilienM 
```
For now, let's try without a new CRD and see how it goes. I'll keep this comment open as it's still relevant in the last KEP's iteration.
```

so maybe we can check https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1765/files first

docs/proposals/20231122-security-groups.md
EmilienM: I can name it `allNodesSecurityGroupRules`, I don't mind.

docs/proposals/20231122-security-groups.md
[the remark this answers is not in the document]
EmilienM: I've pushed https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1774 which change a bit the logic. I hope you'll fine it convenient. A new controller is not required and not easy to implement because managed security groups are managed in the cluster and applied to the machine.

mdbooth:
As discussed at the last office hours, I think this is the biggest single API stabilisation task we have for v1beta1. Specifically, we need to remove the hard-coded Calico rules, but allow:
* CNI rules to be used in managed security groups
* Safely upgrading clusters which are currently using the hard-coded Calico rules

I would very much like to get to v1beta1 asap, so to make that simpler I'd like to propose reducing the scope of the change here as much as possible. 

```yaml
managedSecurityGroups:
  # Enable the management of security groups with the default rules (kubelet, etcd, etc.)
  # The default stays false
  enabled: true
  # Allow to extend the default security group rules for the Bastion,
  # the Control Plane and the Worker security groups
  additionalBastionSecurityGroupRules:
    - direction: ingress
      ethertype: IPv4
      portRangeMin: 1234
      portRangeMax: 1234
      protocol: tcp
  additionalControlPlaneSecurityGroupRules:
    - direction: ingress
      ethertype: IPv4
      portRangeMin: 1234
      portRangeMax: 1234
      protocol: tcp
  additionalWorkerSecurityGroupRules:
    - direction: ingress
      ethertype: IPv4
      portRangeMin: 1234
      portRangeMax:1234
      protocol: tcp
  # Allow to provide rules that will be applied to all nodes
  allNodesSecurityGroupRules:
    - direction: ingress
      ethertype: IPv4
      portRangeMin: 1234
      portRangeMax: 1234
      protocol: tcp
```

I would remove `additionalSecurityGroups` for now for 2 reasons. Firstly, in this proposal they are nested under `managedSecurityGroups`, but they're not managed. I think it would be confusing to write:

```yaml
managedSecurityGroups:
  enabled: false
  additionalSecurityGroups:
  - ...
```

Does the user expect these to be created when managed security groups is disabled? `additionalSecurityGroups` are unmanaged, so I'd have them separate.

Secondly, adding a new `additionalSecurityGroups` field would be a backwards compatible change, so we can do it in v1beta1. Lets punt on anything we don't strictly need right now.

We discussed having a `useLegacyCalicoRules` field, but it occurs to me we don't need it: we can just add the Calico rules directly to `allNodesSecurityGroupRules` during upgrade. This means that the legacy cruft only lives in the legacy API code. We can document what the rules are, and others can add to the documentation the required rules for other CNIs.

In the first instance, we can also remove additional controller complexity by not changing the set of security groups attached to nodes. This means we don't have an upgrade problem, and don't need to be able to dynamically attach and detach security groups. Instead, at least in the first instance, we can simply replicate rules in `allNodesSecurityGroupRules` to `control-plane` and `workers`.

Lastly, we make extensive use of `RemoteGroupID` in our current rules. We need to allow that to be expressed in the new API, but it obviously can't take the UUID of a managed security group. I suggest that in the first instance we add a `remoteManagedGroups` field to security group rules which references one of `bastion`, `control-plane`, or `worker`. As a fully worked example, I would expect an upgraded configuration containing upgraded Calico rules to be:

```yaml
managedSecurityGroups:
  enabled: true
  allNodesSecurityGroupRules:
  - description: BGP (calico)
    direction: ingress
    ethertype: IPv4
    portRangeMin: 179
    portRangeMax: 179
    protocol: tcp
    remoteManagedGroups:
    - control-plane
    - worker
  - description: IP-in-IP (calico)
    direction: ingress
    ethertype: IPv4
    protocol: 4
    remoteManagedGroups:
    - control-plane
    - worker
```

These 2 rules would become 4 rules, as we would create a duplicate for each of the `remoteManagedGroups`. The resulting 4 rules would be added to both `control-plane` and `worker`.

I suggest that this change alone could also be the minimum change, and we could add `additional<X>SecurityGroupRules` later as a backwards-compatible change.

lentzi90:
Thank you for that thought through comment! I completely agree!

EmilienM:
Cool, I'm working on a change here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1826 to address what Matt has asked for.

dulek:
I'll just note one thing here - in complicated envs Neutron has performance issues with remote_group_id as it generates a lot of flows. It's advised to use CIDRs directly if possible. This _probably_ isn't a problem for CAPO scale, but I'm mentioning it here for the record.

EmilienM:

EmilienM:
this is still highly WIP, I need to update it based on our recent work. Please don't review it yet.

EmilienM:
/close
This has been inactive for a very long time and I don't see myself working on that in the coming months unless something changes. Until then I'll close this PR.

PR #1760, opened 2023-11-28
🌱 Remove Containerfile and Dockerfile experimental

mdbooth:
Podman can now build the Dockerfile without modifications, so we don't need to maintain a separate Containerfile.

Docker can build the Dockerfile without dockerfile:1.1-experimental, so we remove that too.


dulek:
CI failure looks valid.

pierreprinetti:
/lgtm
Thanks!

PR #1763, opened 2023-11-29
✨ IPAM provider for floating ips

bilbobrovall:
Creates an IPAM provider for openstack floating ips that allocates floating ips on demand or uses pre allocated ips and in combination with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1762 lets us get floating ips for worker nodes 

Second part of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1750

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+			return ctrl.Result{}, nil
+		}
+
+		infraCluster, err := r.getInfraCluster(ctx, cluster, &claim)
+		if err != nil {
+			return ctrl.Result{}, errors.New("error getting infra provider cluster")
+		} else if infraCluster == nil {
+			log.Info("infra cluster is not ready yet")
+			return ctrl.Result{}, nil
+		}
```
mdbooth, line 129: We do this here because we call CAPO's GetOrCreateFloatingIP below, which takes an infraCluster argument, which it uses to get the cluster's external network. Doing this requires us to label the IPClaim with the cluster name.

I feel like this adds complexity and reduces the flexibility of this API. How about instead of all this we add a `FloatingIPNetwork` parameter to the pool spec. This would be a `NetworkFilter` which would be required to resolve to a single network, and whose resolved UUID we would write to the pool status.

The implications of this would be:
* A FloatingIPPool can only serve IPs from a single network
* That network must be specified on the pool

It would make things much easier to reason about, though. We would be able to get rid of all references to the cluster from the pool and claims.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	networkingService, err := networking.NewService(scope)
+	if err != nil {
+		scope.Logger().Error(err, "Failed to create networking service") // TODO Remove log
+		return "", err
+	}
+
+	fp, err := networkingService.GetOrCreateFloatingIP(pool, openStackCluster, clusterName, "")
+	if err != nil {
+		return "", fmt.Errorf("get or create floating IP: %w", err)
+	}
```
mdbooth, line 273: I think the unconditional existence check we do in setIPStatuses() is redundant because we check again here. This check is better because it only happens on actual allocation, not in steady state.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	ip := fp.FloatingIP
+
+	// TODO: setStatus should probably solve this stuff
+	pool.Status.ClaimedIPs = append(pool.Status.ClaimedIPs, ip)
+	pool.Status.IPs = append(pool.Status.IPs, ip)
+	if err := r.Client.Status().Update(ctx, pool); err != nil {
```
mdbooth, line 279: Again, we don't want to be doing mid-reconcile updates. I appreciate the desire to write our allocated state immediately, but we will sanity check this in setIPStatuses() at the start of the next reconcile anyway, so if we did crash without writing status we'd still be safe.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+
+	return r.Client.Status().Update(ctx, pool)
+}
+
+func (r *OpenStackFloatingIPPoolReconciler) getIP(ctx context.Context, scope scope.Scope, pool *infrav1.OpenStackFloatingIPPool, openStackCluster *infrav1.OpenStackCluster, clusterName string) (string, error) {
+	if len(pool.Status.AvailableIPs) > 0 {
+		ip := pool.Status.AvailableIPs[0]
+		pool.Status.AvailableIPs = pool.Status.AvailableIPs[1:]
+		pool.Status.ClaimedIPs = append(pool.Status.ClaimedIPs, ip)
+		if err := r.Client.Status().Update(ctx, pool); err != nil {
```
mdbooth, line 258: mid-reconcile update

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+					},
+					Address: ip,
+				},
+			}
+
+			if !contains(pool.Spec.PreAllocatedFloatingIPs, ip) && pool.Spec.ReclaimPolicy == infrav1.ReclaimDelete {
+				controllerutil.AddFinalizer(ipAddress, infrav1.DeleteFloatingIPFinalizer)
+			}
+
+			if err = r.Client.Create(ctx, ipAddress); err != nil {
```
mdbooth, line 168: This one is ok! Because of the way you've structured the reconciler with setIPStatuses() first, the fact of this creation means we don't need any of the other updates on the pool object. The pool will also be reconciled again automatically, because we have a watch on ip addresses, so the creation will trigger the reconcile automatically.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+				return ctrl.Result{}, err
+			}
+
+			claim.Status.AddressRef.Name = ipAddress.Name
```
mdbooth, line 172: This could do with some robustification: we would never recover if we wrote the ipaddress but failed to update the claim. I think the simplest thing to do would be to attempt to fetch an ip address with the expected name first, and only call getIP and do the create if it doesn't exist already.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+			if err = r.Client.Status().Update(ctx, &claim); err != nil {
```
mdbooth, line 173: It's ok to update the claim here, too.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+				log.Error(err, "Failed to update IPAddressClaim status", "claim", claim.Name, "ip", ip)
+				return ctrl.Result{}, err
+			}
+			scope.Logger().Info("Claimed IP", "ip", ip)
+		}
+	}
+	if err = r.setIPStatuses(ctx, scope, pool); err != nil {
```
mdbooth, line 180: I'm not sure I'd bother with this. Any operation which altered ip statuses would result in us being reconciled again immediately, in which case we'd update this on the next reconcile anyway.

mdbooth (commented):
This is fantastic work, thanks!

I've just taken a first pass at the pool controller, so I haven't quite gotten all the way through, yet. There's a few things to think about here. I'll continue to review here, but if it would be helpful I can also make time for a face to face chat.

controllers/openstackfloatingippool_controller.go
bilbobrovall: :+1: I agree that's probably a cleaner approach, that would make using a single pool for multiple clusters cleaner aswell or atleast reusing it for a new cluster.

controllers/openstackfloatingippool_controller.go
bilbobrovall: I agree this is not robust enough, trying to fetching the ipamv1.IPAddress is probably a good start but that would not protect against orphan floating ips in openstack in case getIP allocates a new ip and the creation of the ipamv1.IPAddress fails.

I was thinking about throwing in a defer that would delete the IP in that case, but that felt a bit too hacky.. But atleast an orphaned fip would not break the next reconciliation

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+
+	networkingService, err := networking.NewService(scope)
+	if err != nil {
+		return err
+	}
+
+	floatingIPs, err := networkingService.GetAllFloatingIPs()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
bilbobrovall: Agreed that every reconcile is way too much, this was fix after testing and accidentally manually deallocating ips the pool thought was available.

The problem with checking at allocation is, that it would only be a real problem when it's in AvailableIPs and as long as it has AvailableIPs it would use them before allocating new one I would be fine with just checking this when fullfilling a claim with an IP from AvailableIPs, the only downside would be that available ips would not update automatically if they were removed from openstack until someone try to claim them.

But then again this entire check might be overkill, if you delete ips allocated by the pool maybe you shouldn't be surprised if you get broken ips.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	pool.Status.IPs = ips
+
+	for _, ip := range pool.Status.IPs {
+		if !contains(pool.Status.ClaimedIPs, ip) {
+			pool.Status.AvailableIPs = append(pool.Status.AvailableIPs, ip)
+		}
+	}
+
+	return r.Client.Status().Update(ctx, pool)
```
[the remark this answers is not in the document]
bilbobrovall: I thought reconciles are not triggered with `.Status()` but even then I agree that all these updates should not be necessary when using the patchHelper, I'll have another look at it and check if it's works out just removing all these updates.

controllers/openstackfloatingippool_controller.go
bilbobrovall: I wouldn't say we check it here, at the top of getIP we pop the first ip in availableIPs (which is assumed to be allocated), this just allocates new ips when needed.

But skipping the early return and passing the ip here instead might add the necessary sanitation if we remove it from the status in case of an 404

bilbobrovall:
@mdbooth I've just pushed fixes for most of your comments.
The one thing I'm not very happy with is [the time between an ip getting created in openstack and the IPAddress crd being created](https://github.com/elastx/cluster-api-provider-openstack/blob/openstack_floating_ip_pool/controllers/openstackfloatingippool_controller.go#L131-L168) as it opens up for some race conditions and is the reason I'm running the controller with `MaxConcurrentReconciles: 1`.

What I would like to do is something like trying to create the IPAddress before actually getting one from openstack, to make sure that only one reconcile tries to get an IP from openstack for a claim and then updating the IPAddress after getting an ip from openstack but I can't do that because the `IPAddress.Spec` [is immutable](https://sourcegraph.com/github.com/kubernetes-sigs/cluster-api/-/blob/exp/ipam/internal/webhooks/ipaddress.go?L76-78), any ideas? But then again it might be fine without concurrency

mdbooth:
> @mdbooth I've just pushed fixes for most of your comments. The one thing I'm not very happy with is [the time between an ip getting created in openstack and the IPAddress crd being created](https://github.com/elastx/cluster-api-provider-openstack/blob/openstack_floating_ip_pool/controllers/openstackfloatingippool_controller.go#L131-L168) as it opens up for some race conditions and is the reason I'm running the controller with `MaxConcurrentReconciles: 1`.
> 
> What I would like to do is something like trying to create the IPAddress before actually getting one from openstack, to make sure that only one reconcile tries to get an IP from openstack for a claim and then updating the IPAddress after getting an ip from openstack but I can't do that because the `IPAddress.Spec` [is immutable](https://sourcegraph.com/github.com/kubernetes-sigs/cluster-api/-/blob/exp/ipam/internal/webhooks/ipaddress.go?L76-78), any ideas? But then again it might be fine without concurrency

I don't think we have to worry about this. Even with `MaxConcurrentReconciles > 1` controller-runtime will only ever schedule one reconciliation for a particular resource at a time, but it can schedule reconciliations for different resources concurrently. See https://github.com/kubernetes-sigs/controller-runtime/issues/616.

bilbobrovall:
> > @mdbooth I've just pushed fixes for most of your comments. The one thing I'm not very happy with is [the time between an ip getting created in openstack and the IPAddress crd being created](https://github.com/elastx/cluster-api-provider-openstack/blob/openstack_floating_ip_pool/controllers/openstackfloatingippool_controller.go#L131-L168) as it opens up for some race conditions and is the reason I'm running the controller with `MaxConcurrentReconciles: 1`.
> > What I would like to do is something like trying to create the IPAddress before actually getting one from openstack, to make sure that only one reconcile tries to get an IP from openstack for a claim and then updating the IPAddress after getting an ip from openstack but I can't do that because the `IPAddress.Spec` [is immutable](https://sourcegraph.com/github.com/kubernetes-sigs/cluster-api/-/blob/exp/ipam/internal/webhooks/ipaddress.go?L76-78), any ideas? But then again it might be fine without concurrency
> 
> I don't think we have to worry about this. Even with `MaxConcurrentReconciles > 1` controller-runtime will only ever schedule one reconciliation for a particular resource at a time, but it can schedule reconciliations for different resources concurrently. See [kubernetes-sigs/controller-runtime#616](https://github.com/kubernetes-sigs/controller-runtime/issues/616).

:+1:  I think I got  that from the sample [cluster-api-ipam-provider-in-cluster](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/internal/controllers/ipaddressclaim.go#L105) but looking over it again I realized it's because they use a separate controller `For` IPAddressClaims that creates the IPs. I felt like that would only add extra complexity in our implementation though. And since it looks like we don't need `MaxConcurrentReconciles: 1` we can atleast process claims from different pools in parallel while the `For IPAddressClaim` approach could only process one claim at a time without implementing a per pool lock.

And about these e2e-test, what rate of false positives should I expect? My current plan was running /retest a couple of times and if I don't get a pass assume I've broken something and investigate, but that feels suboptimal..

controllers/openstackfloatingippool_controller.go
bilbobrovall: I’ll have a look at this soon, I realized when looking at the other pr for the is machine controller, it is watching and would need the entire claim to be updated. I’ll need to have a look at the ipam proposal again though, if it mentions if the ipamprovider only needs to update claim.Status I think I should handle it in the machine controller instead so we support any ipamprovider that follows the spec

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+			if err := r.Client.Get(ctx, client.ObjectKey{Name: claim.Name, Namespace: claim.Namespace}, ipAddress); err == nil {
+				// IPAddress already exists, another reconciler is working on it
+				log.Info("IPAddress already exists, another reconciler is working on it")
+				continue
+			}
+
+			ip, err := r.getIP(scope, pool)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
```
mdbooth, line 132: I think you mentioned that you were concerned about the gap between this `getIP()` and the creation of the `IPAddress` containing it below, as a failure there would result in leaking the floating IP. Yes, it will. Unfortunately I can't currently think of anything which isn't more than a little Heath Robinson to do this robustly.

To do this robustly I think we would need to be able to identify after a failure that a floating IP which is not in the pool was intended to be in the pool. Currently we identify it because it is in an IPAddress owned by the pool. If it's not then we can't identify it.

The ideal solution would be to tag the floating IP, but unfortunately Neutron doesn't allow us to tag a floating IP on creation, only after creation. This means that we would still have a gap between floating IP creation and floating IP tagging where we could leak it. In practise I suspect that the creation of an IPAddress object below is likely much faster and more reliable than the tagging of a floating IP, so I suspect this would make reliability worse.

Other options are using the floating IP description to identify it as a member of the pool, or creating it in a multi-step process attached to a special port owned by the pool for the purpose:
* Create FIP attached to pool's 'special' port
* Tag port
* Unattach FIP from port

That way we can always determine the set of FIPs owned by the pool in OpenStack as the set of tagged FIPs plus any FIP attached to the pool's 'special' port. I believe this would be atomic, but it's extremely janky. I'm not sure I'd recommend it in the first instance.

We should probably tag all the FIPs anyway to aid cleanup if the control plane is lost for any reason.

How about we:
* Describe the potential for resource leakage in a big, comprehensive comment
* defer() the tagging of the FIP immediately after getIP so we prioritise the creation of the IPAddress, but we'll still try to tag the newly created FIP even if we exit with an error.
* Wrap the Create() call below in a client-go.util.retry.OnError(), or something better that takes a context
* update getIP() to look for an existing tagged FIP which isn't allocated to a member of the pool before creating a new one.

controllers/openstackfloatingippool_controller.go
mdbooth: See above. Lets consider wrapping this in some retry logic to reduce the chances of a resource leak if we can't create the IPAddress.

controllers/openstackfloatingippool_controller.go
mdbooth: We should jump here if IPAddress already existed above.

controllers/openstackfloatingippool_controller.go
mdbooth: My previous comment may not have been clear: I think this update is good and you don't need to change anything.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+			if err = r.Client.Status().Update(ctx, &claim); err != nil {
+				log.Error(err, "Failed to update IPAddressClaim status", "claim", claim.Name, "ip", ip)
+				return ctrl.Result{}, err
+			}
+			scope.Logger().Info("Claimed IP", "ip", ip)
+		}
+	}
+	if err = r.setIPStatuses(ctx, pool); err != nil {
+		return ctrl.Result{}, err
+	}
```
mdbooth, line 181: This is redundant: any change which could have resulted in a change to IP statuses will result in another reconcile, in which case the first thing we will do is call setIPStatuses(). We don't need to do this twice in the same reconcile.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+					Address: ip,
+					Prefix:  32,
+				},
+			}
+
+			// If the ReclaimPolicy is Delete, we add the finalizer to the IPAddress if they are not pre-allocated
+			// this is to ensure that the IPAddress is deleted from openstack when the claim is deleted
+			if pool.Spec.ReclaimPolicy == infrav1.ReclaimDelete && !contains(pool.Spec.PreAllocatedFloatingIPs, ip) {
+				controllerutil.AddFinalizer(ipAddress, infrav1.DeleteFloatingIPFinalizer)
+			}
```
mdbooth, line 165: I'd prefer we always added a Finalizer. Lets put the cleanup logic explicitly in the delete flow of the controller rather than implicitly in the k8s GC.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&infrav1.OpenStackFloatingIPPool{}).
+		Watches(
+			&ipamv1.IPAddressClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.iPAddressClaimToPoolMapper),
+		).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
+		}).
```
mdbooth, line 404: As discussed, I think you can safely remove this.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+	}()
+
+	if err := r.reconcileFloatingIPNetwork(scope, pool); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.setIPStatuses(ctx, pool); err != nil {
+		return ctrl.Result{}, err
+	}
```
mdbooth, line 107: Observation: setIPStatuses() is already fetching and iterating over a full list of IPAddress objects in this pool. We could probably move the IPAddress controller into this controller.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+		// If the IP exist return it, else we continue and try to allocate it if we fail to allocate it we will mark it as failed
+		if fp != nil {
+			return fp.FloatingIP, nil
+		}
+	}
+
+	// ip could be empty meaning we want to get a new one or the IP
+	// if ip is not empty it got and ip from availableIPs that does not exist in openstack
+	// we try to allocate it, if we fail we mark it as failed and skip it next time
+	fp, err := networkingService.CreateFloatingIPForPool(pool, ip)
```
dulek, line 292: This will only work with OpenStack admin credentials. I don't see `CreateFloatingIPForPool` called with `""` as the second parameter. This feels wrong, I think we want this to be able to reserve a new IP automatically, even if it's Neutron choosing the address.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+	for _, item := range b {
+		delete(m, item)
+	}
+	result := make([]string, 0, len(m))
+	for item := range m {
+		result = append(result, item)
+	}
+	return result
+}
```
dulek, line 245: I wonder if there's a potential to use this: https://github.com/kubernetes/utils/blob/master/set/set.go

dulek (changes requested):

controllers/ipaddress_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	networkingService, err := networking.NewService(scope)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !ipAddress.ObjectMeta.DeletionTimestamp.IsZero() {
+		// If the IPAddress has more than one finalizer, it has not been released by machine yet and we should not delete it.
+		if controllerutil.ContainsFinalizer(ipAddress, infrav1.DeleteFloatingIPFinalizer) && len(ipAddress.GetFinalizers()) == 1 {
```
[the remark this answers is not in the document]
bilbobrovall: I'm not 100% sure about how that use case would look and it might only be the comment that needs to be changed. My reasoning was that if there's another finalizer on the address it's probably being used or something wants to keep it and does not want it deleted from openstack thus waiting for the "DeleteFloatingIPFinalizer" to be the last finalizer left feels reasonable.
As we are trying to keep the infra provider and ipam provider as separate as possible and follow the api spec I feel like expecting a finalizer thats not managed by the ipam provider would counteract those things

controllers/ipaddress_controller.go
```diff
@@ -0,0 +1,128 @@
...
+			if err = networkingService.DeleteFloatingIP(pool, ipAddress.Spec.Address); err != nil {
```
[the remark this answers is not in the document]
bilbobrovall: Relates to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1763/files#r1424084068 I'll move the conditional logic from the pool controller to here :+1:

controllers/openstackfloatingippool_controller.go
bilbobrovall: We could, but we would need to bump k8s.io/utils and  k8s.io/utils/pointer is deprecated in favor of k8s.io/utils/ptr so that would involve some refactoring and should probably be done in a separate PR

pkg/cloud/services/networking/floatingip.go
```diff
@@ -70,6 +71,25 @@ func (s *Service) GetOrCreateFloatingIP(eventObject runtime.Object, openStackClu
...
 }
 
+func (s *Service) CreateFloatingIPForPool(pool *infrav1.OpenStackFloatingIPPool, ip string) (*floatingips.FloatingIP, error) {
+	var fpCreateOpts floatingips.CreateOpts
+
+	fpCreateOpts.FloatingNetworkID = pool.Status.FloatingIPNetwork.ID
+	fpCreateOpts.Description = fmt.Sprintf("Created by cluster-api-provider-openstack OpenStackFloatingIPPool %s", pool.Name)
+	if ip != "" {
+		fpCreateOpts.FloatingIP = ip
+	}
```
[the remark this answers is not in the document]
bilbobrovall: That was mostly copy paste from another function, but I realized we wanted to get rid of all the functionality dependant on an admin credentials and it should be removed now.

controllers/ipaddress_controller.go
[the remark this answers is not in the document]
dulek: Okay, I see now, cool!

controllers/openstackfloatingippool_controller.go
dulek: Agreed!

bilbobrovall:
I've rebuilt quite abit now @mdbooth and @dulek and I think everything should be fixed now.

The big things that needs to be reviews is probably that I removed the ipaddress controller and combined the logic with the setIPStatuses method in a new reconcileIPAddresses in the pool controller and the logic for preventing orphaned ips, I tested It a couple of times and made both the tagging and creation of the IPAddress object fail 80% of the time without getting any orphaned addresses. But it's off course still possible if the controller would crash right after the FIP was created or if both the openstack and kubernetes api went down in the same window. And it looks like the only way around that would be using a temporary port and that would feel abit over engineered imo.

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,471 @@
...
+		scope.Logger().Error(err, "Failed to get floating IPs by tag", "pool", pool.Name)
+		return "", err
+	}
+	for _, taggedIP := range taggedIPs {
+		if contains(pool.Status.AvailableIPs, taggedIP.FloatingIP) || contains(pool.Status.ClaimedIPs, taggedIP.FloatingIP) {
+			continue
+		}
+		scope.Logger().Info("Tagged floating IP found that was not known to the pool, adding it to the pool", "ip", taggedIP.FloatingIP)
+		pool.Status.AvailableIPs = append(pool.Status.AvailableIPs, taggedIP.FloatingIP)
+	}
```
mdbooth, line 326: We probably don't want to do this unconditionally in the IPAddressClaim loop for every unallocated IPAddressClaim. Suggest we could do it only if len(AvailableIPs) == 0?

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,471 @@
...
+							Kind:     pool.Kind,
+							Name:     pool.Name,
+						},
+						Address: ip,
+						Prefix:  32,
+					},
+				}
+
+				// Retry creating the IPAddress object
+				err = wait.ExponentialBackoff(backoff, func() (bool, error) {
```
mdbooth, line 174: This should be cancellable:
```suggestion
				err = wait.ExponentialBackoffWithContext(ctx, backoff, func() (bool, error) {
```

controllers/openstackfloatingippool_controller.go
```diff
@@ -0,0 +1,471 @@
...
+		conditions.MarkFalse(pool, infrav1.OpenstackFloatingIPPoolReadyCondition, infrav1.OpenStackErrorReason, clusterv1.ConditionSeverityError, "Failed to create floating IP: %v", err)
+		if ip != "" {
+			pool.Status.FailedIPs = append(pool.Status.FailedIPs, ip)
+		}
+		return "", err
+	}
+	defer func() {
+		tag := pool.GetFloatingIPTag()
+
+		err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
mdbooth, line 355: As above, please use ExponentialBackoffWithContext so this is cancellable.

mdbooth (commented):
I know you need to move forward with this and there are a also number of other folks (myself included) really looking forward to it. I suspect it may still need a few rounds to stabilise, but I don't want to hold it up.

My suggestion is that we move the pool API to v1alpha1 so we don't have to try to hold it to the same stability guarantees we're urgently trying to get the rest of the API to. If it helps, I already did this in this branch: https://github.com/shiftstack/cluster-api-provider-openstack/tree/openstack_floating_ip_pool. Feel free to cherry pick the top commit. It's purely mechanical, so there's no need to co-author me when you squash commits before merging.

There's one more commit in there which I've submitted separately as https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1825. You'll need that so that `make generate` doesn't panic, but I'd just rebase on top of it.

Once we've done that I'm inclined to merge this fairly quickly. We've declared the API as subject to change, and the codebase is fairly independent. My preference is to merge it and iterate on it.

@lentzi90 @jichenjc @dulek thoughts?

dulek:
I can agree with Matt here, this doesn't need to be 100% perfect. I don't understand what v1alpha1 changes though.

lentzi90 (commented):
Nice!
This looks quite good already and I agree with @mdbooth that we can probably take it in already under the v1alpha1 API version. Found a few nits though

api/v1alpha1/openstackfloatingippool_types.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	// ReclaimDelete is the reclaim policy for floating ips.
+	ReclaimDelete ReclaimPolicy = "Delete"
+	// ReclaimRetain is the reclaim policy for floating ips.
+	ReclaimRetain ReclaimPolicy = "Retain"
+)
+
+// OpenStackFloatingIPPoolSpec defines the desired state of OpenStackFloatingIPPool.
+type OpenStackFloatingIPPoolSpec struct {
+	PreAllocatedFloatingIPs []string `json:"preAllocatedFloatingIPs,omitempty"`
```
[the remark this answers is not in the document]
dulek: Little formatting fixes, plus I think this should have `+optional`:

```
	// PreAllocatedFloatingIPs is a list of preallocated floating IPs precreated in OpenStack that should be used by this pool.
	// These are used before allocating new ones and are not deleted from OpenStack when the pool is deleted.
```

dulek:
This looks pretty good to me!

api/v1alpha1/openstackfloatingippool_types.go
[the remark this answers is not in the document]
mdbooth: @dulek FYI I've been recommending against `+optional`s because they're already the default. I don't have a strong position on it, though. Is there specific guidance somewhere I'm ignoring?

mdbooth:
Once the last remaining review feedback is addressed (is it just the docstring?) please can you squash these commits and I'll approve it.

bilbobrovall:
> Once the last remaining review feedback is addressed (is it just the docstring?) please can you squash these commits and I'll approve it.

Great, fixed and squashed :+1: , I'll be looking into rebasing https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1762 when this is merged, how should I think about the apiversion on that one I guess I should just move it to alpha8?

mdbooth:
/approve

I'll leave lgtm to somebody else.

lentzi90 (commented):
/lgtm

api/v1alpha1/openstackfloatingippool_types.go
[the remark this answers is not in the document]
dulek: Okay, so just the formatting fixes! :)

PR #1765, opened 2023-11-29
✨ WIP - Reconcile Machines & Managed Security Groups

EmilienM:

Before this patch, if Managed Security Groups would be enabled,
CAPO would create them and the OpenStack Machines would get these
security groups at instance creation time.

Now, we are updating the ports of the machines to ensure they match the
desired security groups that we want assigned.

For now we only have Control Plane & Worker Security groups but in the
future we'll have Bastion, CNI and also an interface to add more
Security Groups.

We'll need to reconcile the machines to ensure they have the security
group assigned.

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha5/openstackmachine_types.go
+++ b/api/v1alpha5/openstackmachine_types.go
@@ -110,6 +110,9 @@ type OpenStackMachineStatus struct {
 	// +optional
 	InstanceState *InstanceState `json:"instanceState,omitempty"`
 
+	// SecurityGroups is the list of security groups IDs associated with this machine.
+	SecurityGroups []string `json:"securityGroups,omitempty"`
+
 	FailureReason *errors.MachineStatusError `json:"failureReason,omitempty"`
 
 	// FailureMessage will be set in the event that there is a terminal problem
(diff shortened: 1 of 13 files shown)


EmilienM:
* I'm missing the watch on the controller to watch for OpenStackMachine events, this will likely make CI failing I guess
* We probably need a dedicated controller for managed security groups, with watches on both cluster & machines

api/v1alpha5/openstackmachine_types.go
```diff
@@ -110,6 +110,9 @@ type OpenStackMachineStatus struct {
 	// +optional
 	InstanceState *InstanceState `json:"instanceState,omitempty"`
 
+	// SecurityGroups is the list of security groups IDs associated with this machine.
+	SecurityGroups []string `json:"securityGroups,omitempty"`
```
jichenjc, line 114: um... I think for new features we tend to only on latest API ?

EmilienM: yeah this was just for getting a build. Like I wrote in a comment above, I couldn't get the conversion working yet so I did this.

BTW I think this whole patch will go away, I was just experimenting a bit.
Now I believe a new controller is much preferred.

PR #1766, opened 2023-12-01
🌱  Make it possible to use a pre-created volume type in e2e tests 

### LINKED ISSUE #1700: E2e: Enable bring your own OpenStack
/kind feature

Our e2e tests currently require OpenStack admin credentials. My understanding is that it is only required for creating the volume type here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c73d55dd47006207a15c93bbae69811c97274827/test/e2e/suites/e2e/e2e_test.go#L861-L880

If possible, I would like to change this so that the volume type can be pre-existing and configured like the [other OpenStack variables](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/c73d55dd47006207a15c93bbae69811c97274827/test/e2e/data/e2e_conf.yaml#L175-L189). This would make it possible to run the e2e tests without OpenStack admin credentials, e.g. against a public OpenStack cloud.

This should be a pretty good first issue.
/help

maxrantil:
Fixes #1700 

**TODOs**:
- [ X] squashed commits

/hold

### DIFF
--- a/hack/ci/cloud-init/controller.yaml.tpl
+++ b/hack/ci/cloud-init/controller.yaml.tpl
@@ -150,6 +150,14 @@
         openstack aggregate add host "$aggregateid" "$host"
     done
 
+    # Create the volume type
+    VOLUME_TYPE_NAME="test-volume-type"
+    if openstack volume type create --description "Test volume type" --public "${VOLUME_TYPE_NAME}" &> /dev/null; then
+        echo "Volume type '${VOLUME_TYPE_NAME}' created successfully."
+    else
+        echo "Error: Failed to create volume type '${VOLUME_TYPE_NAME}'."
+    fi
+
     # the flavors are created in a way that we can execute at least 2 e2e tests in parallel (overall we have 32 vCPUs)
     openstack flavor delete m1.tiny
     openstack flavor create --ram 512 --disk 1 --ephemeral 1 --vcpus 1 --public --id 1 m1.tiny --property hw_rng:allowed='True'
--- a/test/e2e/suites/e2e/e2e_suite_test.go
+++ b/test/e2e/suites/e2e/e2e_suite_test.go
@@ -52,7 +52,6 @@ func TestE2E(t *testing.T) {
 
 var _ = SynchronizedBeforeSuite(func() []byte {
 	data := shared.Node1BeforeSuite(e2eCtx)
-	createTestVolumeType(e2eCtx)
 	return data
 }, func(data []byte) {
 	shared.AllNodesBeforeSuite(e2eCtx, data)
--- a/test/e2e/suites/e2e/e2e_test.go
+++ b/test/e2e/suites/e2e/e2e_test.go
@@ -28,10 +28,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
-	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumetypes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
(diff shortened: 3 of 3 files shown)


maxrantil:
/cc @lentzi90 
Is this what you had in mind for the issue? Im not sure about the conf variable name that is being used. Should we use a new variable name?

lentzi90 (commented):
Great!
I think we could even do so that if the volume type does not exist we just error out directly, without trying to create it. For the CI, we can pre-create it [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/8a49b9f23c3a8bd2f41c4e071213edc8bc8ede7b/hack/ci/cloud-init/controller.yaml.tpl#L153-L166) together with the flavors.

/ok-to-test

dulek:
Can you also rename the PR to say it's about a volume type? I got really confused at first.

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -150,6 +150,22 @@
         openstack aggregate add host "$aggregateid" "$host"
     done
 
+    VOLUME_TYPE_NAME="test-volume-type"
+    # Create the volume type and check if it's successful
+    if openstack volume type create --description "Test volume type" --public "$VOLUME_TYPE_NAME"; then
+        echo "Volume type '$VOLUME_TYPE_NAME' created successfully.
+        # Fetch the volume type ID
+        export VOLUME_TYPE_ID=$(openstack volume type list 2>/dev/null | grep " $VOLUME_TYPE_NAME " | awk '{print $2}')
```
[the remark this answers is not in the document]
maxrantil: Thank you for the feedback! I will for sure change it after your suggestion when I get it to work correctly. I have some extra verbose during development only.

test/e2e/suites/e2e/e2e_suite_test.go
```diff
@@ -52,7 +52,7 @@ func TestE2E(t *testing.T) {
 
 var _ = SynchronizedBeforeSuite(func() []byte {
 	data := shared.Node1BeforeSuite(e2eCtx)
-	createTestVolumeType(e2eCtx)
+	getVolumeType(e2eCtx)
```
lentzi90, line 55: Is this a sanity check or do we actually need it?
I would prefer to get rid of it

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -891,23 +890,43 @@ func isCloudProviderInitialized(taints []corev1.Taint) bool {
...
+		if volumeType.Name == volumeTypeName {
+			return volumeType.ID, nil
+		}
+	}
+
+	return "", fmt.Errorf("volume type '%s' not found", volumeTypeName)
+}
+
+func getVolumeType(e2eCtx *shared.E2EContext) {
 	providerClient, clientOpts, _, err := shared.GetAdminProviderClient(e2eCtx)
```
lentzi90, line 914: Do we need admin privileges? I was hoping we can get rid of them completely so we can run the tests also when we do not have admin access

lentzi90 (commented):
This looks good overall. I'm not understanding why we have the `getVolumeType` though, and we should preferably not use admin credentials at all

test/e2e/suites/e2e/e2e_test.go
maxrantil: No it seems to work fine when using the function `GetTenantProviderClient` instead. I got lost in the code there and forgot about this aspect of the issue. Thanks for reminding me.

test/e2e/suites/e2e/e2e_suite_test.go
maxrantil: As I understand it, we need this if you're considering using a different volume type than our default `test-volume-type`. However, my tests show that removing the function doesn't cause issues. But, do we want the flexibility to specify a different volume type in the config file?

lentzi90: You mean `createTestVolumeType`? I'm wondering what the point is with `getVolumeType`? My understanding is that it just checks that the volume is there but I may have missed something. Does it get the volume type ID so that can be used later?

maxrantil: You are right, `getVolumeType` was only for confirming the volume type.

test/e2e/suites/e2e/e2e_test.go
maxrantil: No need for neither.

maxrantil:
You are right @lentzi90 the function is now completely gone.

/cc @jichenjc @mdbooth @dulek

lentzi90 (commented):
Great job!
/approve

PR #1768, opened 2023-12-04
🐛 Fall back to cluster identityRef in absence of machine identityRef

### LINKED ISSUE #1759: No obvious failure if `identityRef` is omitted from `OpenStackMachineTemplate`
/kind bug

Create an `OpenStackMachineTemplate` that omits the `spec.template.spec.identityRef` field. The machine fails to be created and we see the following error from the controller:

```
E1127 16:20:23.342436       1 controller.go:324] "Reconciler error" err="auth option failed for cloud : Missing input for argument [auth_url]" controller="openstackmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackMachine" OpenStackMachine="***/***" namespace="***" name="***" reconcileID=***
```

Ideally we could attempt to retrieve credentials from the cluster. In any case, if no identity ref is provided we should provide a clear failure rather than silently ignoring the issue and failing later.

None.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.7.0
- Cluster-API version:
- OpenStack version: Wallaby
- Minikube/KIND version: N/A
- Kubernetes version (use `kubectl version`): v1.28.2

stephenfin:

The `identityRef` attribute is marked as optional but without it we have no ability to talk to the cloud. In a future API version, we may wish to make this a required attribute but for now, provide the ability to retrieve credentials from the cluster in the absence of the machine.

Fixes #1759

~I'm not sure if we can do this. As things stand, this will cause a change in behavior (you can create a Machine without specifying an `identityRef`) between two different CAPO deployments regardless of API version selected. It might be better to instead explode early (rather than exploding later) and make this field required in API v1alpha8/v1beta1? Open to input here.~

(We discussed the above :point_up_2: on [Slack](https://kubernetes.slack.com/archives/CFKJB65G9/p1701773862641579) and agreed it wasn't an issue)

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha7/openstackmachine_types.go
+++ b/api/v1alpha7/openstackmachine_types.go
@@ -89,7 +89,8 @@ type OpenStackMachineSpec struct {
 	// The server group to assign the machine to
 	ServerGroupID string `json:"serverGroupID,omitempty"`
 
-	// IdentityRef is a reference to a identity to be used when reconciling this cluster
+	// IdentityRef is a reference to a identity to be used when reconciling this cluster.
+	// If not specified, the identity ref of the cluster will be used instead.
 	// +optional
 	IdentityRef *OpenStackIdentityReference `json:"identityRef,omitempty"`
 }
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
@@ -3849,7 +3849,8 @@ spec:
                         type: string
                       identityRef:
                         description: IdentityRef is a reference to a identity to be
-                          used when reconciling this cluster
+                          used when reconciling this cluster. If not specified, the
+                          identity ref of the cluster will be used instead.
                         properties:
                           kind:
                             description: Kind of the identity. Must be supported by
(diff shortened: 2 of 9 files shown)


jichenjc:
>>>I'm not sure if we can do this. As things stand, this will cause a change in behavior (you can create a Machine without specifying an identityRef) between two different CAPO deployments regardless of API version selected. It might be better to instead explode early (rather than exploding later) and make this field required in API v1alpha8/v1beta1? Open to input he

stephenfin:
> > > > I'm not sure if we can do this. As things stand, this will cause a change in behavior (you can create a Machine without specifying an identityRef) between two different CAPO deployments regardless of API version selected. It might be better to instead explode early (rather than exploding later) and make this field required in API v1alpha8/v1beta1? Open to input here
> 
> maybe we can talk at original issue, but I think it's better to work as you proposed (fall back to cluster identityRef) but this won't need a API change to v1alpha8/v1beta1 as it's optional anyway per your proposal above?

Yeah, I discussed this with [@mdbooth on Slack](https://kubernetes.slack.com/archives/CFKJB65G9/p1701773862641579) and it sounds like an non-issue. We can probably proceed with this as-is, assuming there are no other issues?

mdbooth (commented):
This is an easy win. Lets get it out of the way.

/approve
/lgtm
/hold cancel

PR #1769, opened 2023-12-04
✨ Allow AZs to be Omitted at Runtime

spjmurray:
CAPI appears to make implicit CP scheduling decisions based on what it's told is available by CAPO on an "LRU" basis. It also assumes an infinite sized AZ, so problems begin when the "next" AZ cannot accommodate the VM. We could manually specify an AZ that aggregates all machines explcitly, however this is another mechanism that disables scheduling by CAPI altogether and allows Nova to do what it does, along with a soft-AA rule. However, switching from CAPI scheduling to Nova scheduling is impossible as the field is immutable, so allow this.  Testing shows existing scheduled clusters undergo no topology changes, which will be due to the KCPM not taking action, but you can force the changes with a rolling upgrade of some variety. Crucially, if a cluster with CAPI scheduling gets stuck, we can modify to Nova scheduling and it should pick up the new specification and get past the hurdle.

As above

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1alpha8/openstackcluster_webhook.go
+++ b/api/v1alpha8/openstackcluster_webhook.go
@@ -130,6 +130,11 @@ func (r *OpenStackCluster) ValidateUpdate(oldRaw runtime.Object) (admission.Warn
 	old.Spec.ControlPlaneAvailabilityZones = []string{}
 	r.Spec.ControlPlaneAvailabilityZones = []string{}
 
+	// Allow the scheduling to be changed from CAPI managed to Nova and
+	// vice versa.
+	old.Spec.ControlPlaneOmitAvailabilityZone = false
+	r.Spec.ControlPlaneOmitAvailabilityZone = false
+
 	// Allow change to the allowAllInClusterTraffic.
 	old.Spec.AllowAllInClusterTraffic = false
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/ok-to-test

lentzi90 (commented):
This seems reasonable to me
/lgtm

/cc @mdbooth

spjmurray:
Yes Sir Mr Robot 🖖🏻

spjmurray:
So looking at the first set of test results, I do see one of the tests failing due to quota limits being exceeded, that may affect the other one that's happening in parallel, but the machine is "running" apparently, just no logs from it or anything.  Need to dig a bit deeper...

lentzi90:
> So looking at the first set of test results, I do see one of the tests failing due to quota limits being exceeded, that may affect the other one that's happening in parallel, but the machine is "running" apparently, just no logs from it or anything. Need to dig a bit deeper...

Thanks for mentioning! This may be why the tests have been quite flaky recently :slightly_frowning_face: 
Maybe we need to look at ways to increase the quota or limit which tests run in parallel...

PR #1770, opened 2023-12-04
 🐛 Verify success of parsing OpenStack cloud cacert

### LINKED ISSUE #1771: Misleading requirement to set OPENSTACK_CLOUD_CACERT_B64 when endpoint is https
/kind bug

In case the OpenStack cloud is configured with a known certificate that is part of the certificate bundle supplied with the OS, it's possible that the `cacert` is not set on the `clouds.yaml `file, which will make the `OPENSTACK_CLOUD_CACERT_B64` to contain [the default value of empty string encoded](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/v0.8.0/templates/env.rc#L171-L175). When connecting to OpenStack API the value of `OPENSTACK_CLOUD_CACERT_B64` [would be used and would cause failure to connect to it](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/v0.8.0/pkg/scope/provider.go#L232-L235). 

A validation should be performed on the cert and if not valid attempt using the host's root CA.

MaysaMacedo:

If no certificate is set to `OPENSTACK_CLOUD_CACERT_B64` env variable, it defaults to an empty string encoded. This can cause failure to connect to OpenStack API when it uses a certificate bundle that is supplied by the OS, instead of a private certificate. This commit fixes the issue by checking if the certificate was parsed successfully and if not attempts to use the host's root CA.

Fixes #1771

/hold

### DIFF
--- a/pkg/scope/provider.go
+++ b/pkg/scope/provider.go
@@ -165,7 +165,11 @@ func NewProviderClient(cloud clientconfig.Cloud, caCert []byte, logger logr.Logg
 	}
 	if caCert != nil {
 		config.RootCAs = x509.NewCertPool()
-		config.RootCAs.AppendCertsFromPEM(caCert)
+		ok := config.RootCAs.AppendCertsFromPEM(caCert)
+		if !ok {
+			// If no certificates were successfully parsed, set RootCAs to nil to use the host's root CA
+			config.RootCAs = nil
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

stephenfin:
/lgtm
Nice catch

PR #1774, opened 2023-12-05
✨ Reconcile SecurityGroups <-> Instances separately

EmilienM:

Originally, the openstackmachine controller was adding the SecurityGroups to the instanceSpec.
Now, the openstackmachine controller will first create the instances
without security groups. Then we have a loop that will reconcile the
desired security groups that we want to apply in the cluster[1] with the
ones that are added to the instance's ports.

[1] for now, we have 2 sets of rules: the one that can be added to
  `OpenStackMachine.Spec.SecurityGroups` and the one managed
  by `OpenStackCluster.Spec.ManagedSecurityGroups`.

**TODOs**:

- [ ] Handle what to do with OpenStackMachineSpec.Ports is defined with a SecurityGroupFilter. Probably I need to rework all work in Status and puts Ports in Status.
- [ ] For an update to applied security groups, think about how we could indicate in the cluster status as a summary of how many machines have been updated.
- [ ] add unit tests to at least the openstackmachine_controller
- [ ] Handle "default" security group applied to all created ports by passing an empty slice of strings if no security groups is required when creating the port.
- [ ] update documentation
- [ ] squashed commits


api/v1alpha7/openstackmachine_types.go
```diff
@@ -107,6 +107,11 @@ type OpenStackMachineStatus struct {
 	// +optional
 	InstanceState *InstanceState `json:"instanceState,omitempty"`
 
+	// SecurityGroups is the list of security groups that were applied to the ports
```
EmilienM, line 110: Note for myself: in comment I'll explain that we only populate that once all the rules have been applied.

Note for reviewers: do we want to have another field `SecurityGroupsReady` once the reconciler has finished its job successfully? Now, let's say we managed to successfully reconcile once and `Status.SecurityGroups` was populated once, then somehow the reconciler fails but we still have `Status.SecurityGroups`, we have [mark](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1774/files#diff-e0a0984c0e26655878e2f94623fbf8cddb4d8a89e3443c1296dfaadc8802838aR339) the instance as NotReady. Is that good enough?

controllers/openstackmachine_controller.go
```diff
@@ -334,6 +334,12 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 		return ctrl.Result{}, err
 	}
 
+	err = r.reconcileSecurityGroupsToInstance(scope.Logger(), openStackCluster, machine, openStackMachine, instanceStatus, computeService, networkingService)
+	if err != nil {
+		conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.SecurityGroupApplyFailedReason, clusterv1.ConditionSeverityError, "Applying security groups failed: %v", err)
```
EmilienM, line 339: See question [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1774/files#r1416543856).

controllers/openstackmachine_controller.go
```diff
@@ -430,6 +436,56 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+func (r *OpenStackMachineReconciler) reconcileSecurityGroupsToInstance(logger logr.Logger, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine, openStackMachine *infrav1.OpenStackMachine, instanceStatus *compute.InstanceStatus, computeService *compute.Service, networkingService *networking.Service) error {
+	logger.Info("Reconciling security groups to instance")
+
+	desiredSecurityGroups := []*infrav1.SecurityGroup{}
+
+	// These security groups are currently not managed by CAPO and provided by the user
+	// Later we'll have an API to manage them.
+	for _, sg := range openStackMachine.Spec.SecurityGroups {
+		if sg.ID == "" {
+			// lookup security group by name
```
EmilienM, line 448: Note for myself: I can remove that comment

controllers/openstackmachine_controller.go
```diff
@@ -430,6 +436,56 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+			securityGroup, err := networkingService.GetSecurityGroupByName(sg.Name)
+			if err != nil {
+				return fmt.Errorf("get security group %q: %w", sg.Name, err)
```
EmilienM, line 451:
```suggestion
				return fmt.Errorf("get security group by name %q: %w", sg.Name, err)
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,67 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+		if err != nil {
+			return fmt.Errorf("error getting port %s: %v", portID, err)
+		}
+
+		// check if port already has desired security groups
+		for _, securityGroupID := range desiredSecurityGroupIDs {
+			if !contains(port.SecurityGroups, securityGroupID) {
+				logger.Info("Port doesn't have desired security group", "portID", portID, "securityGroupID", securityGroupID)
+				portUpdateNeeded = true
+				break
```
EmilienM, line 428: Note for myself: I'll remove the `break` so we get all logs in the loop. We know we need to update the port anyway at this point.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,67 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+			_, err := s.client.UpdatePort(portID, portOpts)
+			if err != nil {
+				return fmt.Errorf("error updating port %s: %v", portID, err)
+			}
+		}
+	}
+
+	appliedSecurityGroups := make([]infrav1.SecurityGroup, len(desiredSecurityGroupIDs))
+	for i, securityGroupID := range desiredSecurityGroupIDs {
+		securityGroup, err := s.client.GetSecGroup(securityGroupID)
```
EmilienM, line 454: Note for myself: I could probably get rid of this call if we directly pass `[]infrav1.SecurityGroup` to this function instead of the security groups IDs that we desire. I'll take a look.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,67 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+		if err != nil {
+			return fmt.Errorf("error getting security group %s: %v", securityGroupID, err)
+		}
+		appliedSecurityGroups[i] = *convertOSSecGroupToConfigSecGroup(*securityGroup)
+	}
+	logger.Info("Updating openStackMachine.Status.SecurityGroups", "securityGroups", appliedSecurityGroups)
+	openStackMachine.Status.SecurityGroups = appliedSecurityGroups
```
EmilienM, line 461: Note to the reviewers:

Here we update Status, if we have applied the desired security groups successfully. However if for some reason the managed control plane security groups hasn't been created yet (and the instance is being created), the desired security groups won't have the control plane one, so here we'll be satisfy and update Status while in reality we want a machine with the control plane security group applied.

So I'm wondering whether we should wait before for the end of managed security groups reconciliation before calling `ReconcileSecurityGroupsToInstanceAttachedInterfaces`.

controllers/openstackmachine_controller.go
```diff
@@ -430,6 +436,56 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+	logger.Info("Reconciling security groups to instance")
+
+	desiredSecurityGroups := []*infrav1.SecurityGroup{}
+
+	// These security groups are currently not managed by CAPO and provided by the user
+	// Later we'll have an API to manage them.
+	for _, sg := range openStackMachine.Spec.SecurityGroups {
+		if sg.ID == "" {
+			// lookup security group by name
+			securityGroup, err := networkingService.GetSecurityGroupByName(sg.Name)
```
mdbooth, line 449: I'm not sure about this for a couple of reasons:
* We're making an OpenStack API call to determine if we need to update (i.e. in a 'hot path')
* The non-uniqueness of names in OpenStack

I like the way we handle these in the cluster controller for network and subnet resources: we 'resolve' resources specified by filter during initial reconciliation and write their UUID to the status. Subsequently we just refer to the status.

If we applied that here, we'd have a new list of security group uuids in the status which is separate to `AppliedSecurityGroups` which we would resolve *before* machine creation. This list would be as immutable as the machine spec, so we would not need to consider updates to it.

So:
```
if len(spec.SecurityGroups) != len(status.MachineSecurityGroups) {
  status.MachineSecurityGroups = resolve(spec.SecurityGroups)
}

... subsequently only ever use status.MachineSecurityGroups
```

In this way we would only look them up once, and we wouldn't break if the user did anything silly like renaming one or creating another one with the same name.

controllers/openstackmachine_controller.go
```diff
@@ -430,6 +436,56 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+
+	if openStackCluster.Spec.ManagedSecurityGroups {
+		if util.IsControlPlaneMachine(machine) && openStackCluster.Status.ControlPlaneSecurityGroup != nil {
+			desiredSecurityGroups = append(desiredSecurityGroups, openStackCluster.Status.ControlPlaneSecurityGroup)
+		} else if openStackCluster.Status.WorkerSecurityGroup != nil {
+			desiredSecurityGroups = append(desiredSecurityGroups, openStackCluster.Status.WorkerSecurityGroup)
+		}
+	}
+
+	if len(desiredSecurityGroups) > 0 {
```
mdbooth, line 468: This isn't right. It should be something like:
```
if normalised(desiredSecurityGroups) != normalised(status.AppliedSecurityGroups) {
```

where normalised() is some function which makes sets of security uuids comparable, e.g. an alphanumeric sort.

i.e. we should reconcile because they are different to what has been applied, not just because they exist. This will effectively *always* reconcile security groups, even when they haven't changed.

pkg/cloud/services/compute/instance.go
```diff
@@ -688,6 +689,10 @@ func (s *Service) DeleteInstance(openStackCluster *infrav1.OpenStackCluster, eve
 	return s.deleteInstance(eventObject, instanceStatus.InstanceIdentifier())
 }
 
+func (s *Service) GetAttachedInterfaces(instanceID string) ([]attachinterfaces.Interface, error) {
```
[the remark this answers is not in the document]
EmilienM: yeah I'm re-using it in the controller.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,67 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+				portUpdateNeeded = true
+				break
+			}
+		}
+		// check if port has security groups that are not desired
+		for _, securityGroupID := range port.SecurityGroups {
+			if !contains(desiredSecurityGroupIDs, securityGroupID) {
+				logger.Info("Port has security group that is not desired", "portID", portID, "securityGroupID", securityGroupID)
+				portUpdateNeeded = true
+				break
```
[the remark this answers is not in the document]
mdbooth: Hmm. Is it a bug that the default security group is attached to the ports?

api/v1alpha7/openstackmachine_types.go
EmilienM: > For machine creation I think it's clear that we would only set ready after security groups have been applied.

Exactly. That's what I meant when writing that code:

```golang
err = r.reconcileSecurityGroupsToInstance(scope.Logger(), openStackCluster, machine, openStackMachine, instanceStatus, computeService, networkingService)
if err != nil {
	conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.SecurityGroupApplyFailedReason, clusterv1.ConditionSeverityError, "Applying security groups failed: %v", err)
	return ctrl.Result{}, fmt.Errorf("add security groups to instance: %w", err)
}
```

> For an update to applied security groups it's less clear, and the CAPI definition doesn't indicate if setting it to false after having set it to true is meaningful or allowed. I would not be inclined to set it to false during a rollout of new managed security groups. I think this would be better indicated in the cluster status as a summary of how many machines have been updated.

ok I'll try to think how the state of how many machines have been updated can be reflected in the `OpenStackClusterStatus`.

controllers/openstackmachine_controller.go
EmilienM: > We're making an OpenStack API call to determine if we need to update (i.e. in a 'hot path')

I think the network/subnet/router implementation in the `openstackcluster_controller` [does it too](https://github.com/shiftstack/cluster-api-provider-openstack/blob/f9f1bae50f425eda850b145664f5d4352b437a92/controllers/openstackcluster_controller.go#L443-L444).

> we'd have a new list of security group uuids in the status which is separate to `AppliedSecurityGroupIDs` which we would resolve before machine creation. This list would be as immutable as the machine spec, so we would not need to consider updates to it.

Just to make sure I understand it 100%, this Status field would only be for the Security Groups specified via `OpenStackMachine.Spec.SecurityGroup` right? If yes, I agree that makes sense and I'll make the change.

> In this way we would only look them up once, and we wouldn't break if the user did anything silly like renaming one or creating another one with the same name.

Very good point.

> `if len(spec.SecurityGroups) != len(status.MachineSecurityGroups)`

Here I'm confused whether you want to compare the list of just their length. But I think you want to compare length indeed, since the `status.MachineSecurityGroups` is immutable (such as the `OpenStackMachine.Spec` too anyway).

Most of this comment is just making sure I understand what you have in mind correctly, thanks :pray: !

pkg/cloud/services/networking/securitygroups.go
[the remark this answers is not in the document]
EmilienM: Ports that are created when port security is enabled will have the `default` security group itself managed by Neutron.

I see two options:
* We keep it and look for its ID (once and early to avoid many Neutron API calls) and put it in `desiredSecuritygroupIDs`, therefore it's not a problem anymore.
* We remove it in this loop

I personally prefer option 1 for safety & because that's default OpenStack behaviours here.

cc @stephenfin with your OpenStack dev hat, any preference?

controllers/openstackmachine_controller.go
mdbooth: I don't think that would work because DeepEqual() won't return true for 2 slices with the same elements in different orders, which is going to be the primary thing you need to normalise. Do we need to address duplicate entries at all?

You can use the slices package for this, btw: https://pkg.go.dev/slices. slices.Sort() will sort them, and slices.Equal() will compare them, and both are safer and faster to use than reflection.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,56 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+		// check if port has security groups that are not desired
+		for _, securityGroupID := range port.SecurityGroups {
+			if !contains(desiredSecurityGroupIDs, securityGroupID) {
+				logger.Info("Port has security group that is not desired", "portID", portID, "securityGroupID", securityGroupID)
+				portUpdateNeeded = true
+			}
+		}
+
+		if portUpdateNeeded {
+			logger.Info("Updating port", "portID", portID, "securityGroups", desiredSecurityGroupIDs)
```
mdbooth, line 439: Maybe `V(3)`?

For reference, see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +402,56 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
+		}
+	}
+	return false
+}
+
+func (s *Service) ReconcileSecurityGroupsToInstanceAttachedInterfaces(logger logr.Logger, openStackMachine *infrav1.OpenStackMachine, attachedInterfaces []attachinterfaces.Interface, desiredSecurityGroupIDs []string) error {
+	for _, iface := range attachedInterfaces {
+		portUpdateNeeded := false
+		portID := iface.PortID
+		port, err := s.client.GetPort(portID)
```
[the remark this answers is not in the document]
EmilienM: very good idea, I always forget about the `deviceID` :facepalm:

pkg/cloud/services/networking/securitygroups.go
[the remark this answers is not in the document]
stephenfin: tbh, until today I didn't know that adding a port would add the default security group. What is the source of these ports? If they are ports that we are creating ourselves then we can instruct neutron not to assign a security group to the port. From openstackclient this is done by e.g.:

```
openstack port create --no-security-group ...
```

which under the hood sets the `security_group_ids` field to `[]` (empty list) when creating the port. Could we do the same thing?

If these ports are coming from elsewhere then it's possible to have other security groups than the default one assigned to the port (`openstack port create --security-group <sec-group> ...`) so I'm not sure how we can ignore everything? Do we forbid specifying ports with security groups defined?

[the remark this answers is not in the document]
stephenfin: Given the above, we probably need to decide what's authoritative: the security groups on the port, or the security ports defined on the `Machine`. I assume the latter, but if this is the case then we're going to modify something that the user gave us. If we go with the former, we're either modifying or ignoring something *else* that the user gave us.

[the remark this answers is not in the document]
EmilienM: > What is the source of these ports?

This is where we pre-build ports before creating an instance:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7961f21443918d50f2567a724b5ceeec4b445428/pkg/cloud/services/compute/instance.go#L177-L222

We don't pass any security groups.

> Could we do the same thing?

I don't think so because we don't provide a list but a list of `SecurityGroupFilter`, so I suspect that nothing will be passed to Neutron API if we pass an empty list of filters.

> we probably need to decide what's authoritative:

I want CAPO to have the authority if possible: if users provide a security group for the Machine, then we add it to the port in the reconcile (not at instance creation) - so instance + initial ports have no SGs in theory.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -399,6 +401,38 @@ func (s *Service) createRule(r infrav1.SecurityGroupRule) (infrav1.SecurityGroup
...
 }
 
+func (s *Service) ReconcileSecurityGroupsToInstanceAttachedPorts(openStackMachine *infrav1.OpenStackMachine, attachedPorts []string, desiredSecurityGroupIDs []string) error {
+	for _, portID := range attachedPorts {
+		portUpdateNeeded := false
+		port, err := s.client.GetPort(portID)
+		if err != nil {
+			return fmt.Errorf("error getting port %s: %v", portID, err)
+		}
+
```
EmilienM, line 411: Instead of doing https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1774/commits/1593e43e66b1f8a912c158e6df15b354f6d7cc10 - we could:

* Put a new struct which maps ports and their security group IDs in OpenStackMachine.Status so we do the lookup once
* we don't want to add these SG IDs to the slices in the openstackmachine_controller that we created for SG IDs to reconcile because they are specific to ports and not applied for all ports (unlike ManagedSecurityGroups & Spec.SecurityGroups)
* Here in the reconcile, if a port is found in the Status with SG IDs, we simply add them to the desired SGs. 

This is a bit complex but this is to address the 3 current options that we have to attach security groups to machines:
* OpenStackCluster.Spec.ManagedSecurityGroups
* OpenStackMachine.Spec.SecurityGroups
* OpenStack.Machine.Spec.Ports.SecurityGroupFilters

EmilienM: Another option is we don't reconcile the security groups from ports defined in `OpenStack.Machine.Spec.Ports` because we consider them immutable. That feels odd to me, and I can't find yet a good option here that doesn't bring complexity.

pkg/cloud/services/networking/securitygroups.go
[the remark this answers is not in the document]
EmilienM: I checked with @pierreprinetti and the good news is that `port.CreateOpts.SecurityGroups` in Gophercloud is `*[]string` so if we pass an empty slice of strings, the port should be created with no security group. I'll test it.

EmilienM:
I'm taking a break from this PR as I figured that I really want the ports to be managed *outside* of the instance creation, see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1788
That will simply the workflow:

Reconcile Ports -> Reconcile Security groups -> Reconcile Instances

EmilienM:

PR #1779, opened 2023-12-07
✨ New API for Server Groups

stephenfin:

Allow users to (a) specify server groups by name or (b) defer management of server groups to CAPO.

*Note*
> This is a proposal to allow me test out ideas before writing a KEP. It's
> intentionally incomplete so we can focus on the API first.

n/a

n/a

**TODOs**:

* [x] Implement non-API code
* [x] Run `make generate`
* [ ] Propose KEP?

/hold


api/v1alpha7/types.go
```diff
@@ -215,6 +215,11 @@ type AdditionalBlockDevice struct {
 	Storage BlockDeviceStorage `json:"storage"`
 }
 
+type ServerGroupFilter struct {
+	ID string `json:"id,omitempty"`
+	Name string `json:"name,omityempty"`
+}
+
```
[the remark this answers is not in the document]
stephenfin: Unfortunately yes, we will need to do this client side. Listing server groups is project specific by default, however, so this shouldn't cost too much.

api/v1alpha7/openstackmachine_types.go
```diff
@@ -87,8 +87,15 @@ type OpenStackMachineSpec struct {
 	AdditionalBlockDevices []AdditionalBlockDevice `json:"additionalBlockDevices,omitempty"`
 
 	// The server group to assign the machine to
+	//
+	// Deprecated: Use serverGroup instead. serverGroupID will be silently ignored if serverGroup is set.
```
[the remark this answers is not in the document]
EmilienM: Yes see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1782

api/v1alpha7/openstackmachine_types.go
```diff
@@ -87,8 +87,15 @@ type OpenStackMachineSpec struct {
...
 	ServerGroupID string `json:"serverGroupID,omitempty"`
 
+	// The server group to assign the machine to.
+	// If specified, this will override the managed server group for the machine if server groups are managed.
+	// +optional
+	ServerGroup ServerGroupFilter `json:"serverGroup,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: Yes, see what I did in the controller for https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1774

api/v1alpha7/managed_server_groups.go
```diff
@@ -0,0 +1,38 @@
...
+	// This overrides Policy if specified.
+	// +kubebuilder:validation:Enum="";affinity;soft-affinity;anti-affinity;soft-anti-affinity
+	// +optional
+	ControlPlanePolicy string `json:"controlPlanePolicy,omitempty"`
+
+	// WorkerPolicy defines the policy that should be used for worker machines
+	// This overrides Policy if specified.
+	// +kubebuilder:validation:Enum="";affinity;soft-affinity;anti-affinity;soft-anti-affinity
+	// +optional
+	WorkerPolicy string `json:"workerPolicy,omitempty"`
```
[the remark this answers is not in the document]
stephenfin: I've deferred this to a separate PR as it's rather more involved than what we've done here.

api/v1alpha7/openstackmachine_types.go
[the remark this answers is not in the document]
stephenfin: Sweet. We should probably combine `Image` and `ImageID` to a single `Image` field with a `ImageFilter` type while we're at it too...

pkg/cloud/services/compute/instance.go
```diff
@@ -562,8 +568,18 @@ func applyServerGroupID(opts servers.CreateOptsBuilder, serverGroupID string) se
 	return opts
 }
 
-// Helper function for getting image id from name.
-func (s *Service) getImageIDFromName(imageName string) (string, error) {
+// Helper function for getting image ID from name or ID.
```
jichenjc, line 571: >>>Helper function for getting image ID from name or ID.

this seems confusing

pkg/cloud/services/compute/instance.go
```diff
@@ -584,16 +600,63 @@ func (s *Service) getImageIDFromName(imageName string) (string, error) {
...
 	}
 
-	return "", nil
+	// if we have an ID, use that
+	if serverGroupFilter.ID != "" {
+		serverGroup, err := s.getComputeClient().GetServerGroup(serverGroupFilter.ID)
+		if err != nil {
+			return "", err
+		}
+		if serverGroupFilter.Name != "" && serverGroupFilter.Name != serverGroup.Name {
```
jichenjc, line 621: should we really have this check? or just safely ignore the Name ? I think the logic here want to make sure they are consistent..

stephenfin: So while I think this is silly check and I think providing both a name and ID is even sillier, but if someone does it then I think we should respect their request fully. To not do so would be confusing.

pkg/cloud/services/compute/instance.go
stephenfin: I agree :smile: However, despite what GitHub says I didn't write this. This is being picked up as a moved line when in fact I simply renamed the function from `getImageID` to `GetImageID` and folded `getImageIDFromName` into this. It's also technically true.

api/v1alpha7/openstackmachine_types.go
```diff
@@ -107,6 +113,12 @@ type OpenStackMachineStatus struct {
 	// Addresses contains the OpenStack instance associated addresses.
 	Addresses []corev1.NodeAddress `json:"addresses,omitempty"`
 
+	// ImageID contains the ID of the image this machine was created with.
+	ImageID string `json:"imageID,omitempty"`
```
mdbooth, line 117: Why are we also adding `ImageID` here? It seems like a good addition, but does it need to happen in this PR?

pkg/cloud/services/compute/instance.go
mdbooth: I'm with @jichenjc on this one. Can we not just add an admission validation that if you specify ID you can't specify anything else?

If the machine creation fails because the specified ID was invalid that should be reported in the machine status.

mdbooth (commented):
I think this is mostly there. I'd like to:
* Rebase on to [v1alpha8](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1782)
* Consequently, take the opportunity the remove the ServerGroupID field
* Remove changes to ImageID in status unless I'm missing some reason it's required for Server Groups (it's good but I'd prefer a separate PR)
* Remove additional validation of SecurityGroup specified by ID

dulek:
I tried to implement Matt's remarks based on his own version of this code. I think I'm still failing to get the conversions or the conversions tests right, will continue on that.

controllers/openstackmachine_controller.go
```diff
@@ -463,7 +466,7 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
 		ConfigDrive:            openStackMachine.Spec.ConfigDrive != nil && *openStackMachine.Spec.ConfigDrive,
 		RootVolume:             openStackMachine.Spec.RootVolume,
 		AdditionalBlockDevices: openStackMachine.Spec.AdditionalBlockDevices,
-		ServerGroupID:          openStackMachine.Spec.ServerGroupID,
+		ServerGroupID:          openStackMachine.Status.ReferencedResources.ServerGroupID,
```
EmilienM, line 469: This isn't set anywhere, if I read the PR correctly. Status of a machine is usually set after a machine is created. I'll take a deeper look.

pkg/cloud/services/compute/instance_test.go
```diff
@@ -118,6 +119,117 @@ func Test_getPortName(t *testing.T) {
...
+			mockCtrl := gomock.NewController(t)
+			mockScopeFactory := scope.NewMockScopeFactory(mockCtrl, "", logr.Discard())
+
+			s, err := NewService(mockScopeFactory)
+			if err != nil {
+				t.Fatalf("Failed to create service: %v", err)
+			}
+			tt.expect(mockScopeFactory.ComputeClient.EXPECT())
+
+			got, err := s.GetServerGroupID(tt.serverGroupID, tt.serverGroupFilter)
```
EmilienM, line 221: `GetServerGroupID` takes one arg (sgFilter only).

api/v1alpha5/conversion.go
```diff
@@ -220,6 +220,8 @@ func Convert_v1alpha5_OpenStackMachineSpec_To_v1alpha8_OpenStackMachineSpec(in *
 
 	if in.ServerGroupID != "" {
 		out.ServerGroup = &infrav1.ServerGroupFilter{ID: in.ServerGroupID}
+	} else {
+		out.ServerGroup = &infrav1.ServerGroupFilter{}
```
dulek, line 224: Hm, do we want to set it if ServerGroupID was empty? Feels counter-intuitive to me, but I'm pretty new to all these conversion things.

api/v1alpha5/conversion.go
```diff
@@ -451,7 +453,7 @@ func Convert_v1alpha8_OpenStackMachineSpec_To_v1alpha5_OpenStackMachineSpec(in *
 		return err
 	}
 
-	if in.ServerGroup != nil {
+	if in.ServerGroup != nil && in.ServerGroup.ID != "" {
```
dulek, line 456: Why do we care? If `in.ServerGroup.ID == ""`, then we can still set `out.ServerGroupID`, it'll have a default value anyway.

api/v1alpha7/conversion.go
```diff
@@ -266,6 +268,34 @@ func Convert_v1alpha8_OpenStackMachineStatus_To_v1alpha7_OpenStackMachineStatus(
 }
 
 func Convert_v1alpha8_BastionStatus_To_v1alpha7_BastionStatus(in *infrav1.BastionStatus, out *BastionStatus, s apiconversion.Scope) error {
-	// ReferencedResources have no equivalent in v1alpha7
+	// ReferencedResources have no equivalent in v0alpha7
```
dulek, line 271: v0alpha7?

controllers/openstackmachine_controller.go
```diff
@@ -145,6 +145,12 @@ func (r *OpenStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return reconcile.Result{}, err
 	}
 
+	// Resolve and store referenced resources
+	err = compute.ResolveReferencedMachineResources(scope, &openStackMachine.Spec, &openStackMachine.Status.ReferencedResources)
```
EmilienM, line 149: Could you please explain why we need this? I'm sure this was explained somewhere but I can't find it in the review. Why don't we just resolve it in `machineToInstanceSpec` directly or somewhere around these lines?

I guess my questions are:
* Why so early in the reconcile, and why not in `machineToInstanceSpec` or somewhere else around it
* Are we creating a new pattern with this "resolved referenced machine resources"? If yes we need to document this.

api/v1alpha5/conversion.go
EmilienM: it's the default, since ServerGroup is not a pointer (otherwise default would have been nil). If you remove the else, the output will be ServerGroup: nil. Try it!

api/v1alpha5/conversion.go
EmilienM: `out.ServerGroupID` is already an empty string by default indeed. Yes we can remove that conditional.

api/v1alpha7/conversion.go
EmilienM: oops

api/v1alpha5/conversion.go
EmilienM: Meh I read-proof myself and I'm wrong this is a pointer. However I believe that's because a pointer that we need to define it.

pkg/cloud/services/compute/servergroup.go
```diff
@@ -0,0 +1,64 @@
+/*
```
EmilienM, line 1: we'll want unit tests for this one.

EmilienM, line 1: we'll want unit tests for this one.

pkg/cloud/services/compute/referenced_resources.go
```diff
@@ -0,0 +1,39 @@
+/*
```
EmilienM, line 1: I wish we could put it out of compute package so it's not tight to compute.
Also unit tests.

EmilienM (changes requested):
Some comments inline.
Also missing docs about new API

controllers/openstackmachine_controller.go
dulek: Okay, yeah, this is a new pattern. Documenting it is a good idea, I'll try to find a place to do it. Anyway doing it here gives us one advantage - it declutters `machineToInstanceSpec()` and we call `machineToInstanceSpec()` in the delete flow too. If we'd put all the resources related to the machine in the status, we should be able to make the delete flow simpler too.

pkg/cloud/services/compute/servergroup.go
dulek: Done, moved them into separate file, but they were in `instance_test.go`.

pkg/cloud/services/compute/referenced_resources.go
dulek: Test added, found a bug.

Where would you propose to put this module? I took a look, but nowhere it really fits better than here, as it's MachineSpec passed there.

dulek:
Added docs as well.

dulek:
/retest

> Port 10483ccf-f19f-4fb8-b837-93c0d95911df on network 49934bc0-9e63-4fe9-a621-aab8b3b70a1d not bound, no agent of type Linux bridge agent registered on host capo-e2e-worker

This doesn't seem to be caused by this patch.

api/v1alpha8/types.go
```diff
@@ -357,6 +363,11 @@ type APIServerLoadBalancer struct {
 	Provider string `json:"provider,omitempty"`
 }
 
+// ReferencedMachineResources contains resolved references to resources required by a machine.
```
EmilienM, line 366: I wish we added more doc to this one. We need to clarify why we have that, how devs can use it for future resources etc.

api/v1alpha8/types.go
```diff
@@ -357,6 +363,11 @@ type APIServerLoadBalancer struct {
...
+type ReferencedMachineResources struct {
+	ServerGroupID string `json:"serverGroupID,omitempty"`
```
EmilienM, line 368: docstring please

controllers/openstackmachine_controller.go
EmilienM: I like the idea. :+1:

EmilienM: The only thing that bothers me is that we call the compute package, while in the future I suspect we'll have networking & storage resources. Could we please find a common place.

pkg/cloud/services/compute/referenced_resources.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package compute
+
+import (
+	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha8"
+	"sigs.k8s.io/cluster-api-provider-openstack/pkg/scope"
+)
+
+func ResolveReferencedMachineResources(scope scope.Scope, spec *infrav1.OpenStackMachineSpec, resources *infrav1.ReferencedMachineResources) error {
```
EmilienM, line 24: please document the function. Thanks

EmilienM (approved):
beside a few nits which can be fixed later, I think this is good to go (feel free to do it in this PR though)

/lgtm

api/v1alpha8/types.go
stephenfin: Would it make sense to drag this whole feature out to a separate PR?

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+becomes
+
+```yaml
+serverGroup:
+  id: "e60f19e7-cb37-49f9-a2ee-0a1281f6e03e"
+```
+
+It is now possible to specify a `ServerGroupFilter` object to select the external network to use for the cluster. The `ServerGroupFilter` object allows to select the server group by name or by ID.
```
stephenfin, line 41: The references to external network are presumably copy-paste mistakes. I think we can simplify this if you like my suggestion above though. How about:

```suggestion
To select a server group by name instead of ID:
```

api/v1alpha8/types.go
dulek: @stephenfin: I don't see a lot of merit in introducing an empty field in a separate PR. We can discuss it's sense here, and circle back to putting ServerGroupID back onto the MachineStatus if we decide it's better.

dulek: I added some docs.

api/v1alpha7/conversion.go
```diff
@@ -232,3 +233,69 @@ func (r *OpenStackMachineTemplateList) ConvertFrom(srcRaw ctrlconversion.Hub) er
...
+func Convert_v1alpha7_OpenStackMachineSpec_To_v1alpha8_OpenStackMachineSpec(in *OpenStackMachineSpec, out *infrav1.OpenStackMachineSpec, s apiconversion.Scope) error {
+	err := autoConvert_v1alpha7_OpenStackMachineSpec_To_v1alpha8_OpenStackMachineSpec(in, out, s)
+	if err != nil {
+		return err
+	}
+
+	if in.ServerGroupID != "" {
+		out.ServerGroup = &infrav1.ServerGroupFilter{ID: in.ServerGroupID}
+	} else {
+		out.ServerGroup = &infrav1.ServerGroupFilter{}
```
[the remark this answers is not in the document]
dulek: Somehow tests fail when this is set. I have hard time debugging these, I could use pairing with someone who would teach me how to debug conversion tests failures.

dulek:
I addressed the docs and comments issue. I'm not sure how to debug the conversions though.

api/v1alpha7/conversion_test.go
```diff
@@ -60,6 +60,29 @@ func TestFuzzyConversion(t *testing.T) {
...
+					v1alpha8OpenStackCluster.Status.Bastion.ReferencedResources = infrav1.ReferencedMachineResources{}
+				}
+			},
+
+			func(v1alpha8ServerGroupFilter *infrav1.ServerGroupFilter, c fuzz.Continue) {
+				c.FuzzNoCustom(v1alpha8ServerGroupFilter)
+
+				// None of the following fields have ever been set in v1alpha7
+				v1alpha8ServerGroupFilter.Name = ""
+			},
```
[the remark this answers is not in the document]
EmilienM: For reference, @mdbooth found a problem in the conversion functions: RestoreField must be provided to a HashedFieldRestorer. AFIK he's working on a patch now to address this so we can apply that in ongoing PRs where I wrote these unwanted fuzzer functions.

api/v1alpha7/conversion_test.go
```diff
@@ -60,6 +60,28 @@ func TestFuzzyConversion(t *testing.T) {
...
```
mdbooth, line 84: None of these should be required. I'll debug.

mdbooth:
I've pushed a new version which makes changes only to the conversion code. It:

* Removes all the fuzzer test workarounds in v1alpha7
* Adds a missing restorer for ReferencedMachineResources in v1alpha7 (which was why the workarounds were required)
* Updates the way we store ServerGroupID in v1alpha6

The last one is interesting. In v1alpha6 `Instance` had an unused `ServerGroupID` field which was removed in v1alpha7. We were storing the ServerGroupID in there. This works well for ServerGroup, but I think we're about to add more fields to this struct. As this trick is only likely to work for ServerGroupID I think it'll end up being more confusing than it's worth, so I've switched it to just doing a regular restore.

EmilienM (approved):
/lgtm

PR #1785, opened 2023-12-08
🐛 Don't apply worker SG to control plane machines

### LINKED ISSUE #1784: Managed worker security group will be applied to control plane machines in absence of managed control plane SG
/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/70494b6a5bf1c4cad0d0f29683df3f0852de7822/controllers/openstackmachine_controller.go#L502-L506

From code inspection, it appears that if a machine *is* a control plane machine, a control plane machine security group *is not* specified, but a worker machine security group *is* specified, the worker machine SG will be applied to the control plane machine.

Worker SGs should only be applied to control plane machines and vice versa.

None.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 70494b6a5bf1c4cad0d0f29683df3f0852de7822
- Cluster-API version: n/a
- OpenStack version: n/a
- Minikube/KIND version: n/a
- Kubernetes version (use `kubectl version`): n/a
- OS (e.g. from `/etc/os-release`): n/a

stephenfin:

Currently, if a worker machine security group is specified but a control plane machine security group is not, the worker machine SG will be be applied to both worker *and* control plane machines. Correct this mistake.

Signed-off-by: Stephen Finucane <stephenfin@redhat.com>

Fixes #1784

None.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -499,15 +499,21 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
 	instanceSpec.SecurityGroups = openStackMachine.Spec.SecurityGroups
 	if openStackCluster.Spec.ManagedSecurityGroups {
 		var managedSecurityGroup string
-		if util.IsControlPlaneMachine(machine) && openStackCluster.Status.ControlPlaneSecurityGroup != nil {
-			managedSecurityGroup = openStackCluster.Status.ControlPlaneSecurityGroup.ID
-		} else if openStackCluster.Status.WorkerSecurityGroup != nil {
-			managedSecurityGroup = openStackCluster.Status.WorkerSecurityGroup.ID
+		if util.IsControlPlaneMachine(machine) {
+			if openStackCluster.Status.ControlPlaneSecurityGroup != nil {
+				managedSecurityGroup = openStackCluster.Status.ControlPlaneSecurityGroup.ID
+			}
+		} else {
+			if openStackCluster.Status.WorkerSecurityGroup != nil {
+				managedSecurityGroup = openStackCluster.Status.WorkerSecurityGroup.ID
+			}
 		}
 
-		instanceSpec.SecurityGroups = append(instanceSpec.SecurityGroups, infrav1.SecurityGroupFilter{
-			ID: managedSecurityGroup,
-		})
+		if managedSecurityGroup != "" {
+			instanceSpec.SecurityGroups = append(instanceSpec.SecurityGroups, infrav1.SecurityGroupFilter{
+				ID: managedSecurityGroup,
+			})
+		}
 	}
 
 	instanceSpec.Ports = openStackMachine.Spec.Ports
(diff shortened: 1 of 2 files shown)


controllers/openstackmachine_controller.go
```diff
@@ -499,15 +499,21 @@ func machineToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, machine *
...
 	if openStackCluster.Spec.ManagedSecurityGroups {
 		var managedSecurityGroup string
-		if util.IsControlPlaneMachine(machine) && openStackCluster.Status.ControlPlaneSecurityGroup != nil {
-			managedSecurityGroup = openStackCluster.Status.ControlPlaneSecurityGroup.ID
-		} else if openStackCluster.Status.WorkerSecurityGroup != nil {
-			managedSecurityGroup = openStackCluster.Status.WorkerSecurityGroup.ID
+		if util.IsControlPlaneMachine(machine) {
+			if openStackCluster.Status.ControlPlaneSecurityGroup != nil {
+				managedSecurityGroup = openStackCluster.Status.ControlPlaneSecurityGroup.ID
+			}
```
mdbooth, line 505: I think in practise this would mean we're in some unusual state and we shouldn't reconcile at all until ControlPlaneSecurityGroup is set. However:
* This is clearly better and safer than the current code
* @EmilienM is currently working on a refactor which will be able to fix this up after the fact even if we were briefly in a weird state

So I'm happy to merge this.

controllers/openstackmachine_controller_test.go
```diff
@@ -89,8 +89,9 @@ func getDefaultOpenStackMachine() *infrav1.OpenStackMachine {
 			ServerMetadata: map[string]string{
 				"test-metadata": "test-value",
 			},
-			ConfigDrive:   pointer.Bool(true),
-			ServerGroupID: serverGroupUUID,
+			ConfigDrive:    pointer.Bool(true),
+			SecurityGroups: []infrav1.SecurityGroupFilter{},
```
mdbooth, line 93: nit: I think you're setting SecurityGroups to an empty list to avoid appending to nil here? If so, you don't need to do that: nil is a valid empty slice. Appending to it will create a slice in the same way that a new slice is created if you exceed the capacity of the existing slice: https://go.dev/play/p/Nhln2sbvEGD

It's also safe and idiomatic to iterate over a nil slice in a for loop: https://go.dev/play/p/E5aqZHV51L0. This means, e.g. you don't need to write an additional guard like:
```go
if (slice != nil) {
    for _, elem := range slice {
        ...
    }
}
```

mdbooth (approved):

controllers/openstackmachine_controller_test.go
stephenfin: Good to know, but the reason I was dumber: I has set `SecurityGroups` to an empty list in the test cases, and I modified the default machine to avoid having to also create a "custom" machine in the test case, which in turn meant I had to modify the default instance spec :sweat_smile: With the benefit of hindight, I could have simply not set `SecurityGroups` in those two additional test cases and thus avoided modifying anything in the defaults. lmk if you want me to respin.

mdbooth: Not worth a respin just for this imho.

PR #1790, opened 2023-12-15
✨ New API: OpenStackCluster.Spec.ExternalNetwork

EmilienM:

This PR converts `OpenStackCluster.Spec.ExternalNetworkID` to `OpenStackCluster.Spec.ExternalNetwork`.

The field `externalNetworkID` has been renamed to `externalNetwork` and isn't a string anymore but a reference to a `NetworkFilter` object.

```yaml
externalNetworkID: "e60f19e7-cb37-49f9-a2ee-0a1281f6e03e"
```

becomes

```yaml
externalNetwork:
  id: "e60f19e7-cb37-49f9-a2ee-0a1281f6e03e"
```

It is now possible to specify a `NetworkFilter` object to select the external network to use for the cluster. The `NetworkFilter` object allows to select the network by name, by ID or by tags.

```yaml
externalNetwork:
  name: "public"
```

If a network is provided, it'll be added to `OpenStackCluster.Status.ExternalNetwork`. If the network can't be found, an error will be returned.
If no network is provided, CAPO will try to find network marked "External" and add it to `OpenStackCluster.Status.ExternalNetwork`. If it can't find a network marked "External",
`OpenStackCluster.Status.ExternalNetwork` will be set to nil.
If more than one network is found, an error will be returned.

It is now possible for a user to specify that no external network should be used by setting `DisableExternalNetwork` to `true`:

```yaml
disableExternalNetwork: true
```

- [X] squashed commits
- [X] API conversion
- [X] includes documentation
- [X] adds unit tests


EmilienM:
Ready for review, CI all green.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,43 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# v1alpha7 compared to v1alpha8
+
+> ⚠️ v1alpha8 has not been released yet.
+
+## API Changes
+
+This only documents backwards incompatible changes. Fields that were added to v1alpha77777777 are not listed here.
```
EmilienM, line 19: Typo here, will fix it

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+becomes
+
+```yaml
+externalNetwork:
+  id: "..."
+```
+
+It is now possible to specify a `NetworkFilter` object to select the external network to use for the cluster. The `NetworkFilter` object allows to select the network by name, by ID or by tags.
```
EmilienM, line 38: Note for myself, the code doesn't allow to get the network by tags. Also I'll need to add a few validations to ensure we get one returned network.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,48 @@
...
+
+```yaml
+externalNetwork:
+  name: "..."
+```
+
+If a network is provided, it'll be added to `OpenStackCluster.Status.ExternalNetwork`.
+If no network is provided, CAPO will try to find network marked "External" and add it to `OpenStackCluster.Status.ExternalNetwork`.
+If the network can't be found, `OpenStackCluster.Status.ExternalNetwork` will be set to nil.
```
dulek, line 47: This implies that when network is provided but isn't found, the CAPO will silently ignore the error. I believe that's wrong and if I read the code correctly it currently has that behavior.

dulek (changes requested):

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
EmilienM: yes exactly :(

api/v1alpha7/conversion_test.go
```diff
@@ -60,6 +60,32 @@ func TestFuzzyConversion(t *testing.T) {
 			func(status *OpenStackClusterStatus, c fuzz.Continue) {
 				c.FuzzNoCustom(status)
 			},
+
+			func(v1alpha8Cluster *infrav1.OpenStackCluster, c fuzz.Continue) {
+				c.FuzzNoCustom(v1alpha8Cluster)
+
+				// None of the following status fields have ever been set in v1alpha6
```
EmilienM, line 67: I need to update this comment here and in v5 too, bad copy paste

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
EmilienM: yeah I'm fine returning an error.

EmilienM: Do you know why we strictly need an external network? If a cluster is deployed on a provider network and reachable from outside already, this should not be needed, right? (ie no needs of floating ips, etc).

dulek: I'm pretty sure we have it for two reasons:

1. If we're creating a router, we need to point it to an external network to allow connectivity.
2. We need a network to create FIPs in (and later on to pass it to CCM so it creates FIPs there).

My main point here is that currently the code silently ignores when user specified a non-existing network. IMO it can just go over it when user haven't specified it, but this isn't really a good user experience if we don't inform users that their input was wrong.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,48 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [v1alpha7 compared to v1alpha8](#v1alpha7-compared-to-v1alpha8)
+  - [Migration](#migration)
```
MaysaMacedo, line 6: looks like this section is missing from the docs

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
EmilienM: But:

1. Router creation is optional, right?
2. Why do we need FIPs? What about when deploying the cluster on a pre-existing network that is already routed (e.g. provider network)

EmilienM: Ok we agreed offline that:
* external network is not strictly required
* if the user provides an external network we will change the behaviour to return a failure if that network wasn't found
* investigate how a user can specify that no external network is desired

pkg/cloud/services/networking/network.go
```diff
@@ -101,10 +106,13 @@ func (s *Service) ReconcileExternalNetwork(openStackCluster *infrav1.OpenStackCl
 
 	switch len(networkList) {
 	case 0:
-		// Not finding an external network is fine
-		openStackCluster.Status.ExternalNetwork = &infrav1.NetworkStatus{}
-		s.scope.Logger().Info("No external network found - proceeding with internal network only")
-		return nil
+		if iTrue {
+			// Not finding an external network is fine if ExternatNetwork is not set
```
dulek, line 110: Typo.

pkg/cloud/services/networking/network.go
```diff
@@ -101,10 +106,13 @@ func (s *Service) ReconcileExternalNetwork(openStackCluster *infrav1.OpenStackCl
 
 	switch len(networkList) {
 	case 0:
-		// Not finding an external network is fine
-		openStackCluster.Status.ExternalNetwork = &infrav1.NetworkStatus{}
-		s.scope.Logger().Info("No external network found - proceeding with internal network only")
-		return nil
+		if iTrue {
```
dulek, line 109: Neat trick. We could maybe name the var `isAutodetecting` or something to make it simpler to understand.

dulek (approved):
/lgtm

Looks good to me, minor comments inline, no need to spin the CI again just for them.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,60 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# v1alpha7 compared to v1alpha8
+
+> ⚠️ v1alpha8 has not been released yet.
+
+## Migration
+
+All users are encouraged to migrate their usage of the CAPO CRDs from older versions to `v1alpha6`. This includes yaml files and source code. As CAPO implements automatic conversions between the CRD versions, this migration can happen after installing the new CAPO release.
```
dulek, line 19: v1alpha8

api/v1alpha8/openstackcluster_types.go
```diff
@@ -65,10 +65,14 @@ type OpenStackClusterSpec struct {
...
-	// to get public internet to the VMs.
+
+	// ExternalNetwork is the OpenStack Network to be used to get public internet to the VMs.
 	// +optional
-	ExternalNetworkID string `json:"externalNetworkId,omitempty"`
+	ExternalNetwork NetworkFilter `json:"externalNetwork,omitempty"`
+
+	// +kubebuilder:validation:Enum:="";auto;none
+	// +kubebuilder:default:=auto
+	ExternalNetworkType ResourceType `json:"externalNetworkType,omitempty"`
```
MaysaMacedo, line 75: Could you add a comment describing the purpose of this field?

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,60 @@
...
+`OpenStackCluster.Status.ExternalNetwork` will be set to nil.
+If more than one network is found, an error will be returned.
+
+It is now possible for a user to specify that no external network should be used by setting `externalNetworkType` to `none`:
+
+```yaml
+externalNetworkType: none
+```
+
+Valid values for `externalNetworkType` are `auto` and `none`. If `externalNetworkType` is not specified, it defaults to `auto`.
```
MaysaMacedo, line 60: Might be worth detailing what "auto" means in the context of `externalNetworkType`

config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
```diff
@@ -5240,9 +5240,33 @@ spec:
...
+                  notTagsAny:
+                    type: string
+                  projectId:
+                    type: string
+                  tags:
+                    type: string
+                  tagsAny:
+                    type: string
+                type: object
+              externalNetworkType:
```
MaysaMacedo, line 5264: There is pattern for naming the API fields that mean to disable something, which I think is what this  does right?
might be worth to rename it to `disableExternalNetwork`.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1alpha8/openstackcluster_types.go#L91

pkg/cloud/services/networking/network_test.go
```diff
@@ -194,3 +195,228 @@ func Test_ReconcileNetwork(t *testing.T) {
...
+				Spec: infrav1.OpenStackClusterSpec{
+					ExternalNetwork: infrav1.NetworkFilter{
+						ID: fakeNetworkID,
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "not reconcile external network when external network type is none",
```
MaysaMacedo, line 312:
```suggestion
			name: "not reconcile external network when external network is disabled",
```

MaysaMacedo:
lgtm besides the last nit I mentioned

api/v1alpha8/openstackcluster_types.go
```diff
@@ -65,10 +65,16 @@ type OpenStackClusterSpec struct {
...
+	// ExternalNetwork is the OpenStack Network to be used to get public internet to the VMs.
+	// +optional
+	ExternalNetwork NetworkFilter `json:"externalNetwork,omitempty"`
+
+	// DisableExternalNetwork determines whether or not to attempt to connect the cluster
+	// to an external network. This allows for the creation of clusters when connecting
+	// to an external network is not possible or desirable, e.g. if using a provider network.
 	// +optional
-	ExternalNetworkID string `json:"externalNetworkId,omitempty"`
+	DisableExternalNetwork bool `json:"disableExternalNetwork"`
```
stephenfin, line 77: Question: given `ExternalNetwork` is optional, couldn't we simply suggest that leaving this undefined/explicitly set to `null` indicates that the external network should be disabled?

*Later:* Okay, an unset `externalNetwork` value already carries a special meaning. I wonder if we can change that meaning. I discuss this later.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,58 @@
...
+
+## API Changes
+
+This only documents backwards incompatible changes. Fields that were added to v1alpha8 are not listed here.
+
+### `OpenStackCluster`
+
+#### Change to externalNetworkID
+
+The field `externalNetworkID` has been renamed to `externalNetwork` and isn't a string anymore but a reference to a `NetworkFilter` object.
```
stephenfin, line 29:
```suggestion
The field `externalNetworkID` has been renamed to `externalNetwork` and isn't a string anymore but a reference to a `NetworkFilter` object.

The field `externalNetworkID` has been renamed to `externalNetwork` and is now a `NetworkFilter` object rather than a string ID.
The `NetworkFilter` object allows selection of a network by name, by ID or by tags.
```

to mirror what we're saying in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1796

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,58 @@
...
+It is now possible to specify a `NetworkFilter` object to select the external network to use for the cluster. The `NetworkFilter` object allows to select the network by name, by ID or by tags.
+
+```yaml
+externalNetwork:
+  name: "public"
+```
+
+If a network is provided, it'll be added to `OpenStackCluster.Status.ExternalNetwork`. If the network can't be found, an error will be returned.
+If no network is provided, CAPO will try to find network marked "External" and add it to `OpenStackCluster.Status.ExternalNetwork`. If it can't find a network marked "External",
+`OpenStackCluster.Status.ExternalNetwork` will be set to nil.
```
stephenfin, line 51: Ah, okay, this is the reason for the new `disableExternalNetwork` knob.

Given we're introducing the ability to filter external networks by name as well as a new API version, could we drop this feature and instead instead users specify an network (or omit it). This warrants a big warning in the migration docs since it'll be a larger (though hopefully easy to spot) change.

stephenfin (changes requested):
I'm wondering if we can get rid of the auto-configuration of external network feature (anti-feature?). Does anyone rely on this?

api/v1alpha8/openstackcluster_types.go
EmilienM: > couldn't we simply suggest that leaving this undefined/explicitly set to null indicates that the external network should be disabled?

I thought about it too but we can't because to remain backward compatible, we have to support that users using the default (not set), CAPO will try to find an external network for them.
If we are ok breaking this then indeed we don't need `DisableExternalNetwork`.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
EmilienM: I'm fine with it. Waiting for @mdbooth's feedback and if he agrees I can propose this change.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -28,6 +28,9 @@ This only documents backwards incompatible changes. Fields that were added to v1
 
 The field `externalNetworkID` has been renamed to `externalNetwork` and isn't a string anymore but a reference to a `NetworkFilter` object.
```
stephenfin, line 29:
```suggestion
```

I think you want to remove this line?

pkg/cloud/services/networking/network_test.go
```diff
@@ -194,3 +195,228 @@ func Test_ReconcileNetwork(t *testing.T) {
...
+					ListNetwork(external.ListOptsExt{
+						ListOptsBuilder: networks.ListOpts{},
+						External:        &isAutodetecting,
+					}).
+					Return([]networks.Network{}, nil)
+			},
+			want: &infrav1.OpenStackCluster{
+				Spec: infrav1.OpenStackClusterSpec{},
+				Status: infrav1.OpenStackClusterStatus{
+					ExternalNetwork: &infrav1.NetworkStatus{},
```
mdbooth, line 345: In this and the test above these should probably be nil.

pkg/cloud/services/networking/network_test.go
mdbooth, line 1: Great tests, thanks!

api/v1alpha7/conversion_test.go
```diff
@@ -60,6 +60,34 @@ func TestFuzzyConversion(t *testing.T) {
...
+				// None of the following fields have ever been set in v1alpha7
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.Name = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.Description = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.Tags = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.TagsAny = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.NotTags = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.NotTagsAny = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.ExternalNetwork.ProjectID = ""
+				v1alpha8ClusterTemplate.Spec.Template.Spec.DisableExternalNetwork = false
+			},
```
mdbooth, line 90: As with the ServerGroup PR, this doesn't look right to me. Lets investigate.

api/v1alpha7/conversion.go
```diff
@@ -232,3 +233,31 @@ func (r *OpenStackMachineTemplateList) ConvertFrom(srcRaw ctrlconversion.Hub) er
...
+	if err != nil {
+		return err
+	}
+
+	if in.ExternalNetwork.ID != "" {
+		out.ExternalNetworkID = in.ExternalNetwork.ID
+	}
+
+	return nil
+}
```
mdbooth, line 263: Could we have some test coverage of these conversions? I think I'd probably test at the OpenStackCluster level as that's the most relevant unit here, which would also make it easy to expand to cover other conversions.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
mdbooth: I don't have any hard data, but I suspect there are a bunch of folks who like the default. Removing it requires the user to be explicit about something that was previously implicit, so we wouldn't be able to convert it.

An alternative might be instead of `disableExternalNetwork` to have `useDefaultExternalNetwork`, i.e. you have to opt-in to the current default.

The disableExternalNetwork knob also allows folks with multiple external networks to not use any of them, which isn't currently possible.

I don't have a strong opinion, tbh. I'm ok with @EmilienM's version.

Soliciting opinions from @lentzi90 @mnaser @mkjpryor.

api/v1alpha7/conversion.go
EmilienM: Just to make sure I understand what you suggest: adding unit tests that check that conversion produced an expected OpenStackCluster object. For example like it was done in v1alpha6 for PortOpts convert stuffs?

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
mkjpryor: Personally, I quite like the fact that if there is exactly one external network then that gets used without further intervention.

I do love the fact that I will now be able to select my external network by Neutron tag now though, like I can with my cluster network.

EmilienM:
/hold
I've added cherry-picks to fix conversion, until Matt's PR has merged.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
lentzi90: I agree with @mkjpryor , this is convenient and nice

dulek (approved):
/lgtm from my side too.

mdbooth:
I just pushed an update which fixes conversion of the new fields. The changes are:
* For OpenStackCluster we use the existing spec restorer, and add restoration of the lost fields in `restorev1alpha8ClusterSpec`
* We add a filter to the cluster spec restorer to fix the overlap with the bastion restorer.
* We change OpenStackCluster template to just have a single big restorer for the spec, as we don't need to worry about frequent updates to a template

mdbooth:
/approve

I'm happy for this to merge once somebody else has looked it over.

pkg/cloud/services/networking/network.go
```diff
@@ -70,28 +70,33 @@ func (c createOpts) ToNetworkCreateMap() (map[string]interface{}, error) {
 	return gophercloud.BuildRequestBody(c, "network")
 }
 
+// ReconcileExternalNetwork will try to find an external network and set it in the cluster status.
+// The external network can be specified in the cluster spec or will be searched for if not specified.
+// OpenStackCluster.Status.ExternalNetwork will be set to nil if one of these conditions are met:
+// - no external network was given in the cluster spec and no external network was found
+// - the user has set OpenStackCluster.Spec.ExternalNetworkType to 'none'.
```
MaysaMacedo, line 77: this doc looks outdated

PR #1796, opened 2023-12-18
✨ Change API for OpenStackMachine.Spec.Image

### LINKED ISSUE #1795: Convert `Image` & `ImageUUID` into a single param using an `ImageFilter` instead
/kind feature

Right now, the image for an `OpenStackMachine` can be defined either its name via `Spec.Image` or via `Spec.ImageUUID`.
We could simply this by using a single parameter which would be of type `ImageFilter`.

EmilienM:

Right now, the image for an OpenStackMachine can be defined either
its name via Spec.Image or via Spec.ImageUUID.

Now we will use a single parameter which would be of type ImageFilter.

Fixes #1795

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests


EmilienM:
Something is wrong in the conversion, the bastion instanceSpec gets wrong API:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1796/pull-cluster-api-provider-openstack-e2e-full-test/1737146179717697536/artifacts/clusters/clusterctl-upgrade-5awqmp/clusterctl-upgrade-7iukbs-cluster-template.yaml

```yaml
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha6
kind: OpenStackCluster
metadata:
  name: clusterctl-upgrade-7iukbs
  namespace: clusterctl-upgrade
spec:
  allowAllInClusterTraffic: true
  apiServerLoadBalancer:
    enabled: true
  bastion:
    enabled: true
    instance:
      flavor: m1.tiny
      image:
        name: cirros-0.6.1-x86_64-disk
```

EmilienM:
I need to move `test/e2e/data/kustomize/common-patches/cni/patch-cluster.yaml` to versioned patches so it's not applied on v1alpha8.

pkg/cloud/services/compute/instance.go
```diff
@@ -562,38 +561,22 @@ func applyServerGroupID(opts servers.CreateOptsBuilder, serverGroupID string) se
...
+func (s *Service) getImageID(image infrav1.ImageFilter) (string, error) {
+	allImages, err := s.getImageClient().ListImages(image.ToListOpt())
 	if err != nil {
 		return "", err
 	}
 
 	switch len(allImages) {
 	case 0:
-		return "", fmt.Errorf("no image with the Name %s could be found", imageName)
+		return "", fmt.Errorf("no images could be found")
```
jichenjc, line 573: should we add the filter info into list?

jichenjc (commented):
the others lgtm

pkg/cloud/services/compute/instance.go
```diff
@@ -562,38 +561,22 @@ func applyServerGroupID(opts servers.CreateOptsBuilder, serverGroupID string) se
...
-// Helper function for getting image id from name.
-func (s *Service) getImageIDFromName(imageName string) (string, error) {
-	var opts images.ListOpts
-
-	opts.Name = imageName
-
-	allImages, err := s.getImageClient().ListImages(opts)
+// Helper function for getting image ID from name or ID.
+func (s *Service) getImageID(image infrav1.ImageFilter) (string, error) {
+	allImages, err := s.getImageClient().ListImages(image.ToListOpt())
```
jichenjc, line 566: should we follow https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1779/files#diff-2669f4e6453b138e99121b1203a91f87fa3bc88468c8873c1912c0492566075fR28 ?

pkg/cloud/services/compute/instance.go
EmilienM: good idea, i'm doing this.

pkg/cloud/services/compute/instance.go
EmilienM: good idea, i'm doing this.

dulek (approved):
/lgtm

MaysaMacedo:
lgtm

pkg/cloud/services/compute/instance.go
```diff
@@ -562,38 +561,26 @@ func applyServerGroupID(opts servers.CreateOptsBuilder, serverGroupID string) se
 	return opts
 }
 
-// Helper function for getting image id from name.
-func (s *Service) getImageIDFromName(imageName string) (string, error) {
-	var opts images.ListOpts
-
-	opts.Name = imageName
+// Helper function for getting image ID from name or ID.
```
stephenfin, line 564:
```suggestion
// Helper function for getting image ID from name, ID, or tags.
```

api/v1alpha8/types.go
```diff
@@ -22,6 +22,12 @@ type OpenStackMachineTemplateResource struct {
 	Spec OpenStackMachineSpec `json:"spec"`
 }
 
+type ImageFilter struct {
+	ID   string   `json:"id,omitempty"`
```
stephenfin, line 26:
```suggestion
	// The ID of the desired image. If this is provided, the other filters will be ignored.
	ID   string   `json:"id,omitempty"`
```

api/v1alpha8/types.go
```diff
@@ -22,6 +22,12 @@ type OpenStackMachineTemplateResource struct {
...
+	Name string   `json:"name,omitempty"`
```
stephenfin, line 27:
```suggestion
	// The name of the desired image. If specified, the combination of name and tags must return a single matching image or an error will be raised.
	Name string   `json:"name,omitempty"`
```

api/v1alpha8/types.go
```diff
@@ -22,6 +22,12 @@ type OpenStackMachineTemplateResource struct {
...
+	Tags []string `json:"tags,omitempty"`
```
stephenfin, line 28:
```suggestion
	// The tags associated with the desired image. If specified, the combination of name and tags must return a single matching image or an error will be raised.
	Tags []string `json:"tags,omitempty"`
```

api/v1alpha8/openstackmachine_types.go
```diff
@@ -45,13 +45,9 @@ type OpenStackMachineSpec struct {
 	// The flavor reference for the flavor for your server instance.
 	Flavor string `json:"flavor"`
 
-	// The name of the image to use for your server instance.
+	// The image to use for your server instance.
 	// If the RootVolume is specified, this will be ignored and use rootVolume directly.
```
stephenfin, line 49: If I understand correctly, this is no longer true: we will now use the image requested by the filter when creating the root volume? If so, how about

```suggestion
	// If the rootVolume is specified, this be used when creating the root volume.
```

stephenfin (changes requested):

api/v1alpha8/openstackmachine_types.go
EmilienM: I'm not sure why we had this comment before, since the image ID was not ignored and used when creating the rootVolume:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/98f154bafb2b51318a7d3eb6655a206a491c9620/pkg/cloud/services/compute/instance.go#L451

So your suggestion is valid now is also valid without this change. I'll accept it.

api/v1alpha8/openstackmachine_types.go
```diff
@@ -45,13 +45,9 @@ type OpenStackMachineSpec struct {
...
 
-	// The name of the image to use for your server instance.
-	// If the RootVolume is specified, this will be ignored and use rootVolume directly.
-	Image string `json:"image,omitempty"`
-
-	// The uuid of the image to use for your server instance.
-	// if it's empty, Image name will be used
-	ImageUUID string `json:"imageUUID,omitempty"`
+	// The image to use for your server instance.
+	// If the rootVolume is specified, this be used when creating the root volume.
```
stephenfin, line 49:
```suggestion
	// If the rootVolume is specified, this will be used when creating the root volume.
```

pkg/cloud/services/compute/instance_test.go
```diff
@@ -119,73 +119,61 @@ func Test_getPortName(t *testing.T) {
...
 			expect: func(m *mock.MockImageClientMockRecorder) {
-				m.ListImages(images.ListOpts{Name: "test-image"}).Return(
-					[]images.Image{{ID: imageIDA, Name: "test-image"}},
+				m.ListImages(images.ListOpts{Name: imageName}).Return(
+					[]images.Image{{ID: imageID, Name: imageName}},
 					nil)
 			},
-			want:    imageIDA,
 			wantErr: false,
 		},
```
stephenfin, line 149: Do we want a test case for filtering by tags?

stephenfin (changes requested):
One further nit and a question.

pkg/cloud/services/compute/instance_test.go
EmilienM: I can add, it doesn't hurt

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,53 @@
...
+## API Changes
+
+This only documents backwards incompatible changes. Fields that were added to v1alpha8 are not listed here.
+
+### `OpenStackMachine`
+
+#### ⚠️ Change to image
+
+The field `image` is now an `ImageFilter` object rather than a string name.
+The `ImageFilter` object allows to select the image by name, by ID or by tags.
```
stephenfin, line 27: Sorry, this was my mistake.

```suggestion
The `ImageFilter` object allows selection of an image by name, by ID or by tags.
```

stephenfin (approved):

api/v1alpha6/conversion_test.go
```diff
@@ -88,6 +88,38 @@ func TestFuzzyConversion(t *testing.T) {
 					status.ExternalNetwork.APIServerLoadBalancer = nil
 				}
 			},
+
+			func(v1alpha8Machine *infrav1.OpenStackMachine, c fuzz.Continue) {
```
EmilienM, line 92: This is not correct.
@mdbooth found a problem in the conversion functions: `RestoreField` must be provided to a `HashedFieldRestorer`.

I'll wait for him to provide a fix on some other PRs and apply it here.

EmilienM:
CI is red until we get https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1812

mdbooth (commented):
Please could you add the resolved image reference to `ReferencedMachineResources` now we've got it?

controllers/openstackcluster_controller.go
```diff
@@ -386,14 +393,19 @@ func reconcileBastion(scope scope.Scope, cluster *clusterv1.Cluster, openStackCl
 	return nil
 }
 
-func bastionToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, clusterName string) *compute.InstanceSpec {
+func bastionToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, clusterName string, computeService *compute.Service) (*compute.InstanceSpec, error) {
 	name := fmt.Sprintf("%s-bastion", clusterName)
+
+	imageID, err := computeService.GetImageID(openStackCluster.Spec.Bastion.Instance.Image)
```
EmilienM, line 399: I've put that here because if the user changes the bastion spec with a new image, the instanceSpec will change and the bastion will be re-created.

EmilienM:
@lentzi90 when time permits, please review this one.

Thanks!

lentzi90 (commented):
Looks good!
One question before I approve: What is the deal with the bastion patches in the e2e test kustomizations? Are we adding/enabling bastion there now? Was it disabled before?
Why are we enabling it now?

The reason I'm asking (besides just trying to understand it) is that I worry that the increased resource usage might cause issues in the CI.

mdbooth:
> Looks good! One question before I approve: What is the deal with the bastion patches in the e2e test kustomizations? Are we adding/enabling bastion there now? Was it disabled before? Why are we enabling it now?
> 
> The reason I'm asking (besides just trying to understand it) is that I worry that the increased resource usage might cause issues in the CI.

Good catch! With a quick scan I can't currently see any e2e bastion tests, which feels like an obvious oversight. @EmilienM I'm guessing this is why you added them? Unfortunately I don't think this is safe to do for the reason @lentzi90 mentions. We should probably have a single, separate e2e bastion test.

Given the relatively limited scope of this PR I think I'd probably be ok opening a separate issue for that and just removing the bastion hosts in the e2e tests here. @lentzi90 what do you think?

lentzi90:
> Given the relatively limited scope of this PR I think I'd probably be ok opening a separate issue for that and just removing the bastion hosts in the e2e tests here. @lentzi90 what do you think?

Sounds good to me! :+1:

EmilienM:
This was a mistake on my side, I'll fix it. Nice catch and thanks for your review!

EmilienM:
> Given the relatively limited scope of this PR I think I'd probably be ok opening a separate issue for that and just removing the bastion hosts in the e2e tests here.

Ok I'll do it.

EmilienM:
I checked and the bastion is used to run `CollectMachineLog` as it goes through the bastion to reach the machines. But this function isn't called anywhere :shrug: ...

EmilienM:
@lentzi90 @mdbooth I have a preference to address this via another PR, which I'm working on now.

EmilienM:
@lentzi90 @mdbooth ready for review again. Thanks!

test/e2e/shared/common.go
```diff
@@ -214,51 +214,54 @@ func (o OpenStackLogCollector) CollectMachineLog(ctx context.Context, management
...
-			// don't do this for now, it just takes to long
-			// {
-			//	title: "systemd",
-			//	cmd:   "journalctl --no-pager --output=short-precise | grep -v  'audit:\\|audit\\['",
-			// },
-			{
-				title: "kern",
-				cmd:   "journalctl --no-pager --output=short-precise -k",
+	// If the machine has a bastion, collect logs from there.
+	if openStackCluster.Status.Bastion != nil {
```
lentzi90, line 218: I'm confused about this. I mean the code makes sense to me but I'm confused about if we have bastions or not. Yesterday I thought that we didn't have bastions when I saw you added the bastion patches, but now when I see this and realize that we cannot collect logs without bastions, then we must have bastions currently because we definitely do have these logs!

What I see now in the tests run on this PR is that there are no longs for the machines (e.g. [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1796/pull-cluster-api-provider-openstack-e2e-test/1745523472215314432/artifacts/clusters/bootstrap/e2e-4p833k/machines/)). But in our periodic tests (e.g. [this](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-openstack-e2e-test-main/1745503729274064896)) there [are logs](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-openstack-e2e-test-main/1745503729274064896/artifacts/clusters/bootstrap/e2e-6urfd4/machines/) for some machines at least.
So now it seems to me that we do have bastions in at least some tests and this PR is perhaps removing them and consequently prevents log collection.

mdbooth: I feel like I'm on the same rollercoaster. Weeeeeeeeeee.

lentzi90 (commented):
/approve
The kustomize patches could be optimized a bit but I don't want to hold this back just because of that :slightly_smiling_face:

EmilienM:
rebased

PR #1799, opened 2023-12-20
📖  Document BYO dual-stack network

MaysaMacedo:
This commit adds a section to the book explaining the modification of the `OpenStackClusterSpec.Subnet` API field.

### DIFF
--- a/docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
+++ b/docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
@@ -12,6 +12,7 @@
       - [Changes to image](#change-to-image)
       - [Removal of imageUUID](#removal-of-imageuuid)
       - [Change to floatingIP](#change-to-floatingip)
+      - [Change to subnet](#change-to-subnet)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -156,3 +157,23 @@ spec:
   bastion:
     floatingIP: "1.2.3.4"
 ```
+
+#### ⚠️ Change to subnet
+
+In v1alpha8, `Subnet` of `OpenStackCluster` is modified to `Subnets` to allow specification of two existent subnets for the dual-stack scenario.
+
+```yaml
+  subnet:
+    id: a532beb0-c73a-4b5d-af66-3ad05b73d063
+```
+
+In v1alpha8, this will be automatically converted to:
+
+```yaml
+  subnets:
+    - id: a532beb0-c73a-4b5d-af66-3ad05b73d063
+```
+
+`Subnets` allows specifications of maximum two `SubnetFilter` one being IPv4 and the other IPv6. Both subnets must be on the same network. Any filtered subnets will be added to `OpenStackCluster.Status.Network.Subnets`.
+
+When subnets are not specified on `OpenStackCluster` and only the network is, the network is used to identify the subnets to use. If more than two subnets exist in the network, the user must specify which ones to use by defining the `OpenStackCluster.Spec.Subnets` field.


MaysaMacedo:
depends on https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1789

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,45 @@
...
+
+All users are encouraged to migrate their usage of the CAPO CRDs from older versions to `v1alpha8`. This includes yaml files and source code. As CAPO implements automatic conversions between the CRD versions, this migration can happen after installing the new CAPO release.
+
+## API Changes
+
+This only documents backwards incompatible changes. Fields that were added to v1alpha8 are not listed here.
+
+### `OpenStackCluster`
+
+#### Change to subnet
```
EmilienM, line 27: We need to document how it behaves with the network defined in the OpenStackCluster.Spec too.

MaysaMacedo: done

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,47 @@
...
+In v1alpha8, this will be automatically converted to:
+
+```yaml
+  subnets:
+    - id: a532beb0-c73a-4b5d-af66-3ad05b73d063
+```
+
+Subnets allows specifications of `SubnetFilter`. Any filtered subnets will be added to `OpenStackCluster.Status.Network.Subnets`.
+
+When subnets are not specified and only the network is, the network is used to identify the subnets to use. If more than two subnets exist in the network, the user must specify which ones to use by defining the `OpenStackCluster.Spec.Subnets` field.
```
dulek, line 45: You should also add that if two subnets are provided, they need to be IPv4 and IPv6. Also IIUC subnets need to be on the same network? That's also worth indicating.

dulek (changes requested):

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
MaysaMacedo: done.

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -0,0 +1,47 @@
...
+```yaml
+  subnets:
+    - id: a532beb0-c73a-4b5d-af66-3ad05b73d063
+```
+
+`Subnets` allows specifications of maximum two `SubnetFilter` one being IPv4 and the other IPv6. Both subnets must be on the same network. Any filtered subnets will be added to `OpenStackCluster.Status.Network.Subnets`.
+
+When subnets are not specified on `OpenStackCluster` and only the network is, the network is used to identify the subnets to use. If more than two subnets exist in the network, the user must specify which ones to use by defining the `OpenStackCluster.Spec.Subnets` field.
+
+Converting from v1alpha8 to v1alpha7 or older releases using two subnets will not be allowed.
```
mdbooth, line 47: No need to include this: we don't support downgrades anyway, and as discussed the user *can* continue to request an old version.

mdbooth (commented):
Please can you merge this into #1789? I don't think we'd ever want to merge these separately.

MaysaMacedo:
Added this docs directly to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1789

PR #1807, opened 2024-01-02
:bug: Fix patching OpenstackMachine's immutable spec during reconcile

### LINKED ISSUE #1787: Reconcile attempts forbidden OpenstackMachine spec patch
/kind bug

Starting from a working workload cluster with a machine with the following "okay" state:
`Machine` status: TLDR; `all True`
`OpenstackMachine`:
```
status:
  conditions:
  - lastTransitionTime: "2023-12-11T14:46:30Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2023-12-11T14:46:30Z"
    status: "True"
    type: InstanceReady
  instanceState: ACTIVE
  ready: true
```
Delete a VM (in horizon) underlying a workload cluster. 
___
## User view
- a new VM is created, the `OpenstackMachine`'s `status` is kept **exactly** the same (timestamps also) - which I interpreted as "okay, never changed to false"
- `Machine` status is not ok though:
```
status:
  conditions:
  - lastTransitionTime: "2023-12-11T14:56:27Z"
    reason: NodeNotFound
    severity: Error
    status: "False"
    type: NodeHealthy
```
- CAPO error logs:
```
controller.go:324] "Reconciler error" err="admission webhook \"validation.openstackmachine.infrastructure.cluster.x-k8s.io\" denied the request: OpenStackMachine.infrastructure.cluster.x-k8s.io \"test4-capi-md-0-hwp75\" is invalid: spec: Forbidden: cannot be modified" controller="openstackmachine" controllerGroup="infrastructure.cluster.x-k8s.io" ...
```
- can only be helped by deleting the `Machine`
## Observations
The webhook log is coming from [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1efeaf76b071ba64386ceecab91c4308356fa8e8/api/v1alpha7/openstackmachine_webhook.go#L113) therefore `OpenstackMachine`'s `spec` (which is immutable by design) is being changed.

This happens during regular `OpenstackMachine` reconcile in [getOrCreate](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1347c0064b8224303b7e06e2db8e7cb340ba3c3d/controllers/openstackmachine_controller.go#L331) func. Instance is fetched by name, is recreated because it doesn't exist and `instanceStatus` is returned and assigned to the `openstackMachine` var. This [defer](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/1347c0064b8224303b7e06e2db8e7cb340ba3c3d/controllers/openstackmachine_controller.go#L137) makes sure all changes to the object are patched to the `OpenstackMachine` CR. This then causes:
1. the webhook admission error, because a `spec.providerID` & `spec.instanceID` update is attempted
2. the `OpenstackMachine` is not up to date with what the actual state of Openstack is

- No errors on CAPO side.
- `OpenstackMachine` CR being up to date (instead there is `instanceID` of the old resource from Openstack present which seems very misleading) 

- Originally discussed in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1741 making new issue based on [slack convo](https://kubernetes.slack.com/archives/CFKJB65G9/p1702302439713059?thread_ts=1702035038.998529&cid=CFKJB65G9).
- maybe this scenario should be handled the same as a other errors on Openstack side ("Not reconciling machine in failed state.")

**Environment:**

- Cluster API Provider OpenStack version: v0.8.0
- Cluster-API version: v1.5.3
- OpenStack version: yoga
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): 1.27.3
- OS (e.g. from `/etc/os-release`): ubuntu-20.04

strudelPi:
**What this PR does / why we need it**:  When a previously existing(OpenstackMachine has instanceID ref) VM vanishes (cannot be fetched from Openstack) the reconcile loop attempts recreate and update the OpenstackMachine spec (the instanceID and providerID) which is immutable. This PR disables the attempt to recreate (consistent behaviour with other broken states [example](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/dc2d0baf7a53508d2fa5ec45d605868c9d927f09/controllers/openstackmachine_controller.go#L291))

Fixes #1787

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -329,7 +329,7 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 	}
 
 	instanceStatus, err := r.getOrCreate(scope.Logger(), cluster, openStackCluster, machine, openStackMachine, computeService, userData)
-	if err != nil {
+	if err != nil || instanceStatus == nil {
 		// Conditions set in getOrCreate
 		return ctrl.Result{}, err
 	}
@@ -439,6 +439,12 @@ func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *cl
 	}
 
 	if instanceStatus == nil {
+		if openStackMachine.Spec.InstanceID != nil {
+			logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
+			openStackMachine.SetFailure(capierrors.UpdateMachineError, errors.New("virtual machine no longer exists"))
+			return nil, nil
+		}
 		instanceSpec := machineToInstanceSpec(openStackCluster, machine, openStackMachine, userData)
 		logger.Info("Machine does not exist, creating Machine", "name", openStackMachine.Name)
 		instanceStatus, err = computeService.CreateInstance(openStackMachine, openStackCluster, instanceSpec, cluster.Name)


controllers/openstackmachine_controller.go
```diff
@@ -439,12 +439,17 @@ func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *cl
...
 
 	if instanceStatus == nil {
-		instanceSpec := machineToInstanceSpec(openStackCluster, machine, openStackMachine, userData)
-		logger.Info("Machine does not exist, creating Machine", "name", openStackMachine.Name)
-		instanceStatus, err = computeService.CreateInstance(openStackMachine, openStackCluster, instanceSpec, cluster.Name)
-		if err != nil {
-			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceCreateFailedReason, clusterv1.ConditionSeverityError, err.Error())
-			return nil, fmt.Errorf("create OpenStack instance: %w", err)
+		if openStackMachine.Spec.InstanceID != nil {
+			logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
```
MaysaMacedo, line 443: setting `conditions.MarkFalse` for `InstanceReadyCondition` might be useful

strudelPi: Yeah, definitely. Good catch :-)

controllers/openstackmachine_controller.go
```diff
@@ -439,6 +439,11 @@ func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *cl
 	}
 
 	if instanceStatus == nil {
+		if openStackMachine.Spec.InstanceID != nil {
+			logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
+			return nil, nil
```
dulek, line 445: I wonder if we shouldn't return error in that case? It's not a successful reconciliation loop. On the other hand retrying it doesn't really make sense either.

dulek (changes requested):

controllers/openstackmachine_controller.go
strudelPi: I think the behaviour should be analogous to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6020a9e69aecf97c43505ae2a577a4d3be8bd903/controllers/openstackmachine_controller.go#L292 state.

strudelPi:
I just realized that the same use case of vanishing machines is a trouble when a delete on that machine is attempted. I think the finalizer makes it so that the delete hangs forever. Will need to check on that further.

controllers/openstackmachine_controller.go
dulek: Makes perfect sense, thanks!

controllers/openstackmachine_controller.go
```diff
@@ -439,6 +439,11 @@ func (r *OpenStackMachineReconciler) getOrCreate(logger logr.Logger, cluster *cl
 	}
 
 	if instanceStatus == nil {
+		if openStackMachine.Spec.InstanceID != nil {
+			logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
```
mdbooth, line 444: Please can you also call `openStackMachine.SetFailure()`? Deletion of the VM is one of the few cases we need to treat as unrecoverable.

mdbooth:
> I just realized that the same use case of vanishing machines is a trouble when a delete on that machine is attempted. I think the finalizer makes it so that the delete hangs forever. Will need to check on that further.

This should be ok. In general we treat 404 as success on delete: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/5f306c3cee9926d440b1d5131e0d4b7e889dd4df/pkg/cloud/services/compute/instance.go#L766-L769

mdbooth (commented):
This is a nasty bug, thanks!

I've approved, but I'd appreciate it if you would make the requested change to call SetFailure() before merging. You'll still need LGTM from somebody.

/approve

mdbooth:
Thanks for the update. Lets wait for a second reviewer to leave lgtm.

PR #1812, opened 2024-01-04
🐛 Move flatcar-sysext to v1alpha8

EmilienM:
This was added via v1alpha7 but the latest api is v1alpha8 and to keep
our CI consistent we will move the manifests to v1alpha8.

### DIFF
--- a/kustomize/v1alpha8/flatcar-sysext/patch-flatcar.yaml
+++ b/kustomize/v1alpha8/flatcar-sysext/patch-flatcar.yaml
@@ -1,0 +1,185 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+kind: OpenStackCluster
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
+  apiServerLoadBalancer:
+    $patch: delete
+---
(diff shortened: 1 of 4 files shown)


kustomize/v1alpha8/flatcar-sysext/patch-flatcar.yaml
```diff
@@ -166,20 +166,22 @@ spec:
...
-apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
 kind: OpenStackMachineTemplate
 metadata:
   name: ${CLUSTER_NAME}-md-0
 spec:
   template:
     spec:
-      image: ${FLATCAR_IMAGE_NAME}
+      image:
```
MaysaMacedo, line 176: Shouldn't this field only be updated on the PR that is adding support for it?

EmilienM: yes, :facepalm:

mdbooth:
We should leave the v1alpha7 templates in place and copy them to v1alpha8 with the required changes.

Issue #1704, opened 2023-10-02
Our control plane security groups are too open


EmilienM:
/kind bug

**What steps did you take and what happened:**
* A lot of default rules are too opinionated (specific to CNI)
* `additionalPorts` allows to create more rules, but automatically allow TCP and UDP, which is too opinionated.

**What did you expect to happen:**

We should make the security group rules protocols explicit (ideally in v1beta1).

jichenjc:
there are multiple issues talked about security group 
such as CNI https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1346

what's the proposed suggestion of this issue?

jichenjc:
yes, I think there are proposal to split the Sec group creation from code and make them into some kind of yaml file
so you are able to choose predefined yaml file as sec group (rules) and able to customerize as well
but it's unfortunately not implemented yet

mdbooth:
The code will be gratefully received though!

@dulek I like the idea of a cluster spec flag which would disable creation of the calico rules when using managed security groups. That would also give us a route to deprecating these hard-coded rules.

wwentland:
I like the approach taken in the AWS provider for CNIs, who [define a variety of resources](https://cluster-api-aws.sigs.k8s.io/crd/index.html) to tailor cluster security groups to your needs. 

Specifically, the following look interesting:

- [CNISpec](https://cluster-api-aws.sigs.k8s.io/crd/index.html#infrastructure.cluster.x-k8s.io/v1beta1.CNISpec)
- [CNIIngressRules](https://cluster-api-aws.sigs.k8s.io/crd/index.html#infrastructure.cluster.x-k8s.io/v1beta1.CNIIngressRules)
- [CNIIngressRule](https://cluster-api-aws.sigs.k8s.io/crd/index.html#infrastructure.cluster.x-k8s.io/v1beta1.CNIIngressRule)

This looks like the following in practice:

```
apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
kind: AWSCluster
metadata:
  name: foo
spec:
  network:
    cni:
      cniIngressRules:
      - description: cilium-health
        fromPort: 4240
        protocol: tcp
        toPort: 4240
      - ...
``` 

They also utilise `IngressRules` in various places and work quite well in "bring your own infrastructure" cases where users have externally managed security groups.

mdbooth:
[...]

PR #1826, opened 2024-01-15
✨ AllNodes security groups API

EmilienM:

New API for all nodes managed security group API.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests
  - [x] adds e2e tests

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1274 https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1704
/hold


api/v1alpha7/conversion.go
```diff
@@ -63,6 +63,11 @@ var v1alpha8OpenStackClusterRestorer = conversion.RestorerFor[*infrav1.OpenStack
 			return &c.Status.Bastion.ReferencedResources
 		},
 	),
+	"allNodesSecurityGroup": conversion.UnconditionalFieldRestorer(
+		func(c *infrav1.OpenStackCluster) **infrav1.SecurityGroup {
+			return &c.Status.AllNodesSecurityGroup
```
EmilienM, line 68: @mdbooth can you explain how you envision the upgrade path where we add the Calico rules during conversion? Will that happen in these conversion.go? I haven't seen it before in CAPO I think.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -167,9 +175,65 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+				return fmt.Errorf("allNodesRule: remoteManagedGroups must be either \"controlplane\", \"worker\" or \"bastion\"")
+			}
+		}
+	}
+
+	// default to the managed groups if no remoteManagedGroups are set
+	// Here we want to figure out what to do with defaults, several options:
+	// - pass a map of managed groups with their suffix and IDs
+	// - use a keyword like "all" if no remoteManagedGroups are set and later in getAllNodesRules replace it with the managed groups IDs passed in the first arg
+	// - don't provide a default and force the user to set remoteManagedGroups
```
EmilienM, line 229: ping @mdbooth ^

EmilienM: I have a preference to allow empty value and provide a default that includes all managed security groups IDs that were created (we collect them in a slice and then iterate over them to create rules).

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -167,9 +175,65 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+	if len(rule.RemoteManagedGroups) == 0 {
+		rule.RemoteManagedGroups = []string{controlPlaneSuffix, workerSuffix, bastionSuffix}
```
EmilienM, line 231: this block will be removed, it doesn't belong here but it'll be done in `getAllNodesRules`.

api/v1alpha7/conversion.go
```diff
@@ -383,6 +390,12 @@ func Convert_v1alpha7_OpenStackClusterSpec_To_v1alpha8_OpenStackClusterSpec(in *
 		}
 	}
 
+	if in.ManagedSecurityGroups {
+		out.ManagedSecurityGroups = &infrav1.SecurityGroupsSpec{
+			Enabled: true,
```
mdbooth, line 395: Add Calico security group rules here.

api/v1alpha7/conversion.go
```diff
@@ -97,6 +102,8 @@ func restorev1alpha8ClusterSpec(previous *infrav1.OpenStackClusterSpec, dst *inf
 	dst.ExternalNetwork.NotTagsAny = previous.ExternalNetwork.NotTagsAny
 
 	dst.DisableExternalNetwork = previous.DisableExternalNetwork
+
+	dst.ManagedSecurityGroups = previous.ManagedSecurityGroups
```
mdbooth, line 106: Be explicit here: restore `allNodes...` instead of the whole struct.

api/v1alpha5/conversion.go
```diff
@@ -30,6 +30,28 @@ var _ ctrlconversion.Convertible = &OpenStackCluster{}
 
 const trueString = "true"
 
+// legacyCalicoSecurityGroupRules are the legacy security group rules for calico
```
EmilienM, line 33: v1alpha5 conversion is really best effort at this time. Let me know if we need to do more.

EmilienM:
I also want to push e2e tests on this one, shouldn't be a big deal.

EmilienM:
I have on idea how e2e job could pass since we don't create the Calico security group rules :smile: - I'll investigate.

EmilienM:
I'm using https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1834 to test whether the default rules were actually sufficient when Calico is the CNI.

* If the PR passes the tests, it means it's sufficient and in this case here I'll add allNodes rules to allow Calico on new clusters and add tests to check security groups after upgrades to check Calico rules are present.
* If the PR doesn't pass the tests, we have a bigger problem where the default don't work with Calico when it's supposed to, so I'll investigate a bit.

EmilienM:

One more time to run CI jobs to check that for the upgrade job, Calico rules were created by API conversion and not the regular kustomize for fresh installs.

EmilienM:
Observations on CI behaviours:

* Upgrade: Calico SG rules [were created](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1826/pull-cluster-api-provider-openstack-e2e-full-test/1748452229083303936/artifacts/clusters/clusterctl-upgrade-dk0lyt/resources/clusterctl-upgrade/OpenStackCluster/clusterctl-upgrade-ihf2ju.yaml) when upgrading to v1alpha8. I think we're good here. We can see from the SG rule Description field that the rule was added during conversion.
* Fresh install: Calico SG rules [were created](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1826/pull-cluster-api-provider-openstack-e2e-test/1748452096883036160/artifacts/clusters/bootstrap/resources/e2e-f43q40/OpenStackCluster/cluster-e2e-f43q40.yaml) when installing a fresh cluster on v1alpha8. I think we're good here. We can see from the SG rule Description field that the rule was added by e2e kustomize patches.

cc @mdbooth @jichenjc @lentzi90 I think this is ready for review.

I might submit a minor patch to the e2e tests to ensure that the SG rule exist (what I verified manually above) and that's it. Also let me know if doc is not enough.

Thanks for your help!

EmilienM:
> Fresh install: Calico SG rules [were created](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1826/pull-cluster-api-provider-openstack-e2e-test/1748452096883036160/artifacts/clusters/bootstrap/resources/e2e-f43q40/OpenStackCluster/cluster-e2e-f43q40.yaml) when installing a fresh cluster on v1alpha8. I think we're good here. We can see from the SG rule Description field that the rule was added by e2e kustomize patches.

Done in the last iteration. We'll see how that works.

However I couldn't find in our tests I can do a similar check for the upgrades.

EmilienM:
/hold cancel

Ready for review!

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -176,4 +176,52 @@ In v1alpha8, this will be automatically converted to:
...
\ No newline at end of file
+When subnets are not specified on `OpenStackCluster` and only the network is, the network is used to identify the subnets to use. If more than two subnets exist in the network, the user must specify which ones to use by defining the `OpenStackCluster.Spec.Subnets` field.
+
+#### ⚠️ Change to managedSecurityGroups
+
+The field `managedSecurityGroups` is now a `SecurityGroupsSpec` object rather than a boolean.
+The `SecurityGroupsSpec` object has a `enabled` boolean to enable/disable the managed security groups feature.
+
+Also, we can now add security group rules that authorize traffic from all nodes via `allNodesSecurityGroupRules`.
+It is list of security groups rules that should be applied to selected nodes.
```
lentzi90, line 187:
```suggestion
It is a list of security groups rules that should be applied to selected nodes.
```

test/e2e/data/kustomize/common-patches/with-calico-sg-rules/patch-cluster.yaml
lentzi90, line 1: Nice!

api/v1alpha8/types.go
```diff
@@ -291,31 +291,61 @@ type LoadBalancer struct {
...
+	Direction      string `json:"direction,omitempty"`
+	EtherType      string `json:"etherType,omitempty"`
+	PortRangeMin   int    `json:"portRangeMin,omitempty"`
+	PortRangeMax   int    `json:"portRangeMax,omitempty"`
+	Protocol       string `json:"protocol,omitempty"`
+	RemoteGroupID  string `json:"remoteGroupID,omitempty"`
+	RemoteIPPrefix string `json:"remoteIPPrefix,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=bastion;controlplane;worker
```
EmilienM, line 331: add a test for that

api/v1alpha8/openstackcluster_webhook.go
```diff
@@ -63,6 +63,14 @@ func (r *OpenStackCluster) ValidateCreate() (admission.Warnings, error) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	if r.Spec.ManagedSecurityGroups.Enabled && r.Spec.ManagedSecurityGroups.AllNodesSecurityGroupRules != nil {
+		for _, rule := range r.Spec.ManagedSecurityGroups.AllNodesSecurityGroupRules {
+			if rule.RemoteManagedGroups == nil {
```
EmilienM, line 68: * don't check nil, but use length
* remove that check
* check mutually exclusive fields: remoteIPPrefix, remoteSecurityGroupID and remoteManagedGroups can't be set together.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -153,6 +156,10 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 			securityGroupsList, err := shared.DumpOpenStackSecurityGroups(e2eCtx, groups.ListOpts{Tags: clusterName})
 			Expect(err).NotTo(HaveOccurred())
 			Expect(securityGroupsList).To(HaveLen(3))
+
+			calicoSGRules, err := shared.DumpCalicoSecurityGroupRules(e2eCtx, openStackCluster)
+			Expect(err).NotTo(HaveOccurred())
```
EmilienM, line 161: comment here why 8

api/v1alpha8/types.go
```diff
@@ -291,31 +291,61 @@ type LoadBalancer struct {
...
+	return string(m)
+}
+
+// SecurityGroupsSpec defines the desired state of security groups and rules for the cluster.
+type SecurityGroupsSpec struct {
+	// Enabled defines whether security groups should be created.
+	Enabled bool `json:"enabled,omitempty"`
+
+	// AllNodesSecurityGroupRules defines the rules that should be applied to all nodes.
+	AllNodesSecurityGroupRules []SecurityGroupRuleSpec `json:"allNodesSecurityGroupRules,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: yeah I was thinking this as well but I couldn't think of a key too. I'm really not sure about `name` :thinking: - maybe @mdbooth has some idea?

docs/book/src/topics/crd-changes/v1alpha7-to-v1alpha8.md
```diff
@@ -176,4 +176,51 @@ In v1alpha8, this will be automatically converted to:
...
+    portRangeMin: 179
+    portRangeMax: 179
+    protocol: tcp
+    description: "Allow BGP between control plane and workers"
+  ```
+
+#### ⚠️ Calico CNI
+
+Historically we used to create the necessary security group rules for Calico CNI to work. This is no longer the case.
+Now the user needs to create the security group rules manually by using the `managedSecurityGroups.allNodesSecurityGroupRules` feature.
```
MaysaMacedo, line 223: This gave me an idea that the user needed to pre-create those rules, so I propose the following instead:

```suggestion
Now the user needs to request creation of the security group rules by using the `managedSecurityGroups.allNodesSecurityGroupRules` feature.
```

api/v1alpha8/types.go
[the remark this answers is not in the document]
mdbooth: It's a tricky one. I agree that users are going to want to merge this. Adding a 'fake' name to use as a merge key, though... I guess it's not completely gross.

@JoelSpeed are you aware of any prior art?

api/v1alpha8/types.go
mdbooth: I think kubebuilder adds this to the CRD and *waves hands*? Can you check the CRD?

api/v1alpha8/types.go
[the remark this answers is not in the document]
JoelSpeed: Is there a field on openstack for this security group rule that must be unique? Does the description have to be unique? Or maybe the combination of source/destination/port/protocol?

You can use more than key for your `listMapKey`

Example of expected keys https://github.com/openshift/api/blob/4f00b4f16b634830bb16b432eec91f6e514f2981/machine/v1/types_controlplanemachineset.go#L403-L408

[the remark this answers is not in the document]
mdbooth: > Is there a field on openstack for this security group rule that must be unique? Does the description have to be unique? Or maybe the combination of source/destination/port/protocol?

There's no single unique field which is actually used by OpenStack. Using a combination of fields is an interesting idea. Unfortunately I think the set of key fields would have to be all fields. That could work, but it would mean that adding a field to the struct would become a breaking change.

Summarising:

Adding `name` requires the user to create a new field which is only used by the CAPO API, but can be easily added/updated/deleted by merge.

Using all struct fields doesn't require an extraneous field. Users can easily add new rules by merging. Deleting requires specifying the whole rule. Updating isn't possible directly, but can be achieved with delete and add. The API is no longer extensible without updating the merge key, which I think is a breaking change (e.g. it would break existing delete 

When it comes to merging, adding a fake merge key certainly feels like the better option. Is it worth requiring users to always create a name field?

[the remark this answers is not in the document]
JoelSpeed: > When it comes to merging, adding a fake merge key certainly feels like the better option. Is it worth requiring users to always create a name field?

You wouldn't be the first to do so. Though it's a shame there's no small combination of fields that can identify a rule as unique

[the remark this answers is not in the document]
mdbooth: @EmilienM I think the resolution here is that we'll add a fake, required name attribute and make it the merge key.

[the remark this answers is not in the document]
lentzi90: +1 for that :slightly_smiling_face:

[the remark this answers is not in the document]
EmilienM: sounds good; thanks @JoelSpeed for your guidance, super appreciated as always :-)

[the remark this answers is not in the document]
EmilienM: I've addressed this feedback on a separated commit, but I'll squash once this PR is approved.
Note: I've added webhook validations for various fields. I know this has caused some concerns before, so let me know in the review if you want me to drop them.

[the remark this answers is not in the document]
EmilienM: @mdbooth FYI2 also I've removed `omitempty` and added docstrings + some kubebuilder.

api/v1alpha8/types.go
EmilienM: cool, I'm working on that next.

EmilienM: @mdbooth where do you expect tests for this one? in `controllers/suite_test.go` ?

EmilienM:
Please review when time permits.

huxcrux:
I have performed some manual testing and this PR is working just fine. I really like that CAPO is no longer centered around a single CNI. However I really think we should add default rules in the sample manifests or add a section in the book that describes how to get started with at least one CNI (maybe just take calico since it has been used historically?)

An added benefit from documenting this is that others could add default values for other CNIs.

I have verified and when running the following rules, Calico(v3.27.0), Openstack cloud-provider(1.29.0), cinder CSI(1.29.0) and metrics server(0.6.4) my test cluster passes the sonobuoy(0.57.1) conformance test. (I know this has not been the case previously, however I do sadly not remember the reason at this time.)

```yaml
  managedSecurityGroups:
    enabled: true
    allNodesSecurityGroupRules:
      - name: "BGP (Calico)"
        description: "BGP (calico)"
        direction: "ingress"
        etherType: "IPv4"
        portRangeMin: 179
        portRangeMax: 179
        protocol: "tcp"
        remoteManagedGroups:
        - controlplane
        - worker
      - name: "IP-in-IP (calico)"
        description: "IP-in-IP (calico)"
        direction: "ingress"
        etherType: "IPv4"
        protocol: "4"
        remoteManagedGroups:
        - controlplane
        - worker
```

api/v1alpha8/types.go
```diff
@@ -291,39 +291,151 @@ type LoadBalancer struct {
...
-		r.PortRangeMin == x.PortRangeMin &&
-		r.PortRangeMax == x.PortRangeMax &&
-		r.Protocol == x.Protocol &&
-		r.RemoteGroupID == x.RemoteGroupID &&
-		r.RemoteIPPrefix == x.RemoteIPPrefix)
+// SecurityGroupsSpec defines the desired state of security groups and rules for the cluster.
+type SecurityGroupsSpec struct {
+	// enabled defines whether security groups should be created.
+	// +optional
+	Enabled bool `json:"enabled"`
```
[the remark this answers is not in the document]
EmilienM: I think @mdbooth has strongs opinions whether pointers should be used.
Note that this proposal follows what we did for the bastion where we also have `enabled: <bool>`.

Let's see if other reviewers have other opinions.

EmilienM:
> I have performed some manual testing and this PR is working just fine. I really like that CAPO is no longer centered around a single CNI. However I really think we should add default rules in the sample manifests or add a section in the book that describes how to get started with at least one CNI (maybe just take calico since it has been used historically?)

Sure I can add some doc.

> An added benefit from documenting this is that others could add default values for other CNIs.
> 
> I have verified and when running the following rules, Calico(v3.27.0), Openstack cloud-provider(1.29.0), cinder CSI(1.29.0) and metrics server(0.6.4) my test cluster passes the sonobuoy(0.57.1) conformance test. (I know this has not been the case previously, however I do sadly not remember the reason at this time.)

Cool: 2 questions:
* Calico with BGP?
* Did you set `allowAllInClusterTraffic` to True? (the whole purpose of this PR is also to set it to False).

huxcrux:
> > I have performed some manual testing and this PR is working just fine. I really like that CAPO is no longer centered around a single CNI. However I really think we should add default rules in the sample manifests or add a section in the book that describes how to get started with at least one CNI (maybe just take calico since it has been used historically?)
> 
> Sure I added some doc, let me know.
> 
I think this is enough, just to help new users getting started :)

> > An added benefit from documenting this is that others could add default values for other CNIs.
> > I have verified and when running the following rules, Calico(v3.27.0), Openstack cloud-provider(1.29.0), cinder CSI(1.29.0) and metrics server(0.6.4) my test cluster passes the sonobuoy(0.57.1) conformance test. (I know this has not been the case previously, however I do sadly not remember the reason at this time.)
> 
> Cool: 2 questions:
> 
>     * Calico with BGP?
> 
>     * Did you set `allowAllInClusterTraffic` to True? (the whole purpose of this PR is also to set it to False).

Yes I used calico with bgp and IP-in-IP and with allowAllInClusterTraffic set to false

api/v1alpha8/types.go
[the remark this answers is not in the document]
huxcrux: Maybe an on/off toggle can make it easier for new users. However I have no strong opinions in this matter and would think boot solutions are working fine :)

api/v1alpha8/types.go
huxcrux: The enum is visible in CRD. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1826/files#diff-1a379f1e3a8a3141f9200c4629eb4c10e79ef7440e43c2e24287bc36de424d6bR5528-R5536

Enum is just a part of the API spec meaning you can still create an instance of the resource that uses other objects. The filtering is done in Kubernetes during an apply/patch. Sure this could be done in a mock test however I think it would make sense as an e2e test?

api/v1alpha8/types.go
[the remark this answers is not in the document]
mdbooth: Hmm.

```yaml
managedSecurityGroups: {}
```
vs
```yaml
managedSecurityGroups:
  enabled: true
```

I feel like `allNodesSecurityGroupRules` will pretty much always be populated, though, so maybe:
```yaml
managedSecurityGroups:
  allNodesSecurityGroupRules:
  - ...
  - ...
```
vs
```yaml
managedSecurityGroups:
  enabled: true
  allNodesSecurityGroupRules:
  - ...
  - ...
```

In that context the `enabled: true` does look a bit redundant and omitting it is likely a trip hazard.

@EmilienM If you're ok with this change I think I'd be ok with it.

kustomize/v1alpha8/default/cluster-template.yaml
```diff
@@ -28,7 +28,8 @@ spec:
     kind: Secret
   apiServerLoadBalancer:
     enabled: true
-  managedSecurityGroups: true
+  managedSecurityGroups:
+    enabled: true
```
mdbooth, line 32: Please could we add the Calico rules here? We'll likely tweak this at some point to compose them in with kustomize, but for now I think it's simplest just to include them in the default templates.

test/e2e/data/kustomize/common-patches/with-calico-sg-rules/patch-cluster.yaml
mdbooth, line 1: As commented above, please can we move this into the published templates instead? That would also mean it won't be required here.

controllers/openstackcluster_controller.go
```diff
@@ -448,7 +448,7 @@ func bastionToInstanceSpec(openStackCluster *infrav1.OpenStackCluster, clusterNa
 	}
 
 	instanceSpec.SecurityGroups = openStackCluster.Spec.Bastion.Instance.SecurityGroups
-	if openStackCluster.Spec.ManagedSecurityGroups {
+	if openStackCluster.Spec.ManagedSecurityGroups.Enabled {
```
mdbooth, line 451: If we changed this into a struct pointer, I think all of these would just become nil checks. That looks fairly safe as far as pointers go, given the current usage.

mdbooth (commented):
I think this is about ready to go, pending resolution of the `enabled` question.

api/v1alpha8/openstackcluster_webhook.go
```diff
@@ -63,6 +63,23 @@ func (r *OpenStackCluster) ValidateCreate() (admission.Warnings, error) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	if r.Spec.ManagedSecurityGroups.Enabled && len(r.Spec.ManagedSecurityGroups.AllNodesSecurityGroupRules) > 0 {
+		for _, rule := range r.Spec.ManagedSecurityGroups.AllNodesSecurityGroupRules {
+			if rule.Name == "" {
+				allErrs = append(allErrs, field.Required(field.NewPath("spec", "managedSecurityGroups", "allNodesSecurityGroupRules", "name"), "name is required"))
+			}
```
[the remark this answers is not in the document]
EmilienM: Just to double check, is that comment just for "name" validation? Or also for other fields. I think we need to validate other fields, since they aren't required.

api/v1alpha8/openstackcluster_types.go
```diff
@@ -120,10 +120,10 @@ type OpenStackClusterSpec struct {
 	// ManagedSecurityGroups determines whether OpenStack security groups for the cluster
 	// will be managed by the OpenStack provider or whether pre-existing security groups will
 	// be specified as part of the configuration.
-	// By default, the managed security groups have rules that allow the Kubelet, etcd, the
-	// Kubernetes API server and the Calico CNI plugin to function correctly.
-	// +optional
-	ManagedSecurityGroups bool `json:"managedSecurityGroups"`
+	// By default, the managed security groups have rules that allow the Kubelet, etcd and the
```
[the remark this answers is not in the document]
EmilienM: I don't see where a comma is missing, I have it between Kubelet and etcd and not after because there is an "and". Please just tell me where it's missing and I'll add a comma.

mdbooth:
> > I think this is about ready to go, pending resolution of the `enabled` question.
> 
> I kind of disagree. Let's say I really want CAPO to handle the security groups for the control plane, workers and bastion, but I don't need CNI rules because $reason (e.g. I use the OpenStackMachine.Spec to define security groups for CNI), then how am I supposed to do? With `enabled` I can have it turned on.

If we made this change we would still be able to do this, as noted in my comment above:
```yaml
managedSecurityGroups: {}
```

I don't think that's as clear as:
```yaml
managedSecurityGroups:
  enabled: true
```

However, I think this is likely to be an unusual use of this API because any configuration with managed security groups is going to need CNI rules, which are no longer included.
```yaml
managedSecurityGroups:
  enabled: true
  allNodesSecurityGroupRules:
  - ...
  - ...
```

In this case `enabled: true` feels redundant, and it's even a potential trip hazard for the user if they forget to set it because then they won't get managed security groups despite having specified them.

For me, the question is which to optimize for, unless anybody has a suggestion which does both well? I don't feel very strongly about this, but now I've seen it I would lean towards the struct pointer.

api/v1alpha8/openstackcluster_webhook.go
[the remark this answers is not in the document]
mdbooth: This comment was specific to `name`.

api/v1alpha7/conversion.go
```diff
@@ -509,6 +559,12 @@ func Convert_v1alpha7_OpenStackClusterSpec_To_v1alpha8_OpenStackClusterSpec(in *
 		}
 	}
 
+	if in.ManagedSecurityGroups {
+		out.ManagedSecurityGroups = &infrav1.SecurityGroupsSpec{
+			AllNodesSecurityGroupRules: legacyCalicoSecurityGroupRules,
+		}
```
mdbooth, line 565: This isn't correct, because there's an interaction with `allowAllInClusterTraffic`. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1878

We should probably unit test this conversion.

mdbooth (approved):

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
 }
 
-// SecurityGroup represents the basic information of the associated
+// SecurityGroupStatus represents the basic information of the associated
 // OpenStack Neutron Security Group.
-type SecurityGroup struct {
-	Name  string              `json:"name"`
-	ID    string              `json:"id"`
-	Rules []SecurityGroupRule `json:"rules"`
+type SecurityGroupStatus struct {
```
lentzi90, line 306: Why the rename for this? It seems strange to me. Based on the name I would expect this to be a status sub-resource of a SecurityGroup. It also feels a bit weird that the `ControlPlaneSecurityGroup` is of type `SecurityGroupStatus` :raised_eyebrow:

lentzi90 (commented):
Great job with this!
One question inline

huxcrux:
I have run thru a fe manual tests again after all changes and everything seems to be working as expected :+1:

Due to the above comment I'll skip the lgtm label to avoid a merge prior to it being answered

api/v1alpha8/types.go
EmilienM: We needed to differentiate the Status and the Spec. Her we have the ID, so it's a Status; which is why ControlPlaneSecurityGroup is a status type, because it contains the ID and the rules.

@mdbooth help me here :D we wrote this together I think.

huxcrux: This object seems to only be used in as status meanwhile the old object seems to have been used for also defining the default rules. [See the code here]( https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1826/files/685696707236c7ec0987adfbb3ea4c5e119719ca#diff-b443ce0ed208632e1331c31b89e3a4f7bf99dd949c15839f920545d55cd5e96eL79-L116)

lentzi90: Why not just `SecurityGroup`? Then the ControlPlaneSecurityGroup would have the type SecurityGroup and you would still have the separate struct for SecurityGroupSpec

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+	// name of the security group
+	// +optional
+	Name string `json:"name"`
+
+	// id of the security group
+	// +optional
+	ID string `json:"id"`
```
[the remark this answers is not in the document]
EmilienM: Re-reading https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required - I should indeed add omitempty if we decide to keep it optional or make it required.

api/v1alpha8/types.go
mdbooth: I think this rename may be a vestige of an intermediate state of this patch when we had `SecurityGroup` both in the spec and the status and we realised they needed to be different. This is no longer the case, so I think you're right and it would be clearer to change it back.

Also we should likely rename `SecurityGroupsSpec` to `ManagedSecurityGroups` and move it out of `types.go` into `openstackcluster_types.go`.

api/v1alpha8/types.go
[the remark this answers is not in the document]
mdbooth: I think @MaysaMacedo is correct. Both name and id should be required.

api/v1alpha8/types.go
EmilienM: ok i'll make that change now

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+// Remote* fields are mutually exclusive.
+type SecurityGroupRuleSpec struct {
+	// name of the security group rule.
+	// It's used to identify the rule so it can be patched and will not be sent to the OpenStack API.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// description of the security group rule.
+	// +optional
+	Description string `json:"description"`
```
mdbooth, line 333: I think `omitempty` would be safe here, but `*string` would be most 'guidelines compliant'. I'm not sure how I feel about that, yet.

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+
+	// direction in which the security group rule is applied. The only values
+	// allowed are "ingress" or "egress". For a compute instance, an ingress
+	// security group rule is applied to incoming (ingress) traffic for that
+	// instance. An egress rule is applied to traffic leaving the instance.
+	// +kubebuilder:validation:enum=ingress;egress
+	// +optional
+	Direction string `json:"direction"`
```
mdbooth, line 341: https://docs.openstack.org/api-ref/network/v2/index.html#create-security-group-rule

This is required.

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+
+	// etherType must be IPv4 or IPv6, and addresses represented in CIDR must match the
+	// ingress or egress rules.
+	// +kubebuilder:validation:enum=IPv4;IPv6
+	// +optional
+	EtherType string `json:"etherType"`
```
mdbooth, line 347: Bah, this is genuinely optional. I think this must become the dreaded `*string`, with `omitempty`.

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+
+	// portRangeMin is a number in the range that is matched by the security group
+	// rule. If the protocol is TCP or UDP, this value must be less than or equal
+	// to the value of the portRangeMax attribute.
+	// +optional
+	PortRangeMin int `json:"portRangeMin"`
```
mdbooth, line 353: `*int` :(

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+
+	// portRangeMax is a number in the range that is matched by the security group
+	// rule. The portRangeMin attribute constrains the portRangeMax attribute.
+	// +optional
+	PortRangeMax int `json:"portRangeMax"`
```
mdbooth, line 358: Ditto.

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+
+	// protocol is the protocol that is matched by the security group rule.
+	// +optional
+	Protocol string `json:"protocol"`
```
mdbooth, line 362: `*string`. If you're feeling keen you could add validation, but without CEL I don't think it would be maintainable:

> The IP protocol can be represented by a string, an integer, or null. Valid string or integer values are any or 0, ah or 51, dccp or 33, egp or 8, esp or 50, gre or 47, icmp or 1, icmpv6 or 58, igmp or 2, ipip or 4, ipv6-encap or 41, ipv6-frag or 44, ipv6-icmp or 58, ipv6-nonxt or 59, ipv6-opts or 60, ipv6-route or 43, ospf or 89, pgm or 113, rsvp or 46, sctp or 132, tcp or 6, udp or 17, udplite or 136, vrrp or 112. Additionally, any integer value between [0-255] is also valid. The string any (or integer 0) means all IP protocols. See the constants in neutron_lib.constants for the most up-to-date list of supported strings.

Yeah, probably best just to say it's the wild west.

api/v1alpha8/types.go
```diff
@@ -301,39 +301,147 @@ type LoadBalancer struct {
...
+	// +optional
+	RemoteIPPrefix string `json:"remoteIPPrefix"`
+
+	// remoteManagedGroups is the remote managed groups to be associated with this security group rule.
+	// You can specify either remoteGroupID or remoteIPPrefix or remoteManagedGroups.
+	// +optional
+	RemoteManagedGroups []ManagedSecurityGroupName `json:"remoteManagedGroups"`
+}
+
+type SecurityGroupRuleStatus struct {
```
mdbooth, line 380: Implicitly copy all the type changes recommended in the spec struct here, too.

mdbooth (commented):
Sorry, I had not taken another objective look at the details of the security group rules API in light of other things we've discussed recently. Thanks @lentzi90 and @MaysaMacedo for the keen eyeballs.

api/v1alpha8/types.go
EmilienM: I'll rename SecurityGroupsSpec to ManagedSecurityGroups but I don't understand why you want to move it out of types.go.

api/v1alpha8/types.go
EmilienM: lol :cowboy_hat_face:

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -157,19 +208,70 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+
+		if len(rule.RemoteManagedGroups) > 0 {
+			if rule.RemoteGroupID != "" {
+				return nil, fmt.Errorf("remoteGroupID must not be set when remoteManagedGroups is set")
+			}
+
+			for _, rg := range rule.RemoteManagedGroups {
+				if _, ok := remoteManagedGroups[rg.String()]; !ok {
+					return nil, fmt.Errorf("remoteManagedGroups: %s is not a valid remote managed security group", rg)
+				}
```
dulek, line 249: Aren't you already doing these validations in `validateRemoteManagedGroups()`?

dulek (commented):
Just one remark, otherwise this looks good!

pkg/cloud/services/networking/securitygroups.go
EmilienM: good spot

EmilienM:
/hold

I've spotted an issue where Security Groups never stop reconciling :(

EmilienM:
found it and fixed it :+1: 
/hold cancel

mdbooth:
> found it and fixed it 👍 

Comparing pointers instead of their values! It's obvious once you've seen it, but I could probably stare at it for an hour and not spot it.

Thanks for the extra testing diligence. Do you think there's an automated test we could have written which would have caught this?

EmilienM:
> > found it and fixed it 👍
> 
> Comparing pointers instead of their values! It's obvious once you've seen it, but I could probably stare at it for an hour and not spot it.
> 
> Thanks for the extra testing diligence. Do you think there's an automated test we could have written which would have caught this?

done

api/v1alpha5/conversion.go
```diff
@@ -554,3 +566,58 @@ func Convert_v1alpha5_Bastion_To_v1alpha8_Bastion(in *Bastion, out *infrav1.Bast
 	in.Instance.FloatingIP = out.FloatingIP
 	return nil
 }
+
+func Convert_v1alpha8_SecurityGroupStatus_To_v1alpha5_SecurityGroup(in *infrav1.SecurityGroupStatus, out *SecurityGroup, s conversion.Scope) error { //nolint:revive
+	out.ID = in.ID
+	out.Name = in.Name
+	out.Rules = make([]SecurityGroupRule, len(in.Rules))
+	for i, rule := range in.Rules {
+		out.Rules[i] = SecurityGroupRule{
```
bilbobrovall, line 575: Couldn't this be written as ?
```suggestion
		out.Rules[i] = SecurityGroupRule{
			ID:                      rule.ID,
			Direction:          rule.Direction,
			Description:      pointer.StringDeref(rule.Description, ""),
			....
```

On a related out of scope note, we should probably upgrade k8s.io/utils and start using k8s.io/utils/ptr that uses generics instead

EmilienM: yeah, I started doing it and it's a big change. Here is an issue to track it: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1896

api/v1alpha8/openstackcluster_webhook.go
```diff
@@ -63,6 +63,20 @@ func (r *OpenStackCluster) ValidateCreate() (admission.Warnings, error) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	if r.Spec.ManagedSecurityGroups != nil && len(r.Spec.ManagedSecurityGroups.AllNodesSecurityGroupRules) > 0 {
```
bilbobrovall, line 66: Does the len > 0 accomplish anything? :thinking:

EmilienM: Not much indeed in this case.

EmilienM:
```
E0220 15:43:45.873177       1 controller.go:329] "Reconciler error" err="failed to reconcile load balancer: load balancer \"k8s-clusterapi-cluster-self-hosted-021hq8-self-hosted-95qimh-kubeapi\" with id 24695b59-be36-446c-8348-611cbe8d090e is not active after timeout: timed out waiting for the condition" controller="openstackcluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackCluster" OpenStackCluster="self-hosted-021hq8/self-hosted-95qimh" namespace="self-hosted-021hq8" name="self-hosted-95qimh" reconcileID="91718218-6629-47d3-b426-28a4249b30cf"
```

lentzi90 (commented):
Thanks for this!
/lgtm

Issue #1684, opened 2023-09-26
OpenStackMachineSpec.ServerMetadata should not be a map


mdbooth:
/kind feature

OpenStackMachineSpec.ServerMetadata should not be a map. To comply with k8s API conventions it should be a list of name/value pairs with listType=map.

EmilienM:
@mdbooth I was about to look at this one but I have a doubt. Should we still try to improve the API here, while in Gophercloud this is exposed as `Metadata map[string]string` ?

mdbooth:
How would you improve it?

This is a breaking chance, btw, so it depends on another version bump. We're going to have to do that eventually, and before long, but we may not want to do it immediately.

EmilienM:
/remove-lifecycle stale
this is getting merged

PR #1828, opened 2024-01-16
⚠️ Convert ServerMetadata from a map to a list

mdbooth:
While we're changing it, we also add validation that keys and values don't exceed 255 characters.

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1684

TODO:
- [ ] Docs

/hold


config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
```diff
@@ -5269,11 +5269,28 @@ spec:
...
+                                kubebuilder:validation:MaxLength:=255
+                              type: string
+                          required:
+                          - key
+                          - value
+                          type: object
+                        type: array
+                        x-kubernetes-list-map-keys:
+                        - key
+                        x-kubernetes-list-type: map
```
dulek, line 5293: Will this have more consequences, like e.g. not allowing some characters because K8s doesn't allow them, but OpenStack does?

api/v1alpha8/openstackmachine_types.go
```diff
@@ -91,6 +93,16 @@ type OpenStackMachineSpec struct {
...
 
+type ServerMetadata struct {
+	// Key is the server metadata key
+	// kubebuilder:validation:MaxLength:=255
+	Key string `json:"key"`
+
+	// Value is the server metadata value
+	// kubebuilder:validation:MaxLength:=255
+	Value string `json:"value"`
+}
```
jichenjc, line 104: I think we may have other K/V settings , maybe something like Metadata then others reuse it?

config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
mdbooth: I don't think there are any constraints on map keys. [The documentation](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy) only says:
> The key fields must be scalars.
They don't have the same constraints as names.

EmilienM:
/hold cancel
doc was pushed.

I'll ask another reviewer to check it before lgtm.

MaysaMacedo (commented):
/lgtm

Nice work, Matt

EmilienM:
It's failing on conversion now, Matt changed something with fuzzer, it'll need adjustment. I'll have a look tomorrow.

api/v1alpha8/openstackmachine_types.go
jichenjc: ok, just check in case

EmilienM:
Thanks Matt for fixing conversion.

PR #1833, opened 2024-01-18
🌱 Enhance Tilt integration with CAPO using a ClusterClass template

### LINKED ISSUE #1767: Develop User-Friendly ClusterClass for Enhanced Tilt-CAPO Integration
/kind feature

Develop a ClusterClass template to improve Tilt integration with CAPO, leveraging the existing CAPI setup. This template should simplify cluster creation and fully utilize CAPO's features. Additionally, refine CAPO's current Tilt setup to support standalone development tests, akin to CAPI.

Focus on user-friendly design for the ClusterClass and document the development process for future enhancements. Reference CAPI's Tilt Developer [Guide](https://cluster-api.sigs.k8s.io/developer/tilt) and experiment with the docker provider for functionality insights.

maxrantil:
This PR introduces a ClusterClass template to enhance Tilt integration with CAPO, leveraging the existing CAPI setup. It simplifies cluster creation.

Fixes #1767

**TODOs**:
- [x] squashed commits
- if necessary:
  - [x] includes documentation

/hold

### DIFF
--- a/docs/book/src/development/development.md
+++ b/docs/book/src/development/development.md
@@ -50,9 +50,104 @@ make docker-build docker-push
 
 After generating `infrastructure-components.yaml`, replace the `us.gcr.io/k8s-artifacts-prod/capi-openstack/capi-openstack-controller:v0.3.4` with your image.
 
+## Testing Cluster Creation using the 'dev-test' ClusterClass with Tilt
+
+This guide demonstrates how to create a Kubernetes cluster using ClusterClass, specifically designed for a development environment. It includes initializing a Kind cluster, configuring secrets, and applying ClusterClass to deploy your cluster with Tilt.
+
+### Creating a Kind Cluster
+
+Create a Kind cluster. This process involves the initialization of Cluster API providers for OpenStack.
+
+```bash
+kind create cluster
+export CLUSTER_TOPOLOGY=true
+clusterctl init --infrastructure openstack
+```
+
+### Secret Configuration
+
+CAPO needs a clouds.yaml file in order to manage the OpenStack resources needed for the Cluster. This should be supplied as a secret named `${CLUSTER_NAME}-cloud-config`. You can create this secret for example with:
+
+```bash
+kubectl create secret generic ${CLUSTER_NAME}-cloud-config --from-file=clouds.yaml
+```
+
+You can also make Tilt create it for you when you run `tilt up`, more on that in the next section.
+
 ## Developing with Tilt
 
-We have support for using [Tilt](https://tilt.dev/) for rapid iterative development. Please visit the [Cluster API documentation on Tilt](https://cluster-api.sigs.k8s.io/developer/tilt.html) for information on how to set up your development environment. 
+We have support for using [Tilt](https://tilt.dev/) for rapid iterative development. Please visit the [Cluster API documentation on Tilt](https://cluster-api.sigs.k8s.io/developer/tilt.html) for information on how to set up your development environment.
+
+Default values align with the devstack setup, as documented below. When specifying the `KUBERNETES_VERSION`, ensure an image named `ubuntu-2204-kube-{{ KUBERNETES_VERSION }}` is available in your environment, corresponding to that Kubernetes version.
+For using Tilt with ClusterClass, update your `tilt-settings.yaml` file as described:
+
+```yaml
+kustomize_substitutions:
+  CLUSTER_TOPOLOGY: "true"
+  # Define the name of your cluster (e.g., "dev-test")
+  CLUSTER_NAME: "<cluster_name>"
+  # Desired Kubernetes Version for Your Cluster (e.g., "v1.28.5")
+  KUBERNETES_VERSION: "<kubernetes_version>"
+  # [Optional] SSH Keypair Name for Instances in OpenStack (Default: "")
+  OPENSTACK_SSH_KEY_NAME: "<openstack_keypair_name>"
+  # [Optional] Control Plane Machine Flavor (Default: m1.medium)
+  OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR: "<openstack_control_plane_machine_flavor>"
+  # [Optional] Node Machine Flavor (Default: m1.small)
+  OPENSTACK_NODE_MACHINE_FLAVOR: "<openstack_node_machine_flavor>"
+  # [Optional] OpenStack Cloud Environment (Default: capo-e2e)
+  OPENSTACK_CLOUD: "<openstack_cloud>"
+
+additional_kustomizations:
+  secret_kustomization: /path/to/kustomize/secret/configuration
+```
+
+Ensure the specified path (`/path/to/kustomize/secret/configuration`) contains both the `clouds.yaml` file and a `kustomization.yaml` file. The `kustomization.yaml` should define the necessary resources, such as a Kubernetes secret, using the `clouds.yaml` file.
+
+For example, if you want to automatically create a secret named `dev-test-cloud-config` with the content of your `clouds.yaml` every time you do `tilt up`, you could do the following.
+
+Create a folder to hold the kustomization.
+We will use `/tmp/capo-dev` as example here.
+
+Add the `clouds.yaml` file that you want to use to the folder.
+It could look something like this:
+
+```yaml
+clouds:
+  capo-e2e:
+    auth:
+      username: demo
+      password: secretadmin
+      # If using application credentials you would have something like this instead:
+      # auth_type: v3applicationcredential
+      # application_credential_id: abc123
+      # application_credential_secret: 456def
+      user_domain_id: default
+      auth_url: https://example.com/identity
+      domain_id: default
+      project_name: demo
+    verify: false
+    region_name: RegionOne
+```
+
+Create a kustomization file named `kustomization.yaml` in the same folder:
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+# Do not add random hash to the end of the secret name
+generatorOptions:
+  disableNameSuffixHash: true
+secretGenerator:
+- files:
+  - clouds.yaml
+  name: dev-test-cloud-config
+  type: Opaque
+```
+
+If you now add `/tmp/capo-dev` to the `additional_kustomizations`, tilt will automatically apply 
+the secret when you do `tilt up`.
+
+To check that the kustomization produces the desired output, do `kustomize build /tmp/capo-dev`.
 
 ## Running E2E tests locally
 
--- a/templates/cluster-template-development.yaml
+++ b/templates/cluster-template-development.yaml
@@ -1,0 +1,13 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
+  topology:
+    class: dev-test
+    version: ${KUBERNETES_VERSION}
+    workers:
+      machineDeployments:
+      - class: default-worker
+        name: md-0
+        replicas: 1
--- a/templates/clusterclass-dev-test.yaml
+++ b/templates/clusterclass-dev-test.yaml
@@ -1,0 +1,128 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: ClusterClass
+metadata:
+  name: dev-test
+spec:
+  controlPlane:
+    ref:
+      apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+      kind: KubeadmControlPlaneTemplate
+      name: ${CLUSTER_NAME}-control-plane-template
+    machineInfrastructure:
+      ref:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+        kind: OpenStackMachineTemplate
+        name: ${CLUSTER_NAME}-control-plane-machine-template
+  infrastructure:
+    ref:
+      apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+      kind: OpenStackClusterTemplate
+      name: ${CLUSTER_NAME}-openstackcluster-template
+  workers:
+    machineDeployments:
+      - class: default-worker
+        template:
+          bootstrap:
+            ref:
+              apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+              kind: KubeadmConfigTemplate
+              name: ${CLUSTER_NAME}-default-worker-bootstraptemplate
+          infrastructure:
+            ref:
+              apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+              kind: OpenStackMachineTemplate
+              name: ${CLUSTER_NAME}-default-worker-machine-template
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
+metadata:
+  name: ${CLUSTER_NAME}-default-worker-bootstraptemplate
+spec:
+  template:
+    spec:
+      files: []
+      joinConfiguration:
+        nodeRegistration:
+          kubeletExtraArgs:
+            cloud-provider: external
+            provider-id: openstack:///'{{ instance_id }}'
+          name: '{{ local_hostname }}'
+---
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+kind: KubeadmControlPlaneTemplate
+metadata:
+  name: ${CLUSTER_NAME}-control-plane-template
+spec:
+  template:
+    spec:
+      kubeadmConfigSpec:
+        clusterConfiguration:
+          apiServer:
+            extraArgs:
+              cloud-provider: external
+          controllerManager:
+            extraArgs:
+              cloud-provider: external
+        files: []
+        initConfiguration:
+          nodeRegistration:
+            kubeletExtraArgs:
+              cloud-provider: external
+              provider-id: openstack:///'{{ instance_id }}'
+            name: '{{ local_hostname }}'
+        joinConfiguration:
+          nodeRegistration:
+            kubeletExtraArgs:
+              cloud-provider: external
+              provider-id: openstack:///'{{ instance_id }}'
+            name: '{{ local_hostname }}'
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+kind: OpenStackClusterTemplate
+metadata:
+  name: ${CLUSTER_NAME}-openstackcluster-template
+spec:
+  template:
+    spec:
+      apiServerLoadBalancer:
+        enabled: true
+      cloudName: ${OPENSTACK_CLOUD:=capo-e2e}
+      dnsNameservers:
+      - 8.8.8.8
+      identityRef:
+        kind: Secret
+        name: ${CLUSTER_NAME}-cloud-config
+      managedSecurityGroups: true
+      nodeCidr: 10.6.0.0/24
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+kind: OpenStackMachineTemplate
+metadata:
+  name: ${CLUSTER_NAME}-control-plane-machine-template
+spec:
+  template:
+    spec:
+      cloudName: ${OPENSTACK_CLOUD:=capo-e2e}
+      flavor: ${OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR:=m1.medium}
+      identityRef:
+        kind: Secret
+        name: ${CLUSTER_NAME}-cloud-config
+      image:
+        name: ubuntu-2204-kube-${KUBERNETES_VERSION}
+      sshKeyName: ${OPENSTACK_SSH_KEY_NAME:=""}
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
+kind: OpenStackMachineTemplate
+metadata:
+  name: ${CLUSTER_NAME}-default-worker-machine-template
+spec:
+  template:
+    spec:
+      cloudName: ${OPENSTACK_CLOUD:=capo-e2e}
+      flavor: ${OPENSTACK_NODE_MACHINE_FLAVOR:=m1.small}
+      identityRef:
+        kind: Secret
+        name: ${CLUSTER_NAME}-cloud-config
+      image:
+        name: ubuntu-2204-kube-${KUBERNETES_VERSION}
+      sshKeyName: ${OPENSTACK_SSH_KEY_NAME:=""}


mdbooth:
Sounds interesting! Could we have a few lines in the documentation on how to use it?

maxrantil:
@mdbooth absolutely! I just wanted @lentzi90 feedback before I did the final version + documentation.

lentzi90 (commented):
Alright let's see if I understood this.

- `openstack_env.rc` reads the e2e_conf.yaml and sets environment variables based on that
- `update_tilt-settings.sh` adds substitutions to the tilt settings that would be used to render the template when creating a Cluster from the tilt GUI

Is that correct?

I'm a bit torn on this. On the one hand it is nice to take the settings directly from the e2e config. I could have multiple different configs and easily generate clusters based on them. And we don't have to keep the e2e config in sync with anything else.

On the other hand, this means that the ClusterClass is basically unusable without an e2e config and/or script to render it. That is not so nice. For this reason I would like to put some default values for at least some of the variables. For example, I suggest putting default values for the image, network, DNS servers, failure domain and machine flavors. For the image I think we should also use a ClusterClass variable to set the version as part of the name.

Finally, I think it would be very nice to be able to use a pre-existing secret instead of generating it as part of the template. This is the hardest variable to fill and we cannot really provide a default, so it makes more sense then to expect the user to provide it as a secret already in Kubernetes.

templates/README.txt
```diff
@@ -2,3 +2,36 @@ To generate the templates in this directory run `make templates` at the top-leve
...
+1. Update Tilt settings using the script in the `templates` directory:
+   ```shell
+   $REPO_ROOT/templates/update_tilt-settings.sh <path-to-tilt-settings.yaml-file>
+   ```
+
+## Apply Network and Cluster Configurations
+1. Apply Calico CNI:
+   ```shell
+   kubectl apply -f $REPO_ROOT/workspaces/clusterclass-tilt-capo/test/e2e/data/cni/calico.yaml
+   ```
```
lentzi90, line 27: This should be in the workload cluster AFAIK.

1. Create kind cluster
2. Initialize CAPI/CAPO
3. Create ClusterClass
4. Create (workload) Cluster
5. Get the workload cluster kubeconfig
6. Apply CNI in the workload cluster

After that you should have a healthy cluster (nodes ready)

maxrantil:
> * `openstack_env.rc` reads the e2e_conf.yaml and sets environment variables based on that
> 
>     * `update_tilt-settings.sh` adds substitutions to the tilt settings that would be used to render the template when creating a Cluster from the tilt GUI
> 
> 
> Is that correct?

Yes, that is correct.

> On the other hand, this means that the ClusterClass is basically unusable without an e2e config and/or script to render it.

The script will be unusable without the e2e config but your can always set the variables yourself so the ClusterClass can still be useful if you do that. I now made the script fail if the `e2e_conf` is not present.

 > For example, I suggest putting default values for the image, network, DNS servers, failure domain and machine flavors. For the image I think we should also use a ClusterClass variable to set the version as part of the name.

Okey I did as you suggested except for for `external_network_id` which will be different on different setups. I also put defaults to `1` for `WORKER_MACHINE_COUNT` & `OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR`.
 
 >Finally, I think it would be very nice to be able to use a pre-existing secret instead of generating it as part of the template. This is the hardest variable to fill and we cannot really provide a default, so it makes more sense then to expect the user to provide it as a secret already in Kubernetes.

This part I am now fully understanding. I added some lines in the README.txt file in `$REPO_ROOT/templates/` on how you just source the `env.rc` file to get the secret. Isn't it so that you could bring your own if you like anyway?
 
 p.s. I know I will need to trim the README a little after all is done. Question: Is the README the right place for that or should it be in any of the files in `$REPO_ROOT/docs/`

templates/README.txt
maxrantil: Gotcha!
What I know, you could apply the `cni` before the ClusterClass and it works too.
But I definitely missed the kubeconfig.

lentzi90:
> The script will be unusable without the e2e config but your can always set the variables yourself so the ClusterClass can still be useful if you do that. I now made the script fail if the e2e_conf is not present.

I expressed myself badly. What I mean is that it would be incredibly inconvenient to use the ClusterClass. Think about the clouds.yaml file that is needed. Who in their right mind has it stored base64 encoded? They would need to encode it and then put it in a variable so that we can then render the template which requires this variable to be present.
In this scenario it makes much more sense to create that secret directly with `kubectl create secret` instead.

> This part I am now fully understanding. I added some lines in the README.txt file in $REPO_ROOT/templates/ on how you just source the env.rc file to get the secret. Isn't it so that you could bring your own if you like anyway?

Not sure I follow? As long as the secret is part of the template I don't see how I could bring my own? It would get overwritten by the secret in the template, and I would still need to have a value set for that environment variable. I could possibly overwrite it after I apply the template but that doesn't seem nice either.

> p.s. I know I will need to trim the README a little after all is done. Question: Is the README the right place for that or should it be in any of the files in $REPO_ROOT/docs/

I think it could fit better in [development.md](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/development/development.md)

maxrantil:
> Think about the clouds.yaml file that is needed. Who in their right mind has it stored base64 encoded? They would need to encode it and then put it in a variable so that we can then render the template which requires this variable to be present.
> In this scenario it makes much more sense to create that secret directly with `kubectl create secret` instead.

Isn't everybody having a clouds.yaml file? To get it encoded you just need to run `source $REPO_ROOT/tempates/env.rc <clouds.yaml path> <$OS_CLOUD>` .  I could take that secret part of the template away but now it follows the same structure as the earlier templates which is good i think.

Thanks for all the clarification about different kinds of secrets. I have been a little confused about that part myself that is also why I have been doing it the same way as the earlier templates.

templates/README.txt
lentzi90: If you are going to create a Cluster from the ClusterClass and apply the CNI in that Cluster definitely not. I think you are talking about applying the CNI in the management cluster. This is not needed.

For creating a Cluster from a ClusterClass, the ClusterClass must exist first. To apply the CNI in the cluster, the Cluster must exist first.

lentzi90:
> Isn't everybody having a clouds.yaml file? To get it encoded you just need to run source $REPO_ROOT/tempates/env.rc <clouds.yaml path> <$OS_CLOUD> . I could take that secret part of the template away but now it follows the same structure as the earlier templates which is good i think.

Yes everybody should have a clouds.yaml. But not everybody runs linux, not everybody has yq installed, not everybody has CAPO cloned so they can easily source this file.

This is not a structure we want to have in all templates IMO. It works well for automated CI but it is not a nice for end users.
I'm hoping we could achieve two things here

1. Convenient defaults. If it is possible to provide a default value do it. Then we have as few required variables as possible. I suggest we base the defaults on the e2e tests so they will at least fit with the devstack setup we use in CI. Then developers can reuse that without changing values.
2. Flexibility in tooling. End users are not expected to use our scripts and they should not have to, but they may want to use our templates. They may manually create secrets, they may create them through External Secrets or some other method. We should enable them to do this while still using our templates.

maxrantil:
Okey I get you and I agree fully. I didn't understand that those earlier templates where there for CI usage. I will modify it so there will only be 2-5 variables to set for the user instead of running any scripts for setting variables.

docs/book/src/development/development.md
```diff
@@ -342,6 +342,76 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
 [hack-ci-gce-project]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/gce-project.sh
 [hack-ci-openstack]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/openstack.sh
 
+## Cluster Creation with ClusterClass and Tilt
+
+This section provides a guide to create a Kubernetes cluster using ClusterClass in the development environment. It covers the process of setting up environment variables, creating a Kind cluster, configuring a secret, applying ClusterClass, deploying the cluster, monitoring its provisioning, obtaining Kubeconfig, and applying CNI to the cluster.
+
+#### Setting Environment Variables
+
+Define necessary environment variables relevant to your cluster and infrastructure. This includes setting variables for SSH key names and Kubernetes version.
```
lentzi90, line 351: Please add an explanation for what these are used for. For example, the kubernetes version will be used to select the image to use for the nodes (so this must be pre-existing and have a specific name), and the ssh key will be added to authorized keys on the nodes and bastion.

docs/book/src/development/development.md
```diff
@@ -342,6 +342,76 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+
+```bash
+kind create cluster
+export CLUSTER_TOPOLOGY=true
+clusterctl init --infrastructure openstack
+```
+
+#### Secret Configuration
+
+Apply a Kubernetes secret that your cluster configuration will reference. Ensure the secret's name aligns with the required format in your configuration.
```
lentzi90, line 372: This is too vague. What secret? What should it be named and what should it contain?
One really nice way to automate this is to use `additional_kustomizations` in the tilt settings! So the developer can add their own kustomization that is automatically applied.

docs/book/src/development/development.md
```diff
@@ -342,6 +342,76 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+
+#### Applying ClusterClass and Cluster Configuration
+
+Substitute variables in your ClusterClass configuration file and apply it to set up the cluster class.
+
+```bash
+envsubst < templates/clusterclass-dev-test.yaml | kubectl apply -f -
+envsubst < templates/cluster-template-development.yaml | kubectl apply -f -
```
lentzi90, line 380: Since we used `clusterctl` above, I would say we should use it also here. It will complain if you forgot some variables and has flags for number of replicas which can be nice.

docs/book/src/development/development.md
```diff
@@ -342,6 +342,76 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+```
+
+#### Setting up CNI and External Cloud Provider
```
lentzi90, line 383: This is good! However, I think it would fit better [in CAPI's quick-start](https://github.com/kubernetes-sigs/cluster-api/blob/28f0335712fc27626d9b03f076cd3b76948a820b/docs/book/src/user/quick-start.md?plain=1#L1444). Could you open a PR to add it there? Then we could link to it from here :slightly_smiling_face: 
Note that we also [have a page](https://cluster-api-openstack.sigs.k8s.io/topics/external-cloud-provider) about the external cloud-provider already, but what you did here would be a nice addition I think!

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,126 @@
...
+  name: dev-test-default-worker-machine-template
+spec:
+  template:
+    spec:
+      cloudName: capo-e2e
+      flavor: m1.small
+      identityRef:
+        kind: Secret
+        name: dev-test-clouds-yaml
+      image: ubuntu-2204-kube-${KUBERNETES_VERSION}
```
lentzi90, line 125: This version should be set using [ClusterClass variables](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass#builtin-variables). Otherwise it is hard coded once you render the ClusterClass, so all Clusters created from it will need to use the same version.

docs/book/src/development/development.md
```diff
@@ -342,6 +342,76 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+#### Using ClusterClass with Tilt for Development
+
+To utilize Tilt for an efficient and streamlined development process, add the following section under `kustomize_substitutions:` in your `tilt-settings.yaml` file:
+
+```yaml
+kustomize_substitutions:
+  CLUSTER_TOPOLOGY: "true"
+  OPENSTACK_SSH_KEY_NAME: "your_ssh_key_name_here"
+  KUBERNETES_VERSION: "v1.28.5"
+  NAMESPACE: "default"
```
lentzi90, line 412: If you use ClusterClass variables to set the version this will no longer be necessary. Then we can set the version directly from the Tilt UI. The namespace can already be set from there.
```suggestion
```

templates/cluster-template-development.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: dev-test
```
lentzi90, line 4:
```suggestion
  name: ${CLUSTER_NAME}
```
Note: We keep the class hard coded, but not the cluster name. For tilt it doesn't matter but in other use-cases it will be more useful if the cluster name can be changed.

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,126 @@
...
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
+kind: OpenStackMachineTemplate
+metadata:
+  name: dev-test-control-plane-machine-template
+spec:
+  template:
+    spec:
+      cloudName: capo-e2e
+      flavor: m1.medium
```
lentzi90, line 106: I think we need to make these variables and add them to the kustomize substitutions

lentzi90 (commented):
Great job @maxrantil ! I think you learned a ton making this :smile: 
It is already usable like this, I tested it just now!
Added some comments below with suggestions though. If you want to split out some of them into follow up PRs I think that would also be fine

docs/book/src/development/development.md
maxrantil: TODO

maxrantil:
I think this is done now @lentzi90 , maybe you mentioned that the docs for Tilt should be in another place. I'll fix that if so. We also talked about having another name for OPENSTACK_SSH_KEY_NAME but in the other templates they use this name so I guess it makes sense to use it here too?

docs/book/src/development/development.md
```diff
@@ -342,6 +342,86 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+
+# [Optional] SSH Keypair Name for Instances in OpenStack (Default: "")
+export OPENSTACK_SSH_KEY_NAME="<openstack_keypair_name>"
+# [Optional] Control Plane Machine Flavor (Default: m1.medium)
+export OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR="<openstack_control_plane_machine_flavor>"
+# [Optional] Node Machine Flavor (Default: m1.small)
+export OPENSTACK_NODE_MACHINE_FLAVOR="<openstack_node_machine_flavor>"
+# [Optional] OpenStack Cloud Environment (Default: capo-e2e)
+export OPENSTACK_CLOUD="<openstack_cloud>"
+
```
lentzi90, line 368:
```suggestion
```

docs/book/src/development/development.md
```diff
@@ -342,6 +342,86 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
 [hack-ci-gce-project]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/gce-project.sh
 [hack-ci-openstack]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/openstack.sh
 
+## Cluster Creation with ClusterClass
+
+This section provides a guide to create a Kubernetes cluster using ClusterClass. It covers the process of setting up environment variables, creating a Kind cluster, configuring a secret, applying ClusterClass and deploying the cluster.
+
+### Setting Environment Variables
+
```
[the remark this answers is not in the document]
lentzi90: Good point. The main difference and benefit here is that we are adding a ClusterClass that is geared specifically at development. It has defaults that fit the devstack setup so that as few variables as possible are needed. I think we need to point this out @maxrantil 
I'll add a few suggestions

docs/book/src/development/development.md
```diff
@@ -342,6 +342,86 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
 [hack-ci-gce-project]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/gce-project.sh
 [hack-ci-openstack]: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/hack/ci/openstack.sh
 
+## Cluster Creation with ClusterClass
```
lentzi90, line 345: Could you move this whole section up to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e89ec9ce14067866bab80c84da5060980a378f36/docs/book/src/development/development.md?plain=1#L52 ?
```suggestion
## Test Cluster Creation with the dev-test ClusterClass
```

docs/book/src/development/development.md
```diff
@@ -342,6 +342,86 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+
+This section provides a guide to create a Kubernetes cluster using ClusterClass. It covers the process of setting up environment variables, creating a Kind cluster, configuring a secret, applying ClusterClass and deploying the cluster.
+
+### Setting Environment Variables
+
+These environment variables are critical for configuring your Kubernetes cluster on OpenStack. cluster_name identifies your cluster within the infrastructure. openstack_ssh_key_name is added to the authorized keys on the nodes and bastion. kubernetes_version selects the pre-existing image used for the nodes. openstack_control_plane_machine_flavor (optional) defines the resource allocation for control plane machines. openstack_node_machine_flavor (optional) sets the resource specifications for worker nodes. openstack_cloud (optional) determines the OpenStack cloud environment for deploying your cluster.
```
lentzi90, line 351: This seems a bit unnecessary since the variables are explained in the snippet below also. I would instead explain here that the defaults should work well with the devstack setup (which is documented below if we move this whole section up). Point out in particular the image names! There *must* be an image named `ubuntu-2204-kube-{{ KUBERNETES_VERSION }}` used for the nodes

docs/book/src/development/development.md
```diff
@@ -342,6 +342,86 @@ $ sudo ln -s /run/user/$(id -u)/podman/podman.sock /var/run/docker.sock
...
+### Applying ClusterClass and Cluster Configuration
+
+Use `clusterctl` to substitute variables in your ClusterClass configuration file and apply it to set up the cluster class and cluster.
+
+```bash
+clusterctl generate cluster ${CLUSTER_NAME} --from templates/clusterclass-dev-test.yaml  | kubectl apply -f -
+clusterctl generate cluster ${CLUSTER_NAME} --from templates/cluster-template-development.yaml  | kubectl apply -f -
+```
+
+#### Using Tilt with CAPO
```
lentzi90, line 398: Merge this with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e89ec9ce14067866bab80c84da5060980a378f36/docs/book/src/development/development.md?plain=1#L53
I missed that we already had a link to it :facepalm:

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: ${CLUSTER_NAME}-control-plane-machine-template
+spec:
+  template:
+    spec:
+      cloudName: ${OPENSTACK_CLOUD:=capo-e2e}
+      flavor: ${OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR:=m1.medium}
+      identityRef:
+        kind: Secret
+        name: ${CLUSTER_NAME}-clouds-yaml
+      image: ubuntu-2204-kube
```
lentzi90, line 141:
```suggestion
      image: cirros-0.6.1-x86_64-disk
```
This is the image we use in the e2e tests and it is automatically added in the devstack

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: ${CLUSTER_NAME}-default-worker-machine-template
+spec:
+  template:
+    spec:
+      cloudName: ${OPENSTACK_CLOUD:=capo-e2e}
+      flavor: ${OPENSTACK_NODE_MACHINE_FLAVOR:=m1.small}
+      identityRef:
+        kind: Secret
+        name: ${CLUSTER_NAME}-clouds-yaml
+      image: ubuntu-2204-kube
```
lentzi90, line 156:
```suggestion
      image: cirros-0.6.1-x86_64-disk
```

templates/clusterclass-dev-test.yaml
lentzi90: Wait I think I was too quick. Somehow I thought this was the bastion image. Please Ignore!

docs/book/src/development/development.md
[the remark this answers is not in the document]
maxrantil: I tried to explain it more clear in the documentation now. See what you think now.

lentzi90 (commented):
Awesome! I'm happy with this now :slightly_smiling_face: 
/approve

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: dev-test
+spec:
+  controlPlane:
+    ref:
+      apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+      kind: KubeadmControlPlaneTemplate
+      name: ${CLUSTER_NAME}-control-plane-template
+    machineInfrastructure:
+      ref:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
```
EmilienM, line 13: the latest version is v1alpha8, let's use it please.

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+    - selector:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7
+        kind: OpenStackMachineTemplate
+        matchResources:
+          controlPlane: true
+      jsonPatches:
+      - op: add
+        path: /spec/template/spec/image
+        valueFrom:
+          template: |
```
EmilienM, line 48: it is now an ImageFilter, https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1alpha8/types.go#L25-L32
so it'll be

```
name: ubuntu-2204-kube-{{ .builtin.controlPlane.version }}
```

templates/clusterclass-dev-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+        matchResources:
+          machineDeploymentClass:
+            names:
+            - default-worker
+      jsonPatches:
+      - op: add
+        path: /spec/template/spec/image
+        valueFrom:
+          template: |
+            ubuntu-2204-kube-{{ .builtin.machineDeployment.version }}
```
EmilienM, line 65: ditto

EmilienM (changes requested):
Minor changes.

templates/clusterclass-dev-test.yaml
EmilienM, line 156: You missed the `-{{ KUBERNETES_VERSION }}`

templates/clusterclass-dev-test.yaml
EmilienM, line 141: You missed the `-{{ KUBERNETES_VERSION }}`

EmilienM: And also in v1alpha8 it's a filter, so add name.

docs/book/src/development/development.md
```diff
@@ -50,9 +50,80 @@ make docker-build docker-push
...
+export CLUSTER_TOPOLOGY=true
+clusterctl init --infrastructure openstack
+```
+
+### Secret Configuration
+
+CAPO needs a clouds.yaml file in order to manage the OpenStack resources needed for the Cluster. This should be supplied as a secret named `${CLUSTER_NAME}-clouds-yaml`. You can create this secret for example with:
+
+```bash
+kubectl create secret generic ${CLUSTER_NAME}-clouds-yaml --from-file=clouds.yaml
```
EmilienM, line 92: note that clouds.yaml secret is named `${CLUSTER_NAME}-cloud-config` in our templates, we might want to keep it consistent.

docs/book/src/development/development.md
```diff
@@ -50,9 +50,80 @@ make docker-build docker-push
...
+  # [Optional] Node Machine Flavor (Default: m1.small)
+  OPENSTACK_NODE_MACHINE_FLAVOR: "<openstack_node_machine_flavor>"
+  # [Optional] OpenStack Cloud Environment (Default: capo-e2e)
+  OPENSTACK_CLOUD: "<openstack_cloud>"
+
+additional_kustomizations:
+  secret_kustomization: /path/to/kustomize/secret/configuration
+```
+
+Ensure the specified path (`/path/to/kustomize/secret/configuration`) contains both the `clouds.yaml` file and a `kustomization.yaml` file. The `kustomization.yaml` should define the necessary resources, such as a Kubernetes secret, using the `clouds.yaml` file.
```
EmilienM, line 126: as a noob I'll love to find out how you did :) but we can add it later.

EmilienM: in fact, we can simply create the secret beforehand

EmilienM (commented):
I finally managed to get tilt working, thanks for that nice tutorial !
I find it super useful. I had minor comments, otherwise LGTM.

Thanks again

templates/clusterclass-dev-test.yaml
lentzi90: > You missed the -{{ KUBERNETES_VERSION }}

The whole thing is actually overwritten with the patch so the variable is not needed. In fact I think we could remove the `image` completely.

templates/clusterclass-dev-test.yaml
lentzi90: Good point! I don't know how I missed that.
The v1alpha8 API is not in any release yet, so using it would require building your own image. However, with tilt that part is automatically done for you anyway, and the whole idea was to have something for development so I agree we should use v1alpha8 :slightly_smiling_face:

docs/book/src/development/development.md
maxrantil: That is right, but it is a bit confusing because we use both clouds.yaml and cloud.conf files. I had it like the templates before but @lentzi90 suggested this change. What do you say Lentzi?

docs/book/src/development/development.md
maxrantil: > as a noob I'll love to find out how you did :) but we can add it later.
:) What do you mean exactly with later? Do you suggest taking the explanation away for now?

docs/book/src/development/development.md
EmilienM: I suggest we try to keep the name consistent across the board even if it's just a name, so my preference goes over using `${CLUSTER_NAME}-cloud-config` everywhere.

docs/book/src/development/development.md
EmilienM: if you're fine adding it in this PR, otherwise we can do it later via another PR, that's what I meant.
Thanks

docs/book/src/development/development.md
lentzi90: Ah I had missed that the other templates used this format. I agree it would be nice with `${CLUSTER_NAME}-cloud-config` to keep things consistent. However, I now realized an issue with this...

We cannot use the `CLUSTER_NAME` variable in the ClusterClass! It is supposed to be used to create multiple clusters with unique names. Actually now that I tested it in tilt, I get an error that the name is invalid for all the objects that has a name starting with `CLUSTER_NAME`. The names end up like `-control-plane-template`, which errors because it starts with `-`.

I think we should put a fixed name on the secret and all the other objects used by the ClusterClass. For example `dev-test`. So the secret would be named `dev-test-clouds-yaml` or `dev-test-cloud-config`. I'm fine with either.

templates/clusterclass-dev-test.yaml
lentzi90: This is what I hinted at when I said it is not part of a release. `clusterctl init` will not use the latest manifests and CRDs directly from the main branch since that would be risky and unstable. It will by default find the latest release and use that. For CAPO that means using v1alpha7.

However, for development that is not so useful since we need to test the latest changes that we are working on. Tilt is already configured to take the latest code and CRDs so if you use that, v1alpha8 will work. If you want to do it with clusterctl init, you need to [use local artifacts](https://cluster-api.sigs.k8s.io/clusterctl/developers#use-local-artifacts). I may be good to mention this, but I would not spend too much time on it. Just link to the CAPI docs for how to do it.

docs/book/src/development/development.md
lentzi90: > in fact, we can simply create the secret beforehand

Yes you can freely choose how to do it. The additional kustomization is just a convenience to make tilt apply it automatically. :slightly_smiling_face: 

It would probably be a good idea to give an example here. Perhaps something like this:

````suggestion
Ensure the specified path (`/path/to/kustomize/secret/configuration`) contains both the `clouds.yaml` file and a `kustomization.yaml` file. The `kustomization.yaml` should define the necessary resources, such as a Kubernetes secret, using the `clouds.yaml` file.

For example, if you want to automatically create a secret named `dev-test-clouds-yaml` with the content of your `clouds.yaml` every time you do `tilt up`, you could do the following.
Create a folder to hold the kustomization.
We will use `/tmp/capo-dev` as example here.
Add the `clouds.yaml` file that you want to use to the folder.
It could look something like this:

```yaml
clouds:
  capo-e2e:
    auth:
      username: demo
      password: secretadmin
      # If using application credentials you would have something like this instead:
      # auth_type: v3applicationcredential
      # application_credential_id: abc123
      # application_credential_secret: 456def
      user_domain_id: default
      auth_url: https://example.com/identity
      domain_id: default
      project_name: demo
    verify: false
    region_name: RegionOne
```

Create a kustomization file named `kustomization.yaml` in the same folder:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
# Do not add random hash to the end of the secret name
generatorOptions:
  disableNameSuffixHash: true

secretGenerator:
- files:
  - clouds.yaml
  name: dev-test-clouds-yaml
  type: Opaque
```

If you now add `/tmp/capo-dev` to the `additional_kustomizations`, tilt will automatically apply the secret when you do `tilt up`.
To check that the kustomization produces the desired output, do `kustomize build /tmp/capo-dev`.
````

templates/clusterclass-dev-test.yaml
EmilienM: Yeah I agree with Lennart here, let's focus on how to use tilt

templates/clusterclass-dev-test.yaml
maxrantil: I have been trying hard to get the patching to work with `v1alpha8` but no luck there. So I have now ended up doing it without patches.

docs/book/src/development/development.md
maxrantil: The name of the secret is now changed to follow the earlier templates.
Regarding the `CLUSTER_NAME`, it is only to add it into a variable in the `tilt-settings.yaml/json` file and then it should work fine. I have updated the docs about it.

templates/clusterclass-dev-test.yaml
maxrantil: Okey the docs and everything else is now focusing on Tilt only.

templates/clusterclass-dev-test.yaml
maxrantil: Are you guys fine without using patching? @EmilienM @lentzi90 , I have a comment a bit down about me trying it without any progress in `v1alpha8`.

mdbooth:
This is wonderful. If you have time we'll want to have the new templates generated from our existing kustomize or they'll start to drift. I wouldn't want to hold it up for that, though.

/lgtm

EmilienM:
@maxrantil I wanted to thank you for your contribution. This really helped me to get into Tilt (and probably changed my life :rofl:) !

maxrantil:
Thank you all for your lovely feedback <3

PR #1834, opened 2024-01-19
e2e: remove "allow-all-in-cluster-traffic" patch

EmilienM:
The default rules should be sufficient for Calico which is used by e2e
clusters; let's remove the allow-all-in-cluster-traffic patch which
should not be needed.


jichenjc:
/approve

test passed, should be good as we don't need allow all (Calico is default)

lentzi90 (commented):
Nice!
/lgtm

PR #1836, opened 2024-01-19
✨ Modify OpenStackCluster.Spec.Network API

MaysaMacedo:
For the BYO scenario, when the `OpenStackCluster.Spec.Network`
is not specified the query to OpenStack will return all the
Networks available in the cloud and fail the reconciliation.
To avoid this, if any Subnets were specified under
`OpenStackCluster.Spec.Subnets` this can used to identify
which Network to use.

Note, this pattern to infer a field based on another field has already been used by the
`OpenStackCluster.Spec.Subnets` and also when no Network is defined in a Port.

**TODOs**:

  - [x] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -23,6 +23,8 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -507,36 +509,41 @@ func reconcileNetworkComponents(scope scope.Scope, cluster *clusterv1.Cluster, o
 	if len(openStackCluster.Spec.ManagedSubnets) == 0 {
 		scope.Logger().V(4).Info("No need to reconcile network, searching network and subnet instead")
 
-		netOpts := openStackCluster.Spec.Network.ToListOpt()
-		networkList, err := networkingService.GetNetworksByFilter(&netOpts)
-		if err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find network: %w", err))
-			return fmt.Errorf("failed to find network: %w", err)
-		}
-		if len(networkList) == 0 {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find any network"))
-			return fmt.Errorf("failed to find any network")
-		}
-		if len(networkList) > 1 {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("found multiple networks (result: %v)", networkList))
-			return fmt.Errorf("found multiple networks (result: %v)", networkList)
-		}
 		if openStackCluster.Status.Network == nil {
 			openStackCluster.Status.Network = &infrav1.NetworkStatusWithSubnets{}
 		}
-		openStackCluster.Status.Network.ID = networkList[0].ID
-		openStackCluster.Status.Network.Name = networkList[0].Name
-		openStackCluster.Status.Network.Tags = networkList[0].Tags
 
-		subnets, err := filterSubnets(networkingService, openStackCluster)
+		err := getCAPONetwork(openStackCluster, networkingService)
 		if err != nil {
 			return err
+		}
+
+		filteredSubnets, err := filterSubnets(networkingService, openStackCluster)
+		if err != nil {
+			return err
+		}
+
+		var subnets []infrav1.Subnet
+		for subnet := range filteredSubnets {
+			filterSubnet := &filteredSubnets[subnet]
+			subnets = append(subnets, infrav1.Subnet{
+				ID:   filterSubnet.ID,
+				Name: filterSubnet.Name,
+				CIDR: filterSubnet.CIDR,
+				Tags: filterSubnet.Tags,
+			})
 		}
 
 		if err := utils.ValidateSubnets(subnets); err != nil {
 			return err
 		}
 		openStackCluster.Status.Network.Subnets = subnets
+
+		// If network is not yet populated on the Status, use networkID defined in the filtered subnets to get the Network.
+		err = populateCAPONetworkFromSubnet(networkingService, filteredSubnets, openStackCluster)
+		if err != nil {
+			return err
+		}
 	} else if len(openStackCluster.Spec.ManagedSubnets) == 1 {
 		err := networkingService.ReconcileNetwork(openStackCluster, clusterName)
 		if err != nil {
@@ -687,18 +694,23 @@ func handleUpdateOSCError(openstackCluster *infrav1.OpenStackCluster, message er
 }
 
 // filterSubnets retrieves the subnets based on the Subnet filters specified on OpenstackCluster.
-func filterSubnets(networkingService *networking.Service, openStackCluster *infrav1.OpenStackCluster) ([]infrav1.Subnet, error) {
-	var subnets []infrav1.Subnet
+func filterSubnets(networkingService *networking.Service, openStackCluster *infrav1.OpenStackCluster) ([]subnets.Subnet, error) {
+	var filteredSubnets []subnets.Subnet
+	var err error
 	openStackClusterSubnets := openStackCluster.Spec.Subnets
-	if openStackCluster.Status.Network == nil {
-		return nil, nil
+	networkID := ""
+	if openStackCluster.Status.Network != nil {
+		networkID = openStackCluster.Status.Network.ID
 	}
-	networkID := openStackCluster.Status.Network.ID
+
 	if len(openStackClusterSubnets) == 0 {
+		if networkID == "" {
+			return nil, nil
+		}
 		empty := &infrav1.SubnetFilter{}
 		listOpt := empty.ToListOpt()
 		listOpt.NetworkID = networkID
-		filteredSubnets, err := networkingService.GetSubnetsByFilter(listOpt)
+		filteredSubnets, err = networkingService.GetSubnetsByFilter(listOpt)
 		if err != nil {
 			err = fmt.Errorf("failed to find subnets: %w", err)
 			if errors.Is(err, networking.ErrFilterMatch) {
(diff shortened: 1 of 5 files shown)


pkg/cloud/services/networking/network.go
```diff
@@ -395,3 +395,28 @@ func (s *Service) ConvertOpenStackSubnetToCAPOSubnet(subnets []infrav1.Subnet, f
 	})
 	return subnets
 }
+
+// ConvertOpenStackNetworkToCAPONetwork converts an OpenStack network to a capo network.
+// It returns the converted subnet.
+func (s *Service) ConvertOpenStackNetworkToCAPONetwork(openStackCluster *infrav1.OpenStackCluster, network *networks.Network) {
```
EmilienM, line 401: `s` is not used, do we really need it?

MaysaMacedo: Nop, I just updated it. Thanks!

pkg/cloud/services/networking/network.go
```diff
@@ -395,3 +395,28 @@ func (s *Service) ConvertOpenStackSubnetToCAPOSubnet(subnets []infrav1.Subnet, f
...
+	openStackCluster.Status.Network.ID = network.ID
+	openStackCluster.Status.Network.Name = network.Name
+	openStackCluster.Status.Network.Tags = network.Tags
+}
+
+// PopulateCAPONetworkFromSubnet gets a network based on the networkID of the subnets and converts it to the CAPO format.
+// It returns an error in case it failed to retrieve the network.
+func (s *Service) PopulateCAPONetworkFromSubnet(subnets []subnets.Subnet, openStackCluster *infrav1.OpenStackCluster) error {
+	if openStackCluster.Status.Network.ID == "" && len(subnets) > 0 {
+		if len(subnets) > 1 && subnets[0].NetworkID != subnets[1].NetworkID {
```
jichenjc, line 411: just curious why this has subnet[0] and [1] ? some case we have 2 subnets and only 2 subnet?

MaysaMacedo: In case a dual-stack cluster is desired, one IPv4 and IPv6 subnet needs to be specified. The API already enforces the max number of subnets to be 2.

mdbooth (commented):
We're doing a very similar dance [in normalizePortTarget](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e89ec9ce14067866bab80c84da5060980a378f36/pkg/cloud/services/compute/instance.go#L54-L122): we are getting a network a subnets from a NetworkFilter and []SubnetFilter and we infer the network from the subnets if it's not specified. Is there any opportunity to refactor that code to be used in both places?

MaysaMacedo:
> We're doing a very similar dance [in normalizePortTarget](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/e89ec9ce14067866bab80c84da5060980a378f36/pkg/cloud/services/compute/instance.go#L54-L122): we are getting a network a subnets from a NetworkFilter and []SubnetFilter and we infer the network from the subnets if it's not specified. Is there any opportunity to refactor that code to be used in both places?

The scenario I'm handling is a bit different because I need the hole Network and not just the ID in order to populate the `OpenStackCluster.status`. Also, the way we filter the subnets is also a bit different because we need to account for when there is no subnets or various subnets. So the end goal for Ports and Network is the same, but there are restrictions that IMHO make the refactor not feasible.

controllers/openstackcluster_controller.go
```diff
@@ -506,26 +507,33 @@ func reconcileNetworkComponents(scope scope.Scope, cluster *clusterv1.Cluster, o
 			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find any network"))
 			return fmt.Errorf("failed to find any network")
 		}
-		if len(networkList) > 1 {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("found multiple networks (result: %v)", networkList))
-			return fmt.Errorf("found multiple networks (result: %v)", networkList)
-		}
 		if openStackCluster.Status.Network == nil {
```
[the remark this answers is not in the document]
MaysaMacedo: ack, I added a condition checking if the filter is empty.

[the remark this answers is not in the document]
EmilienM: > ack, I added a condition checking if the filter is empty.

Where?

controllers/openstackcluster_controller.go
```diff
@@ -709,8 +718,55 @@ func filterSubnets(networkingService *networking.Service, openStackCluster *infr
...
+	openStackCluster.Status.Network.Name = network.Name
+	openStackCluster.Status.Network.Tags = network.Tags
+}
+
+// populateCAPONetworkFromSubnet gets a network based on the networkID of the subnets and converts it to the CAPO format.
+// It returns an error in case it failed to retrieve the network.
+func populateCAPONetworkFromSubnet(networkingService *networking.Service, subnets []subnets.Subnet, openStackCluster *infrav1.OpenStackCluster) error {
+	if openStackCluster.Status.Network.ID == "" && len(subnets) > 0 {
+		if len(subnets) > 1 && subnets[0].NetworkID != subnets[1].NetworkID {
+			return fmt.Errorf("unable to identify the Network to use. NetworkID from Subnets differ: %s, %s", subnets[0].NetworkID, subnets[1].NetworkID)
```
EmilienM, line 740: Can we improve the logging maybe? Here is an example (can be improved):

```suggestion
			return fmt.Errorf("unable to identify the Network to use. NetworkID %s from subnet %s does not match NetworkID %s from subnet %s", subnets[0].NetworkID, subnets[0].ID, subnets[1].NetworkID, subnets[1].ID)
```

controllers/openstackcluster_controller.go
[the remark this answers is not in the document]
MaysaMacedo: @EmilienM No, [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1836/files#diff-39b99bc838c9f76967790a10572cc44f0c41ccacddcd6e1eff934828d4b050a3R756).

controllers/openstackcluster_controller.go
```diff
@@ -709,8 +718,55 @@ func filterSubnets(networkingService *networking.Service, openStackCluster *infr
...
+		network, err := networkingService.GetNetworkByID(subnets[0].NetworkID)
+		if err != nil {
+			return err
+		}
+		convertOpenStackNetworkToCAPONetwork(openStackCluster, network)
+	}
+	return nil
+}
+
+// getCAPONetwork gets a network based on a filter.
```
EmilienM, line 752: nit: add a bit more doc into this function. Yes indeed, we get a network based on the filter and return errors if not found. But if found, we will update the status (by calling convertOpenStackNetworkToCAPONetwork). Also we return no error if Spec.Network is empty because we now try to find the Network specs by the Subnets field.

controllers/openstackcluster_controller.go
```diff
@@ -709,8 +718,55 @@ func filterSubnets(networkingService *networking.Service, openStackCluster *infr
...
+		networkList, err := networkingService.GetNetworksByFilter(&netOpts)
+		if err != nil {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find network: %w", err))
+			return fmt.Errorf("failed to find network: %w", err)
+		}
+		if len(networkList) == 0 {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find any network"))
+			return fmt.Errorf("failed to find any network")
+		}
+		if len(networkList) == 1 {
```
EmilienM, line 767: I'll like to see a switch probably, and return an error if more than one network is found maybe?

controllers/openstackcluster_controller.go
```diff
@@ -709,8 +718,55 @@ func filterSubnets(networkingService *networking.Service, openStackCluster *infr
...
 			}
-			subnets = networkingService.ConvertOpenStackSubnetToCAPOSubnet(subnets, filteredSubnet)
+			filteredSubnets = append(filteredSubnets, *filteredSubnet)
 		}
 	}
-	return subnets, nil
+	return filteredSubnets, nil
+}
+
+// convertOpenStackNetworkToCAPONetwork converts an OpenStack network to a capo network.
```
EmilienM, line 727: more of an code architecture question (ping @mdbooth / @lentzi90), I would have put these functions into network.go instead, as I've seen it elsewhere. However I'm not sure if we have a clear line on that.

controllers/openstackcluster_controller.go
MaysaMacedo: My proposal is intentional. If multiple networks are found, we won't fail and will give another chance to look up the network by the subnet later on on `populateCAPONetworkFromSubnet`.

controllers/openstackcluster_controller.go
```diff
@@ -709,8 +718,55 @@ func filterSubnets(networkingService *networking.Service, openStackCluster *infr
...
+	openStackCluster.Status.Network.Name = network.Name
+	openStackCluster.Status.Network.Tags = network.Tags
+}
+
+// populateCAPONetworkFromSubnet gets a network based on the networkID of the subnets and converts it to the CAPO format.
+// It returns an error in case it failed to retrieve the network.
+func populateCAPONetworkFromSubnet(networkingService *networking.Service, subnets []subnets.Subnet, openStackCluster *infrav1.OpenStackCluster) error {
+	if openStackCluster.Status.Network.ID == "" && len(subnets) > 0 {
+		if len(subnets) > 1 && subnets[0].NetworkID != subnets[1].NetworkID {
+			return fmt.Errorf("unable to identify the Network to use. NetworkID %s from subnet %s does not match NetworkID %s from subnet %s", subnets[0].NetworkID, subnets[0].ID, subnets[1].NetworkID, subnets[1].ID)
```
EmilienM, line 740: the error message can be improved a little bit, because we also check for the length of subnets.
also please s/Network/network/ to be consistent with subnet.

controllers/openstackcluster_controller.go
EmilienM: I'm still convinced that we want to move these new functions into `network.go`.

controllers/openstackcluster_controller.go
EmilienM: :brain: :boom:

controllers/openstackcluster_controller_test.go
```diff
@@ -544,6 +546,45 @@ var _ = Describe("OpenStackCluster controller", func() {
 		Expect(err).To(BeNil())
 		Expect(len(testCluster.Status.Network.Subnets)).To(Equal(2))
 	})
+
+	It("should allow fetch network by subnet", func() {
```
mdbooth, line 550: Great test, thanks.

mdbooth (approved):

MaysaMacedo:
/hold cancel
CI passed

PR #1839, opened 2024-01-24
🐛 controllers: do not return a RequeueAfter and an error at the same time

### LINKED ISSUE #1838: controllers: non-empty reconcile result and error
/kind bug

Since controller-runtime https://github.com/kubernetes-sigs/controller-runtime/pull/2451, a warning will be logged when the OpenStack controllers is preparing the bastion VM, which was returning an error and a non-zero result.

Side note:
~~We also saw infinite retries using the same reconcile ID, this might or might not be related to that issue.
EDIT: I saw it in old jobs (from like a month ago) where the `reconcileID` was re-used thousands of times; I guess this is normal.~~
EDIT2: It's a real issue: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1840

EmilienM:
Since controller-runtime kubernetes-sigs/controller-runtime#2451, a warning will be logged if a controller returns an error and a non-zero result (i.e. a result with a RequeueAfter).
It adjusts the OpenStackCluster controller to return errors first, as an error itself will already trigger a reconciliation (with exponential backoff).

Fixes #1838

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -25,6 +25,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -69,7 +70,7 @@ type OpenStackClusterReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
 
-func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, reterr error) {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Fetch the OpenStackCluster instance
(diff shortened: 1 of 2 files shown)


EmilienM:
/hold
late for me now so I'll check tomorrow our CI logs, to verify that:
1) We don't get the warning anymore
2) We don't use the same reconciler ID thousands of times anymore.

EmilienM:
FYI @mnaser @mdbooth @jichenjc @lentzi90 
This PR (when finalized & reviewed) will need to be backported because https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753 was backported and introduced the issue.

EmilienM:
logs look ok now, no more warning nor error about the missing field "status" field:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1839/pull-cluster-api-provider-openstack-e2e-test/1750120954077384704/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-lpzvt/manager.log

/hold cancel

controllers/openstackcluster_controller.go
```diff
@@ -297,8 +306,11 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
 	}
 
 	result, err := reconcileBastion(scope, cluster, openStackCluster)
-	if err != nil || !reflect.DeepEqual(result, reconcile.Result{}) {
-		return result, err
+	if err != nil {
+		return reconcile.Result{}, err
```
EmilienM, line 310: here I can also comment that in reconcileBastion we might return a result with Requeuing.

I'm honestly not sure we want that but I'm happy to do it.

api/v1alpha8/openstackcluster_types.go
```diff
@@ -167,6 +167,8 @@ type OpenStackClusterSpec struct {
 
 // OpenStackClusterStatus defines the observed state of OpenStackCluster.
 type OpenStackClusterStatus struct {
+	// Ready is true when the cluster infrastructure is ready.
+	// +optional
```
EmilienM, line 171: This sounds odd to me that we don't `omitempty` but I've kept it consistent with `OpenStackMachine.Status.Ready`. I'm happy to do it another way, like add `omitempty` to the `OpenStackMachine.Status.Ready` too or ... ?

jichenjc: maybe you can propose another PR for that purpose later if we decide `omitempty` is needed :)

jichenjc (commented):
/lgtm

dulek (approved):
/lgtm

api/v1alpha8/openstackcluster_types.go
EmilienM: ok, let me split it. I did it within the same PR because I saw the second error right after the warning, so I thought I would fix them together but I think it's best to split.

EmilienM: done via https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1842

EmilienM:
We paired up with @mdbooth and this is an easier way to fix this. I looked in the logs, and indeed no more warnings.

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1839/pull-cluster-api-provider-openstack-e2e-test/1750547077026287616/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-k69mh/manager.log

lentzi90 (commented):
/approve

PR #1829, opened 2024-01-17
🐛 Persist API FloatingIP immediately on creation

mdbooth:
If loadbalancer reconciliation failed at any time after the floating IP was created, we would lead the floating IP and create a new one on the next reconciliation.

Fixes #1827

/hold

Issue #1687, opened 2023-09-27
additionalPorts and allowedCidrs loops and never completes LB provisioning


huxcrux:
/kind bug

**What steps did you take and what happened:**
When creating a cluster and defining booth one or more additionalPorts and at least one CIDR under allowedCidrs this causes the LB to never become fully functional and no cluster to be created.

It seems like there is a check missing for if the LB is out of PENDING_UPDATE causing the API to respond with 

> capo-system/capo-controller-manager-7c78d95c95-7qphw[manager]: I0926 13:49:42.720981       1 recorder.go:104] "events: Failed to create listener k8s-clusterapi-cluster-default-hux-lab1-kubeapi-31991: Expected HTTP response code [201 202] when accessing [POST https://<OPS-URL>:9876/v2.0/lbaas/listeners], but got 409 instead\n{\"faultcode\": \"Client\", \"faultstring\": \"Load Balancer 7d690fcb-ea67-4462-a49e-dda806c8f792 is immutable and cannot be updated.\", \"debuginfo\": null}" type="Warning" object={Kind:OpenStackCluster Namespace:default Name:hux-lab1 UID:d973908d-ccd0-4f5b-9f9f-d628143d021a APIVersion:infrastructure.cluster.x-k8s.io/v1alpha7 ResourceVersion:33567788 FieldPath:} reason="Failedcreatelistener"

I have also tested booth additionalPorts and allowedCidrs one by one and it works as intended, it's just when booth are used at the same time this is occuring.

Also worth noticing is that the allowedCidrs option is being set on the LB and I have verified that the security groups for the LB contains the correct rules meaning this is just a race condition during provisioning

**What did you expect to happen:**
The flow seems to be:
1. Create LB
2. Create Listener
3. If allowedCidrs is defined update the listener with the allowedCidrs list.
4. Create the 2nd listener (this is done without checking if the LB is no longer in PENDING_UPDATE state causing an API failure and CAPO the allocates a ned FIP for the LB and try reconciling listeners again, however since the LB have a FIP connected it will continue to loop in this state)

**Anything else you would like to add:**
A gist with completed output: https://gist.github.com/huxcrux/7c288f1c0b045de67eac1beaf3211e6e
I redacted IPs and Openstack API URIs. Every time FIP attachment fails it's a new FIP that has been allocated.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.8.0
[...]

PR #1842, opened 2024-01-25
🐛 api: openstackcluster.status default to false

### LINKED ISSUE #1841: OpenStackCluster.Status.Ready being required causing issues in reconcile
/kind bug

When fixing https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1838 - I noticed this error in our logs:
```
OpenStackCluster.infrastructure.cluster.x-k8s.io \"cluster-e2e-rha0r3\" is invalid: ready: Required value"
```

This is because the reconcile leaves early when the OpenStackCluster.Status.Ready is not set yet and this field is required.
I think this think should not be required like it's the case for the OpenStackMachine.Status.Ready (omitempty) so we don't have these troubles.

EmilienM:

When a reconcile loop for the bastion is requeued, we have
this error:
```
OpenStackCluster.infrastructure.cluster.x-k8s.io \"cluster-e2e-rha0r3\" is invalid: ready: Required value"
```

The OpenStackMachine.Status is false by default now, so if the status
has not been set to anything, patching the object will not fail with the
previous error.

Fixes #1841

**TODOs**:

- [X] squashed commits

### DIFF
--- a/api/v1alpha8/openstackcluster_types.go
+++ b/api/v1alpha8/openstackcluster_types.go
@@ -167,6 +167,8 @@ type OpenStackClusterSpec struct {
 
 // OpenStackClusterStatus defines the observed state of OpenStackCluster.
 type OpenStackClusterStatus struct {
+	// Ready is true when the cluster infrastructure is ready.
+	// +kubebuilder:default=false
 	Ready bool `json:"ready"`
 
 	// Network contains information about the created OpenStack Network.
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
@@ -5849,6 +5849,8 @@ spec:
                 - name
                 type: object
               ready:
+                default: false
+                description: Ready is true when the cluster infrastructure is ready.
                 type: boolean
               router:
                 description: Router describes the default cluster router


api/v1alpha8/openstackcluster_types.go
```diff
@@ -167,6 +167,8 @@ type OpenStackClusterSpec struct {
 
 // OpenStackClusterStatus defines the observed state of OpenStackCluster.
 type OpenStackClusterStatus struct {
+	// Ready is true when the cluster infrastructure is ready.
+	// +optional
```
EmilienM, line 171: well this is in fact wrong:

https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure#data-types

I'll investigate

EmilienM (commented):
/hold

api/v1alpha8/openstackcluster_types.go
EmilienM: * GCP is setting it as required: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/api/v1beta1/gcpcluster_types.go#L75
* AWS: default to false: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta2/awscluster_types.go#L252
* Azure: optional: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/azurecluster_types.go#L66C1-L66C14

The GCP is doing correctly but I suspect it works because they never patch an empty cluster object while in our case we patch empty cluster when the bastion instance is still being built.

EmilienM: @mdbooth when time permits in a reasonable future we need to get back to this.

lentzi90: This may be a stupid question, but does it help if we set `omitempty`? My understanding would be that no value would then be interpreted as `Ready: false`

huxcrux: I also expected no value to equals false, however this does not seem to be the case. I stumbeled upon this issue and it seems to break the chagnes implemented by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1829

To be honest I think the AWS "solution" looks most appealing. However does not really have an optionion if we could get this resolved.

Currently this however causing Floating IPS to leak and casuing the problem mentioned in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1687

EmilienM:
This one is a good candidate for my list of things I need to run through a debugger and see what's wrong. I'll take a look soon.

EmilienM:
/hold
we agreed we'll go with the AWS way (default to false)

lentzi90 (commented):
Based on the logs this fixes the issue and seems reasonable to me.
/approve

PR #1847, opened 2024-01-30
✨ Add support to set `allocation_pools` for subnet

dulek:
This commit adds API that allows users to set `allocations_pools` in the subnet created by CAPO. This allows the users to restrict the IP address ranges that will be allocated automatically by OpenStack when creating Machines.

Users can utilize this to reserve addresses for VIPs (virtual IPs) or special nodes that will have predefined addresses and will be created later.

Fixes #

**TODOs**:

- [X] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold


EmilienM:
IMO life would be beautiful if we converted `NodeCIDR` into `NodeCIDRs` which is a list of subnet filters or something.

mdbooth:
Another advantage of defining a SubnetSpec struct is that it would be extensible in v1beta1.

jichenjc:
>>>This doesn't feel right. allocationPools feels like a property of a specific created subnet. This is going to break when we (hopefully) do IPI dual-stack.

this is useful in my openstack implementation especially 2 networks are in same CIDR with different allocation pool
I think it's not that helpful if we go IPI or everything under CAPO control, but it might be helpful for some customer
is using their existing env for CAPO (those resource under admin's input)

MaysaMacedo:
> This doesn't feel right. `allocationPools` feels like a property of a specific created subnet. This is going to break when we (hopefully) do IPI dual-stack.
> 
> I was already wondering if `NodeCIDRs` should contain some kind of SubnetSpec rather than just a CIDR. It was already mentioned that this would allow specifying what kind of IPv6 address allocation to use. I would be much happier if `allocationPools` was a property of a SubnetSpec.
> 
> Thoughts? cc @MaysaMacedo

Just sharing what was discussed on the slack thread, having multiple SubnetSpec make sense to me, it would be good to name the field to something more representative of its content.

EmilienM:
please rebase on main

mdbooth:
@dulek I rebased this for you, btw. No conflicts.

dulek:
Thanks!

pkg/cloud/services/networking/network_test.go
```diff
@@ -423,6 +424,260 @@ func Test_ReconcileExternalNetwork(t *testing.T) {
...
+	fakeSubnetID := "d08803fc-2fa5-4179-b9d7-8c43d0af2fe6"
+	fakeCIDR := "10.0.0.0/24"
+	fakeNetworkID := "d08803fc-2fa5-4279-b9f7-8c45d0ff2fe6"
+	fakeDNS := "10.0.10.200"
+
+	tests := []struct {
+		name             string
+		openStackCluster *infrav1.OpenStackCluster
+		expect           func(m *mock.MockNetworkClientMockRecorder)
+		want             *infrav1.OpenStackCluster
```
mdbooth, line 443: The test doesn't check `want`, and if you add a check it fails.

dulek: Fixed!

EmilienM:
@dulek is on vacation, I'll take care of rebase.

EmilienM:
/cc MaysaMacedo
for double check

PR #1852, opened 2024-02-01
(DNM) reproduce for read-after-write issue

EmilienM:
/hold


controllers/openstackcluster_controller.go
```diff
@@ -302,6 +306,11 @@ func reconcileNormal(scope scope.Scope, cluster *clusterv1.Cluster, openStackClu
 		return reconcile.Result{}, err
 	}
 
+	if openStackCluster.Status.Foo == "" {
+		scope.Logger().Info("openStackCluster.Status.Foo is empty")
+		openStackCluster.Status.Foo = "bar"
```
EmilienM, line 311: here I'm writing Foo to "bar" and at some point during the bastion creation, we re-enqueue the reconcilier and I've seen in another PR where the status was empty in a new reconcilier even if it was just written here.

PR #1867, opened 2024-02-07
:sparkles: add TLS configuration flags

tuminoid:

Add `--tls-min-version`, `--tls-max-version` and `--tls-cipher-suites` configuration flags. Same flags can be found in k8s, CAPI, CAPM3 etc.

This enables configuring TLS 1.3 in high security environments, or limiting TLS to 1.2 for debugging purposes as 1.3 is considerably harder to debug.

No.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [x] adds unit tests

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -0,0 +1,111 @@
+/*
```
EmilienM, line 1: thanks for the tests!

EmilienM:
really cool work, thanks!
I was wondering if we could have some doc (maybe in a separate PR) on how to setup CAPO with TLS. Unless it's already covered in some other repositories?

/lgtm

lentzi90 (commented):
> Seems there is a gap in documenting the flags in general.

Good observation. This is something we should work on, but let's do it separately. I think the only docs we have currently is [this tiny little thing](https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration#log-level) about the log level flag... and that is mixed in with how to configure CAPO clusters :see_no_evil: CAPI is not much better either.

/approve

Issue #1869, opened 2024-02-07
OpenstackFloatingIPPool: Can return the same IP for multiple claims


bilbobrovall:
/kind bug

**What steps did you take and what happened:**
Pool can give the same IP multiple times when handling multiple claims in the same reconciliation

**What did you expect to happen:**
Every ip to be unique

PR #1870, opened 2024-02-07
OpenstackFloatingIPPool: Adds popped ip to claimedIPs #1869 

bilbobrovall:
Adds ip to claimedIPs after it's popped from availableIPs to make sure the same IP is not added from tagged IPs when processing multiple claims in the same reconcile. 

Another would be to go back to handling one claim per reconciliation and requeuing to so claimedIPs is always populated from IPAddresses before getIP is called.

Fixes: #1869 

**TODOs**:
- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackfloatingippool_controller.go
+++ b/controllers/openstackfloatingippool_controller.go
@@ -296,8 +296,8 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileIPAddresses(ctx context.Con
 			return err
 		}
 	}
-	unclaimedPreAllocatedIPs := diff(pool.Spec.PreAllocatedFloatingIPs, pool.Status.ClaimedIPs)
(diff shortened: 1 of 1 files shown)


dulek:
Seems to make sense to me, if we work sequentially it should help.

Any consideration if we have multiple reconciliation threads? Do we just assume we don't?

bilbobrovall:
> Seems to make sense to me, if we work sequentially it should help.
> 
> Any consideration if we have multiple reconciliation threads? Do we just assume we don't?

It was discussed a bit in a previous PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1763#issuecomment-1849976633 and it was concluded that there was nothing to worry about since we have a single reconciler for the pool

lentzi90 (commented):
/approve

PR #1877, opened 2024-02-14
✨ Bump Go to 1.22.0

EmilienM:
**What this PR does / why we need it**:

Bump Golang to 1.22.0 freshly released.

PR #1879, opened 2024-02-15
Revert "✨ Bump Go to 1.22.0"

EmilienM:
Reverts kubernetes-sigs/cluster-api-provider-openstack#1877

We need to wait for CAPI & controller-runtime being bumped first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM golang:1.22.0 as builder
+FROM golang:1.21.6 as builder
 WORKDIR /workspace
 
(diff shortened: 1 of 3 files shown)


mdbooth:
Is this just because of the tilt configuration?

EmilienM:
> Is this just because of the tilt configuration?

I have issues with Tilt yes. Also I think it's a bit risky to upgrade Go if our parent projects (CAPI) hasn't, but I'm maybe wrong on good practices here.

EmilienM:
If you think it's fine to use latest go here, I'm happy to hack things on my side to get Tilt working. I'm just worried we want a bit fast for this one, and I thought we would follow CAPI.

PR #1884, opened 2024-02-15
🐛 Don't ignore 'identityRef.spec'

### LINKED ISSUE #1882: `identityRef.kind` is ignored
/kind bug

Identified while working on fix for #1881. The `identityRef` field has a `kind` child field to allow you to specify the kind of object we inspect. Alas, it is completely ignored. We should either remove this field, make it an enum (`["Secret"]`), or otherwise *not* ignore it.

The field should do something.

None.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): `master`
- Cluster-API version: n/a
- OpenStack version: n/a
- Minikube/KIND version: n/a
- Kubernetes version (use `kubectl version`): n/a
- OS (e.g. from `/etc/os-release`): n/a

stephenfin:

We now check the value of `identityRef.spec` for machine template specs and cluster specs. This will allow us to expand the sources of identity ref in the future.

Fixes #1882

Do we want to do this in the API also? Also, there's no explicit test coverage for this. Happy to add it but only if necessary.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/scope/provider.go
+++ b/pkg/scope/provider.go
@@ -59,9 +59,14 @@ func (f *providerScopeFactory) NewClientScopeFromMachine(ctx context.Context, ct
 
 	if openStackMachine.Spec.IdentityRef != nil {
 		var err error
-		cloud, caCert, err = getCloudFromSecret(ctx, ctrlClient, openStackMachine.Namespace, openStackMachine.Spec.IdentityRef.Name, openStackMachine.Spec.CloudName)
-		if err != nil {
-			return nil, err
+		switch openStackMachine.Spec.IdentityRef.Kind {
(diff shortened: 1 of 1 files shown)


mdbooth:
See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1882#issuecomment-1948417448. Following the API guidelines I think we should just remove the kind field.

EmilienM:
We might do https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1903 instead.

PR #1888, opened 2024-02-16
⚠️ Stop serving v1alpha5

mdbooth:
With this change the apiserver will no longer recognise api version v1alpha5. Configurations using v1alpha5 will not work.

As a workaround, the CRD can be edited to temporarily re-enable v1alpha5 so that configurations can be upgraded to a more recent version, but this workaround will also be removed in the next release.


lentzi90 (commented):
/lgtm

mdbooth:
Tests look shiny.

/hold cancel

PR #1895, opened 2024-02-20
🌱 Enforce restricted pod security standards

lentzi90:
We should be able to run with the restricted policy and this will inprove our security.

This enforces the restricted pod security standards profile. It ensures that we do not run containers as root or privileged for example. In a way it can be seen as a validation that we actually configure CAPO according to security best practices.

Fixes #

2. I was a bit unsure what icon to use for this actually but went with :seedling: eventually. It could be seen as breaking for user that modify the CAPO deployment or runs other workload in the namespace, but I would say that is outside of normal usage.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/config/default/namespace.yaml
+++ b/config/default/namespace.yaml
@@ -2,3 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: system
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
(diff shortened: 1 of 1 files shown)


EmilienM:
/lgtm
Thanks!

PR #1899, opened 2024-02-21
🐛 api/additionalPorts: don't create UDP rules

### LINKED ISSUE #1704: Our control plane security groups are too open

EmilienM:

When providing additional ports to open on the control plane security
group, we'll only open TCP from now as it was initially documented.

Opening UDP by default might not be desired for security reasons.

Fixes #1704

/hold


jichenjc:
maybe we should consider add a release notes about this as we changed previous behavior

PR #1900, opened 2024-02-21
🌱 Deprecate v1alpha6

### LINKED ISSUE #1886: Deprecate v1alpha6
/kind feature

We have previously agreed that v1alpha6 would be marked deprecated in v0.10, but should continue to be served for at least 6 months after the release of v0.10 to give users plenty of time to update.

EmilienM:

Mark v1alpha6 version deprecated in our Kubernetes CRDs.

Fixes #1886

### DIFF
--- a/api/v1alpha6/openstackcluster_types.go
+++ b/api/v1alpha6/openstackcluster_types.go
@@ -213,6 +213,7 @@ type OpenStackClusterStatus struct {
 }
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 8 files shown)


lentzi90:
> I have no experience with how deprications have worked prior in CAPO however would it be beneficial to add information regarding deprecation and maybe a removal timeframe to docs or maybe even README?
> 
> Away from the above questions I think this pr looks good :)

Yes we should define a policy for how long we support older API versions and releases. I think we could basically copy [CAPIs policy](https://github.com/kubernetes-sigs/cluster-api/blob/main/CONTRIBUTING.md#support-and-guarantees).

lentzi90 (commented):
v1alpha7 has been around since v0.8 so I think this is reasonable
/approve

PR #1904, opened 2024-02-22
🌱 Reduce cyclomatic complexity of ReconcileLoadBalancer

mdbooth:
This function had become genuinely too complex over time, to the point that even the linter was starting to complain about it when making almost any change.

This change refactors ReconcileLoadBalancer into several smaller logical functions which are much easier to read and reason about. It also revealed some trivial optimisations:
* Only fetch Octavia providers if we need them to create a new loadbalancer
* Only calculate allowed CIDRs once
* Don't re-fetch a loadbalancer to check it's active if it's already active

Apart from just the code cleanup, this has the benefit of not requiring these changes to be mixed into other PRs when the complexity exceeds the linter's threshold.

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -76,42 +76,7 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 		openStackCluster.Status.APIServerLoadBalancer = lbStatus
 	}
 
-	var fixedIPAddress string
-	var err error
-
-	switch {
-	case lbStatus.InternalIP != "":
-		fixedIPAddress = lbStatus.InternalIP
-	case openStackCluster.Spec.APIServerFixedIP != "":
-		fixedIPAddress = openStackCluster.Spec.APIServerFixedIP
-	case openStackCluster.Spec.DisableAPIServerFloatingIP && openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-		fixedIPAddress, err = lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
-		if err != nil {
-			return false, fmt.Errorf("lookup host: %w", err)
-		}
-	}
-
-	providers, err := s.loadbalancerClient.ListLoadBalancerProviders()
-	if err != nil {
-		return false, err
-	}
-
-	// Choose the selected provider if it is set in cluster spec, if not, omit the field and Octavia will use the default provider.
-	lbProvider := ""
-	if openStackCluster.Spec.APIServerLoadBalancer.Provider != "" {
-		for _, v := range providers {
-			if v.Name == openStackCluster.Spec.APIServerLoadBalancer.Provider {
-				lbProvider = v.Name
-				break
-			}
-		}
-		if lbProvider == "" {
-			record.Warnf(openStackCluster, "OctaviaProviderNotFound", "Provider specified for Octavia not found.")
-			record.Eventf(openStackCluster, "OctaviaProviderNotFound", "Provider %s specified for Octavia not found, using the default provider.", openStackCluster.Spec.APIServerLoadBalancer.Provider)
-		}
-	}
-
-	lb, err := s.getOrCreateLoadBalancer(openStackCluster, loadBalancerName, openStackCluster.Status.Network.Subnets[0].ID, clusterName, fixedIPAddress, lbProvider)
+	lb, err := s.getOrCreateAPILoadBalancer(openStackCluster, clusterName)
 	if err != nil {
 		return false, err
 	}
@@ -121,23 +86,20 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	lbStatus.InternalIP = lb.VipAddress
 	lbStatus.Tags = lb.Tags
 
-	if err := s.waitForLoadBalancerActive(lb.ID); err != nil {
-		return false, fmt.Errorf("load balancer %q with id %s is not active after timeout: %v", loadBalancerName, lb.ID, err)
+	if lb.ProvisioningStatus != loadBalancerProvisioningStatusActive {
+		var err error
+		lb, err = s.waitForLoadBalancerActive(lb.ID)
+		if err != nil {
+			return false, fmt.Errorf("load balancer %q with id %s is not active after timeout: %v", loadBalancerName, lb.ID, err)
+		}
 	}
 
 	if !openStackCluster.Spec.DisableAPIServerFloatingIP {
-		var floatingIPAddress string
-		switch {
-		case lbStatus.IP != "":
-			floatingIPAddress = lbStatus.IP
-		case openStackCluster.Spec.APIServerFloatingIP != "":
-			floatingIPAddress = openStackCluster.Spec.APIServerFloatingIP
-		case openStackCluster.Spec.ControlPlaneEndpoint.IsValid():
-			floatingIPAddress, err = lookupHost(openStackCluster.Spec.ControlPlaneEndpoint.Host)
-			if err != nil {
-				return false, fmt.Errorf("lookup host: %w", err)
-			}
+		floatingIPAddress, err := getAPIServerFloatingIP(openStackCluster)
+		if err != nil {
+			return false, err
 		}
+
 		fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterName, floatingIPAddress)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,186 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
+		}
+	}
 
-		listener, err := s.getOrCreateListener(openStackCluster, lbPortObjectsName, lb.ID, port)
+	return false, nil
+}
+
+// getAPIServerVIPAddress gets the VIP address for the API server from wherever it is specified.
+// Returns an empty string if the VIP address is not specified and it should be allocated automatically.
+func getAPIServerVIPAddress(openStackCluster *infrav1.OpenStackCluster) (string, error) {
```
EmilienM, line 143: we could unit test it but I'm not going to nitpick on that. Just make sure the existing tests cover the output of that function and I'm fine.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,186 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
+		return fixedIPAddress, nil
+	}
 
-		pool, err := s.getOrCreatePool(openStackCluster, lbPortObjectsName, listener.ID, lb.ID, lb.Provider)
+	return "", nil
+}
+
+// getAPIServerFloatingIP gets the floating IP from wherever it is specified.
+// Returns an empty string if the floating IP is not specified and it should be allocated automatically.
+func getAPIServerFloatingIP(openStackCluster *infrav1.OpenStackCluster) (string, error) {
```
EmilienM, line 167: ditto

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,186 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
 
-		if err := s.getOrCreateMonitor(openStackCluster, lbPortObjectsName, pool.ID, lb.ID); err != nil {
-			return false, err
+	return "", nil
+}
+
+// getCanonicalAllowedCIDRs gets a filtered list of CIDRs which should be allowed to access the API server loadbalancer.
+// Invalid CIDRs are filtered from the list and emil a warning event.
+// It returns a canonical representation that can be directly compared with other canonicalized lists.
+func getCanonicalAllowedCIDRs(openStackCluster *infrav1.OpenStackCluster) []string {
```
EmilienM, line 192: ditto

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -230,7 +307,41 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
 	return lb, nil
 }
 
-func (s *Service) getOrCreateListener(openStackCluster *infrav1.OpenStackCluster, listenerName, lbID string, port int) (*listeners.Listener, error) {
+// reconcileAPILoadBalancerListener ensures that the listener on the given port exists and is configured correctly.
+func (s *Service) reconcileAPILoadBalancerListener(lb *loadbalancers.LoadBalancer, openStackCluster *infrav1.OpenStackCluster, clusterName string, port int) error {
```
EmilienM, line 311: do we have unit test coverage for this one?

EmilienM (commented):
Awesome change, I just have questions on unit tests coverage.

mdbooth:
I 100% agree about unit test coverage. However, this is basically just code motion so I can separate these changes from the other changes I'm making to the same code without triggering the gocyclo linter. IOW this is just yak shaving for me right now.

If you fancied taking the yak shears and writing some good unit tests while I get back to the API stuff I'd add to the beers I owe you.

pkg/cloud/services/loadbalancer/loadbalancer.go
EmilienM: I wrote it.

pkg/cloud/services/loadbalancer/loadbalancer.go
EmilienM: I wrote it.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,186 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
+	}
+	// The current version is always the last one in the list.
+	octaviaVersion := octaviaVersions[len(octaviaVersions)-1].ID
+
+	return openstackutil.IsOctaviaFeatureSupported(octaviaVersion, openstackutil.OctaviaFeatureVIPACL, lb.Provider), nil
 }
 
-func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackCluster, loadBalancerName, subnetID, clusterName, vipAddress, provider string) (*loadbalancers.LoadBalancer, error) {
+// getOrCreateAPILoadBalancer returns an existing API loadbalancer if it already exists, or creates a new one if it does not.
+func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterName string) (*loadbalancers.LoadBalancer, error) {
```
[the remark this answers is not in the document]
EmilienM: I wrote it.

pkg/cloud/services/loadbalancer/loadbalancer.go
EmilienM: no we don't but it's not an easy one so I've filled https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1906 so we can get back to it at some point.

EmilienM:
/lgtm
I'll let another reviewer take a look.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,190 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
+
+			if openStackCluster.Status.Router != nil && len(openStackCluster.Status.Router.IPs) > 0 {
+				allowedCIDRs = append(allowedCIDRs, openStackCluster.Status.Router.IPs...)
 			}
 		}
 	}
-	lbStatus.AllowedCIDRs = allowedCIDRs
 
-	return false, nil
+	// Filter invalid CIDRs and convert any IPs into CIDRs.
```
MaysaMacedo, line 223: couldn't we keep using `validateIPs` function?

MaysaMacedo:
/hold

@mdbooth holding just to give a chance for you to check my comments, feel free to unhold once you checked them

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -121,23 +86,20 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 	lbStatus.InternalIP = lb.VipAddress
 	lbStatus.Tags = lb.Tags
 
-	if err := s.waitForLoadBalancerActive(lb.ID); err != nil {
-		return false, fmt.Errorf("load balancer %q with id %s is not active after timeout: %v", loadBalancerName, lb.ID, err)
+	if lb.ProvisioningStatus != loadBalancerProvisioningStatusActive {
+		var err error
+		lb, err = s.waitForLoadBalancerActive(lb.ID)
```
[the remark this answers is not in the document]
mdbooth: Because it's a different object, returned from the most recent query. It's probably not necessary, but I figured that as we already had the latest state we might as well use it.

e.g. If we don't do this and for whatever reason we happened to look at ProvisioningStatus again, we would see that it's still not active, because we're still referencing the original version of the object before it became active.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -230,7 +311,45 @@ func (s *Service) getOrCreateLoadBalancer(openStackCluster *infrav1.OpenStackClu
...
+// reconcileAPILoadBalancerListener ensures that the listener on the given port exists and is configured correctly.
+func (s *Service) reconcileAPILoadBalancerListener(lb *loadbalancers.LoadBalancer, openStackCluster *infrav1.OpenStackCluster, clusterName string, port int) error {
+	loadBalancerName := getLoadBalancerName(clusterName)
+	lbPortObjectsName := fmt.Sprintf("%s-%d", loadBalancerName, port)
+
+	if openStackCluster.Status.APIServerLoadBalancer == nil {
+		return fmt.Errorf("APIServerLoadBalancer is not yet available in OpenStackCluster.Status")
+	}
+
+	allowedCIDRs := openStackCluster.Status.APIServerLoadBalancer.AllowedCIDRs
```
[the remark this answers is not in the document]
mdbooth: Yes. It was set directly in ReconcileLoadBalancer before reconciling any listeners.

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: It's a toss-up, but I thought it was too small to be a separate function with only one caller. This function is small enough now, I think it's clearer to inline it here.

Also it doesn't just validate, it also filters, which you wouldn't know just from its name so you'd have to go read the function intentionally to notice that.

mdbooth (commented):
Thanks, @MaysaMacedo. I fixed the typo and responded to the other comments. Let me know what you think.

EmilienM:
Restoring Maysa's LGTM, fixing unit tests and removing /hold.
/lgtm
/hold cancel

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -153,71 +115,190 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
...
-					return false, err
+		if openStackCluster.Status.Network != nil {
+			for _, subnet := range openStackCluster.Status.Network.Subnets {
+				if subnet.CIDR != "" {
+					allowedCIDRs = append(allowedCIDRs, subnet.CIDR)
 				}
-				allowedCIDRs = listener.AllowedCIDRs
+			}
+
+			if openStackCluster.Status.Router != nil && len(openStackCluster.Status.Router.IPs) > 0 {
```
dalees, line 217: This fixes a bug we just hit when specifying both `internalNetwork.networkFilter` and `apiServerLoadBalancer.allowedCIDRs`. Thanks for adding this nil guard!

Issue #1840, opened 2024-01-24
controller/openstackcluster: reconcileID is re-used 1000+ times


EmilienM:
/kind bug

**What steps did you take and what happened:**

Open https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1789/pull-cluster-api-provider-openstack-e2e-test/1736818285917245440/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-z6k2q/manager.log

And `grep` for:
```
reconcileID="12187d68-f3cf-4596-94e8-83666306d5fc"
```
2191 uses of the same reconcile ID.

That's wrong and it wasn't the case in late October:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1696/pull-cluster-api-provider-openstack-e2e-test/1717592179636965376/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-kfgvg/manager.log

I strongly suspect this [PR](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1688) to be the cause, I've bisected CI jobs and this issue started after this PR.

**What did you expect to happen:**

The reconcile loops should be uniquely used and IDs shouldn't be the same. I found this issue when debugging https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1788.

EmilienM:
@kubernetes-sigs/cluster-api-provider-openstack-maintainers please add v1beta1 to milestone.

mdbooth:
I'm confident this is not a v1beta1 issue. I'm going to remove the tag.

EmilienM:
> It looks like when we create a cached provider scope we put the logger in it. The problem is that this logger is reused across multiple reconciles and has different context annotations added for each. We need to find some way to separate the logger from the scope. Perhaps we should pass it around in the context instead.
> 
> The impact of this is that the logs are incorrect. It shouldn't actually be breaking any functionality, but we should fix it asap.

I don't know if that's what you had in mind, but I started something here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1861

PR #1905, opened 2024-02-26
🌱 Reduce cyclomatic complexity of reconcileNetworkComponents

mdbooth:
This change refactors reconcileNetworkComponents into several smaller logical functions which are easier to read and reason about.

It also makes the gocyclo linter happy when making new changes to this code.

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -602,58 +602,12 @@ func reconcileNetworkComponents(scope *scope.WithLogger, cluster *clusterv1.Clus
 	}
 
 	if len(openStackCluster.Spec.ManagedSubnets) == 0 {
-		scope.Logger().V(4).Info("No need to reconcile network, searching network and subnet instead")
-
-		if openStackCluster.Status.Network == nil {
-			openStackCluster.Status.Network = &infrav1.NetworkStatusWithSubnets{}
-		}
-
-		err := getCAPONetwork(openStackCluster, networkingService)
-		if err != nil {
-			return err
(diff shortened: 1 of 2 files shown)


EmilienM:
@mdbooth I've added little tests to it, but I think we need to:
* Split `controllers/openstackcluster_controller.go` into multiple files handling resources in different files and one file for the higher level loops. We shouldn't do that in this PR and I volunteer to do it when I'm back. I've tracked this in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1907.
* Write more tests regarding the network & subnet resources, this is out of scope for this PR so I'm inclined to postpone it (I created https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1908) - I wonder if @MaysaMacedo can help when time permits.

mdbooth:
Lazily hoping that was a flake without checking:

lentzi90:
Testing if our jobs now work in the community prow cluster

mdbooth:
Looks like something related to floating IPs is borked because we run out. Perhaps they're not being deleted, or created in a loop. It's hard to tell, though, because of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1840

mdbooth:
I think I found the problem from code inspection, although this PR also currently includes #1913 for improved logging just in case. The issue was that we were re-executing floating IP creation on every cluster reconcile when not using an Octavia loadbalancer.

mdbooth:
Yay! Looks like that fixed it.

PR #1910, opened 2024-02-27
Ports follow-up

EmilienM:

A follow-up from @mdbooth & @lentzi90's comments on https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1788.

- Remove useless logging for port loop
- API changes for Dependent & Referenced resources


EmilienM:
```
read: connection reset by peer
```

EmilienM:
This doesn't work yet.

/hold

controllers/openstackcluster_controller.go
```diff
@@ -126,20 +126,22 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if openStackCluster.Status.Bastion == nil {
 			openStackCluster.Status.Bastion = &infrav1.BastionStatus{}
 		}
-		changed, err := compute.ResolveReferencedMachineResources(scope, openStackCluster, &openStackCluster.Spec.Bastion.Instance, &openStackCluster.Status.Bastion.ReferencedResources)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if changed {
+		if openStackCluster.Status.Bastion.ReferencedResources == nil {
```
EmilienM, line 129: This is problematic because we adopt `ReferencedResources` once `OpenStackCluster.Status.Network != nil`.

I see multiple solutions:
* Keep `changed` but still convert the fields to pointers to follow API guidelines. This is the simplest option and I think easier to maintain. I'm fairly sure it's going to be working and the LOC change will be tiny.
* Add some logic here that adoption needs to be skipped if Status.Network is nil and in the next loops eventually Network will be created and Adoption will happen *once*. We wouldn't tolerate empty fields in this struct as this is required by the other resources. This is also a valid option, as we'll make adoption run **once**; but with the caveat that we have some weird logic (potentially fragile) where we require something before adoption can happen.

EmilienM: @mdbooth if you can think of something else, I'm all ears :ear: :question: :thinking:

controllers/openstackcluster_controller.go
```diff
@@ -126,20 +126,22 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			openStackCluster.Status.Bastion.ReferencedResources = &infrav1.ReferencedMachineResources{}
+			err := compute.ResolveReferencedMachineResources(scope, openStackCluster, &openStackCluster.Spec.Bastion.Instance, openStackCluster.Status.Bastion.ReferencedResources)
```
mdbooth, line 131: We must not do this because it will write an empty referenced machine resources to the cluster status if resolve fails. This needs to be:

```golang
  referencedResources, err := compute.ResolveReferencedMachineResources(...)
  if err != nil {
    ...
  }
  openStackCluster.Status.Bastion.ReferencedResources = referencedResources
```

controllers/openstackcluster_controller.go
mdbooth: Why don't we just skip this code block if network isn't set? If network isn't set then no bastion has ever been created, so there is also nothing to cleanup.

controllers/openstackcluster_controller.go
```diff
@@ -126,20 +126,22 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 			// If the referenced resources have changed, we need to update the OpenStackCluster status now.
 			return reconcile.Result{}, nil
 		}
 
-		changed, err = compute.ResolveDependentBastionResources(scope, openStackCluster, bastionName(cluster.Name))
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if changed {
+		if openStackCluster.Status.Bastion.DependentResources == nil {
```
mdbooth, line 139: DependentResources is different to ReferencedResources because it's not immutable: we can successively create or adopt dependent resources over multiple reconciles. We must always do adoption here.

EmilienM: but how would you know it has changed and you must leave the reconcile now to write the object?
If you remove the if, we're in infinite loop where we'll always adopt...

EmilienM:
@mdbooth I guess I should close this one right?

mdbooth:
The goals of this PR have been various addressed by other v1beta1 PRs.

PR #1911, opened 2024-02-27
🌱 Enhancements to how controller-gen is invoked

mdbooth:
This PR includes 2 commits which each make a single change to how the Makefile does code generation.

The first adds the following 2 separate targets:
* generate-controller-gen
* generate-conversion-gen

These invoke `controller-gen` and `conversion-gen` respectively. These were previously invoked by the `generate-go` target, but that target is now reduced to simply executing `go generate ./...`.

The `generate` target continues to invoke all code generation.

The advantage of this is the ability to easily run only the generation you need during development.

The second commit changes the behaviour of `generate-conversion-gen`. This produces no change in output against the current codebase, but fixes an issue generating an incorrect package internal package name should conversion-gen want to add a dependency between packages. I encountered this when I made a change which caused it to do this, which has not yet landed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,14 +258,14 @@ generate-controller-gen: $(CONTROLLER_GEN)
 		object:headerFile=./hack/boilerplate/boilerplate.generatego.txt
 
 .PHONY: generate-conversion-gen
+capo_module := sigs.k8s.io/cluster-api-provider-openstack
 generate-conversion-gen: $(CONVERSION_GEN)
 	$(CONVERSION_GEN) \
-		--input-dirs=./api/v1alpha1 \
-		--input-dirs=./api/v1alpha5 \
-		--input-dirs=./api/v1alpha6 \
-		--input-dirs=./api/v1alpha7 \
-		--input-dirs=./api/v1beta1 \
+		--input-dirs=$(capo_module)/api/v1alpha5 \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -245,25 +245,28 @@ modules: ## Runs go mod to ensure proper vendoring.
...
+generate-conversion-gen: $(CONVERSION_GEN)
 	$(CONVERSION_GEN) \
-		--input-dirs=./api/v1alpha1 \
-		--input-dirs=./api/v1alpha5 \
-		--input-dirs=./api/v1alpha6 \
-		--input-dirs=./api/v1alpha7 \
-		--input-dirs=./api/v1beta1 \
+		--input-dirs=$(capo_module)/api/v1alpha5 \
+		--input-dirs=$(capo_module)/api/v1alpha6 \
+		--input-dirs=$(capo_module)/api/v1alpha7 \
```
EmilienM, line 266: I'm not sure why `v1beta1` is missing but I won't block you on that, maybe there is a reason behind.

EmilienM (approved):
/lgtm
/hold
for a question that I have, but feel free to hold cancel if it's all good.

Makefile
mdbooth: Yep, it's deliberately excluded because we don't need to generate conversions for v1beta1 (or v1alpha1).

PR #1912, opened 2024-02-28
:sparkles: Add OpenStackServerGroup CRD and Controller

### LINKED ISSUE #1256: Use a server group to ensure anti-affinity for control plane nodes

dalees:

Implements new CRD for OpenstackServerGroup in v1beta1 to allow managed Server Groups with standard policies, and adds ServerGroupRef to OpenstackMachine that references the new CRD and uses it for VM creation.

Fixes #1256 

This implements comment https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1256#issuecomment-1829999292

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests
- [x] Rebased onto v1beta1 commit (removes v1alpha8)


api/v1alpha8/openstackservergroup_types.go
```diff
@@ -0,0 +1,76 @@
...
+	// ServerGroupFinalizer allows ReconcileOpenStackServerGroup to clean up OpenStack resources associated with OpenStackServerGroup before
+	// removing it from the apiserver.
+	ServerGroupFinalizer = "openstackservergroup.infrastructure.cluster.x-k8s.io"
+)
+
+// OpenStackServerGroupSpec defines the desired state of OpenStackServerGroup
+type OpenStackServerGroupSpec struct {
+	// The name of the cloud to use from the clouds secret
+	// +optional
+	CloudName string `json:"cloudName"`
```
dulek, line 33: This seems a bit weird, we should probably have a reference to an OpenStackCluster instead?

api/v1alpha8/types.go
```diff
@@ -240,6 +240,12 @@ type ServerGroupFilter struct {
 	Name string `json:"name,omitempty"`
 }
 
+type ServerGroupRef struct {
+	// Name of the OpenStackServerGroup resource to be used.
+	// Must be in the same namespace as the resource(s) being provisioned.
+	Name string `json:"name"`
+}
```
dulek, line 247: I think LocalObjectReference [1] should be used as a type [2].

controllers/openstackmachine_controller.go
```diff
@@ -147,12 +147,18 @@ func (r *OpenStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 
-	// Resolve and store referenced resources
+	// Resolve and store referenced OpenStack resources
 	err = compute.ResolveReferencedMachineResources(scope, &openStackMachine.Spec, &openStackMachine.Status.ReferencedResources)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
+	// Resolve referenced resources CAPO resources, using the K8s client
+	err = resolveReferencedClientResources(ctx, r.Client, openStackMachine)
```
dulek, line 157: It feels like it's still a Machine resource. Couldn't we put that into `ResolveReferencedMachineResources` directly? Even if we need to change the arguments of the function.

controllers/openstackservergroup_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	CaCertificates   []byte // PEM encoded ca certificates.
+}
+
+//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackservergroups,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackservergroups/status,verbs=get;update;patch
+
+func (r *OpenStackServerGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, reterr error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Fetch the OpenStackMachine instance.
```
dulek, line 58: This log seems wrong.

controllers/openstackservergroup_controller.go
```diff
@@ -0,0 +1,214 @@
...
+
+	serverGroupName := openStackServerGroup.Name
+
+	computeService, err := compute.NewService(scope)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Get the servergroup by name, even if our K8s resource already has the ID field set.
+	// TODO(dalees): If this returns a 404 do we try to delete with existing UUID? Do we just assume success?
```
dulek, line 143: I think we should look up by ID and only then fallback to looking up by name. IDs are safe in case of duplicate names.

controllers/openstackservergroup_controller.go
```diff
@@ -0,0 +1,214 @@
...
+
+	scope.Logger().Info("Reconciled ServerGroup successfully")
+	return ctrl.Result{}, nil
+}
+
+func (r *OpenStackServerGroupReconciler) getOrCreate(logger logr.Logger, openStackServerGroup *infrav1.OpenStackServerGroup, computeService *compute.Service) (*servergroups.ServerGroup, error) {
+
+	serverGroupName := openStackServerGroup.Name
+
+	serverGroup, err := computeService.GetServerGroupByName(serverGroupName, false)
```
dulek, line 196: Again, we should probably lookup by ID first in case we have duplicate names.

dulek (changes requested):
This looks pretty good, some remarks inline.

api/v1alpha8/openstackservergroup_types.go
dalees: Thank you for the feedback! Yeah, this allows the resource to be reconciled alone, as it's self contained.

However that isn't in any of the use cases, it doesn't seem a limitation to be tied to an existing OpenStackCluster even if the OpenStackServerGroup was only used in MachineSets. It would remove duplication of these creds.

I'll make this change, once the CRD approach is agreed.

controllers/openstackservergroup_controller.go
dalees: Ok, happy to change this - I can see it will lead to less problems if a duplicate named resource was created after this managed one.

api/v1alpha8/openstackservergroup_types.go
dulek: Hm, okay, that's a fair point. The use case to keep all the workers from different clusters in a single ServerGroup makes sense, I see your point.

controllers/openstackservergroup_controller.go
dalees: Agreed, I'll fix this in the next iteration.

api/v1alpha8/types.go
dalees: Would it be an issue that it specifies `omitempty`?

dulek: Ugh, it probably would. Okay, the design of this field is good, we can change internals if we want to later.

controllers/openstackmachine_controller.go
dulek: I see that, sure. Let's see what other reviewers will say here, especially @mdbooth as `ResolveReferencedMachineResources()` is an idea of his.

mdbooth:
@pierreprinetti We agreed this in principal this week. Pinging you because it's similar to something ORC would do.

chess-knight:
Hi, at @SovereignCloudStack we are very interested in this feature. What is the progress here @dalees?

dalees:
> Hi, at @SovereignCloudStack we are very interested in this feature. What is the progress here @dalees?

Hello - pleased to hear of the interest! I'm keen to get this in, and I'm scheduled to revisit this in the next few weeks to get it back into a reviewable state.

api/v1alpha8/openstackservergroup_types.go
dalees: In `v1beta1` CloudName has moved into IdentifyRef, but this review thread is still relevant.

For now, I've chosen not to tie the OpenStackServerGroup to an OpenStackCluster reference and instead hold it's own secret reference. This matches how OpenStackMachine's are designed, and feels reasonable for this resource and avoids circular dependency.

I don't think there's much of a use case for two clusters using a single `OpenStackServerGroup`, that's not the design intention (but is possible with the current implementation).

api/v1alpha8/types.go
dalees: Discussion resolved

controllers/openstackmachine_controller.go
dalees: While rebased onto `v1beta` I have removed `resolveReferencedClientResources` and moved the code into the new `resolveMachineResources`.

controllers/openstackservergroup_controller.go
dalees: The delete operation no longer looks up by name, and now only deletes server groups from OpenStack if there is a stored ID.

dalees:
This change is ready for review, when reviewers have the time :) cc @mdbooth

chess-knight:
Hi @dalees, thank you for pushing this PR. I am also kindly asking you and others about checking the OCCM host-id labelling issue https://github.com/kubernetes/cloud-provider-openstack/issues/2579, what do you think about it? These two features can be nicely combined then, E.g. one can create an anti-affinity server group, and then check the host-id label of k8s nodes to ensure that nodes are distributed on different underlying hypervisors.

mnaser:
@mdbooth this seems like a really good candidate for an ORC-style approach too?

controllers/openstackservergroup_controller.go
okozachenko1203: agree
instead of `GetServerGroupByName` raising error in case of multiple server groups with the same name, i suggest to check `openStackServerGroup.Status.ID`, if it is nil, create a new server group

PR #1914, opened 2024-02-28
⚠️ API cleanup of PortOpts

mdbooth:
Internally, all optional fields become pointers. This change has no direct effect on the CRD, but means that unset values and zero values now have different meanings.

SecurityGroupFilters is renamed to SecurityGroups for consistency with other filter fields used throughout the API. Note that this change thoroughly confuses conversion-gen. Consequently we explicitly disable the conversion of these fields in v1alpha6 and v1alpha5 and do the conversion entirely manually.

/hold


mdbooth:
@lentzi90 @EmilienM This is the first API cleanup change I've extracted. The API change itself is fairly simple. Some of the conversion is a bit hairy, which is mostly specific to this PR.

The new optional package will become a feature though. This allows us to customise the conversion behaviour for specific fields whilst otherwise retaining the defaults in apimachinery.

There will be more like this, but mostly simpler.

EmilienM:
I really like it!
/lgtm

pkg/utils/optional/conversion.go
```diff
@@ -0,0 +1,43 @@
...
+limitations under the License.
+*/
+
+package optional
+
+import (
+	"k8s.io/apimachinery/pkg/conversion"
+)
+
+func Convert_string_To_optional_String(in *string, out *String, _ conversion.Scope) error {
```
EmilienM, line 23: Trivial functions but still could be unit tests. We can do that later.

lentzi90 (approved):
/approve

EmilienM:
/lgtm
restoring lgtm, I just rebased on `main` and ran `make generate`.

pkg/utils/optional/conversion.go
mdbooth: I had a similar thought. I did actually include test coverage for them: notice I added a few 'empty object' conversion tests. These tests will fail if this method isn't doing its job correctly.

mdbooth:
> /lgtm restoring lgtm, I just rebased on `main` and ran `make generate`.

I'm pretty sure this will have fixed it. Not sure why the test run against your most recent push is wedged.

PR #1915, opened 2024-02-29
⚠️Move CloudName into IdentityRef and make cluster IdentityRef required

mdbooth:
This change came from attempting to write validation markers for CloudName and IdentityRef in both the machine and cluster specs.

Firstly I noticed that IdentityRef was marked optional in the cluster spec, but it is certainly required: the cluster cannot be provisioned without cloud credentials. I made IdentityRef required in the cluster spec.

In contrast, IdentityRef is genuinely optional in the machine spec because, if not specified, we will use the credentials defined in the cluster spec.

CloudName on the machine spec is also marked optional. However, it is required if IdentityRef was specified. This is because it refers to the same object as IdentityRef. The most sensible way to to represent this in the API is to put it in the IdentityRef. This means that if IdentityRef is provided, it must be provided completely, including CloudName.

/hold


dulek:
This is so much better design! It looks good to me, but since it already has approved label, I'll leave setting lgtm to the next reviewer.

EmilienM:
Nicely done, I suspect the API doc wasn't generated for some reason?

PR #1916, opened 2024-03-01
🐛 ci: fix upgrades by dropping cloudName

EmilienM:

We missed these files when removing `cloudName` from CI configs.


lentzi90 (commented):
Thanks!
/approve

PR #1922, opened 2024-03-05
feat: add configurable loadbalancer network

oblazek:
Previously when loadbalacer was created it used the same network/subnet as the control plane nodes for the VIP. This was not always the right assumption as some users might want to be able to customize this according to their env.

This commit fixes the above by adding two fields into OpenStackClusterSpec/Status two fields `network` and `subnets` under `APIServerLoadBalancer` so that user can define which network/subnet to use for allocation of the loadbalancer.

```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha8
kind: OpenStackCluster
metadata:
  name: {{ capi_cluster_name }}
spec:
  cloudName: {{ capi_cluster_os_cloud }}
  identityRef:
    kind: Secret
    name: test3-cloud-config
  managedSecurityGroups: true
  apiServerLoadBalancer:
    enabled: true
    network:
      name: vip-net-ulab1
    subnets:
      - name: ulab1-ipv4-1
status:
  apiServerLoadBalancer:
    id: 808239dc-064f-402e-bd7d-ca667157a76c
    internalIP: 10.248.11.161
    ip: ""
    loadBalancerNetwork:
      id: 65f6a910-3968-46c0-8f72-e4f171476670
      name: vip-net-ulab1
      subnets:
      - id: 5b4e2c9b-1ab9-4510-87a4-6d2301e16e77
    name: k8s-clusterapi-cluster-test3-test3-kubeapi
 ```

Fixes: #1809

Signed-off-by: Ondrej Blazek <ondrej.blazek@firma.seznam.cz>

### CODE REVIEW

File: api/v1alpha6/conversion_test.go [hunk 1]
```diff
@@ -102,6 +102,15 @@ func TestFuzzyConversion(t *testing.T) {
...
 			},
 
+			func(status *infrav1.OpenStackClusterStatus, c fuzz.Continue) {
+				c.FuzzNoCustom(status)
+
+				// this field does not exist in v1alpha6
+				if status.APIServerLoadBalancer != nil {
+					status.APIServerLoadBalancer.LoadBalancerNetwork = nil
+				}
+			},
```
dulek, line 112: This sounds wrong, we should probably add restore functions instead of blocking such fuzzing.

api/v1alpha6/conversion_test.go
```diff
@@ -113,6 +122,10 @@ func TestFuzzyConversion(t *testing.T) {
 					c.FuzzNoCustom(&subnet)
 					spec.Subnets = append(spec.Subnets, subnet)
 				}
+
+				// None of these fields exist in v1alpha6
+				spec.APIServerLoadBalancer.Network = nil
+				spec.APIServerLoadBalancer.Subnets = nil
```
dulek, line 128: Same here.

api/v1beta1/types.go
```diff
@@ -599,6 +601,12 @@ type APIServerLoadBalancer struct {
 	AllowedCIDRs []string `json:"allowedCidrs,omitempty"`
 	// Octavia Provider Used to create load balancer
 	Provider string `json:"provider,omitempty"`
+	// Network defines which network should the load balancer be allocated on.
+	//+optional
+	Network *NetworkFilter `json:"network,omitempty"`
+	// Subnet defines which network(s) should the load balancer be allocated on.
+	//+optional
+	Subnets []*SubnetFilter `json:"subnet,omitempty"`
```
dulek, line 609: Do we ever consider this field to have more than 2 elements? I don't think so, it's just for dual stack use case, so we need kubebuilder constraints and a better description here. Also it's "which subnet(s)" and later on the field json serialization has to be plural.

api/v1beta1/types.go
```diff
@@ -413,6 +413,8 @@ type LoadBalancer struct {
 	AllowedCIDRs []string `json:"allowedCIDRs,omitempty"`
 	//+optional
 	Tags []string `json:"tags,omitempty"`
+	//+optional
+	LoadBalancerNetwork *NetworkStatusWithSubnets `json:"loadBalancerNetwork,omitempty"`
```
dulek, line 417: Please add a meaningful description of this field purpose.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -264,9 +264,23 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
...
-	subnetID := openStackCluster.Status.Network.Subnets[0].ID
-	s.scope.Logger().Info("Creating load balancer in subnet", "subnetID", subnetID, "name", loadBalancerName)
+	var vipNetworkID, vipSubnetID string
+	if openStackCluster.Status.APIServerLoadBalancer != nil {
+		if openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork != nil {
+			if openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.ID != "" {
+				vipNetworkID = openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.ID
+			}
+			if len(openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.Subnets) > 0 {
+				vipSubnetID = openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.Subnets[0].ID
```
dulek, line 274: Looks like we assume there can only be one subnet. Let's make sure to document that on the CRD, that only first element is taken into account. Because I assume we don't support dual stack LBs yet in any case, right?

controllers/openstackcluster_controller.go
```diff
@@ -585,6 +585,62 @@ func bastionHashHasChanged(computeHash string, clusterAnnotations map[string]str
...
+func reconcileLoadBalancerNetwork(openStackCluster *infrav1.OpenStackCluster, networkingService *networking.Service) error {
+	if openStackCluster.Spec.APIServerLoadBalancer.Network != nil {
+		lbNetOpts := openStackCluster.Spec.APIServerLoadBalancer.Network.ToListOpt()
+		lbNetList, err := networkingService.GetNetworksByFilter(&lbNetOpts)
+		if err != nil {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find loadbalancer network: %w", err))
+			return fmt.Errorf("failed to find network: %w", err)
+		}
+		if len(lbNetList) == 0 {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find ana loadbalancer network"))
```
dulek, line 597: Typo "ana".

controllers/openstackcluster_controller.go
```diff
@@ -585,6 +585,62 @@ func bastionHashHasChanged(computeHash string, clusterAnnotations map[string]str
...
+
+			// if no match was found, add all subnets within openStackCluster.Status.APIServerLoadBalancer.Network
+			if len(openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.Subnets) == 0 {
+				for _, subnetID := range lbNetList[0].Subnets {
+					openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.Subnets = append(
+						openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.Subnets, infrav1.Subnet{
+							ID: subnetID,
+						})
+				}
+			}
```
dulek, line 637: This behavior feels counter-intuitive. If users specified wrong subnet IDs, we'll silently ignore their mistake. We should rather explicitly fail here.

Also what if 2 subnets were specified and we only found 1? IMO that should be an error too.

dulek (changes requested):

api/v1alpha6/conversion_test.go
mdbooth: Yes. There are several examples in v1alpha6 and v1alpha7.

api/v1beta1/types.go
mdbooth: I think the full set is likely:
* MaxLength=2
* optional
* listType=atomic

Please could you also change the type to `[]SubnetFilter`? I don't think we would ever add a nil value to the list, and the list itself already has a natural nil value.

api/v1alpha6/conversion_test.go
oblazek: thanks will take a closer look..

api/v1beta1/types.go
```diff
@@ -599,6 +601,12 @@ type APIServerLoadBalancer struct {
 	AllowedCIDRs []string `json:"allowedCidrs,omitempty"`
 	// Octavia Provider Used to create load balancer
 	Provider string `json:"provider,omitempty"`
+	// Network defines which network should the load balancer be allocated on.
+	//+optional
+	Network *NetworkFilter `json:"network,omitempty"`
+	// Subnet defines which network(s) should the load balancer be allocated on.
```
mdbooth, line 607: I'm guessing that all these subnets must be in the same network, which is described by `Network`?

Is it permitted to specify Subnets without network?

mdbooth (commented):
I think this is good in principal, but there are a few things to straighten out.

api/v1beta1/types.go
oblazek: Yeah I can totally do that, but regarding the `MaxLength=2` that don't seem right to me. What if we have multiple smaller (e.g. /24s) subnets that the VIP can be assigned from?
Something like:
```
root@octavia1:~# openstack subnet list
+--------------------------------------+-----------------+--------------------------------------+---------------------+
| ID                                   | Name            | Network                              | Subnet              |
+--------------------------------------+-----------------+--------------------------------------+---------------------+
| 340932f1-d742-4762-bd4a-f255dd5609dd | ulab1-ipv4-2    | 1346061d-5bd8-461d-bf43-aa03ed06f2e5 | 10.244.249.0/24     |
| 84e8293d-fffb-4283-b2a7-f8dcda30c3ee | ulab1-ipv4-1    | 1346061d-5bd8-461d-bf43-aa03ed06f2e5 | 10.244.250.0/24     |
| f1c878af-6785-4f02-a551-991e005c8e94 | ulab1-ipv6-1    | 1346061d-5bd8-461d-bf43-aa03ed06f2e5 | 2a02:598:244:1::/64 |
+--------------------------------------+-----------------+--------------------------------------+---------------------+
```

api/v1beta1/types.go
oblazek: yeah that's a good point, tbh we will most likely end up in specifying just the `Network` (by its name) and it won't matter to us which subnet will be chosen as far as I get 1 IPv4 and 1 IPv6 VIP

api/v1beta1/types.go
oblazek: Now that I think about it.. you might be right. I mostly don't care which subnet gets selected as far as I get my IP. We will end up in specifying only the network name/id so that subnet with some IP space is automatically selected for us.

pkg/cloud/services/loadbalancer/loadbalancer.go
oblazek: well the additionalVips are still in beta in gophercloud and only part of  `v2.0.0-beta.1` (or `v2.0.0-beta.2`), I suppose we don't want to upgrade to that, right?

controllers/openstackcluster_controller.go
oblazek: gotcha.. thanks

oblazek:
this should be ready for another round

pkg/cloud/services/loadbalancer/loadbalancer.go
dulek: Yes, I'd rather not upgrade or add support for additionalVIPs here, but I'm really glad the API design assumes it will happen eventually.

api/v1alpha6/openstackcluster_conversion.go
```diff
@@ -287,6 +294,21 @@ func Convert_v1alpha6_OpenStackClusterSpec_To_v1beta1_OpenStackClusterSpec(in *O
...
+}
+
+func Convert_v1beta1_LoadBalancer_To_v1alpha6_LoadBalancer(in *infrav1.LoadBalancer, out *LoadBalancer, s apiconversion.Scope) error {
+	return autoConvert_v1beta1_LoadBalancer_To_v1alpha6_LoadBalancer(in, out, s)
+}
+
+func Convert_v1alpha6_APIServerLoadBalancer_To_v1beta1_APIServerLoadBalancer(in *APIServerLoadBalancer, out *infrav1.APIServerLoadBalancer, s apiconversion.Scope) error {
+	err := autoConvert_v1alpha6_APIServerLoadBalancer_To_v1beta1_APIServerLoadBalancer(in, out, s)
+	out.Subnets = nil
+	out.Network = nil
```
dulek, line 308: I don't think you need to reset these manually? They should be nil already.

api/v1alpha7/openstackcluster_conversion.go
```diff
@@ -289,6 +297,21 @@ func Convert_v1alpha7_OpenStackClusterSpec_To_v1beta1_OpenStackClusterSpec(in *O
...
+}
+
+func Convert_v1beta1_APIServerLoadBalancer_To_v1alpha7_APIServerLoadBalancer(in *infrav1.APIServerLoadBalancer, out *APIServerLoadBalancer, s apiconversion.Scope) error {
+	return autoConvert_v1beta1_APIServerLoadBalancer_To_v1alpha7_APIServerLoadBalancer(in, out, s)
+}
+
+func Convert_v1alpha7_APIServerLoadBalancer_To_v1beta1_APIServerLoadBalancer(in *APIServerLoadBalancer, out *infrav1.APIServerLoadBalancer, s apiconversion.Scope) error {
+	err := autoConvert_v1alpha7_APIServerLoadBalancer_To_v1beta1_APIServerLoadBalancer(in, out, s)
+	out.Subnets = nil
+	out.Network = nil
```
dulek, line 311: Same here.

api/v1beta1/types.go
```diff
@@ -633,6 +639,16 @@ type APIServerLoadBalancer struct {
 	// specified.
 	// +optional
 	Provider optional.String `json:"provider,omitempty"`
+
+	// Network defines which network should the load balancer be allocated on.
+	//+optional
+	Network *NetworkFilter `json:"network,omitempty"`
+	// Subnets define which subnets should the load balancer be allocated on.
+	// It is expected that subnets are located on the network specified in this resource.
```
dulek, line 647: Let's make sure to say that currently only first element is taken into account.

controllers/openstackcluster_controller.go
```diff
@@ -618,6 +672,12 @@ func reconcileNetworkComponents(scope *scope.WithLogger, cluster *clusterv1.Clus
 		return fmt.Errorf("failed to reconcile network: ManagedSubnets only supports one element, %d provided", len(openStackCluster.Spec.ManagedSubnets))
 	}
 
+	err = reconcileLoadBalancerNetwork(openStackCluster, networkingService)
```
dulek, line 675: This isn't exactly "reconcile", more of a "resolve". Not 100% sure, so @mdbooth might have a better opinion.

dulek (commented):
/lgtm

This looks pretty good to me, some small remarks inline.

controllers/openstackcluster_controller.go
mdbooth: Yeah, this is a resolve rather than a reconcile. I'm not going to get hung up on that, though, especially as I'm not 100% convinced of 'resolve' myself.

controllers/openstackcluster_controller.go
```diff
@@ -590,6 +590,60 @@ func bastionHashHasChanged(computeHash string, clusterAnnotations map[string]str
 	return latestHash != computeHash
 }
 
+func reconcileLoadBalancerNetwork(openStackCluster *infrav1.OpenStackCluster, networkingService *networking.Service) error {
```
mdbooth, line 593: Can we pull out:

```golang
lbSpec := openStackCluster.Spec.APIServerLoadBalancer
```

up at the top? It would make some of these lines much shorter.

controllers/openstackcluster_controller.go
```diff
@@ -590,6 +590,60 @@ func bastionHashHasChanged(computeHash string, clusterAnnotations map[string]str
...
+							CIDR: s.CIDR,
+						})
+				}
+			}
+			if !matchFound {
+				handleUpdateOSCError(openStackCluster, fmt.Errorf("no subnet match was found in the specified network (specified subnet: %v, available subnets: %v)", s, lbNetList[0].Subnets))
+				return fmt.Errorf("no subnet match was found in the specified network (specified subnet: %v, available subnets: %v)", s, lbNetList[0].Subnets)
+			}
+		}
+	}
```
mdbooth, line 642: How about making the default explicit and copying in the cluster network in the else branch? Then we could just make the creation code assume it's set. We could also shortcut the lookup at the top of this function if it's already set.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -270,9 +270,26 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
 		return nil, fmt.Errorf("network is not yet available in OpenStackCluster.Status")
 	}
 
-	// Create the VIP on the first cluster subnet
-	subnetID := openStackCluster.Status.Network.Subnets[0].ID
-	s.scope.Logger().Info("Creating load balancer in subnet", "subnetID", subnetID, "name", loadBalancerName)
+	var vipNetworkID, vipSubnetID string
+	if openStackCluster.Status.APIServerLoadBalancer != nil {
+		if openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork != nil {
```
mdbooth, line 275: If we made the suggested change above where we copy in the cluster network when resolving the vip network, this block becomes unconditional. Also, when we add dual stack support here we get it in the non-custom case too, for free. You'll just want to return an error if APIServerLB or APIServerLB.Network are nil in the status, which will also mean we don't need all the nesting.

mdbooth (commented):
A few comments, but this is close. Thanks!

pkg/cloud/services/loadbalancer/loadbalancer.go
oblazek: good call, should be fixed now if I am not mistaken

controllers/openstackcluster_controller.go
oblazek: I think it looks better now.

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/onsi/gomega v1.30.0
 	github.com/prometheus/client_golang v1.17.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/crypto v0.16.0
```
dulek, line 19: Why the deps bump?

oblazek: probably by running some `go mod tidy ` .. will remove

dulek (approved):
/lgtm

PR #1925, opened 2024-03-05
⚠️ Simplify v1alpha6 cluster restorer

mdbooth:
This is a breaking change, although it should not have any negative effects in practise.

The v1alpha6 cluster restorer had become too complex with too many individual fields. This change reduces the number of fields to be consistent with the v1alpha7 restorer.

With this change, an object stored as v1alpha6 with previous v1beta1 annotations could have fields in its spec restored to a semantically equivalent but slightly different serialisation when converted to v1beta1. Differences in the bastion and status are not affected.

In practise this should not matter for 2 reasons:
* The storage version is v1beta1, so the annotation should have been freshly generated in the new format anyway.
* Even if it were not, and the v1alpha6 were under external control, the external controller would observe a difference and restore it to the previous version. It would then be converted again with the new annotation. This would happen only once per object.

Note that if the object is not under external control these differences should not matter anyway.

/hold


EmilienM:
/lgtm

Thanks

PR #1930, opened 2024-03-06
⚠️ OpenStackCluster api general cleanup

mdbooth:
Interestingly, this PR introduces no substantial changes to the CRDs. It primarily affects Go marshalling and documentation. The changes are mostly mechanical.

The PR should hopefully be simpler to review by commit.

The first 2 commits are preparatory, and slightly separate to the other commits. They add some new functionality to the pkg/utils/optional types and update the existing PortOpts to use that functionality.

The rest are structed as changes to single fields, or related sets of fields.

/hold


mdbooth:
This adds a fix for the externalNetwork issue which was causing CI failures. The issue was CI is emitting the following in the cluster spec:

```yaml
  spec:
    externalNetwork:
      id: null
```

The previous logic in ReconcileExternalNetwork correctly treated this as an empty stanza and executed the fallback 'list all external networks' query. The new logic which failed treated this as an explicitly given empty external network filter and executed it, returning all networks including non-external networks. This was an error as there were more than one.

The principal problem here is that we were executing an external network query without the external flag. This is fixed by setting it for both kinds of query (explicit and implicit).

Separately I have also opened https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1932 to fix the weird external network stanza in the CI templates.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -518,12 +518,18 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
...
+		return errors.New("ControlPlaneEndpoint is not yet set in openStackCluster.Spec")
+	}
 
 	loadBalancerName := getLoadBalancerName(clusterName)
 	s.scope.Logger().Info("Reconciling load balancer member", "loadBalancerName", loadBalancerName)
 
 	lbID := openStackCluster.Status.APIServerLoadBalancer.ID
-	portList := []int{int(openStackCluster.Spec.ControlPlaneEndpoint.Port)}
+	var portList []int
+	if openStackCluster.Spec.ControlPlaneEndpoint != nil {
```
EmilienM, line 530: Do we really need the if since we already checked for it at L521?
I guess it's ok. Just in case it was a leftover.

EmilienM:
I've reviewed the commits one by one as suggested and everything looks good to me. I have a minor comment inline, but I think this PR is good to go. Thanks for this huge work!

/lgtm

huxcrux:
This looks good /lgtm

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: No, but I've done similar in a few places. In my first pass at this I didn't double-check any pointers, but after hitting a few panics in various iterations I've decided to err on the side of pretty much always checking. Apart from anything else, this is safer against future cut and paste, where somebody might move either use such that one or the other is no longer guarded.

I'd prefer to fix this through the type system, but I'm not having much joy with that today: https://hachyderm.io/@mattb/112077588227331296

PR #1935, opened 2024-03-08
✨ Add junit output to e2e tests

mdbooth:
Makes e2e output look like this:
image


lentzi90:
Where/how can I see this?

mdbooth:
> Where/how can I see this?

In the e2e output, but only after it finishes.

mdbooth:
This e2e test seems to have become wedged.

Issue #1940, opened 2024-03-12
Port can be created with incorrect name


mdbooth:
/kind bug

This is from code inspection. In `CreatePorts` we generate the port's name by index:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/654d714fd0b0fbdaec2a16489a0d4eb4a5d4f06f/pkg/cloud/services/networking/port.go#L335-L337

However, this index is not absolute. It is relative to the number of ports which were previously created:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/654d714fd0b0fbdaec2a16489a0d4eb4a5d4f06f/controllers/openstackmachine_controller.go#L501-L512

This means that if we declare 2 ports, create 1, fail, then create the other one on the next reconcile, they will both get the index 1.

This is not just untidy, it is also something we rely on in port adoption:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/654d714fd0b0fbdaec2a16489a0d4eb4a5d4f06f/pkg/cloud/services/networking/port.go#L553-L557

PR #1941, opened 2024-03-12
🐛 Fix port name after port creation failure

mdbooth:
CreatePorts was creating a port name based on the index of the port in the *current reconcile*. This could be different to the absolute index of the port if ports had been partially created in a previous reconcile.

We fix this by passing all current state into CreatePorts so it can create an absolute index. This also ensures that partially created ports will be persisted on failure so we don't have to rely on adoption in the next reconcile.

Fixes: #1940 

This PR includes a second commit which refactors and simplifies how we initialises Bastion resources. The motivation for it was that first commit introduces a bug which causes a panic due to an assumption about state having been initialised. The new commit makes this mistake harder to make. It's worth looking at separately.


pkg/cloud/services/networking/port.go
```diff
@@ -58,7 +58,7 @@ func (s *Service) GetPortFromInstanceIP(instanceID string, ip string) ([]ports.P
 	return s.client.ListPort(portOpts)
 }
 
-func (s *Service) CreatePort(eventObject runtime.Object, clusterName string, portName string, portOpts *infrav1.PortOpts, instanceSecurityGroups []string, instanceTags []string) (*ports.Port, error) {
+func (s *Service) CreatePort(eventObject runtime.Object, clusterName string, portName string, portOpts *infrav1.PortOpts, defaultSecurityGroups []string, baseTags []string) (*ports.Port, error) {
```
huxcrux, line 61: Just a minor notice about the name `defaultSecurityGroups`. Since there is a default security group inside Openstack this may cause confusion at first glance.

Maybe there is a good reason for not sticking with the `managedSecurityGroups` name from the openstackmachine_controller?

EmilienM: what about naming it `securityGroups`. We now create ports with SGs, so to me the function arg should be named like this.

EmilienM: good idea :+1:

EmilienM: It's internal API anyway, I'm fine.

EmilienM:
/lgtm
/hold

Feel free to hold cancel when the upgrade passed and when you think it's ready to be merged. Thanks for that work Matt, sorry I missed it in the initial iteration.

mdbooth:
It looks like the full test failed to clean up a couple of ports on deletion. I'm not convinced that was a flake, but running it again to get another data point:

mdbooth:
Oh, we do have logs. The controller paniced: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1941/pull-cluster-api-provider-openstack-e2e-full-test/1768199842766524416/artifacts/clusters/clusterctl-upgrade-y11roz/logs/capo-system/capo-controller-manager/capo-controller-manager-66896777f9-bnz58/manager.log

mdbooth:
#1949 should fix the panic seen in the upgrade test. I don't understand what's triggering it, though.

pkg/cloud/services/networking/port.go
huxcrux: Agree, we can levar it as is after the explanation

mdbooth:
This PR broke cluster deletion when the cluster had a bastion, and the only test which caught it was the upgrade test. This is a bit worrying.

mdbooth:
Note that https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1951 fixes this much more comprehensively by pre-resolving everything about the port prior to creation, including its name.

EmilienM:
Thanks for this work, I'm glad you sorted this out.
/lgtm

Issue #1942, opened 2024-03-12
Deduplicate port adoption code between machines and bastion


mdbooth:
/kind feature

**Describe the solution you'd like**
[AdoptMachinePorts](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/654d714fd0b0fbdaec2a16489a0d4eb4a5d4f06f/pkg/cloud/services/networking/port.go#L530-L578) and [AdoptBastionPorts](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/654d714fd0b0fbdaec2a16489a0d4eb4a5d4f06f/pkg/cloud/services/networking/port.go#L584-L642) are non-trivial substantially identical functions. We should DRY them.

PR #1944, opened 2024-03-12
🌱 Deduplicate AdoptMachinePorts and AdoptBastionPorts

mdbooth:
Both of these methods rely on ReferencedMachineResources and DependentMachineResources, so they can be easily refactored to have a common implementation.

Fixes: #1942 

This PR is based on #1941, so we should merge that first.

TODO:
- [x] Merge #1941 

/hold


pkg/cloud/services/compute/dependent_resources_test.go
EmilienM, line 1: can't we keep the tests or did they become useless?

mdbooth: The function is such a thin wrapper round Adopt() I didn't think it was worth of its own unit tests. They're all basically adoption tests. You don't agree?

mdbooth: Ok, it looks like these were the Adopt tests. I'll resurrect them in port_test.go.

EmilienM: Thanks.

mdbooth: Bah, forgot. Don't let me forget.

mdbooth: I moved them into port_test.go called Test_AdoptPorts, and added a few more.

mdbooth:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1954 was the root cause of most of the full-test CI failures in this PR. This PR includes a change, which we needed to make anyway, which avoids the issue in most cases: we don't need to re-reconcile if we only adopted. However, the underlying issue remains. I'll push a separate PR to fix that.

pkg/cloud/services/compute/dependent_resources.go
```diff
@@ -25,7 +25,7 @@ import (
 func AdoptDependentMachineResources(scope *scope.WithLogger, baseName string, referencedResources *infrav1.ReferencedMachineResources, dependentResources *infrav1.DependentMachineResources) error {
 	networkingService, err := networking.NewService(scope)
 	if err != nil {
-		return false, err
```
dulek: Seems like this change should be in a different commit? It's syntax error otherwise.

dulek (changes requested):

mdbooth:
> Alright, I only found one issue with a change done in the last commit, when it should be done in the middle one. Let's fix that so that reverts would be possible? Or just squash the commits.

Thanks. I'll tidy them up but keep them separate, as I've been *trying* to make smaller, functional commits. When you look at the next one in this series, though, you'll see how that's going 😟

mdbooth:
I've cleaned that up and re-run the non-e2e tests against each commit individually. I also rebased on to main to pick up #1955, which is important for this series.

Issue #1943, opened 2024-03-12
ReferencedResources.Ports contains unresolved security groups


mdbooth:
/kind bug

The purpose of ReferencedMachineResources is to contain unambiguous, fully resolved references to external resources such that the controller can subsequently create resources without further lookups. Unfortunately our referenced ports still contain unresolved references to security groups.

PR #1951, opened 2024-03-15
⚠️ Store []ResolvedPortSpec in ReferencedMachineResources

mdbooth:
The purpose of this change is fix an issue where we are storing unresolved references in ReferencedMachineResources. Specifically we are storing a PortOpts, which is a user-intent struct. PortOpts can contain unresolved references to both subnets and security groups, as well fields requiring additional processing which reference external objects: the port name, description, and tags.

We create a new type, ResolvedPortSpec, which can contain only fully resolved data. This can be seen in the new signature of CreatePorts(), which no longer requires any source of data other than the []ResolvedPortSpec from ReferencedMachineResources, and is now greatly simplified.

Fully resolving the port name also allows a simplification in port adoption.

All of the complexity now moves to ConstructPorts(), which is updated to return []ResolvedPortSpec instead of []PortOpts. ConstructPorts() is updated to resolve security groups, port name, description, and all subnets referenced in FixedIPs.

Fixes: #1943

TODO:
- [x] Merge #1944 

/hold


pkg/cloud/services/networking/port.go
```diff
@@ -393,112 +333,155 @@ func (s *Service) CreatePorts(eventObject runtime.Object, clusterName, baseName
...
+			normalizedPort.SecurityGroups, err = s.GetSecurityGroups(port.SecurityGroups)
+			if err != nil {
+				return nil, fmt.Errorf("error getting security groups: %v", err)
+			}
+		}
 	}
 	return normalizedPorts, nil
 }
 
-// normalizePortTarget ensures that the port has a network ID.
```
dulek, line 469: I think all below should be double-checked by @MaysaMacedo. It looks right to me from dual stack perspective, but this is the delicate part in this PR.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -123,11 +123,10 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 			openStackCluster, err := shared.ClusterForSpec(ctx, e2eCtx, namespace)
 			Expect(err).NotTo(HaveOccurred())
 
-			// Tag: clusterName is declared on OpenStackCluster and gets propagated to all machines
-			// except the bastion host
+			// Tag: clusterName is declared on OpenStackCluster and gets propagated to all machines, including the bastion.
 			allServers, err := shared.DumpOpenStackServers(e2eCtx, servers.ListOpts{Tags: clusterName})
 			Expect(err).NotTo(HaveOccurred())
-			Expect(allServers).To(HaveLen(2))
+			Expect(allServers).To(HaveLen(3))
```
dulek, line 129: Should we document this behavior change in the release notes or v1beta1 docs? It's additive, but someone might be depending on this. Obviously the change itself makes perfect sense to me.

dulek (changes requested):
Comments inline, most of them are more of a question thank a remark.

pkg/cloud/services/networking/port.go
mdbooth: It should be just code motion with changes for context. I didn't intentionally change the logic here.

config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
```diff
@@ -6626,6 +6430,10 @@ spec:
                                 implementations. What type of vNIC is actually available depends on
                                 deployments. If not specified, the Neutron default value is used.
                               type: string
+                          required:
+                          - description
```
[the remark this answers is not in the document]
mdbooth: This is the resolved port spec in the status, which will always have a description: if it didn't have a description we would have generated one. This is expected.

api/v1beta1/types.go
```diff
@@ -260,6 +267,39 @@ type PortOpts struct {
...
+	Description string `json:"description"`
+
+	// Tags applied to the port (and corresponding trunk, if a trunk is configured.)
+	// +listType=set
+	// +optional
+	Tags []string `json:"tags,omitempty"`
+
+	// Trunk specifies whether trunking is enabled at the port level.
+	// +optional
+	Trunk bool `json:"trunk,omitempty"`
```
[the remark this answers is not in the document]
mdbooth: I should make this optional.Bool for consistency.

test/e2e/suites/e2e/e2e_test.go
mdbooth: I created https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1958

mdbooth:
@dulek Comments addressed.

mdbooth:
Nothing obvious, could be a flake.

mdbooth:
Same error.

pkg/utils/names/names.go
```diff
@@ -36,3 +38,7 @@ func GetFloatingAddressClaimName(openStackMachineName string) string {
 func GetOpenStackMachineNameFromClaimName(claimName string) string {
 	return strings.TrimSuffix(claimName, fmt.Sprintf("-%s", FloatingAddressIPClaimNameSuffix))
 }
+
+func ClusterName(cluster *clusterv1.Cluster) string {
```
huxcrux, line 42: Maybe we should call this variable something else? I get this is used for a lot of resources however it is not the clusters name.

Maybe clusterResourceName or something could work?

I really like the initiative to move this to a function since the name are present in a lot of places in the code.

huxcrux (commented):
I looked thru the code and think it looks good. Only one inline comment regarding the function clusterName that actually returns a name used for resources and is not the cluster name.

It can safely be ignored however may cause confusion for someone reading the code. When I see clusterName i almost every time ask myself why cluster.Name is not used since cluster exists almost everywhere already

pkg/utils/names/names.go
mdbooth: I've pushed a separate commit which changes it everywhere. Note that `clusterName` in the e2e tests actually does mean cluster name, so I haven't changed it there.

mdbooth:
I'm not going to re-run the full test on this latest push because it just updates function and variable names.

huxcrux:
/lgtm

Lets hope the full e2e test passes 🤞

mdbooth:
Passed!

/hold cancel

Issue #1956, opened 2024-03-18
Including Security Group Rules make cluster status huge


mdbooth:
It appears we never read them from the status: we always get them from OpenStack. Lets remove them from the status. We can always add them back later if we find a need for it.

PR #1957, opened 2024-03-19
⚠️ Remove security group rules from status

mdbooth:
We were writing the full set of generated security rules to the status. They dominated the size of cluster object and we weren't reading them: we always fetch the current rules from OpenStack during reconciliation.

This also tweaks the rule generation code slightly, which was previously reading security group rules multiple times during a reconciliation. We now do this first and pass the result around.

Fixes: #1956


huxcrux:
This looks good, skipping `/lgtm` to allow the full e2e test to run.

mdbooth:
This is a bug in this patch:

```
  failureMessage: 'failed to reconcile security groups: unable to generate desired
    security group rules: security group k8s-cluster-e2e-bql68j-cluster-e2e-bql68j-secgroup-bastion
    not found'
```

mdbooth:
D'oh! That's so dumb: when I was moving the code which generates `observedSecGroups` I didn't actually add the created groups to the map 🤦

We have test coverage of the various methods called by `ReconcileSecurityGroups`, but none of `ReconcileSecurityGroups` itself, so the local tests missed this. I'll see if I can add something which would have caught this.

mdbooth:
I've fixed the bug and added some unit tests which would have triggered it.

PR #1959, opened 2024-03-19
⚠️ Bastion cleanup

dulek:
This commit works with .spec.bastion:
* makes availablityZone, floatingIP and instance pointers
* adds floatingIP IPv4 validation
* adds validation that if Bastion is enabled, instance cannot be null
* rename instance->spec

Fixes #

**TODOs**:

- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold


api/v1beta1/types.go
```diff
@@ -596,21 +596,26 @@ var (
 	InstanceStateUndefined = InstanceState("")
 )
 
-// Bastion represents basic information about the bastion node.
+// Bastion represents basic information about the bastion node. If you enable bastion, the spec has to be specified.
+// +kubebuilder:validation:XValidation:rule="!self.enabled || has(self.spec)",message="you need to specify the spec if bastion is enabled"
 type Bastion struct {
+	// Enabled means that bastion is enabled. Defaults to false.
 	//+optional
```
mdbooth, line 603: Can we make this required with an explicit default of false?

controllers/openstackcluster_controller.go
```diff
@@ -504,9 +507,9 @@ func bastionAddFloatingIP(openStackCluster *infrav1.OpenStackCluster, clusterNam
 	case openStackCluster.Status.Bastion.FloatingIP != "":
 		// Some floating IP has already been created for this bastion, make sure we re-use it
 		floatingIP = &openStackCluster.Status.Bastion.FloatingIP
-	case openStackCluster.Spec.Bastion.FloatingIP != "":
+	case openStackCluster.Spec.Bastion.FloatingIP != nil && *openStackCluster.Spec.Bastion.FloatingIP != "":
```
mdbooth, line 510: We *should* be able to just use a nil check here. Should...

controllers/openstackcluster_controller_test.go
```diff
@@ -61,6 +61,11 @@ var _ = Describe("OpenStackCluster controller", func() {
 	capiClusterName := "capi-cluster"
 	testClusterName := "test-cluster"
 	testNum := 0
+	instance := infrav1.OpenStackMachineSpec{
```
mdbooth, line 64: nit
```suggestion
	bastionSpec := infrav1.OpenStackMachineSpec{
```

controllers/openstackcluster_controller_test.go
```diff
@@ -194,6 +199,7 @@ var _ = Describe("OpenStackCluster controller", func() {
 		testCluster.Spec = infrav1.OpenStackClusterSpec{
 			Bastion: &infrav1.Bastion{
 				Enabled: false,
+				Spec:    &instance,
```
mdbooth, line 202: You can (should?) omit the Spec here.

controllers/openstackcluster_controller_test.go
```diff
@@ -471,7 +480,10 @@ var _ = Describe("OpenStackCluster controller", func() {
 	It("should delete an existing bastion even if its uuid is not stored in status", func() {
 		testCluster.SetName("delete-existing-bastion")
 		testCluster.Spec = infrav1.OpenStackClusterSpec{
-			Bastion: &infrav1.Bastion{},
+			Bastion: &infrav1.Bastion{
+				Enabled: true,
+				Spec:    &instance,
+			},
```
mdbooth, line 486: I don't think this test is equivalent. The previous test had: `Bastion: &infrav1.Bastion{}`, which implied `Enabled: false`, and we apparently still expect a bastion to be cleaned up in this case. Is that what we were actually testing before? Does it still work that way?

controllers/openstackmachine_controller.go
```diff
@@ -586,7 +586,7 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 		// The other state is normal (for example, migrating, shutoff) but we don't want to proceed until it's ACTIVE
 		// due to potential conflict or unexpected actions
 		scope.Logger().Info("Waiting for instance to become ACTIVE", "id", instanceStatus.ID(), "status", instanceStatus.State())
-		conditions.MarkUnknown(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotReadyReason, "Instance state is not handled: %s", instanceStatus.State())
+		conditions.MarkUnknown(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotReadyReason, "Spec state is not handled: %s", instanceStatus.State())
```
mdbooth, line 589: Over-aggressive search/replace!

```suggestion
		conditions.MarkUnknown(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotReadyReason, "Instance state is not handled: %s", instanceStatus.State())
```

pkg/cloud/services/compute/instance_types.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // InstanceSpec defines the fields which can be set on a new OpenStack instance.
 //
-// InstanceSpec does not contain all of the fields of infrav1.Instance, as not
+// InstanceSpec does not contain all of the fields of infrav1.Spec, as not
```
mdbooth, line 33: This doesn't look correct.

Also, this comment was relevant context when I originally forked this struct from Instance. I think we can just delete it now.

api/v1alpha6/types_conversion.go
```diff
@@ -307,7 +307,7 @@ func Convert_v1beta1_BindingProfile_To_Map_string_To_Interface(in *infrav1.Bindi
 
 /* FixedIP */
 /* AddressPair */
-/* Instance */
+/* Spec */
```
mdbooth, line 310: Nope?

test/e2e/suites/apivalidations/openstackcluster_test.go
mdbooth, line 1: Nice tests!

api/v1beta1/types.go
dulek: Thanks for this, I wasn't sure how to approach this.

controllers/openstackcluster_controller_test.go
dulek: Okay, I needed to actually fix this with a hack on `bastionToInstanceSpec()`, where nil `.spec` will be translated to `&OpenStackMachineSpec{}`.

PR #1960, opened 2024-03-20
🐛 Add terminal failure for lb client init

### LINKED ISSUE #1950: after loadBalancerService client init failed,we should give failed message to openstackcluster status
/kind bug

such as failed to create load balancer service client: No suitable endpoint could be found in the service catalog.

openstack cluster should give an error when this error cannot be automatically recovered.

**Environment:**

latest
- Cluster-API version:
latest
- OpenStack version:
- Minikube/KIND version:
1.20.14

Goend:
After loadBalancerService client init failed,we should give failed message to openstackcluster status.

Fixes #1950 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


controllers/openstackcluster_controller.go
```diff
@@ -745,6 +745,7 @@ func reconcileControlPlaneEndpoint(scope *scope.WithLogger, networkingService *n
 	case openStackCluster.Spec.APIServerLoadBalancer.IsEnabled():
 		loadBalancerService, err := loadbalancer.NewService(scope)
 		if err != nil {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to init load balancer client: %w", err))
```
mdbooth, line 748: There are a couple of problems with this.

Firstly, not all errors returned here would be terminal errors. I think I would want to be matching specifically on an 'endpoint not found' error here. For example, we wouldn't want to go marking all clusters perma-failed just because of a temporary API outage.

Secondly, we unfortunately don't actually have a perma-failed condition for clusters, although I think we should and this would be an example of when we'd use it. We do have this for OpenStackMachine, but not for OpenStackCluster. Before implementing one we'd need to audit most uses of this function.

Goend: fine,I will add it

Goend: Your suggestion is that this change may require reviewing the entire failure handling process of OpenStack Cluster, not just in this one place?

mdbooth: Unfortunately, I think so. We'd have to add a 'failed' condition to start with, which we don't currently have, and we'd need to ensure we don't set it anywhere unintentionally.

I think the effort would be worthwhile, though. We should have this.

Goend: I have got it, I will try my best to complete this PR.

Goend: How do I handle the compatibility of v1alpha6 and v1alpha7 versions? This is my first time submitting code to this community and I need help in this area. Are there any examples or explanations I can refer to?  @mdbooth

api/v1alpha6/openstackcluster_types.go
```diff
@@ -210,6 +210,8 @@ type OpenStackClusterStatus struct {
 	// and/or logged in the controller's output.
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
```
dulek, line 214: You shouldn't modify v1alpha6, only v1beta1 now.

api/v1alpha7/openstackcluster_types.go
```diff
@@ -227,6 +227,8 @@ type OpenStackClusterStatus struct {
 	// and/or logged in the controller's output.
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
```
dulek, line 231: Same here.

api/v1beta1/openstackcluster_types.go
```diff
@@ -276,6 +276,8 @@ type OpenStackClusterStatus struct {
 	// and/or logged in the controller's output.
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
```
dulek, line 280: Once you remove this field from v1alpha* you'll see conversions tests failing. To please them you'll need to write restorers. The idea is that when you're adding new fields and not moving them around, there's no way a backward conversions won't be lossy. In order to make sure that v1beta1->v1alpha6->v1beta1 will return the same, some part of the CRD needs to be restored from information in the annotations. This is what the restorer will do. Please check out a few examples in the conversions code.

dulek (changes requested):

api/v1alpha6/openstackcluster_types.go
Goend: get it.thanks

api/v1beta1/openstackcluster_types.go
Goend: If the test case covers the case you mentioned, then the code is okay. I cannot test this case in my local environment.

EmilienM:
@Goend if you like, I can help you with this PR. Let me know and I can push your patch via another PR with everything fixed.

api/v1alpha6/openstackcluster_conversion.go
```diff
@@ -546,6 +546,29 @@ func restorev1beta1BastionStatus(previous *infrav1.BastionStatus, dst *infrav1.B
 	dst.Resources = previous.Resources
 }
 
+/* Condition status */
```
EmilienM, line 549: Maybe we don't need that?
The way I made it work locally is to add this to `restorev1beta1ClusterStatus`:

```golang
if previous.Conditions != nil {
    dst.Conditions = previous.Conditions
}
```

api/v1beta1/openstackcluster_types.go
EmilienM, line 280: Please add a comment for that, and `+optional`.

controllers/openstackcluster_controller.go
```diff
@@ -972,3 +992,11 @@ func setClusterNetwork(openStackCluster *infrav1.OpenStackCluster, network *netw
 	openStackCluster.Status.Network.Name = network.Name
 	openStackCluster.Status.Network.Tags = network.Tags
 }
+
+func patchCluster(ctx context.Context, patchHelper *patch.Helper, openStackCluster *infrav1.OpenStackCluster, options ...patch.Option) error {
```
EmilienM, line 996: Please add doc for that function.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -46,6 +46,8 @@ const (
 	resolvedMsg     string = "ControlPlaneEndpoint.Host is not an IP address, using the first resolved IP address"
 )
 
+var ErrLoadBalancerNoPoint = "failed to create load balancer service client: No suitable endpoint could be found in the service catalog."
```
EmilienM, line 49: why not a constant?

EmilienM:
/close
This was moved to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2036

PR #1961, opened 2024-03-20
⚠️ Remove InstanceID from machine spec

mdbooth:
This is a status field and should not be included in the spec. ProviderID is the same, but cannot be removed as it is required by Cluster API.

While we're changing it, we change the default printcolumns for OpenStackMachine to report server ID instead of provider ID. We don't need to report provider ID here because we already know everything is an openstack server. New output example:

```
> kc get openstackmachine
NAME             CLUSTER    SERVERID                               INSTANCESTATE   READY   MACHINE          AGE
test-foo-fbvcg   test-foo   b4547d60-0f7e-44ea-b77a-30e951556fe3   ACTIVE          true    test-foo-fbvcg   5m52s
```

/hold


mdbooth:
This needs a docs update, and also an update to the machine printcolumns.

pkg/cloud/services/compute/instance_types.go
```diff
@@ -102,10 +102,7 @@ func (is *InstanceStatus) UpdateBastionStatus(openStackCluster *infrav1.OpenStac
 		openStackCluster.Status.Bastion = &infrav1.BastionStatus{}
 	}
 
-	openStackCluster.Status.Bastion.ID = is.ID()
 	openStackCluster.Status.Bastion.Name = is.Name()
-	openStackCluster.Status.Bastion.SSHKeyName = is.SSHKeyName()
```
MaysaMacedo, line 107: with SSHKeyName being removed, the function at line 91 can also be removed.

MaysaMacedo: Also, it's worth documenting that these fields are being removed

mdbooth: Both addressed. Also noticed that we weren't using `AvailabilityZone()` any more and removed that, too.

api/v1alpha6/openstackmachine_conversion.go
```diff
@@ -342,7 +385,16 @@ func Convert_v1beta1_OpenStackMachineSpec_To_v1alpha6_OpenStackMachineSpec(in *i
 
 /* OpenStackMachineStatus */
 
+func restorev1alpha6MachineStatus(previous *OpenStackMachineStatus, dst *OpenStackMachineStatus) {
+	if dst.InstanceState == nil || *dst.InstanceState == "" {
+		dst.InstanceState = previous.InstanceState
```
MaysaMacedo, line 390: the other restore functions skip the setting of previous into dst when the resource is empty/nil. Should we follow the same pattern here?

docs/book/src/topics/crd-changes/v1alpha7-to-v1beta1.md
```diff
@@ -157,6 +158,10 @@ spec:
         name: "MyOpenStackFloatingIPPool"
 ```
 
+#### Removal of instanceID from spec
+
+The OpenStackMachine controller previously set `instanceID` in the machine spec, in addition to `providerID`. `providerID` remains unchanged as it is required by Cluster API, but `instanceID` is no longer set. If still required, it is now available in the status in `dependentResources`.
```
MaysaMacedo, line 164: dependentResources should be renamed to resources

api/v1beta1/openstackmachine_types.go
```diff
@@ -164,9 +159,9 @@ type OpenStackMachineStatus struct {
 // +kubebuilder:resource:path=openstackmachines,scope=Namespaced,categories=cluster-api,shortName=osm
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this OpenStackMachine belongs"
-// +kubebuilder:printcolumn:name="InstanceState",type="string",JSONPath=".status.instanceState",description="OpenStack instance state"
+// +kubebuilder:printcolumn:name="ServerID",type="string",JSONPath=".status.dependentResources.server.id",description="OpenStack instance ID"
```
MaysaMacedo, line 162: dependentResources should now be resources?

api/v1beta1/openstackmachine_types.go
```diff
@@ -164,9 +159,9 @@ type OpenStackMachineStatus struct {
...
+// +kubebuilder:printcolumn:name="InstanceState",type="string",JSONPath=".status.dependentResources.server.state",description="OpenStack instance state"
```
MaysaMacedo, line 163: ditto

api/v1alpha7/openstackmachine_conversion.go
```diff
@@ -232,6 +275,17 @@ func Convert_v1beta1_OpenStackMachineSpec_To_v1alpha7_OpenStackMachineSpec(in *i
...
 /* OpenStackMachineStatus */
 
+func restorev1alpha7MachineStatus(previous *OpenStackMachineStatus, dst *OpenStackMachineStatus) {
+	if dst.InstanceState == nil || *dst.InstanceState == "" {
+		dst.InstanceState = previous.InstanceState
+	}
+}
+
+func Convert_v1alpha7_OpenStackMachineStatus_To_v1beta1_OpenStackMachineStatus(in *OpenStackMachineStatus, out *infrav1.OpenStackMachineStatus, s apiconversion.Scope) error {
+	// ReferencedResources have no equivalent in v1beta1
```
MaysaMacedo, line 285: we probably want to replace `ReferencedResources` for `Resolved`

api/v1alpha6/openstackmachine_conversion.go
mdbooth: I've been adding that incrementally. As this is a new function I should have added it here.

api/v1beta1/openstackmachine_types.go
mdbooth: Updated and re-tested on a running cluster:

```
> kc get openstackmachine
NAME             CLUSTER    SERVERID                               INSTANCESTATE   READY   MACHINE          AGE
test-foo-c4wdz   test-foo   a5500a2d-d1f8-4b8e-a7bb-2b94977eb893   BUILD                   test-foo-c4wdz   4m26s
```

api/v1beta1/types.go
```diff
@@ -358,14 +358,15 @@ type AddressPair struct {
 }
 
 type BastionStatus struct {
-	ID         string              `json:"id,omitempty"`
```
mdbooth, line 361: Should probably put bastion ID and State back, too.

api/v1beta1/types.go
```diff
@@ -358,14 +358,15 @@ type AddressPair struct {
...
-	SSHKeyName string              `json:"sshKeyName,omitempty"`
-	State      InstanceState       `json:"state,omitempty"`
 	IP         string              `json:"ip,omitempty"`
 	FloatingIP string              `json:"floatingIP,omitempty"`
 	Resolved   ResolvedMachineSpec `json:"resolved,omitempty"`
 	Resources  MachineResources    `json:"resources,omitempty"`
+
+	// Server describes the OpenStack server created for the bastion
+	// +optional
+	Server *ServerStatus `json:"server,omitempty"`
```
mdbooth, line 369: Do we still need this?

api/v1beta1/types.go
dulek: It was the idea that State and ID goes into Server to mimic what's on OpenStackMachineStatus. No strong opinion from my side here.

dulek:
Okay, I think I found the issue.

mdbooth:
I've decided to replace this with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1988, which has a smaller scope.

PR #1962, opened 2024-03-20
🌱 Update FloatingIPPool to use v1beta1

mdbooth:
Also make IdentityRef required, as it doesn't currently permit fallback to cluster credentials.

cc @bilbobrovall @huxcrux 

/hold


mdbooth:
`size/L`. I'm guessing that means either it didn't work, or the .gitattributes change isn't applied to this PR. The non-generated diffstat is actually:

```
api/v1alpha1/openstackfloatingippool_types.go     | 18 +++++-------------
controllers/openstackfloatingippool_controller.go |  7 +++----
pkg/scope/provider.go                             | 12 +++---------
3 files changed, 11 insertions(+), 26 deletions(-)
```

PR #1969, opened 2024-03-23
Add an opt to use AZ for APISeverLoadBalancer

nguyenhuukhoi:
This PR allows to choose AZ for API Load Balancer. It will be useful if we want to use edge sites,

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):

/hold


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -305,6 +305,7 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
 		Description: names.GetDescription(clusterResourceName),
 		Provider:    lbProvider,
 		Tags:        openStackCluster.Spec.Tags,
+		APIServerLoadBalancerAZ: openStackCluster.spec.apiSeverLoadBalancerAZ
```
huxcrux, line 308: I assume this works based on the tags however I assume you have tested to not specify any AZ for your LB just to make sure the CreateLoadBalancer doesn't break if an empty or null value is received

huxcrux (commented):
Thanks for your Pr, this is a good addition to CAPO 👍 

We probably do not want this in older APIs (non v1beta1) however I let one of the approvers make a decision on this. Generally new features is not backported to older API versions unless a bug is found.

When you feel like you are done with this PR it would be nice if you squash your commits to keep the main branch tidy.

/ok-to-test

pkg/cloud/services/loadbalancer/loadbalancer.go
nguyenhuukhoi: Ok. I will check about it.
TY

nguyenhuukhoi:
> Thanks for your PR, this is a good addition to CAPO 👍
> 
> We probably do not want this in older APIs (non v1beta1) however I let one of the approvers make a decision on this. Generally new features is not backported to older API versions unless a bug is found.
> 
> When you feel like you are done with this PR it would be nice if you squash your commits to keep the main branch tidy.
> 
> /ok-to-test
Hello.
We dont add new features on older APIs, Does it is our rule?

pkg/cloud/services/loadbalancer/loadbalancer.go
huxcrux: BTW looks like CI is failing due to a missing comma at the end of this line. I'll add a suggestion below:

```suggestion
		APIServerLoadBalancerAZ: openStackCluster.spec.apiSeverLoadBalancerAZ,
```

nguyenhuukhoi: I changed code.

nguyenhuukhoi:
Hello.
Is it ok if I drop Convert_v1beta1_APIServerLoadBalancer_To_v1alpha7_APIServerLoadBalancer?

nguyenhuukhoi:
> When you feel like you are done with this PR it would be nice if you squash your commits to keep the main branch tidy.

Hello. Could u explain it?

dulek:
> > When you feel like you are done with this PR it would be nice if you squash your commits to keep the main branch tidy.
> 
> Hello. Could u explain it?

We mean that commits should be squashed into one or at most a few logical parts that can be independently reverted. I'm pretty sure this can land as a single commit.

`git rebase -i <first-commit>` will help you do this.

api/v1beta1/types.go
```diff
@@ -647,6 +647,10 @@ type APIServerLoadBalancer struct {
 	// specified.
 	// +optional
 	Provider optional.String `json:"provider,omitempty"`
+
+	// API Server LoadBalancer AvailabilityZone
+	// +optional
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
```
dulek, line 653: If AZ is optional it has to be `optional.String` which resolves to `*string`. This is a K8s API guideline.

dulek (changes requested):

nguyenhuukhoi:
> git rebase -i
Hello.
Could you explain more detail on that?

PR #1987, opened 2024-04-01
🐛 Fix v1alpha7 machine restorer and test

### LINKED ISSUE #1986: v1alpha7 Machine conversion is lossy when controller sets ProviderID or InstanceID
/kind bug

v1alpha7 OpenStackMachine conversion should ignore changes to ProviderID and InstanceID and run the restorers even if they have changed. This is because these spec fields are modified by the openstack machine controller, so if we didn't exclude them, v1alpha7 users would *always* see changes to user-managed fields in the machine spec after conversion.

There's a test for this, `TestMachineConversionControllerSpecFields`, but it's broken in v1alpha7 and doesn't actually check that a lossy field was correctly restored.

mdbooth:
The v1alpha7 machine spec restorer was missing the exclusion for ProviderID and InstanceID was missing, and the test which was supposed to ensure it worked was broken. This fixes both.

Fixes #1986

/hold

### DIFF
--- a/api/v1alpha7/conversion_test.go
+++ b/api/v1alpha7/conversion_test.go
@@ -166,22 +166,24 @@ func TestFuzzyConversion(t *testing.T) {
 func TestMachineConversionControllerSpecFields(t *testing.T) {
 	// This tests that we still do field restoration when the controller modifies ProviderID and InstanceID in the spec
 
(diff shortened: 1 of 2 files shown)


mdbooth:
Out of an abundance of caution:

PR #1988, opened 2024-04-01
⚠️ Move InstanceID from Spec to Status

mdbooth:
This is a status field and should not be included in the spec. ProviderID is the same, but cannot be removed as it is required by Cluster API.

This PR replaces https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1961 with a smaller scope.

TODO:
- [ ] Integrate `restorev1alpha6MachineTemplateMachineSpec` into https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1989

/hold


api/v1beta1/openstackmachine_types.go
```diff
@@ -119,6 +118,10 @@ type OpenStackMachineStatus struct {
 	// +optional
 	Ready bool `json:"ready"`
 
+	// InstanceID is the OpenStack instance ID for this machine.
+	// +optional
+	InstanceID optional.String `json:"instanceID,omitempty"`
```
MaysaMacedo, line 123: Have we considered adding it under the `Resolved` field?

MaysaMacedo (commented):
It lgtm, I just have one question:

api/v1beta1/openstackmachine_types.go
mdbooth: My first PR put it in `Resources`. This made sense, but I've decided against that for now. The reason is I think InstanceID is `fundamental` to the OpenStackMachine object and something an end user might be interested in (e.g. in `printcolumns`), but the other properties are not.

I have an idea that I might move actual machine creation into a separate controller with its own CRD, in which case `Resources` would move to the new controller. In that case I think we'd still want InstanceID to be on *this* object, so I decided to put it here instead.

None of which precludes adding it to Resources in the future too if it turns out that still makes sense.

PR #1989, opened 2024-04-01
🐛 Fix multiple panics in restore functions

mdbooth:
Panics in restore functions are a particular problem because `previous` is unvalidated user-provided input. It should have been generated by a previous conversion, but we have no way to enforce that. We don't need to preserve sensible behaviour given invalid input, but we should not crash.

This PR implements a generic function for fuzzing restore functions, adds v1alpha6 and v1alpha7 tests which calls the fuzzer against all restore functions, and fixes all the discovered panics.

/hold


mdbooth:
In case it's important later: I manually removed the `needs-rebase` label. Not sure why the bot didn't do that itself, but it doesn't currently need a rebase.

PR #1994, opened 2024-04-02
🌱 Bump version artifacts for release-0.10

mdbooth:


dulek:
Looks good, but I wonder why it's needed now and not on 0.10 release.

mdbooth:
> Looks good, but I wonder why it's needed now and not on 0.10 release.

It will be needed on main as soon as a v0.10 tag is created due to the way we generate version numbers and clusterctl gets metadata.

PR #2000, opened 2024-04-04
🐛 Rules are optional in security group rules before v1beta1

### LINKED ISSUE #1999: "status.bastionSecurityGroup.rules: Required value" updating cluster
/kind bug

From @lentzi90 

```
❯ kubectl diff -k CAPO/test-cluster
The OpenStackCluster "lennart-test" is invalid: 
* status.bastionSecurityGroup.rules: Required value
* status.workerSecurityGroup.rules: Required value
* status.controlPlaneSecurityGroup.rules: Required value
```

when doing the following:
```
* Init with CAPO v0.9
* Create a v1alpha7 cluster
* Upgrade CAPO to v0.10.0-alpha.0
* Try to update the OpenStackCluster (e.g. add an IP to the allowedCidrs)
```

mdbooth:
Rules were removed from the security group status in v1beta1. If the status is modified then an up-converted v1alphaN object will not have its security group rules restored on down-conversion. As the CRD requires rules, this results in an API error.

Fixes #1999

/hold

### DIFF
--- a/api/v1alpha5/types.go
+++ b/api/v1alpha5/types.go
@@ -241,7 +241,7 @@ type LoadBalancer struct {
 type SecurityGroup struct {
 	Name  string              `json:"name"`
 	ID    string              `json:"id"`
-	Rules []SecurityGroupRule `json:"rules"`
+	Rules []SecurityGroupRule `json:"rules,omitempty"`
 }
 
 // SecurityGroupRule represent the basic information of the associated OpenStack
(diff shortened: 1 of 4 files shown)


lentzi90 (approved):
Confirmed that this solves the issue with manual testing!
/approve

lentzi90:
For future reference my test was the following:
- Set up a testing environment with Tilt from CAPI
- Apply a v1alpha7 cluster
- Make some change and then do `kubectl diff -k test-cluster`

On this commit I get no issue but when I do the same on main I get the message about the cluster being invalid.

PR #2001, opened 2024-04-04
🌱 Minor improvements to api validation tests

mdbooth:
This comprises 3 small cleanups:
* A makefile tidyup to how the KUBEBUILDER_ASSETS variable is set, which also allows it to be overridden
* Update the apivalidations to delete the objects it creates. Not really important, but might throw something up.
* Fix initialisation of the apivalidations envtest environment so that conversion webhooks work. Doesn't actually add any tests that use it, though.

/hold


lentzi90 (approved):
Very good find!
/approve

PR #2006, opened 2024-04-09
🐛 Don't try to resolve machine on delete if cluster not ready

### LINKED ISSUE #1792: Stuck at deleting cluster
/kind bug

Create cluster with a user without load_balancer_member role.
Delete cluster >> Deleted Failed>> Add load_balancer_member role for this user.
Delete cluster again> DELETE_IN_PROGRESS forever.

Delete cluster successfully.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.8.0
- Cluster-API version: v1.5.4
- OpenStack version: Yoga
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): v1.27.4
- OS (e.g. from `/etc/os-release`): Ubuntu 22.04

mdbooth:
This fixes a bug where if we created a machine for a cluster which never became ready, we would never be able to 'resolve' the machine and therefore never delete it. We address this situation in several layers:

Firstly, we move the point in the machine controller at which we add the finalizer in the first place. We don't add the finalizer until we're writing resolved, so this situation can never occur for newly created machines. This makes sense because we don't create resources until we've observed that both the finalizer has been added and resolved is up to date, so we don't need the finalizer to protect resources which can't have been created yet.

Secondly, we shortcut the delete flow if the cluster is not ready. This is safe for the same reason as above, but is only relevant to machines created before v0.10.

Lastly we surface and restrict the circumstances in which 'Resolved' is required on delete anyway. On closer inspection, this is only required in the very specific circumstance that the machine has volumes defined, and we are deleting it without the machine having been created. To make this more obvious we split volume deletion out of DeleteInstance and only resolve the machine spec in the event that it's required.

2 other factors make this change larger than it might otherwise be.

We hit a cyclomatic complexity limit in reconcileDelete(), requiring a refactor.

We remove the DeleteInstance tests which, after separating out DeleteVolumes, are quite trivial, and replace them with much more comprehensive set of tests for reconcileDelete.

Fixes #1792

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -290,7 +290,18 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 		}
 	}
 
-	if instanceStatus != nil {
+	// If no instance was created we currently need to check for orphaned
+	// volumes. This requires resolving the instance spec.
+	// TODO: write volumes to status resources on creation so this is no longer required.
+	if instanceStatus == nil {
+		instanceSpec, err := bastionToInstanceSpec(openStackCluster, cluster)
+		if err != nil {
+			return err
+		}
+		if err := computeService.DeleteVolumes(instanceSpec); err != nil {
+			return fmt.Errorf("delete volumes: %w", err)
+		}
+	} else {
 		instanceNS, err := instanceStatus.NetworkStatus()
 		if err != nil {
 			return err
@@ -307,11 +318,7 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 			}
 		}
 
-		instanceSpec, err := bastionToInstanceSpec(openStackCluster, cluster)
-		if err != nil {
-			return err
-		}
-		if err = computeService.DeleteInstance(openStackCluster, instanceStatus, instanceSpec); err != nil {
+		if err = computeService.DeleteInstance(openStackCluster, instanceStatus); err != nil {
 			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete bastion: %w", err))
 			return fmt.Errorf("failed to delete bastion: %w", err)
 		}
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -169,18 +169,13 @@ func resolveMachineResources(scope *scope.WithLogger, clusterResourceName string
 		openStackMachine.Status.Resolved = resolved
 	}
 	// Resolve and store resources
-	changed, err := compute.ResolveMachineSpec(scope,
+	return compute.ResolveMachineSpec(scope,
 		&openStackMachine.Spec, resolved,
 		clusterResourceName, openStackMachine.Name,
 		openStackCluster, getManagedSecurityGroup(openStackCluster, machine))
-	if err != nil {
-		return false, err
-	}
-	if changed {
-		// If the resolved machine spec changed we need to start the reconcile again to prevent inconsistency between reconciles.
-		return true, nil
-	}
+}
 
+func adoptMachineResources(scope *scope.WithLogger, openStackMachine *infrav1.OpenStackMachine) error {
 	resources := openStackMachine.Status.Resources
 	if resources == nil {
 		resources = &infrav1.MachineResources{}
@@ -188,7 +183,7 @@ func resolveMachineResources(scope *scope.WithLogger, clusterResourceName string
 	}
 
 	// Adopt any existing resources
-	return false, compute.AdoptMachineResources(scope, resolved, resources)
+	return compute.AdoptMachineResources(scope, openStackMachine.Status.Resolved, resources)
 }
 
 func patchMachine(ctx context.Context, patchHelper *patch.Helper, openStackMachine *infrav1.OpenStackMachine, machine *clusterv1.Machine, options ...patch.Option) error {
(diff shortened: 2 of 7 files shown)


controllers/openstackmachine_controller.go
```diff
@@ -341,6 +342,62 @@ func (r *OpenStackMachineReconciler) reconcileDelete(scope *scope.WithLogger, cl
...
+			return err
+		}
+		return nil
+	}
+
+	// XXX(mdbooth): This looks wrong to me. Surely we should only ever
+	// disassociate the floating IP here. I would expect the API server
+	// floating IP to be created and deleted with the cluster. And if the
+	// delete behaviour is correct, we leak it if the instance was
+	// previously deleted.
```
dulek, line 371: There's an assumption that if `APIServerFloatingIP` was `nil`, then the FIP was created by CAPO [1], but the problem is that this doesn't seem to lookup that FIP, but all the FIPs of that instance. If someone attached additional FIP to an instance, e.g. for debugging purposes it will be deleted too. This is horribly wrong to me.

Also no need to detach FIP at all. If we delete the instance and port is deleted, FIP will be detached.

[1] https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/d33c1bc487fecf5ce6ede3f0d08c0cc3047fb166/api/v1alpha5/openstackcluster_types.go#L82

controllers/openstackmachine_controller.go
```diff
@@ -489,34 +546,50 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+	// Check for orphaned resources previously created but not written to the status
+	if err := adoptMachineResources(scope, openStackMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("adopting machine resources: %w", err)
+	}
+
 	scope.Logger().Info("Reconciling Machine")
+	userData, err := r.getBootstrapData(ctx, machine, openStackMachine)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
dulek, line 592: If this fails we still don't create the machine, why move it down here?

controllers/openstackmachine_controller.go
```diff
@@ -489,34 +546,50 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
...
+	// resources. This means that we never create any resources unless we
+	// have observed that the finalizer and resolved machine resources were
+	// successfully written in a previous transaction. This in turn means
+	// that in the delete path we can be sure that if there are no resolved
+	// resources then no resources were created.
+	if changed {
+		scope.Logger().V(6).Info("Machine resources updated, requeuing")
+		return ctrl.Result{}, nil
+	}
+
```
dulek, line 582: Oh, this is a nice pattern.

dulek (commented):
Bunch of questions inline, but in general it makes a lot of sense to me.

controllers/openstackmachine_controller.go
mdbooth: This still needs to live in the 'create transaction'. There's no point putting it before setting the finalizer because it won't be used. We also don't expect this to fail at this point because we've already checked the bootstrap data is set, so if it does fail it's a presumably just a low-probability ephemeral failure that could happen anywhere.

Here I've just moved it to be local to where the data is actually used.

controllers/openstackmachine_controller.go
mdbooth: This was always the goal of this pattern! It's only the upgrade case which make it at all complicated, and at some point we should be able to delete most of the complications.

controllers/openstackmachine_controller.go
mdbooth: Right. Notice that this is just code motion, though. I'm pretty sure it's broken in exactly the same way it was broken before, it's just slightly easier to see now.

If I haven't regressed this I don't want to fix it in this PR. I added the comment because I noticed it while I was moving it.

controllers/openstackcluster_controller.go
```diff
@@ -290,7 +290,18 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 		}
 	}
 
-	if instanceStatus != nil {
+	// If no instance was created we currently need to check for orphaned
+	// volumes. This requires resolving the instance spec.
+	// TODO: write volumes to status resources on creation so this is no longer required.
```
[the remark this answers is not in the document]
mdbooth: The bastion flow here is not as robust as the machine flow. However, this was the case before and I haven't tried as hard to fix it. The reason is that the bastion delete flow is extremely janky and over-complicated, and what I really want to do is re-design it.

As long as I didn't regress it here I'm ok with not fixing all the edge cases.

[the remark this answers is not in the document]
mdbooth: Incidentally, notice that previously because we only ran any of this when `instanceStatus != nil` we would always leak volume in this case. With this change we no longer leak volumes. Updating volumes to use the same pattern as ports will make this simpler, but that's for another time.

lentzi90 (commented):
Thank you for this!
I did a rerun of my previous test (where I accidentally used a non-existing externalNetworkID). Now it works as I expected. The cluster enters the Failed phase and can easily be deleted :slightly_smiling_face: 
/approve

controllers/openstackcluster_controller.go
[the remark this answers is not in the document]
dulek: I'm just commenting on a comment here, I don't think we'll be able to get rid of the `DeleteVolumes()` even after volumes are saved into the ID.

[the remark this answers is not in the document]
mdbooth: Well... kinda. I eventually want to move them into resources, and at that point they'll go into the Adopt() flow.

nguyenhuukhoi:
Thank you.

PR #2008, opened 2024-04-10
⚠️ Allow explicitly empty volume AZ

### LINKED ISSUE #1649: Compute AZs should not be used as Volume AZs

mdbooth:
Replaces an AvailabilityZone string for volumes with a VolumeAvailabilityZone struct which allows more flexibility in defaulting behaviour. Specifically it enables us to express both the current default behaviour where we take the volume AZ from the Machine, and a new default behaviour where to don't specify a volume AZ at all.

In making this change to both RootVolume and AdditionalBlockDevices we use common code for both APIs. This has the result of updating RootVolume to be consistent with a volume specified via AdditionalBlockDevices.

Fixes #1649

/hold


api/v1beta1/types.go
```diff
@@ -520,13 +523,42 @@ type BlockDeviceVolume struct {
...
+type VolumeAvailabilityZone struct {
+	// From specifies where we will obtain the availability zone for the
+	// volume. The options are "Name" and "Machine". If "Name" is specified
+	// then the Name field must also be specified. If "Machine" is specified
+	// the volume will use the value of FailureDomain, if any, from the
+	// associated Machine.
+	// +kubebuilder:default:=Name
 	// +optional
-	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	From VolumeAZSource `json:"from,omitempty"`
```
dulek, line 555: Looks a bit like an union to me? We probably want validation that if `from=Machine` you cannot set `Name`?

mdbooth: Yep, it's absolutely a discriminated union, although one with a hopefully ergonomic optional default.

The latest version has validation and a complete suite of tests.

mdbooth:
That test failure confirms that the new default behaviour works :)

mdbooth:
Running the e2e tests again because there's some non-determinism in there: if we weren't correctly setting the target AZ, they might land in the right place anyway by accident.

mdbooth:
The merge conflict is only in `instance_test.go`. The e2e-full-test passed before it was detected, so I won't bother running it again.

test/e2e/suites/apivalidations/openstackmachine_test.go
```diff
@@ -89,4 +95,187 @@ var _ = Describe("OpenStackMachine API validations", func() {
...
+		It("should not allow additional volume with name root", func() {
+			machine.Spec.AdditionalBlockDevices = []infrav1.AdditionalBlockDevice{
+				{
+					Name:    "root",
+					SizeGiB: 1,
+				},
+			}
+			Expect(k8sClient.Create(ctx, machine)).NotTo(Succeed(), "Creating a machine with a root named additional block device should fail")
+		})
+		*/
```
dulek, line 137: We keep this on `FIXME`?

dulek (changes requested):
Just one question.

test/e2e/suites/apivalidations/openstackmachine_test.go
mdbooth: I think so, yeah. Should probably create an issue to track it with a pointer to the FIXME code, though.

mdbooth: Done: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2013

api/v1beta1/types.go
```diff
@@ -520,13 +523,44 @@ type BlockDeviceVolume struct {
...
+	// associated Machine.
+	// +kubebuilder:default:=Name
+	// +optional
+	From VolumeAZSource `json:"from,omitempty"`
+
+	// Name is the name of a volume availability zone to use. It is required
+	// if From is "Name". The volume availability zone name may not contain
+	// spaces.
+	// +optional
+	Name *VolumeAZName `json:"name,omitempty"`
```
stephenfin, line 563: I can't put my finger on why, but this feels clunky. I had envisioned the "inherit from compute AZ" behaviour to be either (a) an anti-feature or (b) if actually required something that would be configured either on a deployment- or machine-wide basis. I've no real reason for this though, other than a general dislike of the whole thing, nor do I have an alternate proposal. As such I'm mainly leaving this note here so that I said _something_.

stephenfin (commented):
Only commenting on the API (though the additional test coverage looks good and is appreciated). As noted inline, I don't love the idea of perpetuating this myth that Compute and Block Storage AZs can be considered equivalent. However, it's a widely spread myth that clearly some users rely on so :shrug:

/lgtm

api/v1beta1/types.go
mdbooth: This doesn't preclude adding a machine or cluster-wide default override in the future. However, like you I'm not expecting the `from: Machine` behaviour to be widely used. This is also why I made the discriminator (the From field) optional with an explicit default. This means that the anticipated use case goes from:

```yaml
availabilityZone: foo
```
to
```yaml
availabilityZone:
  name: foo
```

which isn't awful, and gives us a big get-out-of-jail-free card to fix it later without breaking backwards compatibility when we understand it better.

dulek: IMO the situation gets better now, when the default behavior becomes set no AZ at all.

mdbooth:
> @mdbooth: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-cluster-api-provider-openstack-e2e-test	[9aa449c](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/9aa449c186b9cf1c5d2f4cb12bda0c622fe7e598)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2008/pull-cluster-api-provider-openstack-e2e-test/1778673770668494848)	unknown	`/test pull-cluster-api-provider-openstack-e2e-test`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=cluster-api-provider-openstack&pr=2008). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Amdbooth). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues?q=is:issue+is:open) when you hit one in your PR.

@lentzi90 This is worrying. This failure was a flake, and I'm reasonably confident we haven't broken it. However... I'm surprised it would merge a PR with a test failure. I'm assuming this additional test run is against the proposed merge commit, so I get why it runs. But why run it if it's going to merge anyway on failure?

lentzi90:
Oh yes this looks bad. I have no explanation for it. Could there be some higher level config in test-infra that runs it as a post-submit? I remember we removed our post-submits from [here](https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/cluster-api-provider-openstack), but perhaps there is some config telling prow to rerun after merge? :thinking: 
The job is definitely marked as required so I don't see how it could merge except by some bug in prow

liggitt:
https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=cluster-api-provider-openstack&pr=2008 shows a pass of that test at that commit. 

It got tested and merged in a batch merge

IMG_4725

Issue #2020, opened 2024-04-16
Improve the robustness of Bastion creation and deletion


mdbooth:
/kind feature

I've described this more fully in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2021

EmilienM:
I think this was done via OpenStackServer.

PR #2021, opened 2024-04-16
🌱 Add design doc for OpenStackServer

mdbooth:
Design proposal for https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2020

/hold

### DIFF
--- a/design/proposed/2020-openstackserver-controller.md
+++ b/design/proposed/2020-openstackserver-controller.md
@@ -1,0 +1,160 @@
+# Problem statement
+
+We currently have 2 callers of compute.CreateInstance: the machine controller for OpenStackMachine, and the cluster controller for the bastion[^mapo].
+
+[^mapo]: Incidentally, although it is not a primary concern of this project there is also a third consumer in OpenShift's [Machine API Provider OpenStack](https://github.com/openshift/machine-api-provider-openstack).
+
+Of these, the bastion has been most problematic. The principal current problem with the bastion is that, unlike OpenStackMachine, it is mutable because the bastion is not required to have the same lifecycle as the cluster which contains it. The current API allows the user to modify the bastion, and the controller is expected to delete the current one and create a new one. This is a problem in edge cases where the spec is required to determine if resources were previously created and need to be cleaned up, or not. This is a fundamental design problem, but another less fundamental but still very real problem is that trying to integrate the bastion into both the cluster and (effectively) machine creation and deletion flows creates fragile code and has resulted in numerous bugs.
+
+# Potential solutions
+
+## Do nothing
+
+The current solution has sufficed in practice for quite a long time. Additionally, we now have much better test coverage of bastion operations than we have had in the past, so we should hopefully catch issues sooner.
+
+The primary issue with this is the drag it places on bugfixes and new development, because the code flow of bastion deletion in particular can be hard to reason about.
+
+## Create an OpenStackMachine for the bastion
+
+This is an attractive solution because it better models the independent lifecycle of the bastion. When a new bastion is created we create a new OpenStackMachine. If the bastion spec is updated we delete the old one and create a new one, with the old and new objects having their own, independent state.
+
+The problem with this approach is that OpenStackMachine has other contracts. It cannot be created independently, but requires a CAPI Machine object. Other controllers have expectations of Machine objects, for example that they will eventually have a NodeRef.
+
+Creating a Machine/OpenStackMachine object for the bastion could negatively impact other CAPI users.
+
+## Create a new controller for the bastion
+
+This is essentially the same as using OpenStackMachine, except using a new CRD which can be independent of other CAPI objects.
+
+The obvious disadvantage is that it requires a new CRD and controller.
+
+# Proposal: An OpenStackServer CRD and controller
+
+## Philosophy
+
+We will create a new controller which is capable, in principal, of doing all of the OpenStack resource creation tasks common to bastion and OpenStackMachine creation. That is:
+
+* Resolve OpenStack resource parameters to specific resources (e.g. image, networks, subnets, security groups, server groups)
+* Create and delete ports
+* Create and delete volumes
+* Create and delete the server
+
+It will have no dependency on Machine or OpenStackCluster. So things it will explicitly not do include:
+
+* Attaching the Bastion floating IP
+* Creating a loadbalancer member
+* Adding a port for the default cluster network if no ports are provided
+* Referencing an AvailabilityZone from the Machine object
+* Adding default tags from the OpenStackCluster object
+
+## API
+
+Based on OpenStackMachineSpec, with modifications to accomodate the above.
+
+
+```go
+type OpenStackServerSpec struct {
+	AvailabilityZone       string
+	Flavor                 string
+	Image                  ImageParam
+	SSHKeyName             string
+	Ports                  []PortOpts
+	SecurityGroups         []SecurityGroupParam
+	Trunk                  bool
+	Tags                   []string
+	ServerMetadata         []ServerMetadata
+	ConfigDrive            *bool
+	RootVolume             *RootVolume
+	AdditionalBlockDevices []AdditionalBlockDevice
+	ServerGroup            *ServerGroupParam
+	IdentityRef            *OpenStackIdentityReference
+	FloatingIPPoolRef      *corev1.TypedLocalObjectReference
+	UserDataRef            *corev1.TypedLocalObjectReference
+}
+
+type OpenStackServerStatus struct {
+	InstanceID    optional.String
+	InstanceState *InstanceState
+	Addresses     []corev1.NodeAddress
+	Resolved      *ResolvedMachineSpec
+	Resources     *MachineResources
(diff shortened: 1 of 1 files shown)


mdbooth:
@lentzi90 Meta point: I recall you doing something similar before, but I can't find it to re-use it. I'm thinking of writing 2-4 of these in fairly quick succession, so it might be an opportunity to try to make something stick. If we have a better process already I'll switch to that instead.

Hoping to write them for:
* This (obvs)
* External endpoint controllers
* Spine-leaf deployments

The latter likely has tentacles and may turn in to more than one (e.g. separate doc for failure domains).

My thoughts on this are:
* Strictly for 'big' changes only, where 'big' is ill-defined.
* Numeric prefix is the ID of an associated feature request which may just be a stub
* We can and should merge to docs in `proposed` fairly freely without requiring consensus. This makes change tracking easier, and we can always delete it if we end up not planning to implement it after all.
* Consensus is required to move out of the `proposed` directory.

Purposes:
* Design review
* Discussion reference for users or external stakeholders (e.g. CAPI/OpenShift)
* Implementation guide
* A roadmap of sorts

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,145 @@
...
+* Referencing an AvailabilityZone from the Machine object
+* Adding default tags from the OpenStackCluster object
+  
+## API
+
+Based on OpenStackMachineSpec, with modifications to accomodate the above.
+
+
+```go
+type OpenStackServerSpec struct {
```
jichenjc, line 56: is there anything we need to do for existing customer to move to here? transparent or need define those ?

jichenjc (commented):
thanks for the great summary, it's really helpful
just one minor question

design/proposed/2020-openstackserver-controller.md
mdbooth: That's a great point. I should make it clear in the doc that this change will happen automatically and the end user does not need to take any action. I'll update the doc.

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,160 @@
...
+
+* ProviderID is not present: this is a Machine property
+* UserDataRef is added. If present it must exist.
+* AvailabilityZone is added, and refers explicitly
+* It is an error for Ports to be empty. Defaulting the cluster network must be done by the controller creating the object.
+* No managed security group will be added automatically. The managed security group must be added explicitly by the controller creating the object.
+* A Volume specified with `from: Machine` will use the value of `AvailabilityZone` instead of `Machine.FailureDomain`.
+
+It will set the following `Conditions`:
+* `Ready`: `True` when the server is fully provisioned. Once set to `True` will never be subsequently set to `False`.
```
lentzi90, line 110: It sounds strange to me to never change it back, even though we could have `Failed=True`. Is that intentional?
So we could have `Ready=True` and `Failed=True` at the same time :thinking:

lentzi90 (commented):
Looks good overall. I have no objections :slightly_smiling_face:

lentzi90:
The [previous thing](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1565) I did is not merged so that's probably why you didn't find it :smile: 
For that I used date-based naming because I just copied what [CAPI does](https://github.com/kubernetes-sigs/cluster-api/tree/main/docs/proposals). They also have a template if you want to steal it. I have no strong opinion on this though. :smile:

EmilienM:
Awesome proposal :+1:

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,160 @@
...
+	serverObj := fetch OpenStackServer object
+	if serverObj does not exist {
+		substitute tags and default network into bastion spec if required
+		create OpenStackServer object
+		return
+	}
+	if server spec does not match bastion.Spec {
+		deleteOpenStackServer(bastion)
+		return
+	}
```
dulek, line 134: Don't we need to modify the OpenStackCluster here too to force an event?

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,160 @@
...
+	Trunk                  bool
+	Tags                   []string
+	ServerMetadata         []ServerMetadata
+	ConfigDrive            *bool
+	RootVolume             *RootVolume
+	AdditionalBlockDevices []AdditionalBlockDevice
+	ServerGroup            *ServerGroupParam
+	IdentityRef            *OpenStackIdentityReference
+	FloatingIPPoolRef      *corev1.TypedLocalObjectReference
+	UserDataRef            *corev1.TypedLocalObjectReference
```
EmilienM, line 72: so you want a new object for this one? Any reason why not passing a string anymore?

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,160 @@
...
+* OpenStackServer reports Ready
+
+The end-user should not be required to take any action, and the upgrade process should not involve the deletion and recreation of existing OpenStack resources.
+
+### Notes
+
+* ProviderID is not present: this is a Machine property
+* UserDataRef is added. If present it must exist.
+* AvailabilityZone is added, and refers explicitly
+* It is an error for Ports to be empty. Defaulting the cluster network must be done by the controller creating the object.
```
EmilienM, line 105: for ports in the OpenStackServerSpec? Do we want the same API as `OpenStackMachineSpec` ?

EmilienM:
/lgtm
we can iterate later on the design but I think overall it's a solid start.

huxcrux:
I'm a bit late to the party however think the proposal sounds good :+1: 
/lgtm

mdbooth:
I'm going to move this around to conform to the pattern set by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1565

lentzi90:
Do we want to include a paragraph or two about if/how this could be used for [MachinePool support](https://cluster-api.sigs.k8s.io/tasks/experimental-features/machine-pools) in CAPO? From my understanding we would "just" need to add one higher level "pool" object to handle OpenStackServers in order to achieve that.

mdbooth:
> Do we want to include a paragraph or two about if/how this could be used for [MachinePool support](https://cluster-api.sigs.k8s.io/tasks/experimental-features/machine-pools) in CAPO? From my understanding we would "just" need to add one higher level "pool" object to handle OpenStackServers in order to achieve that.

What would this look like? I think MachinePool is for when the cloud has some efficiently-implemented native concept of a scalable set of machines. I don't think we have that in OpenStack?

lentzi90:
> > Do we want to include a paragraph or two about if/how this could be used for [MachinePool support](https://cluster-api.sigs.k8s.io/tasks/experimental-features/machine-pools) in CAPO? From my understanding we would "just" need to add one higher level "pool" object to handle OpenStackServers in order to achieve that.
> 
> What would this look like? I think MachinePool is for when the cloud has some efficiently-implemented native concept of a scalable set of machines. I don't think we have that in OpenStack?

Good point. I think it would be doable, but we would end up with all the logic in CAPO to manage the pool, which is not ideal. Forget that I mentioned it :sweat_smile:

design/proposed/2020-openstackserver-controller.md
```diff
@@ -0,0 +1,160 @@
...
+As the new API is non-trivial we should initially create it in v1alpha1.
+
+### Upgrading an existing deployment
+
+OpenStackServer would be a new API and does not affect any existing API.
+
+In the same way that OpenStackMachine currently has an 'adoption' phase where it will adopt existing resources, OpenStackServer should adopt matching resources which it would have created. On upgrade to a new version of CAPO which manages the bastion with an OpenStackServer object, I would expect the flow to be:
+
+* Cluster controller creates OpenStackServer and waits for it to report Ready
+* OpenStackServer controller looks for existing resources matching its given spec
```
EmilienM, line 94: This part is a bit tricky, we'll need to compare every field out of instanceStatus, I started to play with it and it became messy.
Not against this idea, just making sure we really want to go down that path.

The risk of not doing that is that an instance would be adopted by OpenStackServer that wouldn't be matching 100% the spec if the instance would have been created by the OpenStackServer controller.

mdbooth: I was thinking the adoption behaviour would be *identical* to what we currently do in the OpenStackMachine controller. Literally the same code. We're basically just adopting by name.

design/proposed/2020-openstackserver-controller.md
mdbooth: How about 3 Conditions:
- `Provisioned`: a latch which is never set to false after the server has been provisioned
- `Ready`: an ephemeral 'readiness' state. Can be false due to transient errors.
- `Failed`: permanent failure. No further reconciliation. Never set to false after being set to true.

EmilienM:
/hold cancel
/approve
we can merge the design since the implementation has merged.

EmilienM:
I'm expediting this but that doesn't mean we can't iterate on the design document that was initially proposed.

PR #2024, opened 2024-04-17
⚠️ Don't serve v1alpha6

mdbooth:
With the release of v0.10.0, v1alpha6 is no longer served or e2e tested.

/hold


MaysaMacedo (commented):
Should `OpenStackMachineTemplateList` also be an `unservedversion`?

mdbooth:
Upgrade job passed

/hold cancel

PR #2035, opened 2024-04-24
✨ Add NamePrefix field to OpenStackMachine.Spec

### LINKED ISSUE #2039: Name prefix for OpenStackMachine related resources
/kind feature

- As a security/network operator, I want CAPO resources to be easily identifiable by their name and hostname, so that I can keep track of and verify security policies between environments.
- As a cluster admin, I want control over how CAPO resources are named, so that they can fit in with external resources living in the same OpenStack project.

To address these user stories I propose a new field for the OpenStackMachine (and consequently for the OpenStackMachineTemplate): `.spec.namePrefix`. It would be used as a prefix for all OpenStack resources created by CAPO for this OpenStackMachine.
The field would be optional. If not specified, the current behavior would remain, where resources gets their name based on the OpenStackMachine name.

Why not just name the OpenStackMachine differently?
- It is created and named by CAPI, so we do not have direct control over it.
- It is just implicit that we name the resources after it, so it cannot really be relied upon (unless we make it explicit).
- The OpenStackMachine gets its name from the MachineDeployment or (Kubeadm)ControlPlane. These must be replaced in order to change the naming scheme (and it is anyway not part of the contract). For the KCP that means recreating the cluster.

Why not address this at the CAPI level?
- There is no contract that binds CAPO to name the resources in a certain way, so it does not help to get control over the naming of the OpenStackMachine from CAPI.
- There is an issue (https://github.com/kubernetes-sigs/cluster-api/issues/7030) that may address this. However, making it part of the contract will affect all implementers and probably be very hard to get consensus on. Not to mention that each provider has their own limitations and rules from their infrastructure that may affect what is possible to do.

This feature is something that we didn't notice that we needed until the [breaking change in CAPI v1.7.0](https://github.com/kubernetes-sigs/cluster-api/pull/9833). We had gotten used to the old (and admittedly confusing) behavior where infra machines were named by the infra templates. CAPO (and probably other providers as well) named the resources according to the infra machine and it was then very easy to change the names by switching infra template. Now we are stuck with fixed names based on the KCP and MDs, which cannot be changed. For new clusters, we can of course set the proper names where it matters, but that doesn't address the core issue here: we need control over the naming of the resources. With this feature request, I'm trying to address that so that we can have explicit control over it.

lentzi90:

This adds a new field to the spec of OpenStackMachines: `namePrefix`. It is an optional field that can be specified as a way to influence how OpenStack resources are named. If it is not specified, the OpenStackMachine name will be used (current behavior).

Fixes #2039, related to https://github.com/kubernetes-sigs/cluster-api/issues/10463.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1beta1/openstackmachine_types.go
+++ b/api/v1beta1/openstackmachine_types.go
@@ -98,6 +98,13 @@ type OpenStackMachineSpec struct {
 	// will be assigned to the OpenStackMachine.
 	// +optional
 	FloatingIPPoolRef *corev1.TypedLocalObjectReference `json:"floatingIPPoolRef,omitempty"`
+
+	// NamePrefix is an optional string that will be used as a prefix for the name of OpenStack
+	// resources created based on this OpenStackMachine. A random suffix will be added to the prefix.
+	// If omitted, the name of the OpenStackMachine will be used as name.
+	// +kubebuilder:validation:MaxLength:=255
+	// +optional
+	NamePrefix string `json:"namePrefix,omitempty"`
 }
 
 type ServerMetadata struct {
--- a/api/v1alpha6/openstackmachine_conversion.go
+++ b/api/v1alpha6/openstackmachine_conversion.go
@@ -190,6 +190,8 @@ func restorev1beta1MachineSpec(previous *infrav1.OpenStackMachineSpec, dst *infr
 	dst.ServerGroup = previous.ServerGroup
 	dst.Image = previous.Image
 	dst.FloatingIPPoolRef = previous.FloatingIPPoolRef
+	// NamePrefix is new in v1beta1
+	dst.NamePrefix = previous.NamePrefix
 
 	if len(dst.SecurityGroups) == len(previous.SecurityGroups) {
 		for i := range dst.SecurityGroups {
--- a/api/v1alpha6/zz_generated.conversion.go
+++ b/api/v1alpha6/zz_generated.conversion.go
@@ -1266,6 +1266,7 @@ func autoConvert_v1beta1_OpenStackMachineSpec_To_v1alpha6_OpenStackMachineSpec(i
 		out.IdentityRef = nil
 	}
 	// WARNING: in.FloatingIPPoolRef requires manual conversion: does not exist in peer-type
+	// WARNING: in.NamePrefix requires manual conversion: does not exist in peer-type
 	return nil
 }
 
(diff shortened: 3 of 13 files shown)


mdbooth:
Unfortunately I don't have time to go into this deeply today, but my first impression is that I'm not keen.

It's worth noting that I'm generally supportive of the new CAPI default. It has always been a trip hazard that the Machine and InfraMachine have different names, and making them the same should be simpler for most people.

However, as you point out in the CAPI issue, it has been this way long enough that I think it's reasonable to treat it like a breaking change in practise, regardless of whether it was documented. I hope we can work with the CAPI folks to sort that out.

That said, this change would introduce that same trip hazard in CAPO, just moved one level down. The idea that cloud resources are named after the OpenStackMachine which owns them is simple and elegant. I'm not keen to break it, especially now that we've released v1beta1 and would be left supporting it indefinitely.

My strong preference is to work with CAPI to enable backwards-compatibility for the inadvertent breakage. I took a brief look and the code to support the previous behaviour is still there. I think this can and should be implemented by adding a backwards-compatibility flag (optional, defaults to the new behaviour) to CAPI.

api/v1beta1/openstackmachine_types.go
```diff
@@ -98,6 +98,13 @@ type OpenStackMachineSpec struct {
 	// will be assigned to the OpenStackMachine.
 	// +optional
 	FloatingIPPoolRef *corev1.TypedLocalObjectReference `json:"floatingIPPoolRef,omitempty"`
+
+	// NamePrefix is an optional string that will be used as a prefix for the name of OpenStack
+	// resources created based on this OpenStackMachine. A random suffix will be added to the prefix.
+	// If omitted, the name of the OpenStackMachine will be used as name.
+	// +kubebuilder:validation:MaxLength:=255
```
lentzi90, line 105: Can we do better here? Are there any specific requirements on names in that we could include OpenStack?

lentzi90:
Sure, there is no hurry with this. However, https://github.com/kubernetes-sigs/cluster-api/pull/10511 will only provide a temporary solution. It is targeting only the release branch and will be gone by CAPI v1.8

lentzi90:
This is not needed. There is work on-going for a long term fix in CAPI.

Issue #2034, opened 2024-04-23
Static analyser gets confused about minimum TLS version


mdbooth:
/kind bug

~~The OpenShift build system is complaining at us because `tls-min-version` defaults to TLS 1.2, which has a bunch of known issues. I recommend we default it to TLS 1.3, same as the maximum. In all likelihood nobody is ever going to need to update this: what kube-apiserver isn't going to support TLS 1.3 these days?~~

~~If anybody did really need this there's an option so it remains possible to allow it.~~

Added in #1867

This wasn't the problem at all. The issue was that the static analyser saw that there was a code path (specifically the error path 🤦) where the variable which was used to set the TLS version was set to zero. Lets try to write something not too horrible which makes it easier for a static analyser to verify that tls version can't be set to any unacceptable value.

cc @tuminoid

tuminoid:
Golang since 1.13 (properly since 1.14) has supported TLS 1.3, so some 4 years ago. That'd be taken in k8s 1.19 or so, long out of support by now. Any reasonable kubernetes installation that would take in new CAPO that has TLS 1.3 as default is zero. And you can configure it back to TLS 1.2 as you said.

OTOH, since kube-apiserver is TLS 1.3 enabled and CAPO is TLS 1.3 enabled, they will always negotiate a TLS 1.3 between them. Forcing TLS 1.3 as minimum affects only scenario where malicious user talks to CAPO webhook, and only accepts TLS 1.2. Not exactly sure what the risk here is but I also don't see what the harm would be in bumping the minimum.

Hence, 👍 from me. Would you like me to push a PR for this?

tuminoid:
No, we need to keep the flags, just change the default to min to 1.3, so it can be configured to accept 1.2 if needed.

Before changing it, it would be great to hear why OpenShift is flagging it as insecure. L391 is setting the min version based on config options, which explicitly warn if user enables insecure ciphers that are not enabled by default for TLS1.2 and do not allow TLS 1.1 which would actually be deprecated.

MaysaMacedo:
@tuminoid So, in theory only this should have worked right?
https://github.com/openshift/cluster-api-provider-openstack/pull/305/commits/480ac8bca46b95a40bd4b15b2ba5b66cc8b32d23

tuminoid:
[...]

PR #2037, opened 2024-04-25
:seedling: Refactoring: never assign unacceptable TLS versions

pierreprinetti:

Our downstream security scan is confused by `GetTLSVersion` returning `0` as a value (even if coupled by a non-nil error), which could end up being assigned to the same identifier that (in a non-error context) would set the TLS version.

This patch makes security linting easier by never setting a TLS version outside v1.2 or v1.3, even in case of an unacceptable user input.

This is expected to be a pure refactoring. Please reject this patch if it introduces any change in behaviour.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

Fixes: #2034

### DIFF
--- a/main.go
+++ b/main.go
@@ -370,14 +370,19 @@ func concurrency(c int) controller.Options {
 func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error) {
 	var tlsOptions []func(config *tls.Config)
 
-	tlsMinVersion, err := GetTLSVersion(options.TLSMinVersion)
-	if err != nil {
-		return nil, err
-	}
-
-	tlsMaxVersion, err := GetTLSVersion(options.TLSMaxVersion)
-	if err != nil {
-		return nil, err
+	// To make a static analyzer happy, this block ensures there is no code
+	// path that sets a TLS version outside the acceptable values, even in
+	// case of unexpected user input.
+	var tlsMinVersion, tlsMaxVersion uint16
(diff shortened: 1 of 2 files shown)


mdbooth:
Thanks. Can you please do a couple of things for me before merging?
* Mark this as fixing https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2034 in the description.
* Add a comment in the affected function which explains why the code is a bit weird. Something like: "To make a static analyzer happy, ensure there is no code path...". That'll be a bit more discoverable than the commit message.

pierreprinetti:
> * To make a static analyzer happy, ensure there is no code path

done. PTAL

mdbooth:
/approve

Thank you!

pierreprinetti:
/hold cancel
I think this is ready for a review.

pierreprinetti:
/hold
good that I fixed the tests

EmilienM:
tests still failing. Happy to LGTM once they're fixed.

pierreprinetti:
/hold cancel
:crossed_fingers:

mdbooth:
Unit tests pass, and this PR doesn't do anything which should affect the e2e tests.

/lgtm

EmilienM:
@pierreprinetti do we want/need it in release-0.10?

pierreprinetti:
> @pierreprinetti do we want/need it in release-0.10?

that'd be swell.

In the meantime, the bug has been validated by Snyk and reportedly passed on to their engineering.

Issue #2042, opened 2024-04-30
CAPO Panics when adding ManagedSecurityGroups


grosenba:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
https://kubernetes.slack.com/archives/CFKJB65G9/p1714481726268869?thread_ts=1714372906.792849&cid=CFKJB65G9

`2024/04/30 12:53:40 http: panic serving 100.90.194.192:35079: runtime error: invalid memory address or nil pointer dereference`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

```
2024/04/30 12:53:40 http: panic serving 100.90.194.192:35079: runtime error: invalid memory address or nil pointer dereference
goroutine 5384333 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1898 +0xbe
panic({0x1db33e0?, 0x3646f90?})
	/usr/local/go/src/runtime/panic.go:770 +0x132
sigs.k8s.io/cluster-api-provider-openstack/pkg/webhooks.(*openStackClusterWebhook).ValidateUpdate(0xc0005b0ac0?, {0xc001a62000?, 0xfee?}, {0x2415f10?, 0xc00142f908?}, {0x2415f10, 0xc00142ec88})
	/workspace/pkg/webhooks/openstackcluster_webhook.go:124 +0x3bf
sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*validatorForType).Handle(_, {_, _}, {{{0xc000e3ff80, 0x24}, {{0xc0018c8de0, 0x1f}, {0xc001776560, 0x7}, {0xc001776570, ...}}, ...}})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission/validator_custom.go:102 +0x70b
sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).Handle(_, {_, _}, {{{0xc000e3ff80, 0x24}, {{0xc0018c8de0, 0x1f}, {0xc001776560, 0x7}, {0xc001776570, ...}}, ...}})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission/webhook.go:169 +0x1ee
[...]

PR #2043, opened 2024-04-30
🐛 Fix webhook panic when adding managed security groups

mdbooth:
Also adds fuzz tests which would have caught this and any similar issues.

Fixes: #2042 

/hold


EmilienM:
/lgtm
Thanks for the fix and tests!

mdbooth:
Pushed an update to fix a couple of lint errors.

PR #2049, opened 2024-05-01
🌱 port: no dependency on OpenStackMachine

EmilienM:

It's some work we're preparing to build a new controller in charge of
creating servers and their dependencies (ports, etc). We don't want it
to depent on the OpenStack Machine object.


mdbooth (approved):
That's easier than I'd have expected. I suspect (and hope) that eventually some different order will emerge from this list of arguments.

Issue #2063, opened 2024-05-06
CAPO controller manager panics while creating port


kashifest:
/kind bug

Hi, we are seeing  a invalid memory address or nil pointer dereference while creating a port in openstack with v0.10.1 release

Here is the stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1b74415]
 
goroutine 359 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:116 +0x1e5
panic({0x1db56a0?, 0x3649f90?})
        /usr/local/go/src/runtime/panic.go:770 +0x132
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking.(*Service).CreatePort(0xc00049c678, {0x2418ad0, 0xc0009b1408}, 0xc00041a2c8)
        /workspace/pkg/cloud/services/networking/port.go:195 +0xd55
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/networking.(*Service).CreatePorts(0xc00049c678, {0x2418ad0, 0xc0009b1408}, {0xc00041a2c8, 0x3, 0x1b9c3a5?}, 0xc000867980)
        /workspace/pkg/cloud/services/networking/port.go:336 +0x66
sigs.k8s.io/cluster-api-provider-openstack/controllers.getOrCreateMachinePorts(0xc000e09a40?, 0x24303d8?)
        /workspace/controllers/openstackmachine_controller.go:759 +0x59
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackMachineReconciler).reconcileNormal(0xc00036c180, {0x24303d8, 0xc0008f7530}, 0xc000e09a40, {0xc00037ede0, 0x26}, 0xc0009da288, 0xc0009b1b88, 0xc0009b1408)
        /workspace/controllers/openstackmachine_controller.go:602 +0x307
[...]

PR #2064, opened 2024-05-06
Fix nil pointer issue while creating port

kashifest:
Fixes# https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2063 

/hold

### DIFF
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -192,7 +192,7 @@ func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.Resol
 
 	port, err := s.client.CreatePort(builder)
 	if err != nil {
-		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", port.Name, err)
+		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", portSpec.Name, err)
 		return nil, err
 	}
 


pkg/cloud/services/networking/port.go
```diff
@@ -192,7 +192,7 @@ func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.Resol
 
 	port, err := s.client.CreatePort(builder)
 	if err != nil {
-		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", port.Name, err)
+		record.Warnf(eventObject, "FailedCreatePort", "Failed to create port %s: %v", portSpec.Name, err)
```
EmilienM, line 195: I'm fine with this change because:

* `port` is indeed a pointer so if `CreatePort` returns nil, we can't use it and we'll have to check for it here before.
* We already use `portSpec.Name` in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/349ec37fd8b3ed15a6750f4b3506993d8d287cc4/pkg/cloud/services/networking/port.go#L201-L202

lentzi90 (commented):
Makes sense!
/approve

kashifest:
I realized that I opened the PR towards release branch, corrected the base to main now, so lgtm is lost @EmilienM

kashifest:
I will push a unit test for this in a followup

PR #2071, opened 2024-05-08
🌱 Change AdoptPorts to not be Machine dependent

EmilienM:

Allow to re-use `AdoptPorts` function without the Machine API.
This is useful work for the `OpenStackServer` effort where we need this
function but don't want to rely on the MachineResources.


EmilienM:
This will fail in tests, I'm not sure why yet.

PR #2081, opened 2024-05-15
:bug: Fix sub-ports not deleted with trunks

mquhuy:
Sometimes, when a trunk is associated with sub ports, deletion of the trunk will not lead to deletion of the subports, which is not expected. This PR adds the check to ensure the deletion of the subports in such cases.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90 (commented):
Thanks for this, it looks pretty good!
Please `make lint-update` to fix the linting error.

pkg/clients/networking.go
```diff
@@ -238,6 +239,40 @@ func (c networkClient) DeleteTrunk(id string) error {
...
+      if len(subports) < 1 {
+              return nil
+      }
+
+      var portList []trunks.RemoveSubport
+      for _, element := range subports {
+              var vport trunks.RemoveSubport
+              vport.PortID = element.PortID
+              portList = append(portList, vport)
+      }
```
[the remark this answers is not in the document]
mquhuy: Thank you very much! That's a great suggestion

pkg/clients/networking.go
```diff
@@ -238,6 +239,40 @@ func (c networkClient) DeleteTrunk(id string) error {
 	return mc.ObserveRequestIgnoreNotFound(trunks.Delete(c.serviceClient, id).ExtractErr())
 }
 
+func (c networkClient) RemoveTrunkSubPorts(id string) error {
```
mdbooth, line 242: This would be a significant change of pattern for `clients`. So far we've confined clients to being an extremely thin wrapper round a gophercloud call, essentially so we can mock them.

The disadvantage of this is that we can't mock it. Would you mind moving this logic into trunk.go and calling existing client methods instead?

I'm not completely opposed to going this way, btw. The 'interfaces' we have in `cloud/services/*` are ill-defined at best, chaotic at worst.  However, if we're going to clean it up I'd prefer to do it deliberately.

pkg/cloud/services/networking/trunk.go
```diff
@@ -88,6 +89,29 @@ func (s *Service) DeleteTrunk(eventObject runtime.Object, portID string) error {
...
+			return false, err
+		}
+		return true, nil
+	})
+	if err != nil {
+		record.Warnf(eventObject, "FailedRemoveTrunkSubPorts", "Failed to delete sub ports trunk %s with id %s: %v", trunkInfo[0].Name, trunkInfo[0].ID, err)
+		return err
+	}
+
+	record.Eventf(eventObject, "SuccessfulRemoveTrunkSubPorts", "Removed trunk sub ports %s with id %s", trunkInfo[0].Name, trunkInfo[0].ID)
```
mdbooth, line 114: I note that we'll always generate this event whether we removed any subports or not, but... why have a trunk if you're not going to add subports? I guess this is fine.

mdbooth (changes requested):
Thanks, looks good. Just a request to move some code for consistency.

mquhuy:
Thank you @mdbooth! I have to go now, but I'll make the change asap tomorrow.

pkg/cloud/services/networking/trunk.go
```diff
@@ -77,6 +79,44 @@ func (s *Service) getOrCreateTrunkForPort(eventObject runtime.Object, port *port
...
+		Subports: portList,
+	}
+
+	err = s.client.RemoveSubports(trunkID, removeSubportsOpts)
+	if mc.ObserveRequest(err) != nil {
+		return err
+	}
+
+	for _, subPort := range subports {
+		err := s.client.DeletePort(subPort.PortID)
```
mdbooth, line 111: Just a thought: does deleting the subport also remove it from the trunk? It normally works that way in other contexts, but I haven't checked. If so we could potentially skip the RemoveSubports call above.

mdbooth:
e2e test failures look like an infra-flake. test failures are most likely missing mock calls for the new delete subports flow.

Incidentally, I didn't check but I was assuming that because this is specifically in the Trunk delete flow we've already checked that trunks are expected, and therefore we're not making unnecessary OpenStack API calls for the majority of folks who aren't using trunks.

If this is the case, and the failing tests don't really need trunks, it might be simpler to update the test input object to disable trunks rather than adding trunk calls to the expected mocks.

pkg/cloud/services/networking/trunk.go
mquhuy: I haven't thought about it, but that sounds reasonable

pkg/cloud/services/networking/trunk.go
```diff
@@ -77,6 +79,44 @@ func (s *Service) getOrCreateTrunkForPort(eventObject runtime.Object, port *port
...
+	mc := metrics.NewMetricPrometheusContext("trunk", "deletesubports")
+	subports, err := s.client.ListTrunkSubports(trunkID)
+
+	if mc.ObserveRequestIgnoreNotFound(err) != nil {
+		return err
+	}
+
+	if len(subports) < 1 {
+		return nil
+	}
```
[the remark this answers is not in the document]
mquhuy: I don't think there was any special reason, I'll change it to == 0 then.

pkg/cloud/services/networking/trunk.go
mquhuy: Hi @mdbooth, I have asked around and gave this some thought, and imo it is a good practice to keep the explicit deletion of subports from the trunk. It's possible that even after the ports are deleted, the trunk still has reference to them, hence it'd mess up, at least from the log perspectives. Also, it's more graceful that we first disconnect the trunk from the subports, then remove the sub ports, than directly removing them. Wdyt?

mdbooth: WFM.

mdbooth:
I used this PR as a test for junit formatting if you want to look at something a bit prettier: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2095/pull-cluster-api-provider-openstack-test/1792498916650913792

pkg/cloud/services/networking/trunk.go
```diff
@@ -77,6 +79,44 @@ func (s *Service) getOrCreateTrunkForPort(eventObject runtime.Object, port *port
 	return trunk, nil
 }
 
+func (s *Service) RemoveTrunkSubports(trunkID string) error {
+	mc := metrics.NewMetricPrometheusContext("trunk", "deletesubports")
```
mdbooth, line 83: We really need a metrics strategy. I'd probably leave this out, tbh, but in absence of any related policy I'm not going to object.

mdbooth (approved):
I'm happy to merge this, but I recommend considering how best to test it. Some unit tests covering error cases might be one way. I think (but I haven't checked) we create a port with a trunk in one of the e2e tests. Perhaps we could also add a subport to it, which would cause the test to fail if it isn't cleaned up properly. Just some ideas.

pkg/cloud/services/networking/trunk.go
lentzi90: It looks to me like this is a duplicate. We are already getting the metrics in `networking.go` above.
Can you remove this @mquhuy ?

lentzi90 (commented):
This looks good, except the duplicate metrics

pkg/cloud/services/networking/trunk.go
mquhuy: Yes, sure. Coming up

mquhuy:
> I'm happy to merge this, but I recommend considering how best to test it. Some unit tests covering error cases might be one way. I think (but I haven't checked) we create a port with a trunk in one of the e2e tests. Perhaps we could also add a subport to it, which would cause the test to fail if it isn't cleaned up properly. Just some ideas.

Thank you. I'm wondering if this is the one you were thinking of? https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/port_test.go#L309

I'm not sure if I understand it correctly, but it seems to me the test is only checking if the port could be created. Do you think we should also test if it can be cleaned? If it turned out to be in a bigger scope, I'd be happy to do it in another PR.

lentzi90 (commented):
/lgtm

mdbooth:
> > I'm happy to merge this, but I recommend considering how best to test it. Some unit tests covering error cases might be one way. I think (but I haven't checked) we create a port with a trunk in one of the e2e tests. Perhaps we could also add a subport to it, which would cause the test to fail if it isn't cleaned up properly. Just some ideas.
> 
> Thank you. I'm wondering if this is the one you were thinking of? https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/port_test.go#L309
> 
> I'm not sure if I understand it correctly, but it seems to me the test is only checking if the port could be created. Do you think we should also test if it can be cleaned? If it turned out to be in a bigger scope, I'd be happy to do it in another PR.

Looks like that's only testing port creation. You might want a new unit test for the delete flow. But something similar to that.

For the e2e test I was thinking of this: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/046d6b7c9c7df3f6a514dd452d65922ea7156c92/test/e2e/suites/e2e/e2e_test.go#L447-L455
If we updated the test to add a subport to that trunk we'd presumably get an error if the test didn't clean it up properly.

PR #2086, opened 2024-05-16
🐛 Handle errors returned by GetInstanceStatusByName in machine controller

### LINKED ISSUE #2085: Errors returned by GetInstanceStatusByName should be handled in machine controller.
/kind bug

In a deployment that was running a capo controller with [that fix](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753), we've observed the bug that was supposed to be fixed

The controller was attempting to recreate an existing instance, whereas the openstackMachine had a valid instanceID in its spec:

```                                                                             
I0513 08:01:02.327471       1 openstackmachine_controller.go:473] "Machine does not exist, creating Machine" controller="openstackmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackMachine" OpenStackMachine="wc-rim5/rim5-cluster-md-md0-2fd17ee6b9-qd9kl" namespace="wc-rim5" reconcileID="b5aca506-9b68-4bae-af47-3666f3de3784" openStackMachine="rim5-cluster-md-md0-2fd17ee6b9-qd9kl" machine="rim5-cluster-md0-x2q8n-85dzt" cluster="rim5-cluster" openStackCluster="rim5-cluster" name="free5g6-cluster-md-md0-7f29028a04-2khtx"
```                                                                             

It seems to happen in specific conditions, where nova API always returns 404 errors (because of a mis-configured revese proxy by example)
                                                                                
In that case GetInstanceStatus will return `(nil, nil)` [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6d1b899574b6a0d986182ba6df40c77529c3030a/controllers/openstackmachine_controller.go#L770)                                   

And the error returned by GetInstanceStatusByName will not be handled [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6d1b899574b6a0d986182ba6df40c77529c3030a/controllers/openstackmachine_controller.go#L779) (my bad)
                                                                                
Consequently controller will attempts to create a new instance... Errors returned by GetInstanceStatusByName should be checked at that point.      

**Environment:**

- Cluster API Provider OpenStack version: [75ffe73f88544dbc26c6636a491bca5b3f63c3d4](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/75ffe73f88544dbc26c6636a491bca5b3f63c3d4)

zioc:

Errors returned by GetInstanceStatusByName were ignored up to now, which could lead controller to attempt to recreate a server under specific circumstances (for example if nova API always returns 404 errors - see detailled explanation in related issue)

These errors shouldn't be ignored, otherwise we may consider that server does not exists whereas nova api is misbehaving.

Fixes #2085

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -776,17 +776,22 @@ func (r *OpenStackMachineReconciler) getOrCreateInstance(logger logr.Logger, ope
 	}
 	if instanceStatus == nil {
 		// Check if there is an existing instance with machine name, in case where instance ID would not have been stored in machine status
-		if instanceStatus, err = computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name); err == nil {
-			if instanceStatus != nil {
-				return instanceStatus, nil
-			}
-			if openStackMachine.Status.InstanceID != nil {
-				logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
-				conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
-				openStackMachine.SetFailure(capierrors.UpdateMachineError, errors.New("virtual machine no longer exists"))
-				return nil, nil
-			}
+		instanceStatus, err = computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
+		if err != nil {
+			logger.Info("Unable to get OpenStack instance by name", "name", openStackMachine.Name)
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceCreateFailedReason, clusterv1.ConditionSeverityError, err.Error())
+			return nil, err
 		}
+		if instanceStatus != nil {
+			return instanceStatus, nil
+		}
+		if openStackMachine.Status.InstanceID != nil {
+			logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
+			conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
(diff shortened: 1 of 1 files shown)


mdbooth (approved):
This is an excellent robustification, thanks!

huxcrux:
/lgtm

Awesome fix 🙏

controllers/openstackmachine_controller.go
```diff
@@ -776,17 +776,22 @@ func (r *OpenStackMachineReconciler) getOrCreateInstance(logger logr.Logger, ope
...
-			if instanceStatus != nil {
-				return instanceStatus, nil
-			}
-			if openStackMachine.Status.InstanceID != nil {
-				logger.Info("Not reconciling machine in failed state. The previously existing OpenStack instance is no longer available")
-				conditions.MarkFalse(openStackMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "virtual machine no longer exists")
-				openStackMachine.SetFailure(capierrors.UpdateMachineError, errors.New("virtual machine no longer exists"))
-				return nil, nil
-			}
+		instanceStatus, err = computeService.GetInstanceStatusByName(openStackMachine, openStackMachine.Name)
```
EmilienM, line 779: One thing I thought you would fix too is:
https://github.com/zioc/cluster-api-provider-openstack/blob/821a1a2ef25ac615db5fb26379eb0c4b947ad284/pkg/cloud/services/compute/instance.go#L827-L829

I don't like the fact we don't return an error if more than one server with the same name was found. This is error prone and can lead to cluster issues.

EmilienM: cc @mdbooth for an opinion on ^

mdbooth: Good point, and we have an existing pattern for it.

EmilienM: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2088

PR #2096, opened 2024-05-22
🌱 Add script for API diff check

### LINKED ISSUE #1860: CI: check apidiff once v1beta1 is released
/kind feature

Once we release CAPO v1beta1, our API should remain stable.

One way to check that in CI is to use github.com/joelanford/go-apidiff as it's already done in other providers.
We could add a job that will run apidiff against PR and return an error if the API change is major and not minor.

EmilienM:

* Add target to Makefile to run `make apidiff`
* Add the script itself, copied from other platforms

Fixes #1860


EmilienM:
mhh I need to find a way to skip it for `v1alpha1` stuffs.

EmilienM:
We agreed the following:
* Add a non voting job to run apidiff
* If the job fails, ignore failures if it's because of a change in our go api until we figure things out with our codebase.

mdbooth:
/approve

Wouldn't it be nice if we had a tool like Zuul which could test both this PR and https://github.com/kubernetes/test-infra/pull/32644 together?

PR #2099, opened 2024-05-24
:bug: if the openstackcluster was ready, we don't want to set a terminalError

### LINKED ISSUE #2097: Cluster stays in Failed state after Credentials rotations
/kind bug

The Cluster object stays in failed state after a credentials problem is resolved.

      failureMessage: |
          Failure detected from referenced resource infrastructure.cluster.x-k8s.io/v1beta1, Kind=OpenStackCluster with name "rotate-creds-cnrwq": failed to reconcile external network: failed to get external network: Unable to re-authenticate: Expected HTTP response code [200] when accessing [GET https://url/v2.0/networks/54258498-a513-47da-9369-1a644e4be692], but got 401 instead
          {"error": {"code": 401, "title": "Unauthorized", "message": "The request you have made requires authentication."}}: Resource not found: [POST https://url/v3/auth/tokens], error message: {"error":{"code":404,"message":"Could not find Application Credential: OLD ID.","title":"Not Found"}}

Delete used appcreds and wait for failure to accour
Create new app creds and update all resources
Verify that all actions are possible again, on mgmt cluster or in cluster

Cluster state is good again after fix

Looks like Creds errors are handelt as a terminal error and not as a transient error. 
Seems to be like a problem we had with the Machines too

Code for the machine problem as an idea how to fix: [https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7a19fb6f038[…]a31df228877a446b9b99/controllers/openstackmachine_controller.go](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7a19fb6f038ece8702dea31df228877a446b9b99/controllers/openstackmachine_controller.go#L651-L660)

Code in Openstack cluster controller with the problem: [https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7a19fb6f038[…]a31df228877a446b9b99/controllers/openstackcluster_controller.go](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7a19fb6f038ece8702dea31df228877a446b9b99/controllers/openstackcluster_controller.go#L914-L918)

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v010.3
- Cluster-API version: v1.7.3

qeqar:
f.e if the creds break, but the cluster was ready, the operator can fix it.

Use transient Error if the Openstackcluster was once ready.

Fixes #2097

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -175,7 +175,7 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope 
 		}
 
 		if err = loadBalancerService.DeleteLoadBalancer(openStackCluster, clusterResourceName); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete load balancer: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete load balancer: %w", err), false)
 			return reconcile.Result{}, fmt.Errorf("failed to delete load balancer: %w", err)
 		}
 	}
@@ -183,23 +183,23 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope 
 	// if ManagedSubnets was not set, no network was created.
 	if len(openStackCluster.Spec.ManagedSubnets) > 0 {
 		if err = networkingService.DeleteRouter(openStackCluster, clusterResourceName); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete router: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete router: %w", err), false)
 			return ctrl.Result{}, fmt.Errorf("failed to delete router: %w", err)
 		}
 
 		if err = networkingService.DeleteClusterPorts(openStackCluster); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete ports: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete ports: %w", err), false)
 			return reconcile.Result{}, fmt.Errorf("failed to delete ports: %w", err)
 		}
 
 		if err = networkingService.DeleteNetwork(openStackCluster, clusterResourceName); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete network: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete network: %w", err), false)
 			return ctrl.Result{}, fmt.Errorf("failed to delete network: %w", err)
 		}
 	}
 
 	if err = networkingService.DeleteSecurityGroups(openStackCluster, clusterResourceName); err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete security groups: %w", err))
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete security groups: %w", err), false)
 		return reconcile.Result{}, fmt.Errorf("failed to delete security groups: %w", err)
 	}
 
@@ -271,7 +271,7 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 
 	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
 		if err = networkingService.DeleteFloatingIP(openStackCluster, openStackCluster.Status.Bastion.FloatingIP); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err), false)
 			return fmt.Errorf("failed to delete floating IP: %w", err)
 		}
 	}
@@ -311,14 +311,14 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 			if address.Type == corev1.NodeExternalIP {
 				// Floating IP may not have properly saved in bastion status (thus not deleted above), delete any remaining floating IP
 				if err = networkingService.DeleteFloatingIP(openStackCluster, address.Address); err != nil {
-					handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err))
+					handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err), false)
 					return fmt.Errorf("failed to delete floating IP: %w", err)
 				}
 			}
 		}
 
 		if err = computeService.DeleteInstance(openStackCluster, instanceStatus); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete bastion: %w", err))
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete bastion: %w", err), false)
 			return fmt.Errorf("failed to delete bastion: %w", err)
 		}
 	}
@@ -330,7 +330,7 @@ func deleteBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openStac
 		}
 		for _, port := range bastionStatus.Resources.Ports {
 			if err := networkingService.DeleteInstanceTrunkAndPort(openStackCluster, port, trunkSupported); err != nil {
-				handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete port: %w", err))
+				handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete port: %w", err), false)
 				return fmt.Errorf("failed to delete port: %w", err)
 			}
 		}
@@ -459,7 +459,7 @@ func reconcileBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openS
 
 	err = getOrCreateBastionPorts(openStackCluster, networkingService)
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create ports for bastion: %w", err))
+		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to get or create ports for bastion: %w", err), false)
 		return nil, fmt.Errorf("failed to get or create ports for bastion: %w", err)
 	}
 	bastionPortIDs := GetPortIDs(openStackCluster.Status.Bastion.Resources.Ports)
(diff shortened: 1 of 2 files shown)


EmilienM:
What seems wrong to me, is that if a user changes the bastion spec with a new flavor, that new flavor doesn't exist in Nova, the controller fails to create the bastion and wants to update `FailureReason`, it won't because cluster was `Ready`.
I'm wondering if we should rather set Readiness to false into this function as well?

qeqar:
As this is fixable by fixing the flavor name in the CR, it seems to be the same type of error. So it should not be permanently too.

Setting ready to false is done in no case at the moment, so this would change the behavior, if that is ok, i can put it in the else path, but then i need so find the proper place(s) to set it true again.

EmilienM:
I'll ask @lentzi90 and @mdbooth what they think.

qeqar:
After some additional thinking, would setting ready to false, removes the validly of my if clause.

So first: it need to be set on every error, as a permanent error is more critical an should result in an ready false and second if we have two problems, f.e. some error and we activate the bastion host to have look and then do a flavor typo, it will go directly to a permanent error, as ready was already false.

If we start setting ready to false after it was set to true, we need to find a different switch between permanent and transient errors.

mdbooth:
I've asked about setting `Ready` to `false` here: https://kubernetes.slack.com/archives/C8TSNPY4T/p1716892079243969

controllers/openstackcluster_controller.go
```diff
@@ -912,9 +912,11 @@ func (r *OpenStackClusterReconciler) SetupWithManager(ctx context.Context, mgr c
 }
 
 func handleUpdateOSCError(openstackCluster *infrav1.OpenStackCluster, message error) {
-	err := capierrors.UpdateClusterError
-	openstackCluster.Status.FailureReason = &err
-	openstackCluster.Status.FailureMessage = ptr.To(message.Error())
+	if !openstackCluster.Status.Ready {
```
mdbooth, line 915: I'd prefer to not call this method for a transient error in the first place. This would prevent anything from being fatal after the cluster is ready, which probably isn't right either.

qeqar: Ok moved the ready check outside of the function. 

But there a 4 place i am not sure what to do:

fist case:  
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/controllers/openstackcluster_controller.go#L815-L822
Here some one added a terminal error to the function, but inside the function it never returns with terminalError true

the second case: 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/controllers/openstackcluster_controller.go#L815-L822
In this function someone selected exactly one error already, but did not check against ready. 

In both cases i am not sure if we need to add the ready state check on top?

controllers/openstackcluster_controller.go
```diff
@@ -731,7 +769,9 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
 	if openStackCluster.Spec.Network != nil {
 		network, err := networkingService.GetNetworkByParam(openStackCluster.Spec.Network)
 		if err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find network: %w", err))
+			if !openStackCluster.Status.Ready {
+				handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find network: %w", err))
+			}
```
mdbooth, line 774: Ditto

controllers/openstackcluster_controller.go
```diff
@@ -775,17 +815,24 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
 func reconcileProvisionedNetworkComponents(networkingService *networking.Service, openStackCluster *infrav1.OpenStackCluster, clusterResourceName string) error {
 	err := networkingService.ReconcileNetwork(openStackCluster, clusterResourceName)
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile network: %w", err))
+		if !openStackCluster.Status.Ready {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile network: %w", err))
+		}
```
mdbooth, line 820: Ditto

controllers/openstackcluster_controller.go
```diff
@@ -775,17 +815,24 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
...
 		return fmt.Errorf("failed to reconcile network: %w", err)
 	}
 	err = networkingService.ReconcileSubnet(openStackCluster, clusterResourceName)
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile subnets: %w", err))
+
+		if !openStackCluster.Status.Ready {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile subnets: %w", err))
+		}
```
mdbooth, line 828: Ditto

controllers/openstackcluster_controller.go
```diff
@@ -775,17 +815,24 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
...
 		return fmt.Errorf("failed to reconcile subnets: %w", err)
 	}
 	err = networkingService.ReconcileRouter(openStackCluster, clusterResourceName)
 	if err != nil {
-		handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile router: %w", err))
+		if !openStackCluster.Status.Ready {
+			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to reconcile router: %w", err))
+		}
```
mdbooth, line 835: Ditto

controllers/openstackcluster_controller.go
```diff
@@ -839,7 +886,9 @@ func reconcileControlPlaneEndpoint(scope *scope.WithLogger, networkingService *n
 	case !ptr.Deref(openStackCluster.Spec.DisableAPIServerFloatingIP, false):
 		fp, err := networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterResourceName, openStackCluster.Spec.APIServerFloatingIP)
 		if err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("floating IP cannot be got or created: %w", err))
+			if !openStackCluster.Status.Ready {
+				handleUpdateOSCError(openStackCluster, fmt.Errorf("floating IP cannot be got or created: %w", err))
+			}
```
mdbooth, line 891: Never terminal.

controllers/openstackcluster_controller.go
```diff
@@ -854,7 +903,9 @@ func reconcileControlPlaneEndpoint(scope *scope.WithLogger, networkingService *n
 	// Control plane endpoint is not set, and none can be created
 	default:
 		err := fmt.Errorf("unable to determine control plane endpoint")
-		handleUpdateOSCError(openStackCluster, err)
+		if !openStackCluster.Status.Ready {
+			handleUpdateOSCError(openStackCluster, err)
+		}
```
mdbooth, line 908: Terminal if cluster is not ready.

controllers/openstackcluster_controller.go
```diff
@@ -174,31 +174,41 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope
 		}
 
 		if err = loadBalancerService.DeleteLoadBalancer(openStackCluster, clusterResourceName); err != nil {
-			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete load balancer: %w", err))
+			if !openStackCluster.Status.Ready {
+				handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete load balancer: %w", err))
+			}
```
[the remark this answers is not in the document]
EmilienM: Don't we want it stored in events though?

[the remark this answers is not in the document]
EmilienM: Ah nevermind this is not for events. This is the last error.
I agree this can't be the last error because other resources underneath can't be removed either if this one wasn't (hopefully my reasoning makes sense).

qeqar:
Will have a look on Monday, if i have some time, then i am out for the rest of the week, so could take some time.

controllers/openstackcluster_controller.go
qeqar: ok that is done

controllers/openstackcluster_controller.go
```diff
@@ -503,7 +523,9 @@ func reconcileBastion(scope *scope.WithLogger, cluster *clusterv1.Cluster, openS
 	port, err := computeService.GetManagementPort(openStackCluster, instanceStatus)
 	if err != nil {
 		err = fmt.Errorf("getting management port for bastion: %w", err)
-		handleUpdateOSCError(openStackCluster, err)
+		if !openStackCluster.Status.Ready {
+			handleUpdateOSCError(openStackCluster, err)
+		}
```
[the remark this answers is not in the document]
qeqar: Not sure what to do here? The function is called in the cluster and the machine controller?

so should i rewrite it, that it is terminal in that sub case, or delete the complete check in the cluster controller?

lentzi90 (commented):
Great!
/lgtm

We should probably think about adding conditions for some of the non-fatal errors at some point also

qeqar:
Ok, i am not fully happy with setting the `isFatal` inside the function call, specially the true case. But should be ok for this one.

controllers/openstackcluster_controller.go
```diff
@@ -643,7 +671,9 @@ func resolveLoadBalancerNetwork(openStackCluster *infrav1.OpenStackCluster, netw
 		if lbSpec.Network != nil {
 			lbNet, err := networkingService.GetNetworkByParam(lbSpec.Network)
 			if err != nil {
-				handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find loadbalancer network: %w", err))
+				if !openStackCluster.Status.Ready {
+					handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to find loadbalancer network: %w", err))
+				}
```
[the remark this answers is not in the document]
qeqar: added both error cases as terminal

qeqar:
Ok now re-based with main to get the #2109 error class changes and all occurrences updated

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -82,6 +82,9 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 
 	lb, err := s.getOrCreateAPILoadBalancer(openStackCluster, clusterResourceName)
 	if err != nil {
+		if errors.Is(capoerrors.ErrFilterMatch, err) {
```
lentzi90, line 85: Should this be the other way around?
```suggestion
		if errors.Is(err, capoerrors.ErrFilterMatch) {
```

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -107,6 +110,9 @@ func (s *Service) ReconcileLoadBalancer(openStackCluster *infrav1.OpenStackClust
 
 		fp, err := s.networkingService.GetOrCreateFloatingIP(openStackCluster, openStackCluster, clusterResourceName, floatingIPAddress)
 		if err != nil {
+			if errors.Is(capoerrors.ErrFilterMatch, err) {
```
lentzi90, line 113: Same here

lentzi90 (commented):
Could you please squash the commits into one and check my question below

lentzi90 (commented):
/approve

mdbooth (commented):
/lgtm

Thank you!

EmilienM:
Excellent contribution, thank you!
/hold cancel

PR #2101, opened 2024-05-27
templates: fix clusterclass-dev

EmilienM:

The image API wasn't updated.


lentzi90 (commented):
/approve
Thanks for this! We should really have an e2e test that makes use of this...

PR #2109, opened 2024-06-03
⚠️ More than one instance found is now an error

### LINKED ISSUE #2088: `GetInstanceStatusByName` doesn't return an error when > 1 instance if found
/kind bug

`GetInstanceStatusByName` doesn't return an error when > 1 instance if found and this can lead to cluster problems if we have two instances with the same name. Could be a leftover or a staled resource, that needs to be removed before creating a new one.

An error should be returned.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): latest

EmilienM:

* Move errors functions/vars into capoerrors, so they can easily be
  re-used.
* Now we return an error if multiple instances with the same name were
  found (and not just a warning). There is no safe course of action here.

Fixes #2088


mdbooth:
/approve

I've updated the title because we should highlight this as a breaking change when it's released. It's possible that somebody was previously getting lucky with the old behaviour who will be broken now we've fixed it.

PR #2112, opened 2024-06-04
🐛 instance: re-add error event if server creationg returned an error

### LINKED ISSUE #2111: `FailedCreateServer` error event was removed while it's still being tested
In this [PR](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753), `FailedCreateServer` error event was removed while it's still being tested: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/a63ce561e12131f7853f04a731b3a5ddb1cefa86/test/e2e/suites/e2e/e2e_test.go#L616

I have no idea how our CI didn't catch that :shrug: but for sure I'll take a look.

_Originally posted by @EmilienM in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1753#issuecomment-2148243396_

EmilienM:

When creating a server, Nova can return an error immediately after the
call was made to create it or later (which would be seen when showing the server status with a different API call).

If it's returned immediately, we would like to restore the error event
that we used to have and that we actually test in our e2e when providing
a wrong availability zone.

Fixes #2111

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -119,6 +119,7 @@ func (s *Service) createInstanceImpl(eventObject runtime.Object, instanceSpec *I
 		KeyName:           instanceSpec.SSHKeyName,
 	})
 	if err != nil {
+		record.Warnf(eventObject, "FailedCreateServer", "Failed to create server %s: %v", instanceSpec.Name, err)
 		return nil, fmt.Errorf("error creating Openstack instance: %v", err)
 	}
 


MaysaMacedo:
/lgtm

Looks like the [event was removed only when](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753/files#diff-5e7dbfd52305a21a11fb94fab201062ae2c7b851db93c2a8443a1d82d7aa7936L351) pulling the server status failed and [not as soon as the server is created](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1753/files#diff-5e7dbfd52305a21a11fb94fab201062ae2c7b851db93c2a8443a1d82d7aa7936R333). But in any case the proposed PR is valid.

lentzi90 (commented):
/approve

PR #2114, opened 2024-06-05
🐛 Remove bastion security group when disabling the bastion

### LINKED ISSUE #2113: Bastion security group is not deleted when removing a cluster that had bastion and then disabled
/kind bug

If you have deployed a cluster with managed security groups that has a bastion, then we have a security group created for the bastion.
If the user disables the bastion, the bastion server and ports get deleted but not the security group.
The security group gets removed from the cluster status though.

When removing the cluster, the bastion security group is not deleted and remained a staled resource. 

We should have deleted the bastion security group in the first place when the bastion was disabled.

EmilienM:

We reconcile the security groups before the bastion, because the bastion
needs its security group to be created first when managed security groups are enabled.
When the bastion is disabled, we will try to delete the security group if it exists.
In the first attempt, the security group will still be in-use by the bastion instance
but then the bastion instance will be deleted in the next reconcile loop.
We do that here because we don't want to manage the bastion security group from
elsewhere, that could cause infinite loops between ReconcileSecurityGroups and ReconcileBastion.
Therefore we try to delete the bastion security group as a best effort here
and also when the cluster is deleted so we're sure it will be deleted at some point.

Also, we're trying to remove it when the cluster is deleted in case it
wasn't done before. This doesn't trigger an error if the security group
didn't exist.

Adding e2e tests to cover the scenarios:
* Disabling the bastion should reduce the total number of managed SGs to 2.
* Re-enabling it should make it to 3 SGs.

Fixes #2113

### DIFF
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -56,9 +56,24 @@ func (s *Service) ReconcileSecurityGroups(openStackCluster *infrav1.OpenStackClu
 		workerSuffix:       secWorkerGroupName,
 	}
 
+	secBastionGroupName := getSecBastionGroupName(clusterResourceName)
 	if bastionEnabled {
-		secBastionGroupName := getSecBastionGroupName(clusterResourceName)
 		suffixToNameMap[bastionSuffix] = secBastionGroupName
+	} else {
+		// We reconcile the security groups before the bastion, because the bastion
+		// needs its security group to be created first when managed security groups are enabled.
+		// When the bastion is disabled, we will try to delete the security group if it exists.
+		// In the first attempt, the security group will still be in-use by the bastion instance
+		// but then the bastion instance will be deleted in the next reconcile loop.
+		// We do that here because we don't want to manage the bastion security group from
+		// elsewhere, that could cause infinite loops between ReconCileSecurityGroups and ReconcileBastion.
+		// Therefore we try to delete the bastion security group as a best effort here
+		// and also when the cluster is deleted so we're sure it will be deleted at some point.
+		// https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2113
+		if err := s.deleteSecurityGroup(openStackCluster, secBastionGroupName); err != nil {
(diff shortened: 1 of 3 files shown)


EmilienM:
/hold
I need to fix unit tests

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -56,9 +56,24 @@ func (s *Service) ReconcileSecurityGroups(openStackCluster *infrav1.OpenStackClu
...
+		// needs its security group to be created first when managed security groups are enabled.
+		// When the bastion is disabled, we will try to delete the security group if it exists.
+		// In the first attempt, the security group will still be in-use by the bastion instance
+		// but then the bastion instance will be deleted in the next reconcile loop.
+		// We do that here because we don't want to manage the bastion security group from
+		// elsewhere, that could cause infinite loops between ReconCileSecurityGroups and ReconcileBastion.
+		// Therefore we try to delete the bastion security group as a best effort here
+		// and also when the cluster is deleted so we're sure it will be deleted at some point.
+		// https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2113
+		if err := s.deleteSecurityGroup(openStackCluster, secBastionGroupName); err != nil {
```
MaysaMacedo, line 73: I was looking for ways to avoid often calling neutron in every reconciliation, but I don't think there is any way how to check if the sg should be retrieved and deleted as the sg reference is not present on the cluster obj anymore. Do you see any possible way?

EmilienM: I can't think on a way to not make these calls, unless we expose the risk of race conditions if we would do it once early in the function, to get all SGs, and later in the function consume its data, but seconds could happen between the calls... Not idea either but worth considering the risk anyway, since the deletion isn't fatal and creation would fail if it exists already.

mdbooth: My only slight worry here is that we're potentially short-cutting everything else that this reconciliation does for something we're expecting to fail sometimes.

I've tried to think through all the ordering scenarios and I suspect this is ok. However, I wonder if it would have been safer to add this delete step at the bottom of the reconcile immediately before setting BastionSecurityGroup to nil.

PR #2119, opened 2024-06-08
🌱 e2e: add bastion test case when spec is changed

EmilienM:

We now have a test case where we change the spec of the bastion and
check all is fine.


EmilienM:
This doesn't work right now but will work with my openstackserver pr

I'll cherry pick

/close

PR #2122, opened 2024-06-11
Reconcile with wait after LB cleanup

### LINKED ISSUE #2121: cluster delete with LB: network is removed too soon
/kind bug

Currently, if the loadbalancer is enabled and someone tries to remove the cluster, they'll likely have this error:

```"Failed to delete network k8s-clusterapi-cluster-e2e-whn9ya-cluster-e2e-whn9ya with id 0e5e0e93-5784-4b22-9f7c-0e8d6bfc1337: Expected HTTP response code [202 204] when accessing [DELETE http://10.0.3.15:9696/networking/v2.0/networks/0e5e0e93-5784-4b22-9f7c-0e8d6bfc1337], but got 409 instead\n{\"NeutronError\": {\"type\": \"NetworkInUse\", \"message\": \"Unable to complete operation on network 0e5e0e93-5784-4b22-9f7c-0e8d6bfc1337. There are one or more ports still in use on the network, id for these ports is: de51d5a2-5e16-4873-8e46-a0637d4bc5bc.\", \"detail\": \"\"}}" logger="events" type="Warning" object={"kind":"OpenStackCluster","namespace":"e2e-whn9ya","name":"cluster-e2e-whn9ya","uid":"7071e56c-d34e-4014-8ee4-2edead2623b0","apiVersion":"infrastructure.cluster.x-k8s.io/v1beta1","resourceVersion":"2021"} reason="Faileddeletenetwork"
controller.go:329] "Reconciler error" err=<
failed to delete network: Expected HTTP response code [202 204] when accessing [DELETE http://10.0.3.15:9696/networking/v2.0/networks/0e5e0e93-5784-4b22-9f7c-0e8d6bfc1337], but got 409 instead
{"NeutronError": {"type": "NetworkInUse", "message": "Unable to complete operation on network 0e5e0e93-5784-4b22-9f7c-0e8d6bfc1337. There are one or more ports still in use on the network, id for these ports is: de51d5a2-5e16-4873-8e46-a0637d4bc5bc.", "detail": ""}}
 > controller="openstackcluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackCluster" OpenStackCluster="e2e-whn9ya/cluster-e2e-whn9ya" namespace="e2e-whn9ya" name="cluster-e2e-whn9ya" reconcileID="4155da75-0205-4d52-91ef-9d77fb76979f"
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2099/pull-cluster-api-provider-openstack-e2e-test/1799085067255943168/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-7d957cb6cd-kndmz/manager.log

It's ok because we reconcile later and eventually the ports managed by Octavia are getting removed by Octavia but we could reconcile with a wait to let Octavia a chance to remove the ports itself and avoid this error message which can be confused to our users.

No errors in the logs.

EmilienM:

When deleting the LB, Octavia will then proceed to the management ports
cleanup managed by the provider itself, not CAPO.

The problem is that in CAPO we continue in the reconcile and try to
remove the network but Octavia didn't have a chance yet to remove the
ports so the reconcile loop fails and it retries again later and then is
capable of cleaning up the networks.

Let's be gentle and give time to Octavia so it can remove the ports
itself and reduce the number of false alarms in the logs.

Fixes #2121


EmilienM:
/hold
I want to confirm we don't see the error in the logs.

EmilienM:
/hold cancel
I checked the logs and no more error of that type.

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2122/pull-cluster-api-provider-openstack-e2e-test/1800607078713135104/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-7d957cb6cd-5dgc2/manager.log

EmilienM:
a timeout removing the cluster :/

EmilienM:
/hold
something's wrong I'll investigate

controllers/openstackcluster_controller.go
```diff
@@ -178,6 +179,9 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope
 			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete load balancer: %w", err), false)
 			return reconcile.Result{}, fmt.Errorf("failed to delete load balancer: %w", err)
 		}
+		// Once a loadbalancer is deleted, we need to wait for the Octavia management ports to be cleaned up
+		// by Octavia before we can proceed with the deletion of the network components.
+		return reconcile.Result{RequeueAfter: waitForOctaviaPortsCleanup}, nil
```
mdbooth, line 184: If it's more efficient we could poll here for a short period of time before:
* returning with an error?
* requeuing?

EmilienM: I'm trying to tackle this problem from a different angle; I'll report back my results.

EmilienM:
so my PR fixed the PENDING_DELETE issue. Now i'm going to tackle the management port cleanup.

controllers/openstackcluster_controller.go
EmilienM: i've pushed a separate commit, please re-review.

EmilienM:
This doesn't work in CI, at least it didn't show the new logs. I'll investigate in my local cluster at some point.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -658,32 +665,86 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
...
+
+	lbPortsExist, err := s.checkIfLbPortsExist(openStackCluster)
+	if err != nil {
+		return nil, err
 	}
 
 	if lb == nil {
-		return nil
+		if lbPortsExist {
+			s.scope.Logger().Info("Load balancer ports still exist, waiting for them to be deleted", "name", loadBalancerName)
```
EmilienM, line 719: Looking at the [logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2122/pull-cluster-api-provider-openstack-e2e-test/1800982517210157056/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-7d957cb6cd-dhmz7/manager.log), it seems like we never hit that so I wonder if the [requeue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2122/files#diff-29cb3ac9349ca04685d8b48d0b1bc13a90e76242d0300fea7ed14f4f71ef2eecR769) at the end of the function has given enough time to perform the port cleanup on the Octavia side.

EmilienM:
@mdbooth ready for review when time permits. Maybe over-engineered but it does the job apparently. Feedback is open.

EmilienM:
@mdbooth what to do with this one? It's been open for a long time.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -698,11 +759,14 @@ func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster,
...
-		return err
+		return nil, err
 	}
 
 	record.Eventf(openStackCluster, "SuccessfulDeleteLoadBalancer", "Deleted load balancer %s with id %s", lb.Name, lb.ID)
-	return nil
+
+	// If we have reached this point, that means that the load balancer wasn't initially deleted but the request to delete it didn't return an error.
+	// So we want to requeue until the load balancer and its associated ports are actually deleted.
+	return &ctrl.Result{RequeueAfter: waitForOctaviaLBCleanup}, nil
```
mdbooth, line 769: This is the fix right here. We previously returned nil here and therefore continued while the LB was still in the process of being deleted. With this change we schedule another reconcile, and that reconcile won't allow us to procede until we have observed that the LB no longer exists.

pkg/cloud/services/loadbalancer/loadbalancer.go
mdbooth: See below. This isn't a case of 'enough time'; we're now waiting until the LB is gone, however long that takes. We don't need this check.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -658,32 +665,86 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 	return nil
 }
 
-func (s *Service) DeleteLoadBalancer(openStackCluster *infrav1.OpenStackCluster, clusterResourceName string) error {
+func (s *Service) checkIfLbPortsExist(openStackCluster *infrav1.OpenStackCluster) (bool, error) {
```
mdbooth, line 668: See below: we don't need this any more. In the context of the other fix (waiting until the LB is gone) this results in unnecessary API calls. It also depends on internal Octavia implementation details which aren't part of the API, so it adds additional future failure modes without any benefit. I'd very much prefer we removed this and the new function in `port.go`.

mdbooth (changes requested):
We should remove the unnecessary check, but apart from that this is good to go imho.

mdbooth (commented):
/approve
/lgtm

mdbooth (approved):

PR #2128, opened 2024-06-18
🐛 Expose NodePorts on cluster network instead of 0.0.0.0/0

### LINKED ISSUE #2127: Nodeports open from 0.0.0.0/0 by default, causing unexpected behaviour together with Floating IPs
/kind bug

Nodeports are by default open from 0.0.0.0/0. This has historically not been a problem since nodes previously had an internal IP and was nat:ed over the Openstack router.

However when using Floating IPs this causes nodeports to be exposed from internet. This could lead to some unexpected behaviour. And sadly most end-users does not configure network policies properly meaning some services by mistake are accessible from the internet.

Since nodeport rules are added by default and there is no way of opting out this behaviour cannot be changed by the cluster admin.

Rules are added here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/securitygroups.go#L203
And defined here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/networking/securitygroups_rules.go#L144C1-L165C1

I asked in slack if anyone was aware of any reason for nodeports to be open from 0.0.0.0/0 and got a reply saying it was due to conformance tests.

I think the best case of action would be to only allow nodeports from cluster subnet(s).

Nodeports to not be open from internet without explicitly configuring it.

We may also want to change how rules for the API alongside AdditionalPorts exposed over the API Loadbalancer. However I would say nodeports are more "critical" since I see no reason in having Floating IPs on control-plane node at this time.

I'm currently running a conformance test with the following "change" applied to se if it passes: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/1685c65c303dbcf4bae2959705793c3133bd97fd

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.10.3
- Cluster-API version: 1.7.3
- OpenStack version: 
- Minikube/KIND version:

huxcrux:

This PR changes NodePorts to only be open from cluster network.

Fixes #2127

I assume this is something that would be in 0.11 or 1.0? If so where do we put the changelog?

I also need to make sure what IP ranged we potentially want to use since there is multiple ways of specifying them depending on of they are operator managed or not.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -200,7 +200,7 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
 	workerRules := append([]resolvedSecurityGroupRuleSpec{}, defaultRules...)
 
 	controlPlaneRules = append(controlPlaneRules, getSGControlPlaneHTTPS()...)
-	workerRules = append(workerRules, getSGWorkerNodePort()...)
+	workerRules = append(workerRules, getSGWorkerNodePort(openStackCluster.Spec.ManagedSubnets[0].CIDR)...)
 
 	// If we set additional ports to LB, we need create secgroup rules those ports, this apply to controlPlaneRules only
 	if openStackCluster.Spec.APIServerLoadBalancer.IsEnabled() {
--- a/pkg/cloud/services/networking/securitygroups_rules.go
+++ b/pkg/cloud/services/networking/securitygroups_rules.go
@@ -142,7 +142,7 @@ func getSGControlPlaneHTTPS() []resolvedSecurityGroupRuleSpec {
 }
 
 // Allow all traffic, including from outside the cluster, to access node port services.
-func getSGWorkerNodePort() []resolvedSecurityGroupRuleSpec {
+func getSGWorkerNodePort(cidr string) []resolvedSecurityGroupRuleSpec {
 	return []resolvedSecurityGroupRuleSpec{
 		{
 			Description:  "Node Port Services",
@@ -151,6 +151,7 @@ func getSGWorkerNodePort() []resolvedSecurityGroupRuleSpec {
 			PortRangeMin: 30000,
 			PortRangeMax: 32767,
 			Protocol:     "tcp",
+			RemoteIPPrefix: cidr,
 		},
 		{
 			Description:  "Node Port Services",
@@ -159,6 +160,7 @@ func getSGWorkerNodePort() []resolvedSecurityGroupRuleSpec {
 			PortRangeMin: 30000,
 			PortRangeMax: 32767,
 			Protocol:     "udp",
+			RemoteIPPrefix: cidr,
 		},
 	}
 }


pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +201,15 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
 	controlPlaneRules = append(controlPlaneRules, getSGControlPlaneHTTPS()...)
-	workerRules = append(workerRules, getSGWorkerNodePort()...)
+
+	// Fetch subnet to use for worker node port rules
+	// TODO: add case where subnets are used. However need a way of resolving the subnet to get it's CIDR
+	if openStackCluster.Spec.ManagedSubnets != nil && len(openStackCluster.Spec.ManagedSubnets) > 0 {
+		SubnetCIDR = openStackCluster.Spec.ManagedSubnets[0].CIDR
+	} else {
+		SubnetCIDR = "0.0.0.0/0"
+	}
```
mdbooth, line 211: Can we not use remote group for this? This won't work on deployments which aren't using managed subnets.

huxcrux: I just changed the behaviour to make use of status.Network.Subnet instead to support use cases where bring your own network/subnet is used.

I do not think we could use remote group due to load balancers talking over nodeports?
Load Balancers are created under the "service" project in default openstack meaning it is not added to any security groups inside the project where the cluster is running.

huxcrux: I also left a question in the code that may be good to bring up here.

If no IPv4 subnet is found the code currently defaults to `0.0.0.0/0` to preserve current behaviour.

We could however return an error if no IPv4 subnet is found or potentially fallback to just allowing nodes by using remote groups (this will however break octavia load balancers created using the cloud provider which feels kinda bad).

mdbooth: In other places we just add the rule twice with 2 remote groups, one for the control plane and one for workers. This covers machines even when they aren't using the default cluster network. Is there a reason we can't do that here?

I'd like to reduce our reliance on the cluster's status.Network over time. There are new, interesting deployments that it doesn't play nicely with, e.g. external control planes and spine/leaf.

huxcrux: > In other places we just add the rule twice with 2 remote groups, one for the control plane and one for workers. This covers machines even when they aren't using the default cluster network. Is there a reason we can't do that here?
> 
> I'd like to reduce our reliance on the cluster's status.Network over time. There are new, interesting deployments that it doesn't play nicely with, e.g. external control planes and spine/leaf.

It would break load balancers deployed using the cloud provider (octavia) since they are not in any security group we control (know the ID of). I think it is a really common use case and support out of the box would be nice

huxcrux: Just wanted to follow up on this.

We could add security groups meaning all cluster nodes can contact each other and it will make the conformance test pass. However it would cause LBs created with the cloud provider to fail. I'm not at this point of any way of allowing LBs outside allowing IPs directly towards nodes. 

I also think having a default where LBs "work" is preferable to have a default that works "out of the box".

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -169,6 +170,7 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
 	var secControlPlaneGroupID string
 	var secWorkerGroupID string
 	var secBastionGroupID string
+	var SubnetCIDR string
```
mdbooth, line 173:
```suggestion
	var subnetCIDR string
```

pkg/cloud/services/networking/securitygroups_rules.go
```diff
@@ -142,23 +142,67 @@ func getSGControlPlaneHTTPS() []resolvedSecurityGroupRuleSpec {
...
+			EtherType:     "IPv4",
+			PortRangeMin:  30000,
+			PortRangeMax:  32767,
+			Protocol:      "udp",
+			RemoteGroupID: secControlPlaneGroupID,
+		},
+	}
+}
+
+// Allow all traffic, including from outside the cluster, to access node port services.
```
mdbooth, line 186:
```suggestion
// Allow all traffic from a specific CIDR to access node port services.
```

pkg/cloud/services/networking/securitygroups_rules.go
```diff
@@ -142,23 +142,67 @@ func getSGControlPlaneHTTPS() []resolvedSecurityGroupRuleSpec {
...
+func getSGWorkerNodePortCidr(cidr string) []resolvedSecurityGroupRuleSpec {
```
mdbooth, line 187:
```suggestion
func getSGWorkerNodePortCIDR(cidr string) []resolvedSecurityGroupRuleSpec {
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +202,29 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
 	workerRules := append([]resolvedSecurityGroupRuleSpec{}, defaultRules...)
 
 	controlPlaneRules = append(controlPlaneRules, getSGControlPlaneHTTPS()...)
-	workerRules = append(workerRules, getSGWorkerNodePort()...)
+
+	// Fetch subnet to use for worker node port rules
+	// In the future IPv6 support need to be added here
```
mdbooth, line 207: IPv4-only is icky, but everything else is still IPv4 only and it's not worth fixing it piecemeal so this is fine.

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +202,29 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+	if openStackCluster.Status.Network != nil && len(openStackCluster.Status.Network.Subnets) > 0 {
```
mdbooth, line 208: Iterating over a nil or empty list is safe and idiomatic. The loop below performs this check implicitly anyway.

```suggestion
	if openStackCluster.Status.Network != nil {
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +202,29 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+		for _, subnet := range openStackCluster.Status.Network.Subnets {
+			// Check uf subnet.CIDR is ipv4 subnet
```
mdbooth, line 210:
```suggestion
			// Check if subnet.CIDR is ipv4 subnet
```

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +202,29 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+			_, ipnet, err := net.ParseCIDR(subnet.CIDR)
+			if err != nil {
+				return nil, fmt.Errorf("invalid subnet found during security groups reconcile: %v", err)
+			}
+			if ipnet.IP.To4() != nil {
+				SubnetCIDR = subnet.CIDR
+				break
+			}
```
mdbooth, line 218: This whole block can be replaced with `netutils.IsIPv4CIDRString(subnet.CIDR)`

pkg/cloud/services/networking/securitygroups.go
```diff
@@ -200,7 +202,29 @@ func (s *Service) generateDesiredSecGroups(openStackCluster *infrav1.OpenStackCl
...
+		}
+	}
+	if SubnetCIDR != "" {
+		// If SubnetCIDR is found we allow tcp and udp traffic from said SubnetCIDR, this in order to allow octavia Loadbalancers created by CCM to function properly
+		workerRules = append(workerRules, getSGWorkerNodePortCidr(SubnetCIDR)...)
```
mdbooth, line 223: You could move this inside the block above and also remove the break so you simply add any listed IPv4 subnet. Any other restrictions we have on subnets aren't relevant here, so might as well make it handle whatever's there.

mdbooth:
Please can you also add a code comment in there about why we're explicitly adding this subnet? Namely because this is the default subnet used by CPO as the origin of traffic from Octavia load balancers to load balancer members? Otherwise it risks being deleted by somebody because it looks redundant, even me in 6 months when I've forgotten this discussion again 🙄

mdbooth:
Just needs to be squashed and I'm happy to approve this.

skaramicke:
Well shit. Does it need fixing before merging?

mdbooth:
> Well shit. Does it need fixing before merging?

Sounds like it, yes.

bilbobrovall:
> Well shit. Does it need fixing before merging?

The easiest way is probably getting a patch from `git diff <upstream>/main --patch >nodeports.patch` and if that looks correct, `git reset HEAD~1 --hard`, then rebase, apply patch and commit/push

skaramicke:
I'll do it with Bilbo's assistance first thing tomorrow morning.

mikaelgron:
@bilbobrovall here's what I did if it shines any light on it:
```
git fetch
git reset $(git merge-base main HEAD)
git add -A
git commit -m "Expose NodePorts on cluster network instead of 0
.0.0.0/0"
git push --force
```

`git merge-base main HEAD` yields `6560f8882a2aa7ece3d13d47f2f2badbcba348c3`.

mikaelgron:
I was thinking for a moment that I should've used `git merge-base upstream/main HEAD` instead, but that gives the same ref.

mikaelgron:
@mdbooth Fixed!

PR #2132, opened 2024-06-24
🌱 IdentityRefProvider: DRY obtaining OpenStackIdentityRef

mdbooth:
As we add more resource controllers the pattern of adding another
NewClientScopeFromFoo to ScopeFactory is becoming more cumbersome. This
allows them all to use a common interface.

Summary of changes
* Addition of `infrav1.IdentityRefProvider`
* Implementations of IdentityRefProvider for OpenStackMachine, OpenStackCluster, and OpenStackFloatingIPPool
* Replace `NewClientScopeFrom*` in `scope.Factory` with a single `NewClientScopeScopeFromObject` which takes a list of IdentityRefProvider

The reason it takes a list rather than a single object is so OpenStackMachine can continue to implement a fallback where it will use the IdentityRef from OpenStackCluster if it is not defined on the OpenStackMachine.

/hold


EmilienM:
/lgtm
/approve
very handy, thanks

PR #2133, opened 2024-06-27
Generate applyconfiguration for server-side apply

mdbooth:
This adds generation of applyconfiguration using code-generator from k/k.

While the configuration isn't currently too complex, it's more complex than it
will be eventually due to the different versions of k/k used for
code-generation and CAPO itself. This will get much simpler when CAPO is bumped
to v0.31.


EmilienM:
is it ready for review?

mdbooth:
> is it ready for review?

Yes.

.gitignore
```diff
@@ -146,7 +146,7 @@ kubernetes.tar.gz
 # generated files in any directory
 # TODO(thockin): uncomment this when we stop committing the generated files.
 #zz_generated.*
-zz_generated.openapi.go
+#zz_generated.openapi.go
```
EmilienM, line 149: why not removing it? or comment why we don't want that file.

EmilienM (approved):
I have a minor comment for the gitignore, otherwise LGTM. Looks bit but a lot is generated. From my little knowledge on that topic, I didn't see any red flag.

/approve

PR #2135, opened 2024-07-02
🌱 cloud: Better error message on image lookup failure

stephenfin:

Currently, if cluster-api fails to find an image during machine creation it fails with the following, rather unhelpful error:

    "Reconciler error" err="no images were found with the given image filter: {<nil> 0xc000558880}" ...

Improve our error messages so we get something slightly more useful.

    "Reconciler error" err="no images were found with the given image filter: name=foo, tags=[]" ...

Signed-off-by: Stephen Finucane <stephenfin@redhat.com>

Fixes #

None.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -346,12 +346,20 @@ func (s *Service) GetImageID(image infrav1.ImageParam) (string, error) {
 
 	switch len(allImages) {
 	case 0:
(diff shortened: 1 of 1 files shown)


pierreprinetti:
All the arguments to the error message are nil-checked before use: LGTM!

PR #2139, opened 2024-07-04
:bug: Fix down-conversion of IdentityRef

### LINKED ISSUE #2136: Down-converting v1beta1 to v1alpha6 doesn't set identityRef.kind
/kind bug

Reported against v0.10.3.

```
Error: UPGRADE FAILED: cannot patch "lab-sjc2-orange" with kind OpenStackCluster: OpenStackCluster.infrastructure.cluster.x-k8s.io "lab-sjc2-orange" is invalid: spec.identityRef.kind: Invalid value: "": spec.identityRef.kind in body should be at least 1 chars long && cannot patch "lab-sjc2-orange-control-plane-78d4515c" with kind OpenStackMachineTemplate: OpenStackMachineTemplate.infrastructure.cluster.x-k8s.io "lab-sjc2-orange-control-plane-78d4515c" is invalid: spec.template.spec.identityRef.kind: Invalid value: "": spec.template.spec.identityRef.kind in body should be at least 1 chars long && cannot patch "lab-sjc2-orange-md-0-6c304a66" with kind OpenStackMachineTemplate: OpenStackMachineTemplate.infrastructure.cluster.x-k8s.io "lab-sjc2-orange-md-0-6c304a66" is invalid: spec.template.spec.identityRef.kind: Invalid value: "": spec.template.spec.identityRef.kind in body should be at least 1 chars long
```

This is an upgrade case, originally upgrading from v1alpha5. CAPO has been upgraded to v0.10.3 through v0.9.1, and the helm chart which owns these objects has been updated to v1alpha6.

I'm assuming that the issue here is that the helm chart tries to read v1alpha6 and we don't restore `kind` when we down-convert from v1beta1. The value can't be converted because it's no longer present in v1beta1, but I'm surprised it's not restored. If there's no obvious reason for it we could likely hard-code it to `Secret` in the restore function if it's not already set.

mdbooth:
We were not setting IdentityRef.Kind when down-converting an object with
no previous version annotation, which results in a validation error.

Fixes #2136

TODO:
- [x] Confirm full-test is passing

/hold

### DIFF
--- a/api/v1alpha6/openstackcluster_conversion.go
+++ b/api/v1alpha6/openstackcluster_conversion.go
@@ -365,7 +365,10 @@ func Convert_v1beta1_OpenStackClusterSpec_To_v1alpha6_OpenStackClusterSpec(in *i
 	}
 
 	out.CloudName = in.IdentityRef.CloudName
-	out.IdentityRef = &OpenStackIdentityReference{Name: in.IdentityRef.Name}
+	out.IdentityRef = &OpenStackIdentityReference{}
+	if err := Convert_v1beta1_OpenStackIdentityReference_To_v1alpha6_OpenStackIdentityReference(&in.IdentityRef, out.IdentityRef, s); err != nil {
+		return err
+	}
 
 	if in.APIServerLoadBalancer != nil {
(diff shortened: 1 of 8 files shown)


EmilienM:
/lgtm
/approve
/hold cancel

Thanks

PR #2141, opened 2024-07-11
:seedling: E2E: Add subports to trunked port to test subports deletion

mquhuy:
This is a continuation of https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2081. The idea is to have subports added to trunked port, so that we could verify that subports are deleted with the trunk.

### DIFF
--- a/test/e2e/suites/e2e/e2e_test.go
+++ b/test/e2e/suites/e2e/e2e_test.go
@@ -28,6 +28,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/openstack"
 	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/layer3/routers"
@@ -429,9 +431,10 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 
 			// Note that as the bootstrap config does not have cloud.conf, the node will not be added to the cluster.
 			// We still expect the port for the machine to be created.
+			machineDeployment := makeMachineDeployment(namespace.Name, md3Name, clusterName, "", 1)
 			framework.CreateMachineDeployment(ctx, framework.CreateMachineDeploymentInput{
 				Creator:                 e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
-				MachineDeployment:       makeMachineDeployment(namespace.Name, md3Name, clusterName, "", 1),
+				MachineDeployment:       machineDeployment,
 				BootstrapConfigTemplate: makeJoinBootstrapConfigTemplate(namespace.Name, md3Name),
 				InfraMachineTemplate:    makeOpenStackMachineTemplateWithPortOptions(namespace.Name, clusterName, md3Name, customPortOptions, machineTags),
 			})
@@ -445,9 +448,9 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 				return len(plist)
 			}, e2eCtx.E2EConfig.GetIntervals(specName, "wait-worker-nodes")...).Should(Equal(1))
 
-			port := plist[0]
-			Expect(port.Description).To(Equal("primary"))
-			Expect(port.Tags).To(ContainElement(testTag))
+			primaryPort := plist[0]
+			Expect(primaryPort.Description).To(Equal("primary"))
+			Expect(primaryPort.Tags).To(ContainElement(testTag))
 
 			// assert trunked port is created.
 			Eventually(func() int {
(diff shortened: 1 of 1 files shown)


test/e2e/suites/e2e/e2e_test.go
```diff
@@ -469,6 +471,29 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
 			Expect(trunk.PortID).To(Equal(port.ID))
+			shared.Logf("Trunk ID: %s", trunk.ID)
+
+			// Add subports to the trunked port.
+			shared.Logf("Adding subports to the trunked port")
+			providerClient, clientOpts, _, err := shared.GetTenantProviderClient(e2eCtx)
+			Expect(err).To(BeNil(), "Cannot create providerClient")
+			computeClient, err := openstack.NewComputeV2(providerClient, gophercloud.EndpointOpts{
+				Region: clientOpts.RegionName,
+			})
```
mdbooth, line 482: Not 100% sure, but I think this might be your problem: I'd expect you to need a network client here.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -469,6 +471,29 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+			shared.Logf("Trunk ID: %s", trunk.ID)
+
+			// Add subports to the trunked port.
+			shared.Logf("Adding subports to the trunked port")
+			providerClient, clientOpts, _, err := shared.GetTenantProviderClient(e2eCtx)
+			Expect(err).To(BeNil(), "Cannot create providerClient")
+			networkClient, err := openstack.NewNetworkV2(providerClient, gophercloud.EndpointOpts{
+				Region: clientOpts.RegionName,
+			})
+			Expect(err).To(BeNil(), "Cannot create compute client")
```
MaysaMacedo, line 483: network client

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -469,6 +471,29 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+
+			addSubportsOpts := trunks.AddSubportsOpts{
+				Subports: []trunks.Subport{
+					{
+						SegmentationID:   1,
+						SegmentationType: "vlan",
+						PortID:           port.ID,
```
MaysaMacedo, line 490: It looks like you are using the `port.ID`, which is equal to `trunk.PortID` and consequently it's the Trunk parent port as the subport. So, you need to create a separate Port to be added as the subport.

mquhuy: Thank you very much! That was indeed the issue!

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -445,30 +455,95 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+				},
+			}
+			_, err = trunks.AddSubports(ctx, networkClient, trunk.ID, addSubportsOpts).Extract()
+			Expect(err).To(BeNil(), "Cannot add subports")
+
+			subports, err := trunks.GetSubports(ctx, networkClient, trunk.ID).Extract()
+			Expect(err).To(BeNil())
+			Expect(subports).To(HaveLen(1))
+
+			shared.Logf("Deleting the trunk")
```
MaysaMacedo, line 518: From the CAPO flow, the deletion of a trunk that is associated with a Machine happens when [a Machine is deleted ](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/controllers/openstackserver_controller.go#L198). So, to me it makes sense to try deleting the Machine that uses the trunk and then see if the trunk and subports are deleted.

mquhuy: Hi @MaysaMacedo. Thank you for the suggestion. I've updated the PR now, but it seems that after the server is removed, all of the ports (and the trunk) don't just go away, they simply changed status to `DOWN`.

Could you please have a look and let me know if there's anything I need to change?

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -445,33 +452,131 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+						SegmentationID:   1,
+						SegmentationType: "vlan",
+						PortID:           subPort.ID,
+					},
+				},
+			}
+			_, err = trunks.AddSubports(ctx, networkClient, trunk.ID, addSubportsOpts).Extract()
+			Expect(err).To(BeNil(), "Cannot add subports")
+
+			subports, err := trunks.GetSubports(ctx, networkClient, trunk.ID).Extract()
```
MaysaMacedo, line 522: Instead of retrieving the subports again, could we use the trunk returned at line 519 and check the subports that exist on it?

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -445,33 +452,131 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+			shared.Logf("Waiting for the server to be cleaned")
+
+			computeClient, err := openstack.NewComputeV2(providerClient, gophercloud.EndpointOpts{
+				Region: clientOpts.RegionName,
+			})
+			Expect(err).To(BeNil(), "Cannot create compute client")
+
+			Eventually(
+				func() bool {
+					_, err := servers.Get(ctx, computeClient, serverID).Extract()
```
MaysaMacedo, line 565: Do we really need to fetch the Server again, if the Machine is gone (at line 551)?

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -67,6 +69,8 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
 
 		// Cleanup functions which cannot run until after the cluster has been deleted
 		postClusterCleanup []func(context.Context)
+		// mockCtrl           *gomock.Controller
+		// mockScopeFactory   *scope.MockScopeFactory
```
MaysaMacedo, line 73: leftovers?

test/e2e/suites/e2e/e2e_test.go
mquhuy: Hmm. Aren't we doing that here?

test/e2e/suites/e2e/e2e_test.go
mquhuy: Thanks. Good point, I guess maybe just waiting for the Server gone would be enough.

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -445,33 +448,124 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+
+			Eventually(
+				func() bool {
+					_, err := servers.Get(ctx, computeClient, serverID).Extract()
+					return gophercloud.ResponseCodeIs(err, 404)
+				}, e2eCtx.E2EConfig.GetIntervals(specName, "wait-delete-cluster")...,
+			).Should(BeTrue())
+
+			// Wait here for some time, to make sure the reconciler fully cleans everything
+			time.Sleep(100 * time.Second)
```
jichenjc, line 560: not sure whether we can do like 
sleep 5,10,20,50 etc to make the sleep time smaller?

mquhuy: Thanks @jichenjc . Yes, I can make some experiments

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -445,33 +448,124 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+				ClusterName:       clusterName,
+				Namespace:         namespace.Name,
+				MachineDeployment: *machineDeployment,
+			})
+
+			Expect(machines).To(HaveLen(1))
+
+			machine := machines[0]
+
+			serverID, err := getServerIDFromMachineName(ctx, c, machine.Namespace, machine.Name)
```
MaysaMacedo, line 537: Can you add a log message here? something in the lines of: "fetching server ID"

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -1067,3 +1161,16 @@ func isCloudProviderInitialized(taints []corev1.Taint) bool {
...
+	machine := &infrav1.OpenStackMachine{}
+	key := crclient.ObjectKey{
+		Namespace: namespace,
+		Name:      machineName,
+	}
+	if err := c.Get(ctx, key, machine); err != nil {
+		return "", err
+	}
+
+	return *machine.Status.InstanceID, nil
```
MaysaMacedo, line 1175: from the logs, it looks like this might be nil. Could it be that the machine was not ready yet?

can you add a check that verifies if it's nil? and if so, it waits for the machine to be ready?

MaysaMacedo: another option might be to keep checking the machine to see if it was deleted, instead of checking if the server was deleted this way we can abstract when server will be created/deleted.

test/e2e/suites/e2e/e2e_test.go
mquhuy: Sure thing.

test/e2e/suites/e2e/e2e_test.go
mquhuy: Thanks @MaysaMacedo. We've already got the ports and the trunk, so I think the machine should have been there :-? But I'll check that.

mquhuy: It seems to me that somehow `machine.Status.InstanceID` never got updated, but both the machine and the server were ready. I changed the logic used for fetching the server id. The test passed now xD

PR #2143, opened 2024-07-15
🐛 Ensure tools deps aren't older than CAPO deps

### LINKED ISSUE #2142: Ginkgo version mismatch in E2E tests
/kind bug

Running the e2e tests emits the following warning:
```
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.15.0
  Mismatched package versions found:
    2.17.1 used by e2e
```

mdbooth:
Ginkgo was warning that the ginkgo CLI was older than the ginkgo used in tests. This change ensures that can't happen again.

The change is to add an explicit CAPO dependency in hack/tools with a replace to make it use the version from its parent directory. This forces any dependencies in hack/tools which overlap with CAPO to be updated to the same version.

The resulting go.mod changes to ginkgo and code-generator were made automatically by running `go mod tidy`.

Fixes #2142

/hold


pierreprinetti:
I am rather unfamiliar with the blank-identifier imports. I also would have thought that the local-replaced import in `hack/tools/go.mod` would have been sufficient, wrongly apparently.

PR #2148, opened 2024-07-18
:sparkles: Support Flavor IDs

spjmurray:

If I have a system that uses flavor IDs, beacuse it's unambiguous, I then need to translate it into a flavor name before passing it to CAPO that will then translate it back into an ID using a list API call. Check whether the "name" looks like a UUID, and use it directly for the flavor lookup if you can.  Ideally, we'd make API changes like how images are handled, where you can be explicit about it, but I guess all OpenStack clients do this anyway!

Fixes #

As per usual, this is just a feeler to get feedback for whether my approach, simple and naive as it is, is going to fly!

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1alpha1/openstackserver_types.go
+++ b/api/v1alpha1/openstackserver_types.go
@@ -32,6 +32,7 @@ const (
 )
 
 // OpenStackServerSpec defines the desired state of OpenStackServer.
+// +kubebuilder:validation:XValidation:message="at least one of flavor or flavorID must be set",rule=(has(self.flavor) || has(self.flavorID))
 type OpenStackServerSpec struct {
 	// AdditionalBlockDevices is a list of specifications for additional block devices to attach to the server instance.
 	// +listType=map
@@ -48,8 +49,15 @@ type OpenStackServerSpec struct {
 	ConfigDrive optional.Bool `json:"configDrive,omitempty"`
 
 	// The flavor reference for the flavor for the server instance.
-	// +required
-	Flavor string `json:"flavor"`
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	Flavor *string `json:"flavor,omitempty"`
+
+	// FlavorID allows flavors to be specified by ID.  This field takes precedence
+	// over Flavor.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	FlavorID *string `json:"flavorID,omitempty"`
 
 	// FloatingIPPoolRef is a reference to a FloatingIPPool to allocate a floating IP from.
 	// +optional
--- a/api/v1alpha7/openstackmachine_types.go
+++ b/api/v1alpha7/openstackmachine_types.go
@@ -43,7 +43,13 @@ type OpenStackMachineSpec struct {
 	CloudName string `json:"cloudName"`
 
 	// The flavor reference for the flavor for your server instance.
-	Flavor string `json:"flavor"`
+	// +kubebuilder:validation:MinLength=1
+	Flavor *string `json:"flavor,omitempty"`
+
+	// FlavorID allows flavors to be specified by ID.  This field takes precedence
+	// over Flavor.
+	// +kubebuilder:validation:MinLength=1
+	FlavorID *string `json:"flavorID,omitempty"`
 
 	// The name of the image to use for your server instance.
 	// If the RootVolume is specified, this will be ignored and use rootVolume directly.
--- a/pkg/clients/compute.go
+++ b/pkg/clients/compute.go
@@ -28,7 +28,6 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servergroups"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers"
 	"github.com/gophercloud/utils/v2/openstack/clientconfig"
-	uflavors "github.com/gophercloud/utils/v2/openstack/compute/v2/flavors"
 
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/metrics"
 )
@@ -48,7 +47,7 @@ const NovaMinimumMicroversion = "2.60"
 type ComputeClient interface {
 	ListAvailabilityZones() ([]availabilityzones.AvailabilityZone, error)
 
-	GetFlavorFromName(flavor string) (*flavors.Flavor, error)
+	ListFlavors() ([]flavors.Flavor, error)
 	CreateServer(createOpts servers.CreateOptsBuilder, schedulerHints servers.SchedulerHintOptsBuilder) (*servers.Server, error)
 	DeleteServer(serverID string) error
 	GetServer(serverID string) (*servers.Server, error)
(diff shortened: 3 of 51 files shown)


spjmurray:
Just some further discussion, while I personally feel (from a professional perspective) that a v1beta2 and conversion hook is the correct thing to do from a consistency and ambiguity standpoint, it's likely to be a rather hefty change with wider ranging repercussions throughout the community.

Given my quick/easy/dirty approach here's some precedent from OpenStack proper:

[id or name resolution](https://github.com/openstack/python-novaclient/blob/master/novaclient/utils.py#L239)
[uuid identitfication](https://github.com/openstack/oslo.utils/blob/master/oslo_utils/uuidutils.py#L45)

spjmurray:
Thanks Emilien, allow me to test it does what it says on the tin before approvals...

spjmurray:
Okay, tested a back-port to v0.10.4 because the removal of v1alpha5 CRDs caused havoc without having to regenerate my helm charts 😸   Obviously had to do a rolling upgrade because the flavor changed in the OSMT, but... New machines provisioned fine!

```
spec:
  flavor: f81a2c71-32ab-4f07-a9bf-f4bd5ed97023
  identityRef:
    cloudName: cloud
    name: cluster-022e9cf4-cloud-config
  image:
    id: 02f86403-7146-4dea-b42e-bd6d64e709c3
  providerID: 'openstack:///cd986fed-1e89-44b4-93ce-0ea00c971076'
  serverMetadata:
    - key: clusterID
      value: b1af7f66-c816-4059-a68c-8ed91e5c304a
    - key: organizationID
      value: e9711b20-625f-4b7a-84ee-2fb5ce66389e
    - key: projectID
      value: ede41293-dd82-4449-b804-2dd8b9d2f42f
status:
  addresses:
    - address: 192.168.0.149
      type: InternalIP
    - address: cluster-022e9cf4-d8blc
      type: InternalDNS
  conditions:
    - lastTransitionTime: '2024-07-22T09:22:02Z'
      status: 'True'
      type: Ready
    - lastTransitionTime: '2024-07-22T09:22:02Z'
      status: 'True'
      type: APIServerIngressReadyCondition
    - lastTransitionTime: '2024-07-22T09:21:58Z'
      status: 'True'
      type: InstanceReady
```

I'm _assuming_ the CI jobs will continue to check name based flavors for me.  Although Cloudstrike... 

Ahem, additionally looking at the node labels after the upgrade, OCCM seems to continue to apply the labels as flavor name, so no changes there.either.

pkg/clients/compute.go
```diff
@@ -85,9 +86,20 @@ func (c computeClient) ListAvailabilityZones() ([]availabilityzones.Availability
 	return availabilityzones.ExtractAvailabilityZones(allPages)
 }
 
+// getFlavorIDFromName is a bit of a legacy misnomer, but allows flavors to be specified
```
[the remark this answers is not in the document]
spjmurray: As I say, this is based on pre-existing OpenStack conventions for all the UX stuff (and me being deliberately lazy before this discussion and the inevitable v1beta2, as a CRD change is really the only unambiguous way of doing this 😓 )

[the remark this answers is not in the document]
jichenjc: https://docs.openstack.org/api-ref/compute/#create-flavor

Only alphanumeric characters with hyphen ‘-’, underscore ‘_’, spaces and dots ‘.’ are permitted. If an ID is not provided, then a default UUID will be assigned.

I think even user can give ID, by default it's UUID from openstack layer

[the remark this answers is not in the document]
spjmurray: The only reason I can think of why you'd ever want to specify it is if you are replacing a flavor due to some immutability and want to retain the same ID due to things being linked to it.

mdbooth:
I'm ok with this, and I very much don't want to do a v1beta2 any time soon.

As @EmilienM points out, the ID can be specified. Not only this, it commonly is. I think perhaps devstack does this and folks have copied it?

I would personally just make an extra lookup here and try to fetch by ID. Ideally we would also add it to resolved, but lets not do even that right now.

My suggestion:
- Lookup value as if it it's an ID, return if that works.
- If not, do current lookup

api/v1alpha1/openstackserver_types.go
```diff
@@ -48,8 +48,13 @@ type OpenStackServerSpec struct {
 	ConfigDrive optional.Bool `json:"configDrive,omitempty"`
 
 	// The flavor reference for the flavor for the server instance.
-	// +required
-	Flavor string `json:"flavor"`
+	// +optional
+	Flavor string `json:"flavor,omitempty"`
```
spjmurray, line 52: I think we can do something like:

```
+kubebuilder:validation:XValidation:message="at least one of flavor and flavorID must be set",rule=(has(self.flavor) || has(self.flavorID))
```

on the Spec struct to tighten up the UX...

spjmurray:
As per usual a simple 1 file fix becomes 35... so the core of this adds in the `flavorID` field in a (hopefully what I can tell) backward compatible way where it takes precedence over the `flavor` field and things will be correctly upgraded.  Where I got creative was  "why handle images and flavors any differently?" so I took the liberty to handle them the same way, in the same places, so if you change one in future you should be prompted to change both.  Obviously this isn't my code base, get lost in it very easily, so take the philosophical review away @mdbooth 😓

api/v1alpha7/openstackmachine_types.go
```diff
@@ -43,7 +43,11 @@ type OpenStackMachineSpec struct {
 	CloudName string `json:"cloudName"`
 
 	// The flavor reference for the flavor for your server instance.
-	Flavor string `json:"flavor"`
+	Flavor string `json:"flavor,omitempty"`
```
mdbooth, line 46: We wouldn't normally change old API versions, but I'm inclined to let this slide, tbh.

api/v1alpha1/openstackserver_types.go
mdbooth: Please can you add the above CEL validation both here and on OpenStackMachineSpec, and also add tests for it in `apivalidations`. There are lots of examples in there and they're really easy to write, so shouldn't take long.

Please can you make this a `*string` and give it a MinLength of 1, and the same for FlavorID. This prevents the empty string from being valid, whilst still allowing it to be optional. And the same for 3 versions of OpenStackMachine.

mdbooth (commented):
This looks great. If can tidy up the API as mentioned, and add the validations (and tests) I'll be happy to approve this. Thanks!

mdbooth:
@spjmurray I'd like to cut a new release in the next day or so. Any chance this will make it?

spjmurray:
Done everything else, just working out the CEL testing this second @mdbooth

pkg/cloud/services/compute/instance.go
```diff
@@ -420,6 +405,31 @@ func (s *Service) getImageIDByReference(ctx context.Context, k8sClient client.Cl
...
+	allFlavors, err := s.getComputeClient().ListFlavors()
+	if err != nil {
+		return "", err
+	}
+
+	for _, flavor := range allFlavors {
+		if flavor.Name == *flavorName {
+			return flavor.ID, nil
+		}
+	}
```
mdbooth, line 428: I was going to write that client-side filtering is madness, but:
* That's what we do currently, presumably because
* Nova doesn't appear to support filtering by name anyway 🤯

mdbooth (commented):
Looks good. I'm happy to approve it now as the test failures look like straightforward oversights and it won't merge until they're fixed anyway.

/approve

pkg/cloud/services/compute/instance.go
spjmurray: Oh it drives me mad, it really needs metadata of some variety so you can describe CPU microarchitectures, GPUs and other jazz.  All of which is really useful to an end user!

spjmurray:
The promised land!  @mdbooth when `v1beta2` or `v1` happen, I'll happily do the work to refactor the CRD layout, shouldn't be too hard now

mdbooth:

I have a horrible suspicion we have a genuine issue in the conversion tests, but I don't *think* it's related to this PR.

spjmurray:
Well, the `nil` vs `map[string]string{}` stuff is a regular annoyance for anyone familiar with `reflect.DeepEqual` 😉 .  The bastion one looks somewhat concerning as the OpenstackMachineSpec should require a flavor now, so I may be to blame.

Lemme run `go test -v ./api/v1alpha7/... -run TestFuzzyConversion/for_OpenStackClusterTemplate_with_mutate/spoke-hub-mutate-spoke-hub -count 100  and see what happens...

PR #2150, opened 2024-08-08
🌱 Install tools with go install

mdbooth:
* controller-gen
* golangci-lint
* kustomize
* setup-envtest

This patch follows the model used by kubebuilder scaffolding v4. The advantage of this model is that it independent of any go module.

/hold


mdbooth:
```
                  error: <*exec.Error | 0xc0007d40c0>{
                      Name: "kustomize",
                      Err: <*errors.errorString | 0x3f80fe0>{
                          s: "executable file not found in $PATH",
                      },
                  },
```

Presumably something somewhere is adding hack/tools/bin to PATH? I couldn't immediately find out what.

PR #2153, opened 2024-08-08
Copy InstanceID from OpenStackServer to OpenStackMachine

tsuzu:
* `OpenStackMachine.Status.InstanceID` is always empty if even the server is running
    * Copy the InstanceID from OpenStackServer to OpenStackMachine as soon as the server gets running

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] Add a step to check status of OpenStackMachine in E2E

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -423,6 +423,7 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 
 		// Set properties required by CAPI machine controller
 		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", *openStackServer.Status.InstanceID))
+		openStackMachine.Status.InstanceID = openStackServer.Status.InstanceID
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
 		// If the machine has a NodeRef then it must have been working at some point,


controllers/openstackmachine_controller.go
```diff
@@ -423,6 +423,7 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 
 		// Set properties required by CAPI machine controller
 		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", *openStackServer.Status.InstanceID))
+		openStackMachine.Status.InstanceID = openStackServer.Status.InstanceID
```
EmilienM, line 426: What problem are we trying to solve here? I think we made that on purpose but I'll like you know what your issue is before we revert that change.

tsuzu: Hi!
As an operator of CAPI/CAPO, I have a Kubernetes controller that depends on `OpenStackMachine.Stataus.InstanceID`.
And the disappearance of InstanceID will cause a bug.

Are you planning to delete this field in `v1beta1`?
I understand that `InstanceID` can be still retrieved from `ProviderID`, but I hope the breaking change is delivered with the update of API versions like `v1beta2`.

Thank you.

EmilienM: I guess since we're v1beta1, we shouldn't have remove that field. I won't block this PR.

PR #2156, opened 2024-08-14
🐛 OpenStackServer: move Resolved & Resources from Status to Spec

### LINKED ISSUE #2155: race: deadlock between OpenStackCluster and OpenStackServer
/kind bug

Seen in periodic jobs:

```
I0814 04:05:47.813699       1 openstackserver_controller.go:98] "Reconciling OpenStackServer" controller="openstackserver" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackServer" OpenStackServer="self-hosted-r7wjlk/self-hosted-d5omd6-bastion" namespace="self-hosted-r7wjlk" name="self-hosted-d5omd6-bastion" reconcileID="c6afdfa1-e6ac-4446-a2c8-98552e6959ab"
I0814 04:05:47.813776       1 openstackserver_controller.go:106] "OpenStackServer %s/%s linked to a Cluster that is paused. Won't reconcile" controller="openstackserver" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackServer" OpenStackServer="self-hosted-r7wjlk/self-hosted-d5omd6-bastion" namespace="self-hosted-r7wjlk" name="self-hosted-d5omd6-bastion" reconcileID="c6afdfa1-e6ac-4446-a2c8-98552e6959ab" self-hosted-r7wjlk="self-hosted-d5omd6-bastion"
```

* The move won't progress because OpenStackCluster is not ready
* OpenStackCluster is not ready because it's waiting on the bastion
* The bastion won't progress because the cluster is paused

EmilienM:

To handle the clusterctl move, we need to make sure that the resolved & resources are in spec and not in status because status is not moved.

Fixes #2155


jichenjc:
>>>From now, we'll mark the OpenStackCluster to be Ready even if the
Bastion isn't yet.

not sure this assumption is correct, I will always expect I can use my cluster when you told me it's ready..
but it turn out it's not , I might still wait for Bastion and there is chance the Bastion is not able to be running ..

mdbooth:
> > > > From now, we'll mark the OpenStackCluster to be Ready even if the
> > > > Bastion isn't yet.
> 
> not sure this assumption is correct, I will always expect I can use my cluster when you told me it's ready.. but it turn out it's not , I might still wait for Bastion and there is chance the Bastion is not able to be running

The Bastion isn't a required part of the cluster infrastucture from CAPI's POV. Strictly speaking, the only thing that's required is the control plane endpoint, but we also need networking to be up because of how we provision machines. With this change we're marking the infrastructure ready as soon as it's ready enough for all other components which need it to proceed. No CAPI component depends on the Bastion.

The ready flag also doesn't strictly indicate that the Bastion is currently available in any case, because the Bastion is mutable. If the bastion was added after the cluster became ready, or if the bastion is modified after the bastion became ready, the infrastructure will not be marked not ready just because the Bastion isn't ready.

I believe this behaviour is consistent with our current contracts. It also has the advantage of making provisioning faster, because worker provisioning will start concurrently with provisioning of the bastion, as well as more robust because failure to provision the bastion won't prevent provisioning of all workers.

mdbooth:
```
I0814 16:08:35.460010       1 controller.go:115] "Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference" controller="openstackserver" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackServer" OpenStackServer="self-hosted-3g23fy/self-hosted-91sv3i-md-0-zg7w9-ljkvh" namespace="self-hosted-3g23fy" name="self-hosted-91sv3i-md-0-zg7w9-ljkvh" reconcileID="abe5b559-e0ba-4307-92f9-513bcea08855"
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1ac879b]

goroutine 272 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:116 +0x1e5
panic({0x1cf4da0?, 0x3495480?})
	/usr/local/go/src/runtime/panic.go:770 +0x132
sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.AdoptServerResources(0xc0009371a0, 0x0, 0xc000adf2a8)
	/workspace/pkg/cloud/services/compute/server_resources.go:31 +0x3b
sigs.k8s.io/cluster-api-provider-openstack/controllers.adoptServerResources(0xc0009371a0?, 0xc00075ce40?)
	/workspace/controllers/openstackserver_controller.go:426 +0x85
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackServerReconciler).reconcileDelete(0xc0004b2300, 0xc0009371a0, 0xc000524a08)
	/workspace/controllers/openstackserver_controller.go:215 +0xc6
sigs.k8s.io/cluster-api-provider-openstack/controllers.(*OpenStackServerReconciler).Reconcile(0xc0004b2300, {0x23353b0, 0xc0007c3dd0}, {{{0xc0007825a0?, 0x0?}, {0xc000688870?, 0xc00093bd10?}}})
	/workspace/controllers/openstackserver_controller.go:130 +0x648
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x233a9e8?, {0x23353b0?, 0xc0007c3dd0?}, {{{0xc0007825a0?, 0xb?}, {0xc000688870?, 0x0?}}})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:119 +0xb7
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000432e60, {0x23353e8, 0xc0003fd5e0}, {0x1db8600, 0xc00089c740})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:316 +0x3bc
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc000432e60, {0x23353e8, 0xc0003fd5e0})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:266 +0x1be
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:227 +0x79
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2 in goroutine 46
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.3/pkg/internal/controller/controller.go:223 +0x50c
```
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2156/pull-cluster-api-provider-openstack-e2e-full-test/1823734659926724608/artifacts/clusters/bootstrap/logs/capo-system/capo-controller-manager/capo-controller-manager-7d957cb6cd-q6c8z/manager.log

mdbooth:
Firstly, some context about the CAPI move operation: one of its quirks is that it only moves an object's spec, not its status. Therefore immediately after a move we have an empty status.

The immediate cause of this panic is our assumption in delete that `resolvedResources` is populated. An example flow that could have caused this panic is:
* Create OSServer on cluster A
* Pause OSServer
* Move OSServer to cluster B
* Delete OSServer on cluster B
* Unpause OSServer

In this flow, when we unpause OSServer on cluster B it is marked deleted but it does not have resolved set in the status because status was not moved. This results in the panic when we try to adopt resources in the delete flow, which assumes that resolved is already set.

If we ignore CAPI move for a moment, we assumed it should be impossible for this to happen because we write resolved in the same transaction (or before) we add the finalizer. There are at least 2 problems with this assumption.

The first is pretty trivial: it's theoretically possible we could be called with a deleted object which has simply not been removed yet. The fix for this is simple: we should not execute our delete flow if the object does not have our finalizer. This will be either because we never added it, in which case we know no resources were created and therefore delete does not need to run, or because we removed it, in which case we already executed delete and it does not need to run again. The most likely way we would hit this in practise is if another controller added its own finalizer. This is not the cause of this issue, but we should fix it.

The second is something I gave quite a lot of thought to while working on the image controller but I didn't get around to implementing (although this issue highlights that we should): we should probably move `resolved` and `resources` from the status to the spec.[^1] For the same reason, we should probably copy them from OpenStackMachine to OpenStackServer when initially creating an OpenStackServer for an existing OpenStackMachine.

[^1]: For the image controller, it would be `imageID`.

CAPI move's design decision to only move the spec, not the status, is predicated on an assumption that the status can always be deterministically reconstructed from the spec. However, the reason we added resolved and resources is we kept hitting cases in OpenStack where that is not true, usually because (in descending order of likelihood):
* multiple resources can exist with the same name
* resources can be renamed
* resources can be deleted and recreated with the same name
with the first being overwhelmingly the most problematic.

So we made the decision to store resource uuids for everything we need to act on before acting on them. `resolved` holds the set of resources which we depend on and therefore must exist before we create anything, and `resources` holds the set of resources we created.

Once set, their purpose is that the future behaviour of the controller depends on them, so adding them to the spec isn't so much of a stretch. It is a bit untidy, though, because it is not something a user should set. It would also be less error-prone if we set it with SSA rather than Patch, but I don't think it's worth rewriting the OpenStackServer controller for that.

cc @sbueringer or any other CAPI folks who might be interested in this move use case.

sbueringer:
To be honest, I don't have the time at the moment to look into this

EmilienM:
This is work in progress, it fails in CI. I'll make it work locally for me this week and push it here when ready.

controllers/openstackcluster_controller.go
```diff
@@ -564,6 +564,15 @@ func bastionToOpenStackServerSpec(openStackCluster *infrav1.OpenStackCluster) *i
 	}
 	openStackServerSpec := openStackMachineSpecToOpenStackServerSpec(bastion.Spec, openStackCluster.Spec.IdentityRef, compute.InstanceTags(bastion.Spec, openStackCluster), az, nil, getBastionSecurityGroupID(openStackCluster), openStackCluster.Status.Network.ID)
 
+	if server != nil {
```
EmilienM, line 567: This is the ugly part that I would agree to change/refactor.
Basically if an OpenStackServer already existed (the case for the bastion, which can be created then updated with a new spec), we want to compare Apples with Apples, so we need to get the Resolved & Resources fields too if they were here before. Otherwise it'll reconcile in infinite and the bastion will be deleted/recreated/deleted, etc.

EmilienM:
@jichenjc for review, when time permits.

EmilienM:
another try to make sure I wasn't lucky

EmilienM:
@jichenjc are you fine with this PR?
Thanks

jichenjc:
/approve

yes, sorry for late reply due to attend KubeCon now :)

EmilienM:
> yes, sorry for late reply due to attend KubeCon now :)

No problem, I hope you had fun!

EmilienM:
I'm going to merge this to help with CI.

@mdbooth please review it when you're back from PTO.

/lgtm

EmilienM:
Trying an alternative here https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2167

EmilienM:
@jichenjc please look at https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2167 as a less invasive alternative. Feel free to provide feedback.
cc @mdbooth

jichenjc:
sure, will take a look and comment there, thanks

PR #2159, opened 2024-08-15
🐛 port: don't add any SGs when port security is disabled

### LINKED ISSUE #2158: Impossible to disable portSecurity on a port in OpenStackMachine
/kind bug

if `managedSecurityGroups` is enabled, each port that is created will get the security group that belongs to it (master or worker).
But if we explicitly set `disablePortSecurity: true` on that port, it'll still try to get the CAPO-managed security group and therefore Neutron will fail to create that port. A port with no security can't have security groups in its spec.

We shouldn't create a port that has explicit disablePortSecurity with security groups.

EmilienM:

When port security is disabled on a port, don't add any security group
to the port options.

Skip the security groups when resolving the ports spec.
Report an error when a port tries to be created with both security
groups and disable port security to true.
Adds unit tests coverage for these scenarios.

Fixes #2158

### DIFF
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -170,6 +170,9 @@ func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.Resol
 		createOpts.FixedIPs = fixedIPs
 	}
 	if portSpec.SecurityGroups != nil {
+		if ptr.Deref(portSpec.DisablePortSecurity, false) {
+			return nil, errors.New("security groups cannot be set when port security is disabled")
+		}
 		createOpts.SecurityGroups = &portSpec.SecurityGroups
 	}
 	builder = createOpts
@@ -434,13 +437,15 @@ func (s *Service) normalizePorts(ports []infrav1.PortOpts, clusterResourceName, 
 			return nil, err
 		}
 
-		// Resolve security groups
-		if len(port.SecurityGroups) == 0 {
-			normalizedPort.SecurityGroups = defaultSecurityGroupIDs
-		} else {
-			normalizedPort.SecurityGroups, err = s.GetSecurityGroups(port.SecurityGroups)
-			if err != nil {
-				return nil, fmt.Errorf("error getting security groups: %v", err)
+		// Resolve security groups when port security is not disabled
+		if !ptr.Deref(port.DisablePortSecurity, false) {
+			if len(port.SecurityGroups) == 0 {
+				normalizedPort.SecurityGroups = defaultSecurityGroupIDs
+			} else {
+				normalizedPort.SecurityGroups, err = s.GetSecurityGroups(port.SecurityGroups)
+				if err != nil {
+					return nil, fmt.Errorf("error getting security groups: %v", err)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


pkg/cloud/services/networking/port.go
```diff
@@ -170,6 +170,9 @@ func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.Resol
 		createOpts.FixedIPs = fixedIPs
 	}
 	if portSpec.SecurityGroups != nil {
+		if ptr.Deref(portSpec.DisablePortSecurity, false) {
+			return nil, errors.New("security groups cannot be set when port security is disabled")
+		}
```
mdbooth, line 175: IIRC if you try to add a security group to a port with port security disabled you get an error, right? i.e. It's not just that it ignores the option, but it actually won't add the security groups?

If so, we should be able to safely add API validation for this instead, because we know there is no working configuration with it set.

i.e. We should write this as CEL instead. The tests would be in apivalidations.

mdbooth (commented):
Change looks sound. If my assumption about the error is correct, we should add API validation instead of runtime validation, though.

pkg/cloud/services/networking/port.go
EmilienM: > IIRC if you try to add a security group to a port with port security disabled you get an error, right? i.e. It's not just that it ignores the option, but it actually won't add the security groups?

yeah right. And good points on API validations. I'll check that.

EmilienM: I've added tests in the webhook like other fields and tests in API validations. I've let this check on purpose because:
1. it doesn't hurt, I think
2. it's safe and it makes sense to report an error and not let it through anyway.

pkg/webhooks/openstackserver_webhook.go
```diff
@@ -64,6 +65,12 @@ func (*openStackServerWebhook) ValidateCreate(_ context.Context, objRaw runtime.
 		}
 	}
 
+	for _, port := range newObj.Spec.Ports {
+		if ptr.Deref(port.DisablePortSecurity, false) && len(port.SecurityGroups) > 0 {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "ports"), "cannot have security groups when DisablePortSecurity is set to true"))
+		}
+	}
```
mdbooth, line 72: Is there any reason we can't do this in CEL? We haven't added any new logic to the webhooks in a while and I'd like to phase them out.

pkg/cloud/services/networking/port.go
mdbooth: Ack to leaving the check in.

I'd prefer we didn't add anything new to the webhooks, though, unless we absolutely have to. I take a look to see how easy the CEL is to write.

EmilienM: that was my thought when I looked at CEL and why I took the webhook way instead :confused:

EmilienM:
if you can /approve instead so I can ask @MaysaMacedo to take a look as well. Thanks

EmilienM:
@MaysaMacedo it merged by accident but when time permits, please have a look and let me know if anything is wrong, we'll fix afterwards. Thanks

PR #2168, opened 2024-08-29
:book: docs/configuration: fix security groups example

tormath1:
**What this PR does / why we need it**: it now uses a filter to get the security group by name

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [ ] adds unit tests

/hold


EmilienM (approved):
/lgtm
/approve

thanks!

PR #2174, opened 2024-09-17
⚠️ Convert APIServerPort to uint16

mdbooth:
This problem was highlighted by the updated gosec linter brought in by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2173, which pointed out the unsafe conversion from int to int32. CAPI's control plane endpoint port is an int32. We were using int, which can be larger. We also weren't validating it.

APIServerPort is a TCP port, which is an unsigned 16 bit integer. We were
previously allowing it to be any integer.

This is a breaking change, but as any working configuration would have
to have a value which passes the new validation it should not break any
working configuration.

### DIFF
--- a/api/v1alpha6/openstackcluster_conversion.go
+++ b/api/v1alpha6/openstackcluster_conversion.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1alpha6
 
 import (
+	"math"
 	"reflect"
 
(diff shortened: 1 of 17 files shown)


mdbooth:
Lets merge https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2175 first for tidiness

EmilienM:
/approve
feel free to merge it when you want

PR #2178, opened 2024-09-25
🐛 Ensure SSA patch can't accidentally create a new object

mdbooth:
This fixes a race during image delete where we can get a second reconcile on a deleted object and accidentally create a new object when we only intended to remove the finalizer from the old object. With this change, the second patch will throw an error instead.

/hold

### DIFF
--- a/internal/controllers/image/reconcile.go
+++ b/internal/controllers/image/reconcile.go
@@ -232,7 +232,7 @@ func (r *orcImageReconciler) reconcileDelete(ctx context.Context, orcImage *orcv
 	log.V(4).Info("Image is deleted")
 
 	// Clear owned fields on the base resource, including the finalizer
-	applyConfig := orcapplyconfigv1alpha1.Image(orcImage.Name, orcImage.Namespace)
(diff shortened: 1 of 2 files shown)


EmilienM:
Thanks a lot for the time spent on this one.
I'll approve it right away so when you think it's ready we can merge it.

I suggest we run the tests ~10 times to try reproducing the error in case it happens again. Then we can hold cancel and land it.

/approve

mdbooth:
I just looked through all the CAPO logs of every test run so far against this PR, and unfortunately I haven't seen the error message which would be produced if the fix in this PR had prevented the recreate-after-delete race. IOW I don't think we've actually hit the race this is fixing. Lets give it a few more tries, but eventually I think we'll just have to merge it and see if it makes the failures in the periodics go away.

mdbooth:
Not sure what the multi-AZ test failure was, but it's not related. We don't have artifacts from the failed cluster, which is weird.

mdbooth:
Again nothing. Lets make this the last one:


/test pull-cluster-api-provider-openstack-e2e-full-test pull-cluster-api-provider-openstack-e2e-test

mdbooth:
Nothing. Lets merge this.

/hold cancel

PR #2180, opened 2024-09-26
🐛 Image controller: don't attempt upload when adopting

mdbooth:
When an image is to be adopted, the controller would still attempt to upload content when it observed the image in the queued state after having been created externally. This didn't result in incorrect behaviour because API validation ensures that images for adoption don't specify a source. This meant that no upload could be attempted, but it did result in a spurious error message reported both to the user and in the logs.

/hold


EmilienM:
/approve
/lgtm

@mdbooth good to go I think, thanks for the fix

PR #2187, opened 2024-10-09
feat(manager): :sparkles: add region to openStackMachineSpec.ProviderID field from annotation

### LINKED ISSUE #2183: Multi openstack cloud support with OCCM
/kind feature

**Glossary**

 - **cloud**: I will use "cloud" term to identify an independent openstack cluster, 2 different clouds means 2 different openstack clusters with differents endpoints, credentials, regoin name, keystore ...

**Description**

I would like to be able to create k8s cluster spread across different openstack clusters. I have already underlying network based on VPN tunnels and network interconnections, ready to VMs communications between different clouds.

I am able to create VM in my different clouds with the help of `spec.template.spec.identityRef` from `OpenStackMachineTemplate` which permit to override `idetityRef` define in parent `OpenStackCluster` object.

It works well ! My issue concern integration with OCCM (Openstack Cloud Controller Manager).

The only way yo manage multiple clouds with OCCM seems to be to deploy 1 OCCM instance per cloud by setting env variables `OS_CCM_REGIONAL="true"` and `OS_V1_INSTANCES="true"`.

In this way each OCCM manage his own cloud by adding OS region name (a limitation could be that 2 differents clouds should have different regions names but I don't think that is really an issue). The "feature flag" implementation is [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/instances.go#L371).
OCCM with this env variables manage VMs with providerId format like `openstack://region_name/uuid`
k8s nodes `spec.providerID` is immutable and created by kubeadm in the usecase of CAPO, so I configure my VMs with following configuration :
```yaml
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
kind: KubeadmConfigTemplate
metadata:
  name: pool-region-one
  annotations:
    controlplane.cluster.x-k8s.io/skip-coredns: ""
    controlplane.cluster.x-k8s.io/skip-kube-proxy: ""
spec:
  template:
    spec:
      joinConfiguration:
        nodeRegistration:
          kubeletExtraArgs:
            cloud-provider: external
            provider-id: openstack://region-one/'{{ instance_id }}'
...
```
In that way k8s node are properly provision with the right providerID anc correctly recognize by the right OCCM.

**Missing feature**

The issue is now on CAPO implementation which hardcode providerID in format `openstack:///uuid` in his machine crd object with no possibility to configure endoint field with region name
```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Machine
spec:
  providerID: openstack:///uuid
  ...
```
I explain this scenario in an issue concerning usecase without kubeadm and unset providerID with usage of multiple OCCM in same k8s cluster https://github.com/kubernetes/cloud-provider-openstack/pull/2590#issuecomment-2376822814 , 
**TLDR**: CAPI see k8s node with providerID `openstack://region_name/uuid` but CAPO see machine with providerID `openstack:///uuid` (with same uuid) and it is stuck in this state.

**Code to patch**

I think that CAPO machine crd providerID is hardcoded [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/controllers/openstackmachine_controller.go#L425) but I don't find proper way to get region name here, to template this instruction.

**How to patch it ?**

- I can retreive identityRef fields so technically I can read k8s secret and parse OS config to get region name from configuration.
- Add an optional field in [OpenStackMachineSpec](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1beta1/openstackmachine_types.go#L95) "region_name" which will be added in providerID if it is fill.
- I didn't find better way yet... Maybe @mdbooth or someone could give me a good hint :crossed_fingers:

MatthieuFin:

To made CAPO compatible with OCCM multi-cloud deployment we need to manage providerID format with region code as introduced [here](https://github.com/kubernetes/cloud-provider-openstack/issues/1900) (`openstack://region/instance_uuid`). Use annotation from OpenstackMachine to define region in providerID field.

An empty or undefined annotation permit to be fully backward compatible. Moreover use annotation permit to manage some cluster with new OCCM provider-id string and other cluster with old format by same CAPO instance.

Fixes #2183

- As discussed in issue #2183 I first try to implement k8s secret call to get region name from config [here](https://github.com/MatthieuFin/cluster-api-provider-openstack/commit/04a4e609617d51249142d3328b3c017ed3d9c72a), it works well but it's not backward compatible. And if I setup a feature flag wiath a cli arg I loose the possibility to manage some k8s cluster with new providerID format beside of some k8s cluster with old providerID format managed by same CAPO instance.

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -645,7 +645,7 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 
 		// Set properties required by CAPI machine controller
-		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", instanceStatus.ID()))
+		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", openStackMachine.ObjectMeta.Annotations["cluster.x-k8s.io/openstack.cloud"], instanceStatus.ID()))
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
 		// If the machine has a NodeRef then it must have been working at some point,


MatthieuFin:
I don't understand why `make lint` fail, it fail on last release tag `v0.10.5` too.

EmilienM:
we should submit the PR on main first and then backport.

MatthieuFin:
Alright ! I create PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2193 on main which need [regression bug](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2191) to be fixed to work correctly.

I let this PR open if you wanna merge it for backport on v0.10

PR #2191, opened 2024-10-10
🐛 Re-allow `OpenStackMachine` to use their own `IdentityRef`

EmilienM:
**What this PR does / why we need it**:

This patch fixes a regression where the last `IdentityRef` from a given list where used instead of the first one that was found valid.
In the case of the OpenStackMachine, we pass two refs, one for the Machine and one for the Cluster. Now the provider will return the ref from the Machine if it exists.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2190

PR #2193, opened 2024-10-14
feat(manager): :sparkles: add region to openStackMachineSpec.ProviderID field from crd identityRef

### LINKED ISSUE #2183: Multi openstack cloud support with OCCM

MatthieuFin:

To made CAPO compatible with OCCM multi-cloud deployment we need to manage providerID format with region code as introduced [here](https://github.com/kubernetes/cloud-provider-openstack/issues/1900) (`openstack://region/instance_uuid`). Use annotation from OpenstackMachine to define region in providerID field.

An empty or undefined annotation permit to be fully backward compatible. Moreover use annotation permit to manage some cluster with new OCCM provider-id string and other cluster with old format by same CAPO instance.

Fixes #2183

- As discussed in issue #2183 I first try to implement k8s secret call to get region name from config [here](https://github.com/MatthieuFin/cluster-api-provider-openstack/commit/04a4e609617d51249142d3328b3c017ed3d9c72a), it works well but it's not backward compatible. And if I setup a feature flag wiath a cli arg I loose the possibility to manage some k8s cluster with new providerID format beside of some k8s cluster with old providerID format managed by same CAPO instance.
- This feature need https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2191 merged to work correctly !

### DIFF
--- a/api/v1beta1/identity_types.go
+++ b/api/v1beta1/identity_types.go
@@ -18,6 +18,7 @@ package v1beta1
 
 // OpenStackIdentityReference is a reference to an infrastructure
 // provider identity to be used to provision cluster resources.
+// +kubebuilder:validation:XValidation:rule="(!has(self.region) && !has(oldSelf.region)) || self.region == oldSelf.region",message="region is immutable"
 type OpenStackIdentityReference struct {
 	// Name is the name of a secret in the same namespace as the resource being provisioned.
 	// The secret must contain a key named `clouds.yaml` which contains an OpenStack clouds.yaml file.
@@ -28,6 +29,12 @@ type OpenStackIdentityReference struct {
 	// CloudName specifies the name of the entry in the clouds.yaml file to use.
 	// +kubebuilder:validation:Required
 	CloudName string `json:"cloudName"`
+
+	// Region specifies an OpenStack region to use. If specified, it overrides
+	// any value in clouds.yaml. If specified for an OpenStackMachine, its
+	// value will be included in providerID.
+	// +optional
+	Region string `json:"region,omitempty"`
 }
 
 // IdentityRefProvider is an interface for obtaining OpenStack credentials from an API object
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -422,7 +422,11 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 
 		// Set properties required by CAPI machine controller
-		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", *openStackServer.Status.InstanceID))
+		var region string
+		if openStackMachine.Spec.IdentityRef != nil {
+			region = openStackMachine.Spec.IdentityRef.Region
+		}
+		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", region, *openStackServer.Status.InstanceID))
 		openStackMachine.Status.InstanceID = openStackServer.Status.InstanceID
 		openStackMachine.Status.Ready = true
 	case infrav1.InstanceStateError:
--- a/api/v1alpha6/conversion_test.go
+++ b/api/v1alpha6/conversion_test.go
@@ -91,6 +91,13 @@ func TestFuzzyConversion(t *testing.T) {
 				}
 			},
 
+			func(identityRef *infrav1.OpenStackIdentityReference, c fuzz.Continue) {
+				c.FuzzNoCustom(identityRef)
+
+				// None of the following identityRef fields have ever been set in v1alpha6
+				identityRef.Region = ""
+			},
+
 			func(spec *OpenStackMachineSpec, c fuzz.Continue) {
 				c.FuzzNoCustom(spec)
 
--- a/api/v1alpha6/zz_generated.conversion.go
+++ b/api/v1alpha6/zz_generated.conversion.go
@@ -1085,6 +1085,7 @@ func autoConvert_v1alpha6_OpenStackIdentityReference_To_v1beta1_OpenStackIdentit
 func autoConvert_v1beta1_OpenStackIdentityReference_To_v1alpha6_OpenStackIdentityReference(in *v1beta1.OpenStackIdentityReference, out *OpenStackIdentityReference, s conversion.Scope) error {
 	out.Name = in.Name
 	// WARNING: in.CloudName requires manual conversion: does not exist in peer-type
+	// WARNING: in.Region requires manual conversion: does not exist in peer-type
 	return nil
 }
 
--- a/api/v1alpha7/conversion_test.go
+++ b/api/v1alpha7/conversion_test.go
@@ -89,6 +89,13 @@ func TestFuzzyConversion(t *testing.T) {
 					}
 				}
 			},
+
+			func(identityRef *infrav1.OpenStackIdentityReference, c fuzz.Continue) {
+				c.FuzzNoCustom(identityRef)
+
+				// None of the following identityRef fields have ever been set in v1alpha7
+				identityRef.Region = ""
+			},
 		}
 
 		return slices.Concat(v1alpha7FuzzerFuncs, testhelpers.InfraV1FuzzerFuncs())
--- a/api/v1alpha7/zz_generated.conversion.go
+++ b/api/v1alpha7/zz_generated.conversion.go
@@ -1319,6 +1319,7 @@ func autoConvert_v1alpha7_OpenStackIdentityReference_To_v1beta1_OpenStackIdentit
 func autoConvert_v1beta1_OpenStackIdentityReference_To_v1alpha7_OpenStackIdentityReference(in *v1beta1.OpenStackIdentityReference, out *OpenStackIdentityReference, s conversion.Scope) error {
 	out.Name = in.Name
 	// WARNING: in.CloudName requires manual conversion: does not exist in peer-type
+	// WARNING: in.Region requires manual conversion: does not exist in peer-type
 	return nil
 }
 
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackclusters.yaml
@@ -3482,10 +3482,20 @@ spec:
                               The secret must contain a key named `clouds.yaml` which contains an OpenStack clouds.yaml file.
                               The secret may optionally contain a key named `cacert` containing a PEM-encoded CA certificate.
                             type: string
+                          region:
+                            description: |-
+                              Region specifies an OpenStack region to use. If specified, it overrides
+                              any value in clouds.yaml. If specified for an OpenStackMachine, its
+                              value will be included in providerID.
+                            type: string
                         required:
                         - cloudName
                         - name
                         type: object
+                        x-kubernetes-validations:
+                        - message: region is immutable
+                          rule: (!has(self.region) && !has(oldSelf.region)) || self.region
+                            == oldSelf.region
                       image:
                         description: |-
                           The image to use for your server instance.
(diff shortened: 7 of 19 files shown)


controllers/openstackmachine_controller.go
```diff
@@ -422,7 +422,7 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 
 		// Set properties required by CAPI machine controller
-		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", *openStackServer.Status.InstanceID))
+		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", openStackMachine.ObjectMeta.Annotations["cluster.x-k8s.io/openstack.cloud"], *openStackServer.Status.InstanceID))
```
mdbooth, line 425: As [discussed in slack](https://kubernetes.slack.com/archives/CFKJB65G9/p1728981106884829), lets look into adding this to the API rather than as an annotation. This should be fine as long as any new field is optional, and the behaviour remains unchanged when it is not specified.

Something to think about: we permit the modification of identityRef on objects, but if we added `region` and it became part of the `providerID`, should it be immutable?

Something else to note here is that we need to preserve the immutability of `providerID`. This isn't a problem in the current code, as if we have an instance it's always going to have the same ID. With this change this is no longer true, as the behaviour can change across an upgrade. However we do this, I suggest that we add a check here and only set `ProviderID` if it is currently unset.

mdbooth:
I've spent much of the morning thinking this through and discussing with @stephenfin and Sean Mooney. Based on that, I'd like to define the following contract:

* The format of `providerID` is openstack://`region`/`server uuid`
* If region is defined, it specifically means that the server with `server uuid` was created by and is known to the Nova API in `region`.
* It explicitly does not make any direct assertions about any other OpenStack resources. It refers specifically to the Nova server.
* region may be empty, in which case consumers of providerID which require a region must have some other way to obtain it. The lack of a region cannot be used to make assumptions about region.

It's also worth restating that providerID is immutable, so any change we make to it will not be applied to existing Machines.

In case anybody else was also fuzzy about the details, here's a restatement of my understanding of OpenStack regions: regions are effectively only a key used in the catalog of services returned by keystone during authentication. Region is not used during authentication. Keystone returns all service endpoints, and the client picks one from the desired region.

This means it makes no sense to use a region returned by keystone. Instead we should record the region we used to select and endpoint. i.e. We use the *input* from the API, not the *output* from keystone.

I propose we could add `Region` as an optional, immutable value in `OpenStackIdentityReference`, perhaps:
```go
	// Region specifies an OpenStack region to use. If specified, it overrides
	// any value in clouds.yaml. If specified for an OpenStackMachine, it's
	// value will be included in providerID.
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="region is immutable"
	// +optional
	Region *string `json:"region,omitempty"`
```

I would expect us to use that value explicitly here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/934d61d38a26210b00811c9a3d354e92f32d09f7/pkg/scope/provider.go#L182-L188

as well as when setting `providerID`.

MatthieuFin:
Personally, I prefer don't use the new field `Region` from `OpenStackIdentityReference` in OS client creation https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/934d61d38a26210b00811c9a3d354e92f32d09f7/pkg/scope/provider.go#L182-L188 

If you have 2 providers which declare same region name, in this case you cannot differentiate these regions. If we keep them manage separately you could have `openstack://provider1-region1/instance_uuid` and `openstack://provider2-region1/instance_uuid`. and keep region field in client configuration to `region1` in both cases.

Currently OCCM don't allow this but maybe a day it will be supported.

MatthieuFin:
Ok, that’s good for me, let’s see if we need to explore this topic further later! And you're right, we'll probably never need to replace the region name, at least it's too early to worry about that.

api/v1beta1/identity_types.go
```diff
@@ -28,6 +28,13 @@ type OpenStackIdentityReference struct {
 	// CloudName specifies the name of the entry in the clouds.yaml file to use.
 	// +kubebuilder:validation:Required
 	CloudName string `json:"cloudName"`
+
+	// Region specifies an OpenStack region to use. If specified, it overrides
+	// any value in clouds.yaml. If specified for an OpenStackMachine, its
+	// value will be included in providerID.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="region is immutable"
```
mdbooth, line 35: Please can you add a test to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/suites/apivalidations/openstackmachine_test.go to validate that the field is indeed immutable. I think we only need to test it for a single object, and OpenStackMachine makes sense. We should test:
* The value can be set on creation
* The value cannot be modified after creation
* The value cannot be unset after creation

controllers/openstackmachine_controller.go
```diff
@@ -422,7 +422,7 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 
 		// Set properties required by CAPI machine controller
-		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack:///%s", *openStackServer.Status.InstanceID))
+		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", openStackMachine.Spec.IdentityRef.Region, *openStackServer.Status.InstanceID))
```
mdbooth, line 425: How can we get some test coverage of this?

mdbooth (commented):
This looks great, and thanks for getting this done so quickly. I'm +1 on everything here.

As mentioned we should have apivalidations coverage of all API validations. They're very easy to write and there are lots of examples to copy from.

Ideally I'd also like some e2e test coverage of this. I think we can add this fairly easily by doing some like modifying the MachineDeployment template of the workers of one of the existing e2e scenarios and adding validation in the test suite that providerID is set on the CAPI Machine the way we expect.

Ping me on Slack if I can help with this.

controllers/openstackmachine_controller.go
mdbooth: Whoops! IdentityRef is a pointer and is usually nil.

Incidentally, wouldn't these errors be so much simpler to see if you'd had to write something like `openStackMachine.Spec.IdentityRef->Region` instead? Go automatic dereferencing is insidious.

controllers/openstackmachine_controller.go
```diff
@@ -422,7 +422,7 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
 		conditions.MarkTrue(openStackMachine, infrav1.InstanceReadyCondition)
 
 		// Set properties required by CAPI machine controller
-		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", openStackMachine.Spec.IdentityRef.Region, *openStackServer.Status.InstanceID))
+		openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", (*openStackMachine.Spec.IdentityRef).Region, *openStackServer.Status.InstanceID))
```
mdbooth, line 425: That's still going to panic when IdentityRef is nil. How about something like:

```go
var region string
if openStackMachine.Spec.IdentityRef != nil {
    region = openStackMachine.Spec.IdentityRef.Region
}
openStackMachine.Spec.ProviderID = ptr.To(fmt.Sprintf("openstack://%s/%s", region, *openStackServer.Status.InstanceID))
```

MatthieuFin: Oh, IdentityRef could be nil I did'nt get It !

I'll modify that Thx.

api/v1beta1/identity_types.go
```diff
@@ -18,6 +18,7 @@ package v1beta1
 
 // OpenStackIdentityReference is a reference to an infrastructure
 // provider identity to be used to provision cluster resources.
+// +kubebuilder:validation:XValidation:rule="(!has(self.region) && !has(oldSelf.region)) || self.region == oldSelf.region",message="region is immutable"
```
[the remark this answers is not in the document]
MatthieuFin: yes, `"self == oldSelf"` as you proposed, permit to set a value on an existing machine without region field defined. To do that I had to create rule on top level object because `has(self)` seems not valid syntax.

mdbooth:
This looks great. Can you squash the commits (into a single commit with a single, coherent description) and I'll approve. Thanks!

mdbooth:
`providerID` looks normal on unaffected machines, e.g. https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2193/pull-cluster-api-provider-openstack-e2e-test/1846816338392125440/artifacts/clusters/bootstrap/resources/e2e-tk3rp9/Machine/cluster-e2e-tk3rp9-control-plane-2p6jt.yaml

```yaml
providerID: openstack:///d8a589fa-98ca-4999-92da-bf8d3a2678d3
```

As mentioned we still need some test coverage of this actually working. I'll approve on the basis that:
* it's a new feature, so it has no current users to regress
* our existing test coverage shows it does not regress users who aren't using it
* the API looks good
* manual testing by @MatthieuFin

If we want to continue working in the future, though, we'll need a test.

MatthieuFin:
I'm just currently testing on my side before squash that.

I did'nt write e2e tests but I agree we need it to be future proof.

I'm ok to try writing some tests about that, I'll probably ping you on slack for help on that purpose.

api/v1beta1/identity_types.go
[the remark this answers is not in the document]
mdbooth: Thanks. Turns out I was missing some tests in the image API and this was broken: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2197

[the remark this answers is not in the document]
MatthieuFin: Glad to have been helpful !

PR #2194, opened 2024-10-15
🌱 ORC: Move image controller into ORC

mdbooth:
This is a 'lift and shift' of the ORC image controller from CAPO into ORC, and is the first step towards making ORC:
* Independently buildable
* With no dependencies on CAPO

It effectively just copies all the CAPO code the image controller depends on into ORC, where they can gradually diverge as their requirements differ. We don't yet build a separate ORC manager, but that should now be possible with this change.

/hold


mdbooth:
I separated out https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2199 and https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2200 from this PR, so we should wait for those to merge first and rebase.

pierreprinetti:
Amazing, thank you! 
/lgtm

EmilienM:
This is good to go. Doing a last round of review and I'll merge it.

PR #2199, opened 2024-10-18
🌱 Remove pkg/utils/hash

mdbooth:
This package was hosting 2 exported functions that weren't sharing any code and each had only a single consumer. We instead move the methods to their respective points of use.

Also entirely removes the exported function HashInstanceSpec, which had no consumers.


EmilienM (approved):
/approve

lentzi90 (commented):
/lgtm

PR #2200, opened 2024-10-18
🌱 Fix verification of generated files

mdbooth:
Firstly, we weren't checking the go workspace in hack/codegen.

Secondly, we weren't verifying generated files in the orc directory at all.


EmilienM (approved):
/approve

PR #2203, opened 2024-10-22
🌱 ORC: Generate common parts of resource API

mdbooth:
Adds a very simple code generator which simply uses a go template to generate the common parts of an ORC resource API.

My preference was to do this with generics, but for now this breaks controller-tools: https://github.com/kubernetes-sigs/controller-tools/issues/844

/hold


orc/cmd/resource-generator/data/api.template
```diff
@@ -0,0 +1,161 @@
...
+// +kubebuilder:validation:MaxProperties:=1
+type {{ .Name }}Import struct {
+	// ID contains the unique identifier of an existing OpenStack resource. Note
+	// that when specifying an import by ID, the resource MUST already exist.
+	// The ORC object will enter an error state if the resource does not exist.
+	// +optional
+	// +kubebuilder:validation:Format:=uuid
+	ID *string `json:"id,omitempty"`
+
+	// Filter contains an resource query which is expected to return a single
```
lentzi90, line 35:
```suggestion
	// Filter contains a resource query which is expected to return a single
```

lentzi90 (commented):
/approve
(tiny typo below)

orc/cmd/resource-generator/main.go
```diff
@@ -0,0 +1,90 @@
...
+	SpecExtraValidations []specExtraValidation
+}
+
+var allResources []templateFields = []templateFields{
+	{
+		Name:       "Image",
+		APIVersion: "v1alpha1",
+		SpecExtraValidations: []specExtraValidation{
+			{
+				Rule:    "!has(self.__import__) ? has(self.resource.content) : true",
```
pierreprinetti, line 38: This is a good first step, but going forward we probably want a more structured way to define required fields in the spec...

pierreprinetti (commented):
This is a great first step. Thank you for setting up this generator! I hope its life will be glorious and short, until we can use generics in the API.

orc/cmd/resource-generator/main.go
mdbooth: This particular one is special because the Image controller is a snowflake. This is a controller-specific validation at the *spec* level. I don't expect this to be a common pattern.

orc/cmd/resource-generator/data/api.template
```diff
@@ -0,0 +1,161 @@
...
+// {{ .Name }}Status defines the observed state of an ORC resource.
+type {{ .Name }}Status struct {
+	// Conditions represents the observed status of the object.
+	// Known .status.conditions.type are: "Available", "Progressing"
+	//
+	// Available represents the availability of the OpenStack resource. If it is
+	// true then the resource is ready for use.
+	//
+	// Progressing indicates the state of the OpenStack resource not currently
+	// reflect the desired state, but that reconciliation is progressing.
```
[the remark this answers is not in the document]
mdbooth: It does.

orc/cmd/resource-generator/data/api.template
```diff
@@ -0,0 +1,161 @@
...
+	//
+	// Resource may not be specified if the management policy is `unmanaged`.
+	//
+	// Resource must be specified when the management policy is `managed`.
+	// +optional
+	Resource *{{ .Name }}ResourceSpec `json:"resource,omitempty"`
+
+	// ManagementPolicy defines how ORC will treat the object. Valid values are
+	// `managed`: ORC will create, update, and delete the resource; `unmanaged`:
+	// ORC will import an existing resource, and will not apply updates to it or
```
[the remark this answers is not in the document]
mdbooth: Honestly I can't think of a practical case when I'd use 'nor' outside of formal logic except: neither this not that. It would be weird here.

mdbooth:
Just discovered a templating bug in this PR while adding subnet: the spec ends up on the same line as the previous validation. Will fix.

PR #2210, opened 2024-10-28
Add e2e test to cover region in providerID clusterv1 machine object

MatthieuFin:
Add E2E test to cover region in providerID clusterv1 machine object introduced in issue https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2183 and PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2193

I have write 2 test cases:
1. One cluster object with identityRef.Region field specified in OpenStackCluster `providerID-with-region-default` and no identityRef specified in OpenStackMachineTemplate
2. One cluster object without identityRef.Region field specified in OpenstackCluster `providerID-with-region-override` and identityRef.Region specified in OpenStackMachineTemplate.

I uncounter an error on test case number 1, region name missing from clusterv1.Machine providerID string. That why I add a [2nd commit](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/750e386a0005d7fd5f211a7f2e42fc4649de1434) to fix that usecase and keep behavior of identityRef herited from Cluster Object.

It seems to do the job but it is "semantically" not really fair because I mutate object openStackMachine to append Cluster idenityRef if openStackMachine one's is nil, but this object goal of life is to mirror reality so that's technically a hack should I append a parameter to reconcileMachineState function to pass cluster object or cluster's identityRef object ?

/hold


MatthieuFin:
Outch, I thinked that CAPO_REGION env variable will be set by https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/env.rc#L89 but it seems not, personally I run tests by setting my own CAPO_REGION variable with my region name.
How could I retrieve CI region name to feed CAPO_REGION env variable ?

controllers/openstackmachine_controller.go
```diff
@@ -391,6 +391,9 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 		Address: instanceStatus.Name(),
 	})
 	openStackMachine.Status.Addresses = addresses
+	if openStackMachine.Spec.IdentityRef == nil {
```
EmilienM, line 394: I'm curious why you need to do that here. We already do it here:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2210/files#diff-e0a0984c0e26655878e2f94623fbf8cddb4d8a89e3443c1296dfaadc8802838aR578-R583

I'll like to avoid dupplications.

Makefile
```diff
@@ -176,7 +176,9 @@ e2e-templates: $(addprefix $(E2E_NO_ARTIFACT_TEMPLATES_DIR)/, \
 		 cluster-template-flatcar.yaml \
                  cluster-template-k8s-upgrade.yaml \
 		 cluster-template-flatcar-sysext.yaml \
-		 cluster-template-no-bastion.yaml)
+		 cluster-template-no-bastion.yaml \
+         cluster-template-providerID-with-region-override.yaml \
```
EmilienM, line 180: please fix alignments

test/e2e/data/kustomize/providerID-with-region-default/patch-kubeadm-control-plane-provider-id.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- op: replace
+  path: /spec/kubeadmConfigSpec/initConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+- op: replace
+  path: /spec/kubeadmConfigSpec/joinConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+
```
EmilienM, line 8: extra line not needed

test/e2e/data/kustomize/providerID-with-region-default/patch-kubeadm-config-template-provider-id.yaml
```diff
@@ -0,0 +1,5 @@
+---
+- op: replace
+  path: /spec/template/spec/joinConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+
```
EmilienM, line 5: extra line not needed

test/e2e/data/kustomize/providerID-with-region-default/patch-openstack-machine-template-identityRef.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- op: add
+  path: /spec/template/spec/identityRef
+  value: 
+    cloudName: ${OPENSTACK_CLOUD}
+    name: ${CLUSTER_NAME}-cloud-config
+    region: ${CAPO_REGION}
+
```
EmilienM, line 8: extra line not needed

test/e2e/data/kustomize/providerID-with-region-override/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+- path: patch-kubeadm-config-template-provider-id.yaml
+  target:
+    kind: KubeadmConfigTemplate
+- path: patch-kubeadm-control-plane-provider-id.yaml
+  target:
+    kind: KubeadmControlPlane
+- path: patch-openstack-machine-template-identityRef.yaml
+  target:
+    kind: OpenStackMachineTemplate
+#- path: patch-no-bastion.yaml
```
EmilienM, line 19: commented?

test/e2e/data/kustomize/providerID-with-region-override/patch-kubeadm-config-template-provider-id.yaml
```diff
@@ -0,0 +1,5 @@
+---
+- op: replace
+  path: /spec/template/spec/joinConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+
```
EmilienM, line 5: extra line not needed

test/e2e/data/kustomize/providerID-with-region-override/patch-kubeadm-control-plane-provider-id.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- op: replace
+  path: /spec/kubeadmConfigSpec/initConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+- op: replace
+  path: /spec/kubeadmConfigSpec/joinConfiguration/nodeRegistration/kubeletExtraArgs/provider-id
+  value: "openstack://${CAPO_REGION}/{{ instance_id }}"
+
+
```
EmilienM, line 9: extra line not needed

test/e2e/data/kustomize/providerID-with-region-override/patch-no-bastion.yaml
```diff
@@ -0,0 +1,5 @@
+---
+- op: replace
+  path: /spec/bastion/enabled
+  value: false
+
```
EmilienM, line 5: extra line not needed

test/e2e/data/kustomize/providerID-with-region-override/patch-openstack-machine-template-identityRef.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- op: add
+  path: /spec/template/spec/identityRef
+  value: 
+    cloudName: ${OPENSTACK_CLOUD}
+    name: ${CLUSTER_NAME}-cloud-config
+    region: ${CAPO_REGION}
+
```
EmilienM, line 8: extra line not needed

EmilienM:
Really good start, thanks for writing these tests! A few minor comments inline.

controllers/openstackmachine_controller.go
MatthieuFin: function `getOrCreateMachineServer` build an openStackServer object, that's not the purpose of this context so I didn't to use it like that.
Do you think that [this way](https://github.com/MatthieuFin/cluster-api-provider-openstack/commit/9250c73f899906b73aaf9853b3a6446856defc49) is a better approach ?

EmilienM: No it's not a better approach I think.
However moving the code to a separated function and use it in both places seems a good idea. We share the logic in case it has to change one day.

Thanks for trying that.

MatthieuFin: Obviously it seems to make sens, I'll take a look on this way.

MatthieuFin: IdentityRef lookup logic was also implmented [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/0f945e1/pkg/scope/provider.go#L58), so I tried to factorise code to use same implementation at anyplaces with this last [commit](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2210/commits/dc72b3ce230ee1f5e05c7a56d91d93514dc57c60).

pkg/scope/scope.go
```diff
@@ -43,6 +43,21 @@ func NewFactory(maxCacheSize int) Factory {
...
+func (f *defaultScopeFactory) GetIdentityRefFromObjects(objects ...infrav1.IdentityRefProvider) (*string, *infrav1.OpenStackIdentityReference) {
+	var namespace *string
+	var identityRef *infrav1.OpenStackIdentityReference
+	for _, o := range objects {
+		namespace, identityRef = o.GetIdentityRef()
+		if identityRef != nil {
+			break
+		}
+	}
+	return namespace, identityRef
```
EmilienM, line 60: I'm not trying to be picky but I want to ensure we create code that will be maintainable and debuggable. I don't see error management here. What if the identifyRef wasn't found? Shouldn't we return an error?
Or from the callers, if identityRef is nil, we should return an error?

I think either way we need to stop and send a clean error if we can't find an identityRef. Especially because we use a pointer in one of the cases and I like to see pointer checks before using them.

MatthieuFin: No problem, I ask myself about that too.
I took the choose of that the responsibility of caller since `objects ...infrav1.IdentityRefProvider` could have nil pointer to IdentityRef field, I don't wanna return an error because the value of objects could be nil as a normal behavior.

In my 2 callers functions last parameters are `openStackCluster` type of `infrav1.OpenStackCluster` which is not a pointer https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1beta1/openstackcluster_types.go#L194 so I'm sure that identityRef variable [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2210/commits/dc72b3ce230ee1f5e05c7a56d91d93514dc57c60#diff-e0a0984c0e26655878e2f94623fbf8cddb4d8a89e3443c1296dfaadc8802838aR395) and [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2210/commits/dc72b3ce230ee1f5e05c7a56d91d93514dc57c60#diff-e0a0984c0e26655878e2f94623fbf8cddb4d8a89e3443c1296dfaadc8802838aR579) and correctly defined.

But I'm agree maybe tomorrow it could be different, so I could validate function return from caller for more maintainability.

PR #2219, opened 2024-11-01
dependabot: config changes

EmilienM:

* Add blocks for main and for release-0.11
* Fix the prefix for commit messages
* Separate changes from CAPI and k8s dependencies
* Don't ignore github.com/prometheus/* bumps


lentzi90 (commented):
/approve

PR #2238, opened 2024-11-05
🐛 Drop caBundle from CRDs to support Kubernetes 1.31

reyvonger:
**What this PR does / why we need it:**

Starting with Kubernetes 1.31 it won't be possible anymore to continuously apply CRDs that are setting caBundle to an invalid value (in our case Cg==). The solution is to simply drop the caBundle field (it was never actually required by kube-apiserver).
For more details see: https://kubernetes.slack.com/archives/C0EG7JC6T/p1722441161968339

Let me know if you have any questions.

We'll want to backport this into all supported releases. It's never great to set the caBundle to an invalid value, even before Kubernetes 1.31.

This is an effort driven by CAPI: https://github.com/kubernetes-sigs/cluster-api/pull/10972

### DIFF
--- a/config/crd/patches/webhook_in_openstackclusters.yaml
+++ b/config/crd/patches/webhook_in_openstackclusters.yaml
@@ -10,9 +10,6 @@ spec:
     webhook:
       conversionReviewVersions: ["v1", "v1beta1"]
       clientConfig:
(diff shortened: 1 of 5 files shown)


EmilienM:
@reyvonger please sign the CLA
/ok-to-test

EmilienM:
/lgtm
/approve
/cherry-pick release-0.11

thanks!

PR #2252, opened 2024-11-13
:bug: Ensure tags and trunks for existing ports

mquhuy:
When a port is created, it is tagged and explicitly checked if it needs a trunk, but if the port already exists for some reason, it is taken without any of those checking. This PR fixes that by adding a similar check for the existing ports, similar to the check in new ports.

Fixes #

Technically this rarely happens, as a failure in VM creation will clean up all ports, hence we don't normally have existing ports with same name and network ID as the one we need. Still, if that happens that there exists such a port, a problem may arise.

This is an issue with release-0.9. I'm not sure if other releases may behave in the same manner.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/compute/instance_test.go
+++ b/pkg/cloud/services/compute/instance_test.go
@@ -366,7 +366,6 @@ func TestService_ReconcileInstance(t *testing.T) {
 				Description: portOpts["description"].(string),
 			}, nil
 		})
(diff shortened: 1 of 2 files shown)


lentzi90:
We have merged the fixes in main, release-0.12, release-0.11 and release-0.10. We do not need to fix v0.9 anymore as it is out of support.
Closing

PR #2256, opened 2024-11-15
:bug: Ensure that existing ports also have correct tags and trunks

mquhuy:
When a port is created, it is tagged and explicitly checked if it needs a trunk, but if the port already exists for some reason, it is taken without any of those checking. This PR fixes that by adding a similar check for the existing ports, similar to the check in new ports.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackserver_controller_test.go
+++ b/controllers/openstackserver_controller_test.go
@@ -479,6 +479,7 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
 				listDefaultPortsNotFound(r)
 				createDefaultPort(r)
 				listDefaultServerNotFound(r)
+				listDefaultPortsNotFound(r)
 				createDefaultServer(r)
 			},
 		},
@@ -499,6 +500,7 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
 				},
 			},
 			expect: func(r *recorders) {
+				listDefaultPorts(r)
 				listDefaultPorts(r)
 				listDefaultServerFound(r)
 			},
--- a/controllers/openstackserver_controller_test.go
+++ b/controllers/openstackserver_controller_test.go
@@ -114,6 +114,18 @@ var listDefaultPorts = func(r *recorders) {
 	}, nil)
 }
 
+var listDefaultPortsWithID = func(r *recorders) {
+	r.network.ListPort(ports.ListOpts{
+		Name:      openStackServerName + "-0",
+		ID:        portUUID,
+		NetworkID: networkUUID,
+	}).Return([]ports.Port{
+		{
+			ID: portUUID,
+		},
+	}, nil)
+}
+
 var listDefaultPortsNotFound = func(r *recorders) {
 	r.network.ListPort(ports.ListOpts{
 		Name:      openStackServerName + "-0",
@@ -501,7 +513,7 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
 			},
 			expect: func(r *recorders) {
 				listDefaultPorts(r)
-				listDefaultPorts(r)
+				listDefaultPortsWithID(r)
 				listDefaultServerFound(r)
 			},
 		},
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -124,7 +124,49 @@ func (s *Service) GetPortForExternalNetwork(instanceID string, externalNetworkID
 	return nil, nil
 }
 
-func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.ResolvedPortSpec) (*ports.Port, error) {
+// ensurePortTagsAndTrunk ensures that the provided port has the tags and trunk defined in portSpec.
+func (s *Service) ensurePortTagsAndTrunk(port *ports.Port, eventObject runtime.Object, portSpec *infrav1.ResolvedPortSpec) error {
+	if len(portSpec.Tags) > 0 {
+		if err := s.replaceAllAttributesTags(eventObject, portResource, port.ID, portSpec.Tags); err != nil {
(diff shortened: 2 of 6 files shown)


EmilienM:
no need to retest, the unit test failed

controllers/openstackserver_controller_test.go
```diff
@@ -85,19 +84,6 @@ var defaultPortsStatus = []infrav1.PortStatus{
 	},
 }
 
-var createDefaultPort = func(r *recorders) {
```
lentzi90, line 88: This means that we are now not testing any case where a port is created. All test cases find already existing ports. We need to cover also the case when ports are created.

controllers/openstackserver_controller_test.go
```diff
@@ -477,8 +463,8 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
 			},
 			expect: func(r *recorders) {
 				listDefaultPortsNotFound(r)
-				createDefaultPort(r)
 				listDefaultServerNotFound(r)
+				listDefaultPorts(r)
```
lentzi90, line 467: Please change this so that it finds no ports. That way we keep testing the same case as before, i.e. port creation.
```suggestion
				listDefaultPortsNotFound(r)
```

controllers/openstackserver_controller_test.go
```diff
@@ -499,6 +485,7 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
 				},
 			},
 			expect: func(r *recorders) {
+				listDefaultPorts(r)
```
lentzi90, line 488: This one is fine. The test case is supposed to test adoption :+1:

test/e2e/data/e2e_conf.yaml
```diff
@@ -198,7 +198,7 @@ intervals:
   default/wait-cluster: ["20m", "10s"]
   default/wait-control-plane: ["30m", "10s"]
   default/wait-worker-nodes: ["30m", "10s"]
-  default/wait-delete-cluster: ["5m", "10s"]
+  default/wait-delete-cluster: ["10m", "10s"]
```
lentzi90, line 201: Great find! I thought there was some issue with the code making it stuck. Turns out it was just a too tight deadline :sweat_smile:

pkg/cloud/services/networking/port_test.go
lentzi90, line 1: The test itself should be renamed also to reflect that `CreatePort` was renamed to `EnsurePort`.
Would it be possible to also add a test case for adopting an existing port?

lentzi90 (commented):
This looks pretty much ready! I just have a couple of comments on the tests.

lentzi90 (commented):
/approve
/cc @EmilienM

EmilienM:
To avoid copy/pasting my thoughts, I'll share a thread on Slack about this PR: https://kubernetes.slack.com/archives/CFKJB65G9/p1732192515903409
Before we merge this, I want to make sure we don't have another (long term) way of fixing these kind of issues which really can happen on any kind of OpenStack resources, not just ports and trunks.

pkg/cloud/services/networking/port.go
```diff
@@ -124,7 +124,49 @@ func (s *Service) GetPortForExternalNetwork(instanceID string, externalNetworkID
...
 
-func (s *Service) CreatePort(eventObject runtime.Object, portSpec *infrav1.ResolvedPortSpec) (*ports.Port, error) {
+// ensurePortTagsAndTrunk ensures that the provided port has the tags and trunk defined in portSpec.
+func (s *Service) ensurePortTagsAndTrunk(port *ports.Port, eventObject runtime.Object, portSpec *infrav1.ResolvedPortSpec) error {
+	if len(portSpec.Tags) > 0 {
+		if err := s.replaceAllAttributesTags(eventObject, portResource, port.ID, portSpec.Tags); err != nil {
+			record.Warnf(eventObject, "FailedReplaceTags", "Failed to replace port tags %s: %v", portSpec.Name, err)
+			return err
+		}
+	}
```
mdbooth, line 134: Isn't this going to call replaceAllAttributesTags on every reconcile, regardless of whether they're already set?

pkg/cloud/services/networking/port.go
```diff
@@ -124,7 +124,49 @@ func (s *Service) GetPortForExternalNetwork(instanceID string, externalNetworkID
...
+	return nil
+}
+
+// EnsurePort ensure that a port defined with portSpec Name and NetworkID exists,
+// and that the port has suitable tags and trunk.
+func (s *Service) EnsurePort(eventObject runtime.Object, portSpec *infrav1.ResolvedPortSpec) (*ports.Port, error) {
+	existingPorts, err := s.client.ListPort(ports.ListOpts{
+		Name:      portSpec.Name,
+		NetworkID: portSpec.NetworkID,
+	})
```
mdbooth, line 155: We have the port IDs of ports we've already created. Can we fetch by port ID? We should fall back to fetch by name in case we created a port but failed to write it to the status, but ideally fetch by ID would be the default here.

pkg/cloud/services/networking/port.go
lentzi90: Yes. Fixed now

pkg/cloud/services/networking/port.go
lentzi90: Done

lentzi90:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1940 has been ruled out. It looks like we are genuinely hitting a case where port creation is interrupted in the middle by rate limits. So some ports may be created and tagged, with trunk, some may be missing tags and some may be missing trunk. Cleanup will never work because if we hit rate limits in the middle of creating the ports, the API calls to clean up will also be rate limited.

lentzi90:
Downstream has confirmed that this is working for them (on the backported patch). This is good to go from my perspective.
/hold cancel

PR #2257, opened 2024-11-18
Do not delete bastion floating ip if set in spec

### LINKED ISSUE #2157: Keep Floating IP when deleting bastion host
/kind feature

I want to have the option (or as default behavior) to keep the floating IP of the bastion host when deleting it if it is specified in `openstackcluster.spec.bastion.floatingIP`. There were some discussions about keeping floating IPs in [this issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1615#issuecomment-1721241197), but this does not seem to be the case for bastions atm.

I believe [this](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/fb3779cdb1d72a02f03ddf5eaa21d96c912a2cf5/controllers/openstackcluster_controller.go#L247-L280) is the part that would need to be changed.

anders-elastisys:

This PR makes it so that if an floatingIP is set in the `openstackcluster.spec.bastion.floatingIP`, and it is the same as set in the `status` of the `openstackcluster` resource, it will not be deleted when deleting the cluster.

Fixes #2157 

My issue got stale so I am creating this PR so that hopefully this feature does not get forgotten.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -252,14 +252,14 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 	var statusFloatingIP *string
 	var specFloatingIP *string
 	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
-		// Floating IP set in status
 		statusFloatingIP = &openStackCluster.Status.Bastion.FloatingIP
 	}
 	if openStackCluster.Spec.Bastion.FloatingIP != nil {
-		// Floating IP from the spec
 		specFloatingIP = openStackCluster.Spec.Bastion.FloatingIP
 	}
 
+	// We only remove the bastion's floating IP if it exists and if it's not the same value defined both in the spec and in status.
+	// This decision was made so if a user specifies a pre-created floating IP that is intended to only be used for the bastion, the floating IP won't get removed once the bastion is destroyed.
 	if statusFloatingIP != nil && (specFloatingIP == nil || *statusFloatingIP != *specFloatingIP) {
 		if err = networkingService.DeleteFloatingIP(openStackCluster, openStackCluster.Status.Bastion.FloatingIP); err != nil {
 			handleUpdateOSCError(openStackCluster, fmt.Errorf("failed to delete floating IP: %w", err), false)
@@ -286,7 +286,8 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 
 		for _, address := range addresses {
 			if address.Type == corev1.NodeExternalIP {
-				// If a floating IP is set for the bastion spec, skip deleting it
+				// If a floating IP retrieved is the same as what is set in the bastion spec, skip deleting it.
+				// This decision was made so if a user specifies a pre-created floating IP that is intended to only be used for the bastion, the floating IP won't get removed once the bastion is destroyed.
 				if specFloatingIP != nil && address.Address == *specFloatingIP {
 					continue
 				}
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -249,7 +249,18 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 		return err
 	}
 
+	var statusFloatingIP *string
+	var specFloatingIP *string
 	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		// Floating IP set in status
+		statusFloatingIP = &openStackCluster.Status.Bastion.FloatingIP
+	}
+	if openStackCluster.Spec.Bastion.FloatingIP != nil {
+		// Floating IP from the spec
+		specFloatingIP = openStackCluster.Spec.Bastion.FloatingIP
+	}
+
+	if statusFloatingIP != nil && (specFloatingIP == nil || *statusFloatingIP != *specFloatingIP) {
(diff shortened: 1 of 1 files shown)


mdbooth (commented):
/ok-to-test
/lgtm

@EmilienM I think you're the expert in this code.

controllers/openstackcluster_controller.go
```diff
@@ -244,7 +244,18 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 		return err
 	}
 
+	var statusFloatingIP *string
+	var specFloatingIP *string
 	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		// Floating IP set in status
```
EmilienM, line 250: this comment doesn't bring value, remove it please.

controllers/openstackcluster_controller.go
```diff
@@ -244,7 +244,18 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
...
+		statusFloatingIP = &openStackCluster.Status.Bastion.FloatingIP
+	}
+	if openStackCluster.Spec.Bastion.FloatingIP != nil {
+		// Floating IP from the spec
```
EmilienM, line 254: this comment doesn't bring value, remove it please.

controllers/openstackcluster_controller.go
```diff
@@ -244,7 +244,18 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
...
+		specFloatingIP = openStackCluster.Spec.Bastion.FloatingIP
+	}
+
+	if statusFloatingIP != nil && (specFloatingIP == nil || *statusFloatingIP != *specFloatingIP) {
```
EmilienM, line 258: Here please add a comment on what opinionated decision we make.

```
We only remove the bastion's floating IP if it exists and if it's the not same value defined both in the spec and in status. This decision was made so if a user specifies a pre-created floating IP that is intended to be used for the bastion, the floating IP won't get removed once the bastion is destroyed.
```

Or something like that.

controllers/openstackcluster_controller.go
```diff
@@ -270,6 +281,10 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 
 		for _, address := range addresses {
 			if address.Type == corev1.NodeExternalIP {
+				// If a floating IP is set for the bastion spec, skip deleting it
```
EmilienM, line 284: same here, put more context in this comment.

EmilienM (changes requested):
Thanks for your contribution.

I've put some comments inline.
I don't want to be picky but I honestly think we should have e2e coverage for this PR, simply because the code around bastion has proven to break quite easily and I've spent hours fixing it.

Please have a look at how we test a flavor change for the bastion:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/300f85079c9e336986d9637f2b03dce02c824efb/test/e2e/suites/e2e/e2e_test.go#L277

and see if we can do the same kind of testing:
* Create a FIP using Gophercloud
* Update the cluster spec to provide that FIP in Bastion
* Check that the bastion uses that FIP
* Delete the Bastion, make sure the FIP was not deleted
* Delete the FIP

Let me know if it's too much asked, I can help.
If you want to do the tests as a follow-up, we can also do that and in that case I'll approve this PR.

anders-elastisys:
> Let me know if it's too much asked, I can help. If you want to do the tests as a follow-up, we can also do that and in that case I'll approve this PR.

Thanks for the comments, I will take try to add e2e tests when I get the time, if I am not successful in this I'll let you know.

EmilienM:
> > Let me know if it's too much asked, I can help. If you want to do the tests as a follow-up, we can also do that and in that case I'll approve this PR.
> 
> Thanks for the comments, I will take try to add e2e tests when I get the time, if I am not successful in this I'll let you know.

awesome; thanks. And if we don't manage it to do it in a timely manner, I can merge as is and do the tests afterwards.

anders-elastisys:
@EmilienM I updated the code comments as you suggested https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2257/commits/b0ba0469d39fa8ea1659e5a7d3802b90827881e9, I have not had the time to do the tests sadly, can we do this as a separate task or can someone else add this?

controllers/openstackcluster_controller.go
```diff
@@ -249,7 +249,18 @@ func (r *OpenStackClusterReconciler) deleteBastion(ctx context.Context, scope *s
 		return err
 	}
 
+	var statusFloatingIP *string
+	var specFloatingIP *string
 	if openStackCluster.Status.Bastion != nil && openStackCluster.Status.Bastion.FloatingIP != "" {
+		statusFloatingIP = &openStackCluster.Status.Bastion.FloatingIP
+	}
+	if openStackCluster.Spec.Bastion.FloatingIP != nil {
```
alexandrevilain, line 257:
```suggestion
	if openStackCluster.Spec.Bastion != nil && openStackCluster.Spec.Bastion.FloatingIP {
```
Looks like this is the reason of failing e2e.

alexandrevilain (changes requested):
Hi!

This PR is pretty cool and would help users using Openstack on public cloud providers. In public cloud senarios we don't have enough rights to order again the same floating IP.

We already debated about this issue here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1615#issuecomment-1719670115

@anders-elastisys feel free to ping me if you need help implement e2e tests asked by @EmilienM

lentzi90:
/retitle :bug: Do not delete bastion floating ip if set in spec

PR #2258, opened 2024-11-18
🌱 CI: bump devstack to 2024.2 (Dalmatian)

EmilienM:

This PR bumps the version of OpenStack in our CI to test the latest maintained version.

### DIFF
--- a/scripts/ci-conformance.sh
+++ b/scripts/ci-conformance.sh
@@ -47,7 +47,7 @@ trap cleanup EXIT
 
 apt-get update -y
 # Install requests module explicitly for HTTP calls.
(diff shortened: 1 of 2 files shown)


EmilienM:
The version of setuptools is too new or our version of packaging is too old.
https://review.opendev.org/q/topic:%22fix-setuptools-canonicalize-version%22

I couldn't find how to pin things in the CI job here, but since this isn't urgent I'll look at it later unless someone has an idea.

EmilienM:
We probably want/need https://review.opendev.org/c/openstack/devstack/+/936961 first

PR #2285, opened 2024-11-26
🌱 Dependabot conf changes

EmilienM:
Both main and 0.11:
* Remove `open-pull-requests-limit`.
* Enforce grouping, it opens too many PRs and it's painful to merge (they
  merge conflict from each other every week).

Only for 0.11: allow CAPI bumps but only patch versions.


lentzi90 (commented):
/lgtm

EmilienM:
/approve
let's go with it and see how it works. We might have to adjust again :)

PR #2286, opened 2024-11-26
🌱 Add Image templates and update dev docs

lentzi90:

This adds a template for ORC managed Images. The template is meant as a simple way for developers to get a bastion and node image loaded into their openstack environment. Docs are also updated to reflect this.

The ClusterClass template is also updated to make it more useful. Most notably, this adds variables for identityRef and for specifying the image name.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/templates/images-template.yaml
+++ b/templates/images-template.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: openstack.k-orc.cloud/v1alpha1
+kind: Image
+metadata:
+  name: node-image
+spec:
+  managementPolicy: managed
+  resource:
+    name: ubuntu-2404-kube-v1.31.2
+    content:
+      diskFormat: qcow2
+      download:
+        url: ${NODE_IMAGE_URL:="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2024-11-21/ubuntu-2404-kube-v1.31.2.img"}
+  cloudCredentialsRef:
+    secretName: dev-test-cloud-config
+    cloudName: capo-e2e
+---
+apiVersion: openstack.k-orc.cloud/v1alpha1
+kind: Image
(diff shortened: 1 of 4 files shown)


lentzi90:
I'm working my way up to adding a test that makes use of this template. With this I can at least comfortably use it in local development. Hopefully, next PR will be to add an e2e test :slightly_smiling_face:

templates/images-template.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: node-image
+spec:
+  managementPolicy: managed
+  resource:
+    name: ubuntu-2404-kube-v1.31.2
+    content:
+      diskFormat: qcow2
+      download:
+        url: ${NODE_IMAGE_URL:="https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/ubuntu/2024-11-21/ubuntu-2404-kube-v1.31.2.img"}
```
lentzi90, line 13: I want to point this out to make sure it is ok. I have here linked to the image we (are about to) use in our e2e tests. I assume that no one will use this for any kind of serious work, but I hope it is ok for developers to use it. It is anyway not a secret where we pull it from.
@mdbooth do you know if it will generate any kind of costs for the project?

mdbooth: I believe it does generate costs for the project, and I *think* we're discouraged from doing this. I wonder if it's worth asking in... #sig-k8s-infra? Perhaps it's ok, or perhaps there's a better place to host these things.

mdbooth: Can't we do flatcar images with sysext completely from their public repos? If so, I'd be ok making flatcar our default repo here.

cc @tormath1

PR #2295, opened 2024-11-29
🌱 E2E: Use CAPI production images and skip cert-manager pre-pull

lentzi90:

Use CAPI production images instead of from GCR.

Don't pre-pull cert-manager. We always let these images get out of sync, causing us to pull double instead of saving anything. We end up with different tags in e2e_conf.yaml and in the Makefile. Additionally, CAPI may change what version is installed by clusterctl by default, causing us to pull *two* wrong versions and then install a third if we don't remember to check when bumping CAPI.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,12 +209,9 @@ e2e-image: docker-build
 
 # Pull all the images references in test/e2e/data/e2e_conf.yaml
 test-e2e-image-prerequisites:
-	docker pull gcr.io/k8s-staging-cluster-api/cluster-api-controller:v1.6.0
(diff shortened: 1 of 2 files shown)


EmilienM:
that's a great change, thanks a lot for taking care of that @lentzi90 .
/approve

I think we can backport it as the version of CAPI is the same for now.
/cherry-pick release-0.11

PR #2338, opened 2024-12-18
✨ feat: reconcile pre-existing router

simonostendorf:

Currently only pre existing networks and subnets are reconciled and not a pre existing router.

I think it is a good idea to also reconcile pre-existing routers and put them in the status. This is useful for debugging, reading info from the status (e.g. outgoing router ip of the cluster) and if the router ip is written to the status, it can be used for the lb allowlist.

none

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -695,10 +695,10 @@ func reconcileNetworkComponents(scope *scope.WithLogger, cluster *clusterv1.Clus
 }
 
 // reconcilePreExistingNetworkComponents reconciles the cluster network status when the cluster is
-// using pre-existing networks and subnets which are not provisioned by the
+// using pre-existing networks, subnets and router which are not provisioned by the
 // cluster controller.
 func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingService *networking.Service, openStackCluster *infrav1.OpenStackCluster) error {
-	scope.Logger().V(4).Info("No need to reconcile network, searching network and subnet instead")
+	scope.Logger().V(4).Info("No need to reconcile network, searching network, subnet and router instead")
 
 	if openStackCluster.Status.Network == nil {
 		openStackCluster.Status.Network = &infrav1.NetworkStatusWithSubnets{}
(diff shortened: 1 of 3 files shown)


controllers/openstackcluster_controller.go
```diff
@@ -740,6 +740,33 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
 		setClusterNetwork(openStackCluster, network)
 	}
 
+	// add pre-existing router to status if router spec is defined
```
EmilienM, line 743: I'm not sure about this comment. Either make it better or remove it, thanks.

controllers/openstackcluster_controller.go
```diff
@@ -690,10 +690,10 @@ func reconcileNetworkComponents(scope *scope.WithLogger, cluster *clusterv1.Clus
 }
 
 // reconcilePreExistingNetworkComponents reconciles the cluster network status when the cluster is
-// using pre-existing networks and subnets which are not provisioned by the
+// using pre-existing networks, subnets and router which are not provisioned by the
```
EmilienM, line 693: other words are plural, so either remove "s" on others or add "s" to router. thanks

controllers/openstackcluster_controller.go
```diff
@@ -740,6 +740,33 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
...
+		}
+
+		openStackCluster.Status.Router = &infrav1.Router{
+			Name: router.Name,
+			ID:   router.ID,
+			Tags: router.Tags,
+			IPs:  routerIPs,
+		}
+	}
+
```
EmilienM, line 769: While we are here, I'll like someone (maybe you via this PR, if you don't mind) to document `reconcileProvisionedNetworkComponents` as well. For someone reading these blocks for the first time, it can be confusing why we have these 2 similar functions. Thanks

controllers/openstackcluster_controller.go
```diff
@@ -740,6 +740,33 @@ func reconcilePreExistingNetworkComponents(scope *scope.WithLogger, networkingSe
 		setClusterNetwork(openStackCluster, network)
 	}
 
+	// add pre-existing router to status if router spec is defined
+	if openStackCluster.Status.Router == nil {
+		openStackCluster.Status.Router = &infrav1.Router{}
+	}
```
EmilienM, line 746: what happens if we don't do that?

EmilienM:
Sorry it took me time to review it. I think this is mostly good, I have a few comments inline.
Thanks for taking the time to fix this use case!

controllers/openstackcluster_controller.go
simonostendorf: I didnt add an s to router because its only one router. Network and subnet can be multiple due to the lb and the main network and subnet.

controllers/openstackcluster_controller.go
simonostendorf: I dont remember why I added this. But when I read this now I think it doesnt make sense because it forces the user to provide a router reference otherwise they would get logs that the router couldnt be found over and over again.

controllers/openstackcluster_controller.go
simonostendorf: I added a comment that is similar to the comment for `reconcilePreExistingNetworkComponents`.

So we have the same discussion about plural s or not here, but I will adjust it to match the one on `reconcilePreExistingNetworkComponents` after we discussed how this should look like.

controllers/openstackcluster_controller.go
simonostendorf: thanks for the comment, I removed it because other code snmippets like that are not commented either.

controllers/openstackcluster_controller.go
EmilienM: fair enough !

controllers/openstackcluster_controller.go
EmilienM: yeah this block doesn't make sense to me.

simonostendorf:
@EmilienM I fixed the router spec thing, should I squash the commits before merge?

EmilienM:
> @EmilienM I fixed the router spec thing, should I squash the commits before merge?

please squash, yes. Then I'll merge it.

/approve

simonostendorf:
@EmilienM CI is ready, do you think this can be merged or is something missing?

PR #2352, opened 2025-01-08
:bug: New IPAddressClaim.ObjectMeta.Labels nil map assign panic

mikaelgron:

When an IPAddressClaim is created and the openstackserver has a cluster name label, the capo controller panics because the Labels map is nil by default.

This PR creates an empty Lables map on new IPAddressClaim ObjectMeta.

**TODOs**:

- [X] squashed commits

/hold

### DIFF
--- a/controllers/openstackserver_controller.go
+++ b/controllers/openstackserver_controller.go
@@ -602,6 +602,7 @@ func (r *OpenStackServerReconciler) getOrCreateIPAddressClaimForFloatingAddress(
 				},
 			},
 			Finalizers: []string{infrav1.IPClaimMachineFinalizer},
+			Labels:     map[string]string{},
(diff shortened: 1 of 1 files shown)


mdbooth:
The linter didn't like the formatting:
```
controllers/openstackserver_controller.go:605: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/k-orc/openstack-resource-controller) -s prefix(sigs.k8s.io/cluster-api-provider-openstack) (gci)
			Labels: map[string]string{},
```

Can't immediately see why, but if you run `make lint-update` it will almost certain just fix it to make itself happy.

mikaelgron:
@mdbooth 'twas indentation

PR #2355, opened 2025-01-09
🌱 e2e: bump to ubuntu-2404-kube-v1.31.2

EmilienM:

Let's bump k8s release to v1.31.2 to align with other platforms
and the k8s version of CAPI & CAPO.


lentzi90 (commented):
/lgtm

PR #2376, opened 2025-01-20
📖 Add Image templates, improve dev-test clusterclass and update dev docs

lentzi90:

This adds a template for ORC managed Images. The template is meant as a simple way for developers to get a bastion and node image loaded into their openstack environment. Docs are also updated to reflect this.

The ClusterClass template is also updated to make it more useful. Most notably, this adds variables for identityRef and for specifying the image name. It is possible to choose between the traditional image including kubernetes components, or using e.g. flatcar with a sysext overlay. The sysext overlay patch is only for ignition for now.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [x] includes documentation
  - [ ] adds unit tests

### DIFF
--- a/templates/images-template.yaml
+++ b/templates/images-template.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: openstack.k-orc.cloud/v1alpha1
+kind: Image
+metadata:
+  name: node-image
+spec:
+  managementPolicy: managed
+  resource:
+    name: flatcar_production
+    content:
+      diskFormat: qcow2
+      download:
+        url: ${NODE_IMAGE_URL:="https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_openstack_image.img"}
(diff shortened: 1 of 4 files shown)


templates/images-template.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: bastion-image
+spec:
+  managementPolicy: managed
+  resource:
+    name: ubuntu-22.04
+    content:
+      diskFormat: qcow2
+      download:
+        url: ${BASTION_IMAGE_URL:="https://cloud-images.ubuntu.com/releases/jammy/release/ubuntu-22.04-server-cloudimg-amd64.img"}
```
EmilienM, line 29: nit: to be consistent with other ubuntu images, we could have given the link of 24.04.

EmilienM:
/lgtm
not blocking it

PR #2378, opened 2025-01-20
🌱 dependabot: skip code-generator updates on 0.11

EmilienM:

* Delete `orc/` directory in main, it doesn't exist anymore.
* Skip new major and minor releases for code-generator, it requires k8s
  api v1.31 and beyond which we don't want in release-0.11.


lentzi90 (commented):
/lgtm

PR #2388, opened 2025-01-22
🌱 hack/tools: bump gotestsum and golangci to latest tags

EmilienM:

Bump golangci lint to v1.63.4 and gotestsum to 1.12.0 (latest tags).


lentzi90 (commented):
/lgtm

PR #2424, opened 2025-02-12
🌱 go.mod: stick to 1.23.0

EmilienM:

Specifying the patch version is not needed and can causes problems on systems which haven't updated yet to the 1.23.4.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-openstack
 
-go 1.23.4
+go 1.23.0
 
 require (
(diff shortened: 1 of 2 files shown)


mdbooth:
/lgtm

I wonder if we can document this somewhere. Perhaps a comment in go.mod not to specify a patch version other than `.0` unless we've positively determined that it's required?

EmilienM:
I think it is a best practice to let .0 so we don't need to document that for now.
This was clearly a mistake when I did this though.
I think we're good to go
/approve
/cherry-pick release-0.12

PR #2445, opened 2025-03-04
🌱(deps): pin crypto for stable 0.11

EmilienM:

New versions of crypto require a new version of Go (1.23) and we don't want to bump go in release-0.11.


lentzi90 (commented):
/lgtm

PR #2450, opened 2025-03-05
🌱 ci: add GH workflow to automate golangci-lint bumps

EmilienM:

Add a new GH action to periodically bump the golangci-lint when a new tag is available.

### DIFF
--- a/.github/workflows/update-golangci-lint.yaml
+++ b/.github/workflows/update-golangci-lint.yaml
@@ -1,0 +1,42 @@
+name: Update golangci-lint version
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Neat!
I was going to say that we could use [renovate](https://github.com/renovatebot/renovate) for these things, but honestly this simple workflow is easier to understand and debug.
/lgtm

EmilienM:
wait let me check renovate first!

EmilienM:
Sounds like renovate is not allowed in the kubernetes-sigs organization, I sent a request but I suggest we don't wait. If it's authorized we might use it. This simple workflow would do the job now. Moreover, we might have additional bumps we want to make over time.
/approve

PR #2453, opened 2025-03-05
🌱 Fix workflows/update-golangci-lint

EmilienM:

Improve the workflow and make it work.


EmilienM:
@lentzi90 ready for review

.github/workflows/update-golangci-lint.yaml
```diff
@@ -1,42 +1,56 @@
...
-          default_author: github_actor
-          message: "chore: bump golangci-lint to v${LATEST_VERSION}"
+          commit-message: "This commit updates golangci-lint to version v${{ steps.get_version.outputs.latest_version }}."
+          title: ":seedling: chore: bump golangci-lint to v${{ steps.get_version.outputs.latest_version }}"
+          body: |
+            This PR was generated by GH workflow `update-golangci-lint` and
+            updates golangci-lint to version v${{ steps.get_version.outputs.latest_version }}.
+            Manual cherry-pick is needed on release-0.11.
+
+            /cherry-pick release-0.12
```
lentzi90, line 52: I would maybe leave this out just to avoid having to update the workflow on every new minor release

lentzi90 (commented):
/lgtm

PR #2457, opened 2025-03-05
🌱(deps): pin envsubst

EmilienM:

Latest envsubst requires go 1.24 and we are not ready for it yet.


EmilienM:
/hold
under discussion

mdbooth (approved):

Issue #2474, opened 2025-03-11
Panic after Deleting VM from Horizon


KTAN119:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Panic in the capo logs after deleting VM from horizon.

**What did you expect to happen:**
Panic should not happen and openstackmachine state should be error

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Logs from capo
`E0311 04:09:29.731101       1 signal_unix.go:917] "Observed a panic" controller="openstackmachine" controllerGroup="[infrastructure.cluster.x-k8s.io](http://infrastructure.cluster.x-k8s.io/)" controllerKind="OpenStackMachine" OpenStackMachine="default/21ee23f0-3d47-4b08-b263-3b0449f775f1-knd67-qdvb5" namespace="default" name="21ee23f0-3d47-4b08-b263-3b0449f775f1-knd67-qdvb5" reconcileID="3ad89599-1bd3-4f00-acb5-aa785d50ba84" panic="runtime error: invalid memory address or nil pointer dereference" panicGoValue="\"invalid memory address or nil pointer dereference\"" stacktrace=<
	goroutine 442 [running]:
	[k8s.io/apimachinery/pkg/util/runtime.logPanic({0x251cdd0](http://k8s.io/apimachinery/pkg/util/runtime.logPanic(%7B0x251cdd0), 0xc000de0bd0}, {0x1e79fe0, 0x3786d70})
		/go/pkg/mod/k8s.io/apimachinery@v0.31.5/pkg/util/runtime/runtime.go:107 +0xbc
	[sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile.func1()](http://sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile.func1())
		/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.5/pkg/internal/controller/controller.go:105 +0x112
	panic({0x1e79fe0?, 0x3786d70?})
		/usr/local/go/src/runtime/panic.go:785 +0x132
	[sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*InstanceStatus).NetworkStatus(0x0)](http://sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/services/compute.(*InstanceStatus).NetworkStatus(0x0))
[...]

PR #2475, opened 2025-03-11
Fix panic when OpenStack server is deleted by an external agent

mdbooth:
The panic happened because we did not correctly handle that GetInstanceStatus returns a nil server if the server does not exist, rather than a 404 error.

We had the same oversight in the OpenStackServer controller. It looks like this would have resulted in recreating the server.

Fixes: #2474

/hold

### DIFF
--- a/controllers/openstackserver_controller_test.go
+++ b/controllers/openstackserver_controller_test.go
@@ -31,8 +31,11 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
 	. "github.com/onsi/gomega" //nolint:revive
 	"go.uber.org/mock/gomock"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 
 	infrav1alpha1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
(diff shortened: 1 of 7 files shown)


mdbooth:
@EmilienM Can you take a look at this? I'm slightly uncomfortable with the interactions between the machine and server controllers here. Specifically the machine controller is re-litigating the value of `openStackServer.Status.InstanceState` in `reconcileMachineState`. The problem with that is that in this specific case that value is meaningless[^1] because it refers to the last observed state before it was deleted. We should probably just copy the relevant conditions from server to machine, and let the server controller decide what the status means.

I was also wondering if there's an easy way to add an e2e test for this. What do you think?

[^1]: Incidentally, if we set the status in an SSA transaction InstanceState would have become unset, which would make more sense.

EmilienM:
> @EmilienM Can you take a look at this? I'm slightly uncomfortable with the interactions between the machine and server controllers here. Specifically the machine controller is re-litigating the value of `openStackServer.Status.InstanceState` in `reconcileMachineState`. The problem with that is that in this specific case that value is meaningless[1](#user-content-fn-1-1bbddb0f33f84398ef2d178bc40de289) because it refers to the last observed state before it was deleted. We should probably just copy the relevant conditions from server to machine, and let the server controller decide what the status means.
> 
> I was also wondering if there's an easy way to add an e2e test for this. What do you think?
> ## Footnotes
> 
>     1. Incidentally, if we set the status in an SSA transaction InstanceState would have become unset, which would make more sense. [↩](#user-content-fnref-1-1bbddb0f33f84398ef2d178bc40de289)

as I said on Slack, this PR looks good to me, we probably want to sync the conditions at some point.

controllers/openstackserver_controller_test.go
```diff
@@ -548,3 +551,182 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
...
+				},
+			},
+			setupMocks: func(r *recorders) {
+				r.compute.GetServer(instanceUUID).Return(nil, fmt.Errorf("error"))
+			},
+			wantErr: true,
+			wantCondition: &clusterv1.Condition{
+				Type:    infrav1.InstanceReadyCondition,
+				Status:  corev1.ConditionFalse,
+				Reason:  infrav1.InstanceNotFoundReason,
```
mdbooth, line 596: Out of scope of this change, and with the way we treat these conditions probably not terribly important, but returning InstanceNotFoundReason here for anything other than a 404 is not correct. e.g We shouldn't report InstanceNotFound for an authentication error which prevented us from fetching the instance.

That's a follow-on if we address it at all, though.

controllers/openstackserver_controller_test.go
```diff
@@ -548,3 +551,182 @@ func Test_OpenStackServerReconcileCreate(t *testing.T) {
...
+			setupMocks: func(r *recorders) {
+				r.compute.ListServers(servers.ListOpts{
+					Name: "^" + openStackServerName + "$",
+				}).Return([]servers.Server{}, nil)
+				r.compute.CreateServer(gomock.Any(), gomock.Any()).Return(&servers.Server{ID: instanceUUID}, nil)
+			},
+			wantErr: false,
+			wantServer: &servers.Server{
+				ID: instanceUUID,
+			},
```
mdbooth, line 641: The asymmetry of not setting InstanceReadyCondition here is weird, but it's weirdly asymmetric in the function under test.

controllers/openstackserver_controller_test.go
EmilienM: I'm fixing this here.

controllers/openstackserver_controller_test.go
EmilienM: I know! The condition is set later after getOrCreateServer is called :cry: 
I'll propose a move of that in this PR.

EmilienM: mhh actually we reconcile until the server has been created and then we check its state later in the switch:

https://github.com/shiftstack/cluster-api-provider-openstack/blob/22c9b36db47a84b4aa9a0a0c1d1d90f8f7418287/controllers/openstackserver_controller.go#L367-L392

EmilienM:
I had a look at copying Conditions from OpenStackServer and it's a can of :worm: to be honest, I suggest we tackle this later.

EmilienM:
@mdbooth I think it worked: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2475/pull-cluster-api-provider-openstack-e2e-test/1899941891034583040/artifacts/clusters/bootstrap/resources/e2e-4g1zo5/OpenStackMachine/cluster-e2e-4g1zo5-control-plane-9g925.yaml

```yaml
status:
  addresses:
  - address: 10.6.0.211
    type: InternalIP
  - address: cluster-e2e-4g1zo5-control-plane-9g925
    type: InternalDNS
  conditions:
  - lastTransitionTime: "2025-03-12T22:51:08Z"
    message: server has been unexpectedly deleted
    reason: InstanceDeleted
    severity: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2025-03-12T22:47:14Z"
    status: "True"
    type: APIServerIngressReadyCondition
  - lastTransitionTime: "2025-03-12T22:51:08Z"
    message: server has been unexpectedly deleted
    reason: InstanceDeleted
    severity: Error
    status: "False"
    type: InstanceReady
  failureMessage: server has been unexpectedly deleted
  failureReason: UpdateError
  instanceID: 25be8125-214b-473a-8f02-9b3ad44d424b
  ready: true
```

EmilienM:
//hold cancel

EmilienM:
just in case

lentzi90 (commented):
/lgtm

EmilienM:
/lgtm

I'll let Matt approve it.

mdbooth:
Thanks!

/approve

lentzi90:
/retitle :bug: Fix panic when OpenStack server is deleted by an external agent

PR #2478, opened 2025-03-14
Don't set OSMachine Ready until all config is complete

mdbooth:
With this change, an OpenStackMachine will not be marked Ready until it has been completely initialised, including setting its Addresses and adding it to the APIServer loadbalancer if required.

Also fix a missing early return when the instance has been unexpectedly deleted.

/hold


lentzi90:
/retitle :bug: Don't set OSMachine Ready until all config is complete

PR #2479, opened 2025-03-14
ci: build e2e prerequisites in parallel

mdbooth:
With this change we build the prerequisites of the e2e tests in parallel with the devstack build. We also upload the e2e image to devstack while devstack is still building, as soon as SSH is available.

This seems to result in an approximately 10 minute speed-up.

/hold


mdbooth:
@EmilienM @lentzi90 That last run finished in 54 minutes. 

Prior test runs excluding ones from this PR have been:
* 65
* 68
* 64
* 68
* 70
* 65
* 63

So there's a fair bit of variation there, but 54 minutes is definitely a chunk faster. What do you think?

mdbooth:
For good measure:

EmilienM:
/lgtm
ship it, nice job!

lentzi90:
/retitle :seedling: ci: build e2e prerequisites in parallel

PR #2506, opened 2025-04-02
✨ feat: update ORC to v2

EmilienM:

Update ORC to v2


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/gophercloud/gophercloud/v2 v2.6.0
 	github.com/gophercloud/utils/v2 v2.0.0-20241209100706-e3a3b7c07d26
 	github.com/hashicorp/go-version v1.7.0
-	github.com/k-orc/openstack-resource-controller v1.0.1
+	github.com/k-orc/openstack-resource-controller/v2 v2.0.0
```
mandre, line 14: you'll want v2.0.2.

test/e2e/data/e2e_conf.yaml
```diff
@@ -137,7 +137,7 @@ providers:
 - name: openstack-resource-controller
   type: RuntimeExtensionProvider # ORC isn't a provider but we fake it so it can be handled by the clusterctl machinery.
   versions:
-  - name: v1.0.99
+  - name: v2.0.99
```
lentzi90, line 140: I think we will need to keep two versions here and configure the test to init with v1.0. During the test it will then upgrade ORC and CAPO to the latest version. The issue we get now is that we are trying to run CAPO v1.10 with ORC v2 and those are not compatible

mdbooth: ORC v2 should be compatible with CAPO v1.10. I think the issue we're hitting is that the k8s version in the 'upgrade from' is too old. The error message we see is:

```
Message: "CustomResourceDefinition.apiextensions.k8s.io \"securitygroups.openstack.k-orc.cloud\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[resource].properties[rules].items.x-kubernetes-validations[3].rule: Invalid value: apiextensions.ValidationRule{Rule:\"!has(self.remoteIPPrefix) || (isCIDR(self.remoteIPPrefix) && cidr(self.remoteIPPrefix).ip().family() == 4 && self.ethertype == 'IPv4') || (isCIDR(self.remoteIPPrefix) && cidr(self.remoteIPPrefix).ip().family() == 6 && self.ethertype == 'IPv6')\", Message:\"remoteIPPrefix should be a valid CIDR and match the ethertype\", MessageExpression:\"\", Reason:(*apiextensions.FieldValueErrorReason)(nil), FieldPath:\"\", OptionalOldSelf:(*bool)(nil)}: compilation failed: ERROR: <input>:1:37: undeclared reference to 'isCIDR' (in container '')
```

This is a failure to load an ORC v2 CRD because it's using a CEL expression (`isCIDR`) which this version of k8s doesn't have. ORC explicitly requires k8s v1.29+. We should probably hold off on increasing that requirement for a while, though.

However, I wonder if your point about upgrading ORC at the same time as CAPO during the upgrade test stands 🤔 Ideally we would test that in ORC, but I think there would be no harm in testing our specific use case in the upgrade test. I think we should do this.

lentzi90: Oh ok, that makes sense!

lentzi90: Wait a second. The oldest version we use is v1.30.1. Is ORC requiring v1.31+?

lentzi90: Tested this:
```bash
kind create cluster --image kindest/node:v1.30.10@sha256:4de75d0e82481ea846c0ed1de86328d821c1e6a6a91ac37bf804e5313670e507
kubectl apply -f https://github.com/k-orc/openstack-resource-controller/releases/download/v2.0.2/install.yaml
```
fails with the same error. Looks like ORC v2 doesn't support k8s v1.30. That is a bit much to ask from our users I would say. We should be fine with v1.30 until EOL at least (2025-06-28). We could possible take it in on main but I don't think we should do any releases requiring v1.31+ yet

mdbooth: Hmm, looks like we *should* be installing v1.30.1, which *should* be compatible:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/49d55d77ba308149654b4e2cf8592586aa5ab4eb/test/e2e/data/e2e_conf.yaml#L155-L157

However, I'm in little doubt that the above error is a k8s compatibility issue. I suspect that either:
* The ORC minimum version is wrong and it should really be v1.31, or
* We're somehow installing the wrong k8s version in the CAPO e2e test

The former feels more likely.

mdbooth: And indeed I have reproduced the issue installing ORC on v1.30. We have our minimum version wrong. It should be v1.31.

@EmilienM @lentzi90 How do we feel about v1.31 as a CAPO requirement. Is that going to cause problems? I guess right now there's no reason to *require* users to install ORC v2. If they continue to use ORC v1 the minimum k8s requirement stays at v1.29.

cc @mandre

lentzi90: As I said above, I don't think we should require v1.31 until v1.30 is EOL later this summer. I'm fine having newer versions on main though, but I would not want to do a CAPO release requiring v1.31 while v1.30 is still supported.

EmilienM: Very much agree with @lentzi90 ^

mandre: We should remove the CEL validation that causes the issue. I've pushed https://github.com/k-orc/openstack-resource-controller/pull/336, then we'll need to make a 2.0.3 release.

mandre: [v2.0.3](https://github.com/k-orc/openstack-resource-controller/releases/tag/v2.0.3) restored kubernetes 1.29 compatibility.

EmilienM:
Please review again

mandre (commented):
/lgtm

lentzi90 (commented):
/approve

PR #2511, opened 2025-04-12
✨ Allow changing DNSNameservers in subnet config for OpenstackCluster

### LINKED ISSUE #2468: Allow modifications to dnsNameservers in OpenstackCluster spec post-deployment
/kind feature

Right now this is immutable:
```
spec:
  config
    managedSubnets:
    - cidr: 10.6.0.0/24
      dnsNameservers:
      - 1.1.1.1
      - 8.8.8.8
```
What if I would like to change the DNS servers after I have the cluster up?
In Openstack API it's a simple `openstack subnet` set call, but it's impossible through CAPO.

Currently mutable fields:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/4142407d6bec0474e797850932b0584764b14e60/pkg/webhooks/openstackcluster_webhook.go#L76

pbasov:
Allows changing DNSNameservers in subnet config for OpenstackCluster after the initial deployment

Fixes #2468

/hold

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -201,6 +201,10 @@ func (s *Service) ReconcileSubnet(openStackCluster *infrav1.OpenStackCluster, cl
 	} else if len(subnetList) == 1 {
 		subnet = &subnetList[0]
 		s.scope.Logger().V(6).Info("Reusing existing subnet", "name", subnet.Name, "id", subnet.ID)
+
+		if err := s.updateSubnetDNSNameservers(openStackCluster, subnet); err != nil {
+			return err
+		}
 	}
 
 	openStackCluster.Status.Network.Subnets = []infrav1.Subnet{
@@ -246,6 +250,44 @@ func (s *Service) createSubnet(openStackCluster *infrav1.OpenStackCluster, clust
 	}
 
 	return subnet, nil
+}
+
+// updateSubnetDNSNameservers updates the DNS nameservers for an existing subnet if they differ from the desired configuration
+func (s *Service) updateSubnetDNSNameservers(openStackCluster *infrav1.OpenStackCluster, subnet *subnets.Subnet) error {
+	// Picking the first managed subnet since we only support one for now
+	desiredNameservers := openStackCluster.Spec.ManagedSubnets[0].DNSNameservers
+	currentNameservers := subnet.DNSNameservers
+
+	needsUpdate := false
+	if len(desiredNameservers) != len(currentNameservers) {
+		needsUpdate = true
+	} else {
+		for i, ns := range desiredNameservers {
+			if i >= len(currentNameservers) || ns != currentNameservers[i] {
+				needsUpdate = true
+				break
+			}
+		}
+	}
+
+	if needsUpdate {
+		s.scope.Logger().Info("Updating subnet DNS nameservers", "id", subnet.ID, "from", currentNameservers, "to", desiredNameservers)
+
+		updateOpts := subnets.UpdateOpts{
+			DNSNameservers: &desiredNameservers,
+		}
+
+		updatedSubnet, err := s.client.UpdateSubnet(subnet.ID, updateOpts)
+		if err != nil {
+			record.Warnf(openStackCluster, "FailedUpdateSubnet", "Failed to update DNS nameservers for subnet %s: %v", subnet.ID, err)
+			return err
+		}
+
+		*subnet = *updatedSubnet
+		record.Eventf(openStackCluster, "SuccessfulUpdateSubnet", "Updated DNS nameservers for subnet %s", subnet.ID)
+	}
+
+	return nil
 }
 
 func (s *Service) getNetworkByName(networkName string) (networks.Network, error) {
--- a/pkg/webhooks/openstackcluster_webhook.go
+++ b/pkg/webhooks/openstackcluster_webhook.go
@@ -139,6 +139,62 @@ func (*openStackClusterWebhook) ValidateUpdate(_ context.Context, oldObjRaw, new
 		newObj.Spec.ManagedSecurityGroups.AllowAllInClusterTraffic = false
 	}
 
+	// Allow changes only to DNSNameservers in ManagedSubnets spec
+	if newObj.Spec.ManagedSubnets != nil && oldObj.Spec.ManagedSubnets != nil {
+		// Check if any fields other than DNSNameservers have changed
+		if len(oldObj.Spec.ManagedSubnets) != len(newObj.Spec.ManagedSubnets) {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "managedSubnets"), "cannot add or remove subnets"))
+		} else {
+			// Build maps of subnets by CIDR
+			oldSubnetMap := make(map[string]infrav1.SubnetSpec)
+			newSubnetMap := make(map[string]infrav1.SubnetSpec)
+
+			for _, subnet := range oldObj.Spec.ManagedSubnets {
+				oldSubnetMap[subnet.CIDR] = subnet
+			}
+
+			// Check if all new subnets have matching old subnets with the same CIDR
+			for _, newSubnet := range newObj.Spec.ManagedSubnets {
+				oldSubnet, exists := oldSubnetMap[newSubnet.CIDR]
+				if !exists {
+					allErrs = append(allErrs, field.Forbidden(
+						field.NewPath("spec", "managedSubnets"),
+						fmt.Sprintf("cannot change subnet CIDR from existing value to %s", newSubnet.CIDR),
+					))
+					continue
+				}
+
+				// Check if AllocationPools have changed
+				if !reflect.DeepEqual(oldSubnet.AllocationPools, newSubnet.AllocationPools) {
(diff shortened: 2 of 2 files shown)


pbasov:
I absolutely hated making DNSNameservers mutable in the webhook, it was tricky because subnets is a list (even though we support only ONE per network). Surely there's a better way to manage mutability for the config.

pbasov:
So yeah, why are we doing these webhook gymnastics when kubebuilder has immutable fields?
https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification

lentzi90:
> So yeah, why are we doing these webhook gymnastics when kubebuilder has immutable fields? https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification

Likely because no-one has taken the time to refactor it yet

pbasov:
@lentzi90 @mdbooth finished with the help of @s3rj1k, ready for review, tested manually on internal openstack cloud as well.

pkg/cloud/services/networking/network.go
```diff
@@ -248,6 +253,39 @@ func (s *Service) createSubnet(openStackCluster *infrav1.OpenStackCluster, clust
...
+	// Picking the first managed subnet since we only support one for now
+	desiredNameservers := openStackCluster.Spec.ManagedSubnets[0].DNSNameservers
+	currentNameservers := subnet.DNSNameservers
+
+	var needsUpdate bool
+	if len(desiredNameservers) != len(currentNameservers) {
+		needsUpdate = true
+	} else {
+		needsUpdate = !equality.Semantic.DeepEqual(currentNameservers, desiredNameservers)
+	}
```
mnaser, line 267: Probably not that big of a deal but I'm curious why don't we just drop `needsUpdate` and use the equality all the time.  Is this an optimization or is it simply something I don't know? :)

mnaser (approved):
looks good to me otherwise

lentzi90 (commented):
/approve

pkg/webhooks/openstackcluster_webhook.go
```diff
@@ -193,6 +194,62 @@ func (*openStackClusterWebhook) ValidateUpdate(_ context.Context, oldObjRaw, new
...
+
+				if newSubnet, exists := newSubnetMap[subnet.CIDR]; exists {
+					newSubnetCopy := newSubnet
+					newSubnetCopy.DNSNameservers = nil
+					newSubnets = append(newSubnets, newSubnetCopy)
+				}
+			}
+
+			oldObj.Spec.ManagedSubnets = oldSubnets
+			newObj.Spec.ManagedSubnets = newSubnets
```
mdbooth, line 249: We don't need this any more.

pkg/webhooks/openstackcluster_webhook.go
```diff
@@ -193,6 +194,62 @@ func (*openStackClusterWebhook) ValidateUpdate(_ context.Context, oldObjRaw, new
...
+
+				// Check if AllocationPools have changed
+				if !equality.Semantic.DeepEqual(oldSubnet.AllocationPools, newSubnet.AllocationPools) {
+					allErrs = append(allErrs, field.Forbidden(
+						field.NewPath("spec", "managedSubnets").Child("allocationPools"),
+						"cannot modify allocation pools in existing subnet",
+					))
+				}
+
+				newSubnetMap[newSubnet.CIDR] = newSubnet
```
mdbooth, line 230:
```suggestion
				// DNSNameservers is mutable
				oldSubnet.DNSNameservers = nil
				newSubnet.DNSNameservers = nil
```

mdbooth (commented):
Thanks for this, and especially for the amazing test coverage 👌

I'm not 100% confident of my assertion that order of DNSNameservers doesn't matter. It feels like one of those things that shouldn't matter, but might in practise. Let me know if I'm wrong about that.

Strong preference for an extra unit test covering an ordering-only difference, with the behaviour asserted being whatever we decide it should be.

Suggestion inline for a significant simplification of the webhook. Let me know if that makes sense.

pkg/cloud/services/networking/network.go
s3rj1k: `equality.Semantic.DeepEqual` is used in multiple places, so I guess it makes sense to have a wrapper function to compare slices, something like this

```
func EqualIgnoreOrder[T cmp.Ordered](a, b []T) bool {
	if len(a) != len(b) {
		return false
	}

	if len(a) == 0 {
		return true
	}

	aCopy := slices.Clone(a)
	bCopy := slices.Clone(b)

	slices.Sort(aCopy)
	slices.Sort(bCopy)

	return slices.Equal(aCopy, bCopy)
}
```

pkg/webhooks/openstackcluster_webhook.go
s3rj1k: suggest to change to something like this, what do you think?

```
	// Allow changes only to DNSNameservers in ManagedSubnets spec
	if newObj.Spec.ManagedSubnets != nil && oldObj.Spec.ManagedSubnets != nil {
		// Check subnet count hasn't changed
		if len(oldObj.Spec.ManagedSubnets) != len(newObj.Spec.ManagedSubnets) {
			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "managedSubnets"), "cannot add or remove subnets"))

			return aggregateObjErrors(newObj.GroupVersionKind().GroupKind(), newObj.Name, allErrs)
		}

		// Build map of old subnets by CIDR
		oldSubnetMap := make(map[string]infrav1.SubnetSpec, len(oldObj.Spec.ManagedSubnets))
		for _, subnet := range oldObj.Spec.ManagedSubnets {
			oldSubnetMap[subnet.CIDR] = subnet
		}

		// Validate each new subnet
		for i, newSubnet := range newObj.Spec.ManagedSubnets {
			oldSubnet, exists := oldSubnetMap[newSubnet.CIDR]
			if !exists {
				allErrs = append(allErrs, field.Forbidden(
					field.NewPath("spec", "managedSubnets").Index(i).Child("cidr"),
					fmt.Sprintf("cannot change subnet CIDR to %s", newSubnet.CIDR),
				))

				continue
			}

			// Check AllocationPools haven't changed
			if !equality.Semantic.DeepEqual(oldSubnet.AllocationPools, newSubnet.AllocationPools) {
				allErrs = append(allErrs, field.Forbidden(
					field.NewPath("spec", "managedSubnets").Index(i).Child("allocationPools"),
					"cannot modify allocation pools in existing subnet",
				))
			}
		}

		// If there are errors, return early without modifying objects
		if len(allErrs) > 0 {
			return aggregateObjErrors(newObj.GroupVersionKind().GroupKind(), newObj.Name, allErrs)
		}

		// Clear DNSNameservers from both objects
		for i := range oldObj.Spec.ManagedSubnets {
			oldObj.Spec.ManagedSubnets[i].DNSNameservers = nil
		}

		for i := range newObj.Spec.ManagedSubnets {
			newObj.Spec.ManagedSubnets[i].DNSNameservers = nil
		}
	}
```

pkg/cloud/services/networking/network.go
mdbooth: This is just a nit, and mostly because I've spent time in the past reading Semantic.DeepEqual trying to determine the precise behaviour of its edge cases.

I only mention it at all because my personal preference would be to use something more explicit, especially when the data structure is simple as here. But this works and the PR has great test coverage, so there's no pressing need to update this.

s3rj1k: I am fine replacing if we bring similar helper in, I would hate to replace 1 func call with len,copy,sort,equal checks in multiple places, this can be followup if that makes sense

pkg/webhooks/openstackcluster_webhook.go
mdbooth: I was trying to get rid of the AllocationPools check. I think this is fragile because if we added another field to this struct we'd need to also check it hasn't changed. The chances of that being forgotten are high.

pkg/cloud/services/networking/network_test.go
```diff
@@ -39,6 +39,86 @@ const (
...
+			desiredNameservers: []string{"8.8.8.8", "8.8.4.4"},
+			expect: func(m *mock.MockNetworkClientMockRecorder) {
+				m.UpdateSubnet(subnetID, subnets.UpdateOpts{
+					DNSNameservers: &[]string{"8.8.8.8", "8.8.4.4"},
+				}).Return(&subnets.Subnet{
+					ID:             subnetID,
+					DNSNameservers: []string{"8.8.8.8", "8.8.4.4"},
+				}, nil)
+			},
+		},
```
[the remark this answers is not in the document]
mdbooth: We think that order probably does matter, so please can we have another test where the set of nameservers is identical but the order is different? We should assert that they're updated to be in the specified order.

mdbooth:
This looks great. Can you squash the commits and I'll approve.

s3rj1k:
> This looks great. Can you squash the commits and I'll approve.

Done, thanks for helping out and reviewing

lentzi90 (commented):
/lgtm
/unhold

PR #2512, opened 2025-04-12
✨Allow API Loadbalancer Health Monitor configuration

pbasov:
Right now these values are hardcoded and the timeouts are too long, which can affect cluster poorly during control plane node outage / replacement.

/hold


pbasov:
@lentzi90 @mdbooth ready for review

lentzi90 (commented):
This is great!
A real e2e test of this would be nice, as we discussed in the office hours. Would you be willing to add that also in this PR?
I think [this test](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/3ec8787f5e38d1185b8520390754edfffb1498c7/test/e2e/suites/e2e/e2e_test.go#L731) should give a pretty good idea for how to do it.

s3rj1k:
> This is great! A real e2e test of this would be nice, as we discussed in the office hours. Would you be willing to add that also in this PR? I think [this test](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/3ec8787f5e38d1185b8520390754edfffb1498c7/test/e2e/suites/e2e/e2e_test.go#L731) should give a pretty good idea for how to do it.

PTAL on that e2e test when able, thanks

PR #2517, opened 2025-04-15
[release-0.11] Bump net to v0.36.0

larainema:
Bump net to v0.36.0 to fix [CVE-2025-22870](https://avd.aquasec.com/nvd/cve-2025-22870)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module sigs.k8s.io/cluster-api-provider-openstack
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.23.8
 
 require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module sigs.k8s.io/cluster-api-provider-openstack
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.23.8
```
EmilienM, line 5: I don't think we want that in a release version.
Let's let the bot doing this update on Mondays/Thursdays if possible?

mnaser: I agree that we shouldn't change this.  The reason for the initial change was due to the CVE that was blowing up our scanning :)

https://avd.aquasec.com/nvd/2025/cve-2025-22870/

lentzi90:
Looks like we would really have to bump the go version to get the newer net version. I don't think it is a good idea to do that on the release branch unfortunately

PR #2518, opened 2025-04-16
✨ allow switching from network.filter.name to network.id in OSC spec

okozachenko1203:
This PR allows switching from `spec.network.filter.name` to `spec.network.id` in OpenstackCluster webhook.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/webhooks/openstackcluster_webhook.go
+++ b/pkg/webhooks/openstackcluster_webhook.go
@@ -159,6 +159,13 @@ func (*openStackClusterWebhook) ValidateUpdate(_ context.Context, oldObjRaw, new
 		newObj.Spec.APIServerFloatingIP = nil
 		oldObj.Spec.APIServerFloatingIP = nil
 	}
+	// Allow change from spec.network.filter to spec.network.id only if it matches the current network.
+	if newObj.Spec.network != nil && oldObj.Spec.network != nil {
+		if ptr.Deref(newObj.Spec.network.id, "") == oldObj.Status.network.id {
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/ok-to-test
Seems safe and ok to me
/lgtm

lentzi90:
I think the test failure is due to a type. It should be capital N in `Spec.Network`

mnaser:
@okozachenko1203 do you think we can somehow exercise this in any sort of test to allow validating the change?

mnaser:
Looking at the tests, I see we can use this here:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/webhooks/openstackcluster_webhook_test.go

EmilienM:
I'm not against this PR but why don't we also do Subnets and other resources in that case?

lentzi90 (commented):
Please squash the commits.
@EmilienM also has a good point. Why do it only for network? We could do this for all resources that follow the same pattern

okozachenko1203:
@EmilienM @lentzi90 please review again

lentzi90 (commented):
/approve

mdbooth:
This is great.

I would still really like to ditch the webhooks and replace it all with CEL. I don't think we're currently doing anything that can't be defined in CEL. This would be one of the more interesting examples, though!

/lgtm
/approve

PR #2536, opened 2025-05-06
🌱 Add weekly security scan using govulncheck and Trivy

lentzi90:

This adds a github workflow that runs the govulncheck and trivy.
It is all copied and adapted from CAPI (hence the existing licenses at the top).
We may want to think about how to notify maintainers of this.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/.github/workflows/security-scan.yaml
+++ b/.github/workflows/security-scan.yaml
@@ -1,0 +1,32 @@
+name: Weekly security scan
+
+on:
+  schedule:
+  # Cron for every Monday at 9:12 UTC.
+  - cron: "12 9 * * 1"
+
+# Remove all permissions from GITHUB_TOKEN except metadata.
+permissions: {}
+
+jobs:
+  scan:
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [main, release-0.12, release-0.11, release-0.10]
+    name: Trivy
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # tag=v4.2.2
+      with:
+        ref: ${{ matrix.branch }}
+    - name: Calculate go version
+      id: vars
+      run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+    - name: Set up Go
+      uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # tag=v5.4.0
+      with:
+        go-version: ${{ steps.vars.outputs.go_version }}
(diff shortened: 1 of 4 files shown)


lentzi90:
Oh I wasn't aware of that one actually! I asked around a bit and it turns out we did use it before in CAPM3. Unfortunately we had issues with it around the go version and dependabot updates. I'm open to try it though, perhaps the issues are solved.

Another thing I was thinking about is if we should rather copy what CAPI does. They run govulncheck directly through a [workflow](https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/weekly-security-scan.yaml) and some [make targets](https://github.com/kubernetes-sigs/cluster-api/blob/700814cde497ecabe2b3833b6e12401c233dc5ec/Makefile#L746-L777).
What do you think?

lentzi90:
Thanks! Then I will rework this to follow CAPI :slightly_smiling_face:

lentzi90:
```
 hack/tools/bin/govulncheck-v1.1.4 ./... && R1=$? || R1=$?; \
hack/tools/bin/govulncheck-v1.1.4 -C "hack/tools" ./... && R2=$? || R2=$?; \
if [ "$R1" -ne "0" ] || [ "$R2" -ne "0" ]; then \
	exit 1; \
fi
No vulnerabilities found.
No vulnerabilities found. 
```

The normal test will run govulncheck also but not trivy to scan images.

.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# Remove all permissions from GITHUB_TOKEN except metadata.
+permissions: {}
+
+jobs:
+  scan:
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [main, release-1.10, release-1.9, release-1.8]
```
mnaser, line 16: This needs to be updated to our branches.

mnaser (changes requested):

.github/workflows/security-scan.yaml
lentzi90: Oh! Thanks for catching that!

mnaser (approved):

lentzi90:
> Nice! Running this locally seems to be recommending a bump to Go 1.23.8.

Thanks! Yes, this is also why I pushed https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2539 yesterday :slightly_smiling_face:

lentzi90:
Ok I'm just going to merge this now. The netlify check has been stuck for 4 days

PR #2561, opened 2025-05-26
🌱 E2E: clusterctl-upgrade from v0.10

lentzi90:

This is an attempt to fix the clusterctl-upgrade test that is currently failing on this branch. Starting from v0.9 is probably part of the issue since this is so old an unsupported now.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


EmilienM (approved):
Thanks for working on that

lentzi90:
Looks like we have a mismatch in the CAPI version we use in the e2e_conf.yaml (v1.9.3) and in go.mod (v1.9.7). I suspect this breaks us: https://github.com/kubernetes-sigs/cluster-api/commit/31a8274ea360028462725150d069052c55607131
Because we run the test framework with the newer version, while the controllers are on an older version that probably doesn't do the v1beta2 conditions.
Will update tomorrow.

PR #2570, opened 2025-05-26
🌱 E2E: Image refactor

lentzi90:

This makes the images configurable through the e2e_conf.yaml.
The images are applied together with the cluster-template, instead of in
the JustBefore code.

Fixes #

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


mdbooth (commented):
The way the code is currently written was intended to reduce the number of image creations. Before we used ORC for image downloading we downloaded the images once during initialisation, and effectively referenced them from the Machine specs. When we added ORC for images I was trying hard not to regress the runtime or cost of running the e2e tests, so I was basically trying to do the same thing.

Specifically:
* We define `managed` images explicitly in the `default` namespace, meaning that however many times they are applied they are only actually downloaded once
* We define 'non-core' images at their point of use, so if, e.g. we don't run the flatcar tests, we don't download the flatcar images at all
* We only directly reference `unmanaged` in machine specs because we expect them to have been downloaded exactly once by the `managed` versions in `default`

I don't think we pay ingress for images downloaded from a GCP source, so perhaps the financial motivation for this isn't so strong as long as we have enough disk space for multiple copies of the images. Pretty sure we'd pay ingress for the flatcar download, though, so ideally we'd continue to minimise downloads of that.

That said, I still think these are reasonable goals. What's not great is that the download locations themselves moved in to code, making them hard to modify when running locally. I see this PR moves them back into e2e_conf. How would you feel about keeping that part of this change, and just updating the references centrally?

lentzi90:
It seems to work! I can see the node-image downloaded in one test and just reused in another.
- [Here it has no download attempt](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2570/pull-cluster-api-provider-openstack-e2e-test/1932662960845492224/artifacts/clusters/bootstrap/resources/e2e-xk25cs/Image/node-image.yaml) and became Avaliable immediately when created (check timestamps).
- [Here it has a download attempt](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2570/pull-cluster-api-provider-openstack-e2e-test/1932662960845492224/artifacts/clusters/bootstrap/resources/e2e-21saej/Image/node-image.yaml) and it takes more than a minute from creation until it is Available.

lentzi90:
Checking optional tests.

lentzi90:
gcloud crashed :thinking: 

lentzi90:
The normal e2e tests are working, as mentioned above. I'm having issues with the "full" suite. Specifically the self-hosted test requires the Images to be moved over to the workload cluster, to make it self-hosted. Hopefully I have managed to include the right images and set labels on all of them now...

`pull-cluster-api-provider-openstack-test` fails for unrelated reasons. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2590

lentzi90:
Nailed it! :tada: 
Now we just need to get https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2590 to get rid of the failing govulncheck.

lentzi90:
@mdbooth could you take another look? I have gotten it working now, but this is removing all of the go code that was handling the images before. I realize it is a quite radical change and I hope it is for the better. This should make it more flexible so that we can also at some point use the e2e test suite in other environments than the devstack.

test/e2e/data/kustomize/common-patches/images-without-ref/images.yaml
```diff
@@ -1,13 +1,20 @@
...
   name: node-image
+  labels:
+    clusterctl.cluster.x-k8s.io/move: ""
 spec:
-  import:
-    filter:
-      name: ${OPENSTACK_IMAGE_NAME}
-  managementPolicy: unmanaged
+  resource:
+    name: ${OPENSTACK_IMAGE_NAME}
```
mdbooth, line 9: @mandre note that CAPO is now also switching to the behaviour where we rely on the adoption behaviour of ORC. That's at least 2 users of this behaviour which we never really designed for. I wonder if it's worth formalising 'import or create' in the ORC api.

@lentzi90 Unless @mandre says otherwise I think this is a reasonable use, despite the above. We're also doing this in OpenShift.

mdbooth (commented):
I think this is a reasonable change. As noted it relies on some behaviour in ORC which is not formally defined, but my feeling is that we should try to formally define it in ORC.

nit: Updating an image now requires updating both `*_IMAGE_NAME` and `*_IMAGE_URL` in the config, which is a potential trip hazard. I'll bet you could extract NAME from URL.

/lgtm

test/e2e/data/kustomize/common-patches/images-without-ref/images.yaml
mandre: No objection from me. I agree we'll need to formalize how users can do an `import or create` operation in ORC, since there's a need for it.

lentzi90:
Great!
/hold cancel
/approve

lentzi90:
Let's see if this is possible to backport easily
/cherry-pick release-0.12 release-0.11

PR #2600, opened 2025-06-18
🌱 Bump e2e k8s versions and update templates

lentzi90:

Normal tests run with kubernetes v1.33.1. Upgrades are going from v1.32.5 to v1.33.1.
Templates needed to be updated for v1.32 as the API server has dropped the cloud-provider flag.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/kustomize/v1beta1/default/cluster-template.yaml
+++ b/kustomize/v1beta1/default/cluster-template.yaml
@@ -73,9 +73,6 @@ spec:
           cloud-provider: external
           provider-id: "openstack:///'{{ instance_id }}'"
     clusterConfiguration:
-      apiServer:
-        extraArgs:
-          cloud-provider: external
       controllerManager:
         extraArgs:
           cloud-provider: external
--- a/kustomize/v1beta1/flatcar-sysext/patch-flatcar.yaml
+++ b/kustomize/v1beta1/flatcar-sysext/patch-flatcar.yaml
@@ -38,16 +38,16 @@ spec:
                 mode: 0644
                 contents:
                   remote:
-                    url: https://github.com/flatcar/sysext-bakery/releases/download/latest/kubernetes-${KUBERNETES_VERSION%.*}.conf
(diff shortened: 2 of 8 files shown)


lentzi90:
@tormath1 can you help shed some light on this?
I think the issue is that this URL is no longer correct:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/792395c2cb2d31005b7966a6f397f4547a437426/kustomize/v1beta1/flatcar-sysext/patch-flatcar.yaml#L41

```console
❯ curl -L https://github.com/flatcar/sysext-bakery/releases/download/latest/kubernetes-v1.33.conf
Not Found
```

I can also see that the sysext bakery releases are a bit different now. Is it so that we could use something like this instead?

```
https://github.com/flatcar/sysext-bakery/releases/download/kubernetes/kubernetes-v1.33.conf
```

tormath1:
Hey @lentzi90 that is correct, with the increasing number of folks using sysext-bakery + the increasing number of sysext images, we enhanced a bit the way we provide system extensions.

There is now a documentation page for each sysext (e.g Kubernetes: https://flatcar.github.io/sysext-bakery/kubernetes/) and the "correct" links would now be:
* sysext update configuration: `https://extensions.flatcar.org/extensions/kubernetes/kubernetes-${KUBERNETES_VERSION%.*}.conf`
* sysext image itself: `https://extensions.flatcar.org/extensions/kubernetes-${KUBERNETES_VERSION}-x86-64.raw`

Legacy links was still working for existing system but not for newer releases (e.g Kubernetes 1.33). Sorry for the turbulence here :)

lentzi90:
Awesome, thanks! :slightly_smiling_face: 
I'll update our templates to use the correct URL here also

lentzi90:
Alright I think we are good to go!
/approve
/hold cancel
/cc @mdbooth

lentzi90:
That is a good point! I'll try to backport to get the updated templates on the release branches
/cherry-pick release-0.12 release-0.11

PR #2601, opened 2025-06-18
:seedling: Test ORC upgrades in e2e

### LINKED ISSUE #2568: Test ORC upgrades in e2e
We should test upgrading between ORC versions, just like we do with all CAPI providers in clusterctl-upgrade tests. I think it should work by adding an older version in the [e2e_config.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/data/e2e_conf.yaml) and then specify in the [clusterctl-upgrade test](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/606fcbdbce41befe503c723038437570deb59a06/test/e2e/suites/e2e/clusterctl_upgrade_test.go#L59) what version to start from.

sirpaleet:
Test ORC upgrades in e2e.

Fixes #2568

Edits e2e tests.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


test/infrastructure/openstack-resource-controller/config/upgrade-from/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+labels:
+# openstack-resource-controller is not a provider, but by adding this label
+# we can get this installed by Cluster APIs Tiltfile and by the clusterctl machinery we use in E2E tests.
+- pairs:
+    cluster.x-k8s.io/provider: "runtime-extension-openstack-resource-controller"
+
+resources:
+# latest stable release of major version v1.0
+- https://github.com/k-orc/openstack-resource-controller/releases/download/{go://github.com/k-orc/openstack-resource-controller@v1.0}/install.yaml
```
lentzi90: I think we have to put the version hard-coded here. The `{go://...}` syntax only works in the e2e config

test/e2e/data/e2e_conf.yaml
```diff
@@ -154,6 +154,15 @@ providers:
     replacements:
     - old: "imagePullPolicy: Always"
       new: "imagePullPolicy: IfNotPresent"
+  # This is only for clusterctl upgrade tests, latest stable release of major version v1.0
+  - name: "{go://github.com/k-orc/openstack-resource-controller@v1.0}"
```
lentzi90: Since we anyway have to hard-code the version in the kustomization, I would suggest hard-coding here also. Otherwise it will be confusing when the name is automatically updated but the actually installed version stays the same

lentzi90:
The relevant test for the upgrade is pull-cluster-api-provider-openstack-e2e-full-test when you feel ready :slightly_smiling_face:

test/infrastructure/openstack-resource-controller/config/upgrade-from/kustomization.yaml
sirpaleet: Would it be better to also hard-code the name in e2e_conf.yaml and in clusterctl_upgrade_test.go files so that there would be no confusion of the versions used in the future? This would most likely need some additional thought on error handling as well.

sirpaleet: Missed your other comment, my bad

sirpaleet:
The test passed but brought up some fails with pod watching and UnhandledError related to clusterctl-upgrade and reflector.go, looking into it.

sirpaleet:
> The test passed but brought up some fails with pod watching and UnhandledError related to clusterctl-upgrade and reflector.go, looking into it.

This seems to relate to lost connection and should not be because of these changes.

PR #2611, opened 2025-06-26
🌱 Update to Kubernetes v1.33.2

pierreprinetti:

Is it time for a kube bump yet?

**TODOs**:

- [x squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests


pierreprinetti:
/hold

cluster-api v1.10.3 still [needs apimachinery v0.32.3](https://github.com/kubernetes-sigs/cluster-api/blob/9beb81e3a3efdbfb21779d90d95d2640273f1d4d/go.mod#L45). The bump has been made on the main branch, and we have to wait for them to cut a release.

pierreprinetti:
Util [functions](https://github.com/kubernetes-sigs/cluster-api/blob/abcc6d85c72aeaa652976f7f7bb341f0db9f4073/util/util.go#L160) in cluster-api v1.11 return v1beta2 objects. Do we expect to use v1beta2 objects in the CAPO releases that are meant for k8s v1.33?

CC @mdbooth @lentzi90

mdbooth:
> Util [functions](https://github.com/kubernetes-sigs/cluster-api/blob/abcc6d85c72aeaa652976f7f7bb341f0db9f4073/util/util.go#L160) in cluster-api v1.11 return v1beta2 objects. Do we expect to use v1beta2 objects in the CAPO releases that are meant for k8s v1.33?

Sounds like this is going to be the path of least resistance. Do you think it's worth a separate PR to bump to v1beta2 objects?

pierreprinetti:
> > Util [functions](https://github.com/kubernetes-sigs/cluster-api/blob/abcc6d85c72aeaa652976f7f7bb341f0db9f4073/util/util.go#L160) in cluster-api v1.11 return v1beta2 objects. Do we expect to use v1beta2 objects in the CAPO releases that are meant for k8s v1.33?
> 
> Sounds like this is going to be the path of least resistance. Do you think it's worth a separate PR to bump to v1beta2 objects?

It looks like v1beta2 was introduced in capi v1.11: https://github.com/kubernetes-sigs/cluster-api/tree/v1.10.3/api

pierreprinetti:
[EDIT] this was a long question that is not necessary any more

pierreprinetti:
@mdbooth In the end I think that the best solution would be for cluster-api to continue to provide the v1beta1 util functions we need. I am proposing it here: https://github.com/kubernetes-sigs/cluster-api/pull/12441

This PR temporarily uses my fork of cluster-api to test the potential result.

pierreprinetti:
```plaintext
test/e2e/shared/suite.go:188:56: cannot use OpenStackLogCollector{…} (value of struct type OpenStackLogCollector) as framework.ClusterLogCollector value in argument to framework.WithMachineLogCollector: OpenStackLogCollector does not implement framework.ClusterLogCollector (wrong type for method CollectInfrastructureLogs)
		have CollectInfrastructureLogs(context.Context, "sigs.k8s.io/controller-runtime/pkg/client".Client, *"sigs.k8s.io/cluster-api/api/core/v1beta1".Cluster, string) error
		want CollectInfrastructureLogs(context.Context, "sigs.k8s.io/controller-runtime/pkg/client".Client, *"sigs.k8s.io/cluster-api/api/core/v1beta2".Cluster, string) error 
```

I’m starting to question what I committed to.

mdbooth:
The problem is we're trying to call `framework.WithMachineLogCollector` from CAPI , which takes an argument which implements the `ClusterLogCollector` from CAPI. There's only a v1beta2 version of this interface, and of `framework.WithMachineLogCollector`. I have a horrible suspicion that at the end of this road is a complete re-implementation of the CAPI test framework 😬

pierreprinetti:
@mdbooth 
What do you think about @fabriziopandini's [comment in the cluster-api PR](https://github.com/kubernetes-sigs/cluster-api/pull/12441#issuecomment-3033248034)?

Shall we move to v1beta2 manifests (based on cluster-api v1beta2), move controllers to v1beta2 and implement conversion to support v1beta1?

mdbooth:
> @mdbooth What do you think about @fabriziopandini's [comment in the cluster-api PR](https://github.com/kubernetes-sigs/cluster-api/pull/12441#issuecomment-3033248034)?
> 
> Shall we move to v1beta2 manifests (based on cluster-api v1beta2), move controllers to v1beta2 and implement conversion to support v1beta1?

I personally think this is the path of least resistance. If I'm honest it's what I originally assumed you were doing here.

I don't think we'll need any conversions until we update the OpenStack resources (OpenStackMachine, OpenStackCluster) to the v1beta2 contract. That is: CAPO controllers will *consume* CAPI v1beta2 objects, but will themselves continue to *publish* infra objects conforming to CAPI v1beta1. We've been set a deadline of August next year to move to v1beta2, though, so we should probably get on that.

pierreprinetti:
Closing in favour of #2640

PR #2612, opened 2025-06-26
[release-0.9] 🌱 Move webhooks into pkg/webhooks

stephenfin:
This is a manual backport of #1920.

/hold

### DIFF
--- a/api/v1alpha5/conversion.go
+++ b/api/v1alpha5/conversion.go
@@ -100,6 +100,18 @@ func (r *OpenStackClusterTemplate) ConvertFrom(srcRaw ctrlconversion.Hub) error 
 	return utilconversion.MarshalData(src, r)
 }
 
+var _ ctrlconversion.Convertible = &OpenStackClusterTemplateList{}
+
+func (r *OpenStackClusterTemplateList) ConvertTo(dstRaw ctrlconversion.Hub) error {
+	dst := dstRaw.(*infrav1.OpenStackClusterTemplateList)
+	return Convert_v1alpha5_OpenStackClusterTemplateList_To_v1alpha7_OpenStackClusterTemplateList(r, dst, nil)
+}
+
+func (r *OpenStackClusterTemplateList) ConvertFrom(srcRaw ctrlconversion.Hub) error {
+	src := srcRaw.(*infrav1.OpenStackClusterTemplateList)
+	return Convert_v1alpha7_OpenStackClusterTemplateList_To_v1alpha5_OpenStackClusterTemplateList(src, r, nil)
+}
+
 var _ ctrlconversion.Convertible = &OpenStackMachine{}
 
 func (r *OpenStackMachine) ConvertTo(dstRaw ctrlconversion.Hub) error {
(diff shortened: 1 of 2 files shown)


stephenfin:
fwiw, this backport should allow us to address an issue that is currently blocking work on the migration path from [MAPO to CAPO in OpenShift](https://github.com/openshift/cluster-capi-operator/pull/188) in OpenShift. While I appreciate this is primarily for downstream, I'm proposing it upstream as I hope it should introduce minimal risk while allowing us/me to stay as close to upstream as possible (always my preferred approach, personally). I'm happy to discuss why we need this (tl;dr: MAPO is stuck with CAPO v0.9.x for $reasons) and build a case for doing this here, as needed, though if we really don't want to do this here I'd understand.

stephenfin:
(marking as ready-for-review with a hold to get CI results)

mnaser:
anything that helps an active downstream get closer to upstream is something I am in favour of :)

mandre:
I looks like the problems with the CI are due to a mismatch of versions between go and golangci-lint. By the look of [https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-a[…]-provider-openstack-test/1938281313530286080/prowjob.json](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2612/pull-cluster-api-provider-openstack-test/1938281313530286080/prowjob.json) it seems the job uses [gcr.io/k8s-staging-test-infra/kubekins-e2e:v20250613-876fb90a97-master](http://gcr.io/k8s-staging-test-infra/kubekins-e2e:v20250613-876fb90a97-master) with go 1.24.

But golangci-lint [is at version v1.54.2](https://github.com/shiftstack/cluster-api-provider-openstack/blob/d645938c2af582a36e032eeee6d5d7b41e81ff2f/hack/tools/Makefile#L18) which is not compatible with go 1.24. I wonder if bumping it to 1.64 (first version to support go 1.24) helps.

mandre:
> But golangci-lint [is at version v1.54.2](https://github.com/shiftstack/cluster-api-provider-openstack/blob/d645938c2af582a36e032eeee6d5d7b41e81ff2f/hack/tools/Makefile#L18) which is not compatible with go 1.24. I wonder if bumping it to 1.64 (first version to support go 1.24) helps.

I tried bumping golangci-lint with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2619 and it's now effectively [showing different errors](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2619/pull-cluster-api-provider-openstack-test/1944761943675899904), with `gosec` and `revive` linters, while we [previously had](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/2612/pull-cluster-api-provider-openstack-test/1938281313530286080) `typecheck` errors due to version incompatibilities.

mandre:
It failed to properly bring up the worker nodes so the multiple AZs test failed. Trying again.
/retest

stephenfin:
Hurrah!

stephenfin:
/unhold

This is ready for review now.

Issue #2386, opened 2025-01-22
[Feature Request] Introduce OpenStackClusterIdentity Resource for Centralized Credential Management


bnallapeta:
/kind feature

**Describe the solution you'd like**

The current implementation of identityRef in OpenStackCluster requires secrets containing credentials (e.g., clouds.yaml) to reside in the same namespace as the OpenStackCluster resource. This limitation poses challenges in environments with namespace-based access control, where sensitive credentials need to remain centralized and secure while still being usable by resources in other namespaces.

To address this, I propose introducing an OpenStackClusterIdentity custom resource, similar to [AWSClusterStaticIdentity](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b29e04c5f7673b7516cf7ef9e8b588c13b264b11/api/v1beta2/awsidentity_types.go#L79) or [AzureClusterIdentity](https://capz.sigs.k8s.io/reference/v1beta1-api#infrastructure.cluster.x-k8s.io/v1beta1.AzureClusterIdentity). This resource would allow for centralized credential management while providing fine-grained access control across namespaces.

***Proposed Implementation:***
	1.	New CRD: Create OpenStackClusterIdentity to reference credentials stored in a secret, allowing usage across namespaces with access controls.
	2.	Identity Reference: Update OpenStackCluster.spec.identityRef to reference the new resource instead of secrets directly.
	3.	Access Control: Implement validation in the CAPO controller to restrict usage based on namespaces or roles.

mdbooth:
Even implementing the ability to reference cross-namespace credentials raises architectural issues in CAPO akin to requiring a setuid bit in a posix application. Despite what anybody else may have done, my instinct is to actively avoid doing this. It requires us to robustly implement and enforce our own RBAC scheme. I'm not sure we're up for that.

IOW, far from protecting secrets, I believe this feature would open a potential hole where a bug might expose any secret in the cluster.

However, I'm very interested in the use case. Can you explain why the credentials cannot live in the target namespace alongside the CRs which reference them?

s3rj1k:
@mdbooth Can you please elaborate on "CAPO akin to requiring a setuid bit in a posix application"?

Assuming that I understand use case correctly, to have what other major CAPI providers have, to have similar API (both AWS and Azure have this).

wkonitzer:
[...]

PR #2628, opened 2025-07-22
proposal: add new CRD OpenStackClusterIdentity

bnallapeta:
This PR adds a proposal doc to work on the issue described in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2386

### DIFF
--- a/docs/proposals/20250722-openstackclusteridentity.md
+++ b/docs/proposals/20250722-openstackclusteridentity.md
@@ -3,10 +3,10 @@
 ## Metadata
 
 - **Authors**: @bnallapeta
-- **Reviewers**: CAPO maintainers
+- **Reviewers**: CAPO maintainers (@mdbooth)
 - **Status**: Proposed
 - **Creation Date**: 2025-07-22
-- **Last Updated**: 2025-07-22
+- **Last Updated**: 2025-07-29
 
 ## Summary
 
@@ -53,11 +53,8 @@ type OpenStackClusterIdentitySpec struct {
     // SecretRef references the secret containing OpenStack credentials
     SecretRef OpenStackCredentialSecretReference `json:"secretRef"`
 
-    // AllowedNamespaces defines which namespaces can use this identity
-    // +optional
-    AllowedNamespaces []string `json:"allowedNamespaces,omitempty"`
-
     // NamespaceSelector selects allowed namespaces via labels
+    // All namespaces have a kubernetes.io/metadata.name label containing their name
     // +optional
     NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
 }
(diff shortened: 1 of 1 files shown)


docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+    NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+}
+
+type OpenStackCredentialSecretReference struct {
+    Name      string `json:"name"`
+    Namespace string `json:"namespace"`
+}
+
+type OpenStackClusterIdentityStatus struct {
+    Ready bool `json:"ready"`
```
mdbooth: Both `Ready` and `bool` are very much out of fashion in k8s apis lately. If you dropped `Ready` could you replace it with a Condition?

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+    Conditions clusterv1.Conditions `json:"conditions,omitempty"`
```
mdbooth: Please can you document which Conditions you plan to add and what their semantics would be?

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Enhanced OpenStackCluster:**
+```go
+type OpenStackClusterSpec struct {
+    // ... existing fields unchanged ...
+
+    // ClusterIdentityRef references a cluster-scoped identity (NEW)
+    // Takes precedence over IdentityRef if specified
+    // +optional
+    ClusterIdentityRef *OpenStackClusterIdentityReference `json:"clusterIdentityRef,omitempty"`
```
mdbooth: I wonder if we should add this to `OpenStackIdentityReference` instead, perhaps as a discriminated union? 🤔 This would probably look like a new, optional `type` field whose default value is `Secret`.

This isn't a concrete request for change, btw, just a request for discussion.

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+The scope factory will implement dual-path credential resolution:
+
+```go
+func (f *providerScopeFactory) resolveCredentials(obj infrav1.IdentityRefProvider) {
+    // Priority 1: Check for cluster identity reference
+    if clusterRef := obj.GetClusterIdentityRef(); clusterRef != nil {
+        return f.newScopeFromClusterIdentity(clusterRef)
+    }
+    
+    // Priority 2: Fall back to existing namespace identity behavior
```
mdbooth: Another advantage of extending identityRef would be not needing a fall-back behaviour. The identity ref would always be concretely defined as one type or another.

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+    return f.newScopeFromNamespaceIdentity(obj.GetIdentityRef())
+}
+```
+
+**Key Edge Cases:**
+- Both references specified: Use `clusterIdentityRef`, log warning
+- Identity deletion: Clusters show degraded status, don't fail
+- Permission changes: Dynamic re-validation during reconciliation
```
mdbooth: What permissions are you referring to here?

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,213 @@
...
+    Status OpenStackClusterIdentityStatus `json:"status,omitempty"`
+}
+
+type OpenStackClusterIdentitySpec struct {
+    // SecretRef references the secret containing OpenStack credentials
+    SecretRef OpenStackCredentialSecretReference `json:"secretRef"`
+
+    // AllowedNamespaces defines which namespaces can use this identity
+    // +optional
+    AllowedNamespaces []string `json:"allowedNamespaces,omitempty"`
```
mdbooth: You could (and therefore probably should) drop this and just leave `NamespaceSelector`, btw: all namespaces have a `kubernetes.io/metadata.name` label containing their own name.

For example, `ValidatingWebhook` works this way: https://github.com/kubernetes/kubernetes/blob/e2eed70a296604f179171826d87172a9f2cf3add/pkg/apis/admissionregistration/types.go#L778-L822

mdbooth (commented):
I'm very much in favour of this. A few API-related notes inline.

The more I think about it, the more I'm convinced we should extend identityRef rather than adding an alternative with a fallback. I think the semantics and validation will be much clearer.

Incidentally, on the subject of validations I'd expect this to be validated with CEL. Lets not add anything new to the existing webhooks when it's not required. That's an implementation detail, though, so no need to add CEL expressions to this doc unless you really want to.

bnallapeta:
> I'm very much in favour of this. A few API-related notes inline.
> 
> The more I think about it, the more I'm convinced we should extend identityRef rather than adding an alternative with a fallback. I think the semantics and validation will be much clearer.
> 
> Incidentally, on the subject of validations I'd expect this to be validated with CEL. Lets not add anything new to the existing webhooks when it's not required. That's an implementation detail, though, so no need to add CEL expressions to this doc unless you really want to.

yes, I considered both while thinking about this proposal. Let me put down my thoughts as simple pros vs cons:

**new clusterIdentityRef**
pros:
zero risk to existing APIs
validation becomes easier and cleaner
fallback designed to support both
migration would be easier
naming is in par with other providers - users feel familiar

cons:
dual fields for the same purpose - might be confusing to users
maintenance burden

**extend current identityRef**
pros:
users don't have to learn about a new field, less confusing
cleaner long-term API

cons:
higher migration risk
have to consider migration scenarios for testing



Kindly think this over and let me know.

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Yes, could be accomplished using labels and NamespaceSelector, but I kept this for user experience. It would be more intuitive to use AllowedNamespaces. But I agree that having both of them might feel redundant. 

What do you think? I am open to dropping it if its not needed.

mdbooth:
> cons: higher migration risk have to consider migration scenarios for testing

I don't follow this. Can you explain the additional risk?

For additional context I'm thinking something like this:

## Current

This is the existing syntax. It is identical, with identical semantics. The absence of a `type` field means we default to `Secret`.

```yaml
kind: OpenStackCluster
metadata:
  name: my-cluster
  namespace: team-a
spec:
  identityRef:
    name: fallback-secret
    cloudName: openstack
```

## ClusterIdentity

Specifying a ClusterIdentity. Note the `type` field.

```yaml
kind: OpenStackCluster
metadata:
  name: my-cluster
  namespace: team-a
spec:
  identityRef:
    type: ClusterIdentity
    name: production-openstack
```

## Current with explicit type

This form uses the current `Secret` semantics explicitly.

```yaml
kind: OpenStackCluster
metadata:
  name: my-cluster
  namespace: team-a
spec:
  identityRef:
    type: Secret
    name: fallback-secret
    cloudName: openstack
```

This is a common pattern in kubernetes APIs. It's not strictly a `discriminated union` because that typically discriminates on a single sub-field, but it's close enough without breaking backwards compatibility. This pattern has a number of advantages:
- It is very easy to validate: the validation is limited in scope to a single struct (in this case `identityRef`).
- It is easy to implement without breaking backwards compatibility.
- It is easy to extend in the future.
- It is unambiguous: only a single thing can be expressed.

Incidentally, an older version of the CAPO API actually had a `type` field, but we removed it because we only had a single type. Perhaps that was short-sighted, but it's easy enough to add back.

Another reason in favour of extending the existing struct is that it would propagate everywhere that the struct is used, including OpenStackMachine (and Server). We have had users relying on separate Machine credentials, and there's no reason to restrict which credential types they can use.

docs/proposals/20250722-openstackclusteridentity.md
mdbooth: `allowedNamespaces` is simpler for the user if you just want specific namespaces. However, I'm just on the side of removal because:
- consistency with existing kubernetes APIs
- only having one way to do the thing

If we remove it I would definitely add a godoc comment to the namespace selector about `kubernetes.io/metadata.name`.

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Decided to drop this. Currently, there is nothing like "CredentialsValid" or "AuthenticatedSecret" or similar. No need to over-complicate this. Current code handles this and we can keep it at that.

Conditions will have 'Ready' field. Examples added.

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Clarified in docs.

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Done.

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,384 @@
...
+    // +kubebuilder:validation:XValidation:rule="self == 'Secret' ? has(self.cloudName) : !has(self.cloudName)",message="cloudName required for Secret type, forbidden for ClusterIdentity type"
+    // +kubebuilder:validation:XValidation:rule="has(self.name)",message="name is required"
+    // +optional
+    Type string `json:"type,omitempty"`
+
+    // Name of the secret (type=Secret) or cluster identity (type=ClusterIdentity)
+    // +optional
+    Name string `json:"name,omitempty"`
+    
+    // CloudName required for Secret type, ignored for ClusterIdentity type
```
mdbooth:
```suggestion
    // CloudName required for Secret type, forbidden for ClusterIdentity type
```

Makes changing things simpler later if you disallow the edge cases from the start.

docs/proposals/20250722-openstackclusteridentity.md
```diff
@@ -0,0 +1,384 @@
...
+- `OpenStackMachine` resources  
+- `OpenStackServer` resources
+
+## Conditions
+
+Following the existing CAPO pattern for simple resources (like OpenStackFloatingIPPool), the `OpenStackClusterIdentityStatus` uses a single condition:
+
+- **Ready**: Indicates whether the identity can be used for authentication. This includes secret accessibility, structure validation, and credential validation. Different reason codes provide specific failure details.
+
+### Example Condition States
```
mdbooth: Everything below implies:

* The existence of a ClusterIdentity controller
* That resource controllers consuming a ClusterIdentity must:
  - Wait for their ClusterIdentity to have the Ready condition, and therefore
  - Watch the ClusterIdentity resource

Is that complexity required in the first instance? What if it was just a dumb data struct, and any errors were reported by the resource controllers as normal?

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Done.

docs/proposals/20250722-openstackclusteridentity.md
bnallapeta: Done. I have removed the conditions field for now. We can consider it later if required.

mdbooth:
/lgtm

@lentzi90 @EmilienM Any thoughts? This is quite a big change, but I think it's a good one.

mandre (commented):
The API looks good to me but I think the implementation details is lacking information on how this impacts operations involving ORC. If there is no impact, we should say it.

lentzi90 (commented):
/approve

mdbooth:
@mandre There's definitely an impact on ORC. ORC will need to implement this or something like it.

lentzi90:
/retitle :book: proposal: add new CRD OpenStackClusterIdentity

PR #2629, opened 2025-07-22
🐛 Improve handling of missing load balancer permissions

sd109:

Currently, when a user tries to create a cluster using OpenStack credentials which are missing the load balancer permissions, CAPO adds the finalized to the OpenStackCluster resource then fails to create the load balancer. When the user then tries to delete the cluster, CAPO makes a GET request to the Octavia API to get the load balancer details and receives a 403 (permission denied) response, so the only way to allow the cluster deletion to proceed is to manually remove the finalizer from the OpenStackCluster resource.

This change prevents the above edge case by only attempting to delete the API server load balancer if the load balancer ID is populated in the OpenStackCluster's status field.

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -177,7 +177,7 @@ func (r *OpenStackClusterReconciler) reconcileDelete(ctx context.Context, scope 
 		return reconcile.Result{}, err
 	}
 
-	if (openStackCluster.Spec.APIServerLoadBalancer.IsEnabled() && openStackCluster.Status.APIServerLoadBalancer.ID != "") {
+	if openStackCluster.Spec.APIServerLoadBalancer.IsEnabled() && openStackCluster.Status.APIServerLoadBalancer.ID != "" {
 		loadBalancerService, err := loadbalancer.NewService(scope)
 		if err != nil {
 			return reconcile.Result{}, err


bnallapeta:
@sd109 

Thank you for this PR. The change you've proposed to check the status field certainly fixes the immediate issue. However, there's a more robust way to solve this that avoids a potential race condition (where the LB is created but the status isn't updated).

Instead of avoiding the call to delete the load balancer, I think it's better to make the `DeleteLoadBalancer` function itself more resilient. Could you modify the code so that when `DeleteLoadBalancer` receives a `403 Forbidden error` from the OpenStack API, it logs a warning and returns nil?

This would handle the permission error gracefully, allowing cluster deletion to proceed without relying on the status field being in sync. Once that change is made, adding a unit test for that specific error handling logic would be the final step.

Let me know what you think of this approach.

EmilienM:
Right now, CAPO only supports Octavia when a load balancer is being enabled via `openStackCluster.Spec.APIServerLoadBalancer.IsEnabled`.
If your cloud doesn't have Octavia or you don't want Octavia to handle the API Server load balancing, you shouldn't set this to True because CAPO doesn't support that now.

Matt had in mind a major rework for our load balancer support, having some sort of a way to define a load balancer provider (octavia, metallb, etc) but this hasn't been done.

The reason why I don't want this patch is that you should not have an failure to delete a load balancer if openStackCluster.Spec.APIServerLoadBalancer.IsEnabled was initially set to enabled; simply because if you could have created it, you should be able to disable it.

Let me know if I missed something, I'm happy to discuss further and refine here a fix that works for you.

bnallapeta:
@EmilienM 

While I agree users shouldn't enable LB without proper permissions, OpenStack permission models can be complex (e.g., IAM policies that allow create but not delete, or permissions that change between cluster lifecycle events). A more robust solution would be to make `DeleteLoadBalancer` handle 403 errors gracefully rather than avoiding the call entirely - this way we still attempt deletion as you suggest, but don't leave users stuck with unremovable finalizers when permissions are insufficient.

sd109:
> Right now, CAPO only supports Octavia when a load balancer is being enabled via `openStackCluster.Spec.APIServerLoadBalancer.IsEnabled`. If your cloud doesn't have Octavia or you don't want Octavia to handle the API Server load balancing, you shouldn't set this to True because CAPO doesn't support that now.

@EmilienM agreed, but even with Octavia-enabled clouds we still see problems arise when other tools are used to wrap CAPO functionality, since users are not always explicitly aware that `APIServerLoadBalancer.IsEnabled` is being set for them. 

For example, when using the [magnum-capi-helm driver](https://opendev.org/openstack/magnum-capi-helm) a user might execute an `openstack coe cluster create` command using an app cred which is missing the load-balancer_member role. In such a situation, the cluster creation is stuck in an indefinite `CREATING` state because of the permission denied on the LB create API call and then when a user later tries to delete the Magnum cluster it instead gets stuck in a `DELETING` state indefinitely until the cloud operator manually removes the relevant finaliser from the OpenStackCluster object on the management cluster (which the user cannot do themselves because the CAPI management cluster is a cloud admin responsibility).

We originally proposed a fix on the Magnum driver side [here](https://review.opendev.org/c/openstack/magnum-capi-helm/+/955078), perhaps the previous discussion over there will help to provide some additional context.

lentzi90:
Could you squash the commits?
I think this looks quite ok. Can anyone think of a case when this would lead to issues? For example, could there be a situation when the status does not have the ID, but the LB was actually created?

bnallapeta:
@lentzi90 I have already provided my comment [above](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2629#issuecomment-3144623591). Let me know what you think.

lentzi90:
> @lentzi90 I have already provided my comment [above](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2629#issuecomment-3144623591). Let me know what you think.

Did you have some specific scenario in mind when this could happen? I'm not convinced that it is safer to return `nil` if there is a 403 error. What if the credentials expired or were revoked? If we have the ID in the status it would be better to block deletion at that point IMO.

bnallapeta:
> > @lentzi90 I have already provided my comment [above](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2629#issuecomment-3144623591). Let me know what you think.
> 
> Did you have some specific scenario in mind when this could happen? I'm not convinced that it is safer to return `nil` if there is a 403 error. What if the credentials expired or were revoked? If we have the ID in the status it would be better to block deletion at that point IMO.

As I have mentioned earlier, the only realistic case where an LB exists but status.ID is empty is a narrow race: LB creation succeeded but we crashed before patching status. Outside of that, an empty ID means CAPO never tracked an LB. I don't expect that this would happen in a normal scenario.

But besides that, my point was to centralize the deletion logic behavior in `DeleteLoadBalancer`.

- If status.ID is empty and Octavia returns 403, treat it as non-fatal: return (nil, nil), emit a clear Warning event/condition about missing permissions and “no LB tracked in status,” and let deletion proceed. 
- If status.ID is present and we get 403, return an error and block deletion to avoid leaking a real LB. Operators can restore permissions and retry.

lentzi90:
Ok I see, I thought you wanted to only check the 403 error and not look at the ID. I'm not sure what to think of this really. Maybe the best thing is actually to keep it the way it was.

@sd109 is there no way to disable the API loadbalacer instead of deleting the cluster?

bnallapeta:
@sd109 Could you please take a look at this?

PR #2659, opened 2025-08-18
🌱 Don't run golanci-lint update on PRs

lentzi90:

The workflow is failing in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2647 because the base is not set. This is an attempt to fix that.
However, creating a PR from a workflow triggered by a PR requires more privileges than the normal token has. I suggest we don't run this workflow on PRs at all, rather than giving it more permissions and a custom token.
We can still manually trigger it to run and select on which branch it should run.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests


lentzi90:
/hold
There is something else wrong with this workflow also. I think the reason it doesn't work on the PR is because it starts from the merge commit (not on a branch). This is not an issue in the [periodic job](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/actions/runs/17027496745) though, but it still fails to create PRs.

lentzi90:
Ok now it is at least trying to create the PR! The remaining issue is that the default token doesn't have permission to create PRs when the action is triggered from a PR/push. I think that is a sane policy that we should not tamper with. Perhaps we can do some kind of dry-run on PRs :thinking:

lentzi90:
I have not found any clever way to dry-run this so removing the PR trigger. We can still trigger the workflow manually on specific branches as a way of testing it.
/hold cancel

EmilienM:
/lgtm
/approve

Thanks!

PR #2665, opened 2025-08-20
🌱Add image validation for openstack machine

### LINKED ISSUE #2265: Missing image is hard to debug
/kind bug

1. Create a cluster using an image that doesn't exist.
2. The OpenStackCluster becomes ready (provided no bastion is used or bastion is using an existing image)
3. There is no visible progress and no explanation for what is wrong at the OpenStackMachine level. Worse, there is a misleading message that the OpenStackMachine is waiting for bootstrap data, which is not true.
4. There is no explanation on the OpenStackServer. It is just "not ready"

There should be a condition on the OpenStackServer that would propagate to the OpenStackMachine, with a message explaining that the image cannot be found. The OpenStackMachine should not say that it is waiting for bootstrap data when the data is available.

This issue is also present in v0.10, just that there is no OpenStackServer involved then.

**Environment:**

- Cluster-API version: v1.8.5
- OpenStack version:
- Minikube/KIND version:

smoshiur1237:
This pr will be part of fix for image validation. At this moment invalid images are not reported to the describe log of the openstack machine. The goal is to improve the log situation by adding the image validation for image parameters.
Fixes #2265

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/cluster-api-provider-openstack/pkg/utils/optional"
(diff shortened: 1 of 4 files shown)


smoshiur1237:
/hold 
work in progress

api/v1beta1/types.go
```diff
@@ -75,6 +75,24 @@ func (f *ImageFilter) IsZero() bool {
...
+	// Not possible to validate the image if it is missing
+	if i == nil {
+		errors = append(errors, field.Required(&base, "image is required"))
+		return errors
+	}
+	if i.Filter.Name == nil || i.Filter.Tags == nil {
+		errors = append(errors, field.Required(base.Child("Image filter"), "either name or tags of image are missing"))
+	}
+	if i.ImageRef.Name == "" {
+		errors = append(errors, field.Required(base.Child("ORC mage Regerecne"), "Orc image is missing"))
```
mandre: I first wondered what language that was, but then realized you probably meant:
```suggestion
		errors = append(errors, field.Required(base.Child("ORC Image Reference"), "ORC image is missing"))
```

😅

bnallapeta:
@smoshiur1237 Just checking in on this. Do you have some time to look over the failing tests and fix them? 
(would need a rebase too)

smoshiur1237:
I had a discussion with @lentzi90  and trying to explore different way of validation. The approach we are using may not cover the validation. I will get back on  this soon.

PR #2692, opened 2025-09-08
🌱Add entry of new release branch to depandabot

### LINKED ISSUE #2686: Add dependabot config to release process
We have an [issue template](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/.github/ISSUE_TEMPLATE/new_release.md) for making releases, but it does not mention the [dependabot config](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/.github/dependabot.yml) that needs to be updated.

When a new branch is created, it should be added to the dependabot config. After the release, we should also remove the oldest branch, since it goes out of support at that point.

smoshiur1237:
Adding entry of new release branch to depandabot config. Also removing some dependency on main and release-0.13 config which are no longer needed here because of k8s and go uplift on these branch.
Fixes #2686

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,16 +44,59 @@ updates:
     update-types: ["version-update:semver-major", "version-update:semver-minor"]
   - dependency-name: "sigs.k8s.io/controller-tools"
     update-types: ["version-update:semver-major", "version-update:semver-minor"]
-  # We will need k8s v0.31.3 to bump structured-merge-diff to v4.4.2 (check git history for details).
-  - dependency-name: "sigs.k8s.io/structured-merge-diff/*"
-  # For now we deploy CAPI from an unreleased commit
-  - dependency-name: "sigs.k8s.io/cluster-api*"
-  # These dependencies are skipped because they require a newer version of go:
-  - dependency-name: "github.com/a8m/envsubst"
   labels:
     - "area/dependency"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -54,6 +54,60 @@ updates:
...
+  - dependency-name: "sigs.k8s.io/controller-tools"
+    update-types: ["version-update:semver-major", "version-update:semver-minor"]
+  # We will need k8s v0.31.3 to bump structured-merge-diff to v4.4.2 (check git history for details).
+  - dependency-name: "sigs.k8s.io/structured-merge-diff/*"
+  # These dependencies are skipped because they require a newer version of go:
+  - dependency-name: "github.com/a8m/envsubst"
+  - dependency-name: "github.com/onsi/gomega"
+  labels:
+    - "area/dependency"
+    - "ok-to-test"
```
lentzi90: This should be fine to remove now, also for the main branch above. We have just forgotten to update.
K8s is already at v0.33 and structured-merge-diff at v4.7. We also have a newer go version on main and release-0.13 compared to release-0.12 when we added envsubst and gomega here.
Could you please remove these lines for main and release-0.13?

lentzi90 (commented):
/ok-to-test
/approve
/lgtm

PR #2725, opened 2025-09-19
🌱 Add bastion and allowedCIDR to dev-test clusterclass

### LINKED ISSUE #2693: Add bastion and allowed CIDR to dev-test ClusterClass
The [dev-test ClusterClass](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/clusterclass-dev-test.yaml) does not expose the API loadbalancer allowed CIDR, nor the bastion. This makes it very hard to use.
We should add a variable for allowed CIDR and (a complex variable) for the bastion.

bnallapeta:
Fixes #2693

### DIFF
--- a/templates/cluster-template-development.yaml
+++ b/templates/cluster-template-development.yaml
@@ -24,3 +24,14 @@ spec:
       value: ${ADD_IMAGE_VERSION:=false}
     - name: injectIgnitionSysext
       value: ${INJECT_IGNITION_SYSEXT:=true}
+    - name: allowedCIDRs
+      value: ${OPENSTACK_API_SERVER_ALLOWED_CIDRS:=[]}
+    - name: bastion
+      value:
+        enabled: ${OPENSTACK_BASTION_ENABLED:=false}
+        spec:
+          flavor: ${OPENSTACK_BASTION_FLAVOR:=m1.small}
+          image:
+            filter:
+              name: ${OPENSTACK_BASTION_IMAGE_NAME:=ubuntu-24.04}
+          sshKeyName: ${OPENSTACK_SSH_KEY_NAME:=""}
--- a/templates/clusterclass-dev-test.yaml
+++ b/templates/clusterclass-dev-test.yaml
@@ -73,6 +73,53 @@ spec:
           Use a sysext overlay to add the Kubernetes components to the image.
           This is for use with flatcar and similar images.
         default: false
+  - name: allowedCIDRs
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: array
+        items:
+          type: string
+        description: |
+          List of CIDR ranges allowed to access the API server load balancer.
+          If not specified, all traffic will be allowed.
+        default: []
+  - name: bastion
+    required: false
+    schema:
+      openAPIV3Schema:
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            description: "Enable or disable the bastion host"
+            default: false
+          spec:
+            type: object
+            description: "OpenStackMachineSpec for the bastion host"
+            properties:
+              flavor:
+                type: string
+                description: "OpenStack flavor for the bastion"
+              image:
+                type: object
+                properties:
+                  filter:
+                    type: object
+                    properties:
+                      name:
+                        type: string
+                        description: "Name of the image to use for bastion"
+              sshKeyName:
+                type: string
+                description: "SSH key pair name for bastion access"
+          floatingIP:
+            type: string
+            format: ipv4
+            description: "Floating IP address to assign to the bastion (optional)"
+          availabilityZone:
+            type: string
+            description: "Availability zone for the bastion host"
   patches:
   - name: image
     description: "Sets the OpenStack image that is used for creating the servers."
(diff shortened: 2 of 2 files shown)


templates/cluster-template-development.yaml
```diff
@@ -24,3 +24,16 @@ spec:
...
+    - name: bastion
+      value:
+        enabled: ${OPENSTACK_BASTION_ENABLED:=false}
+        spec:
+          flavor: ${OPENSTACK_BASTION_FLAVOR:=m1.small}
+          image:
+            filter:
+              name: ${OPENSTACK_BASTION_IMAGE_NAME:=ubuntu-2404-kube}
+          sshKeyName: ${OPENSTACK_SSH_KEY_NAME:=""}
+        floatingIP: ${OPENSTACK_BASTION_FLOATING_IP:=""}
```
lentzi90, line 38: This basically makes it mandatory to specify a floating IP, since the default `""` is not valid.
I suggest we leave it out instead. It can still be in the ClusterClass.

```suggestion
```

templates/cluster-template-development.yaml
```diff
@@ -24,3 +24,16 @@ spec:
...
+    - name: allowedCIDRs
+      value: ${OPENSTACK_API_SERVER_ALLOWED_CIDRS:=[]}
+    - name: bastion
+      value:
+        enabled: ${OPENSTACK_BASTION_ENABLED:=false}
+        spec:
+          flavor: ${OPENSTACK_BASTION_FLAVOR:=m1.small}
+          image:
+            filter:
+              name: ${OPENSTACK_BASTION_IMAGE_NAME:=ubuntu-2404-kube}
```
lentzi90, line 36: Let's align this with what we have in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/image-template-bastion.yaml
```suggestion
              name: ${OPENSTACK_BASTION_IMAGE_NAME:=ubuntu-24.04}
```

templates/cluster-template-development.yaml
```diff
@@ -24,3 +24,16 @@ spec:
...
+          sshKeyName: ${OPENSTACK_SSH_KEY_NAME:=""}
+        floatingIP: ${OPENSTACK_BASTION_FLOATING_IP:=""}
+        availabilityZone: ${OPENSTACK_BASTION_AVAILABILITY_ZONE:=""}
```
lentzi90, line 39: I would drop this also and keep it in the ClusterClass. Advanced users can create their own cluster (template) and specify this if they need.

```suggestion
```

docs/book/src/development/development.md
```diff
@@ -138,6 +138,19 @@ kustomize_substitutions:
...
+  # [Optional] Enable bastion host for secure SSH access
+  OPENSTACK_BASTION_ENABLED: "false"
+  # [Optional] Bastion host flavor (when enabled)
+  OPENSTACK_BASTION_FLAVOR: "m1.small"
+  # [Optional] Bastion host image name (when enabled)
+  OPENSTACK_BASTION_IMAGE_NAME: "ubuntu-2404-kube"
+  # [Optional] Bastion host floating IP (when enabled)
+  OPENSTACK_BASTION_FLOATING_IP: ""
+  # [Optional] Bastion host availability zone (when enabled)
+  OPENSTACK_BASTION_AVAILABILITY_ZONE: ""
```
lentzi90, line 152: These are mainly per-cluster variables so I would not put them in the config file.

```suggestion
```

docs/book/src/development/development.md
```diff
@@ -156,6 +169,8 @@ Make sure that the image or sysext overlay you use exists for and works with tha
 **Note:** All clusters created from the dev-test ClusterClass will require a secret named `dev-test-cloud-config` with the `clouds.yaml` to be used by CAPO for interacting with OpenStack.
 You can create it manually or see below how to make Tilt automate it.
 
+**Security Note:** When using `allowedCIDRs`, ensure you include the IP range of your management cluster to maintain CAPO connectivity. When enabling a bastion host, configure appropriate SSH key access and firewall rules.
+
```
lentzi90, line 173: I think we already have this documented in the "book" and would say that it is unnecessary in the development notes.

```suggestion
```

lentzi90 (commented):
Thanks for the PR!

templates/cluster-template-development.yaml
bnallapeta: Done.

templates/cluster-template-development.yaml
bnallapeta: Ah, I was trying to match the instance image from other files. Makes sense to match the bastion image. Updated.

templates/cluster-template-development.yaml
bnallapeta: Done.

docs/book/src/development/development.md
bnallapeta: That's why I put the `Optional` tag there. But sure, removed it.

docs/book/src/development/development.md
bnallapeta: Found it in configuration.md

Removed.

PR #2726, opened 2025-09-21
🐛Remove waitingForServer loop to allow controller to move forward and do further checks

### LINKED ISSUE #2265: Missing image is hard to debug

smoshiur1237:
Here is a situation where openStack machine controller stuck in a loop for waiting for Server and doesn't do further checks for any other error and that is leading to wrong error message. It would solve the reporting of missing image for OpenStack machine.
Fixes #2265

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -361,8 +361,8 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope 
 
 	scope.Logger().Info("Reconciling Machine")
 
-	machineServer, waitingForServer, err := r.reconcileMachineServer(ctx, scope, openStackMachine, openStackCluster, machine)
-	if err != nil || waitingForServer {
+	machineServer, err := r.reconcileMachineServer(ctx, scope, openStackMachine, openStackCluster, machine)
+	if err != nil {
 		return ctrl.Result{}, err
 	}
 
@@ -572,7 +572,7 @@ func openStackMachineSpecToOpenStackServerSpec(openStackMachineSpec *infrav1.Ope
 
 // reconcileMachineServer reconciles the OpenStackServer object for the OpenStackMachine.
 // It returns the OpenStackServer object and a boolean indicating if the OpenStackServer is ready.
-func (r *OpenStackMachineReconciler) reconcileMachineServer(ctx context.Context, scope *scope.WithLogger, openStackMachine *infrav1.OpenStackMachine, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine) (*infrav1alpha1.OpenStackServer, bool, error) {
+func (r *OpenStackMachineReconciler) reconcileMachineServer(ctx context.Context, scope *scope.WithLogger, openStackMachine *infrav1.OpenStackMachine, openStackCluster *infrav1.OpenStackCluster, machine *clusterv1.Machine) (*infrav1alpha1.OpenStackServer, error) {
 	var server *infrav1alpha1.OpenStackServer
 	server, err := r.getOrCreateMachineServer(ctx, openStackCluster, openStackMachine, machine)
 	if err != nil {
@@ -580,13 +580,10 @@ func (r *OpenStackMachineReconciler) reconcileMachineServer(ctx context.Context,
 		// we won't requeue the request so reconcileNormal can add conditions to the OpenStackMachine
 		// and we can see the error in the logs.
 		scope.Logger().Error(err, "Failed to get or create OpenStackServer")
-		return server, false, err
+		return server, err
 	}
-	if !server.Status.Ready {
-		scope.Logger().Info("Waiting for OpenStackServer to be ready", "name", server.Name)
-		return server, true, nil
-	}
-	return server, false, nil
+
+	return server, nil
 }
 
 // getOrCreateMachineServer gets or creates the OpenStackServer object for the OpenStackMachine.


smoshiur1237:
/cc @lentzi90 
Please check if I need to do any other changes

lentzi90:
Adding some notes from our code dive.

The early return when waiting for the OSS to become ready, here
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6d42cbb56e08d01c2b7d8fecc9fd3ae3ad4e6a9c/controllers/openstackmachine_controller.go#L364-L367

means that we never reconcile the OSM state, here
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/6d42cbb56e08d01c2b7d8fecc9fd3ae3ad4e6a9c/controllers/openstackmachine_controller.go#L412

So we never update the status of the OSM while waiting on the OSS to become ready.
The bug here is that the OSS can be in error state and we just keep waiting indefinitely without propagating that information up.

controllers/openstackmachine_controller.go
```diff
@@ -361,8 +361,8 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 
 	scope.Logger().Info("Reconciling Machine")
 
-	machineServer, waitingForServer, err := r.reconcileMachineServer(ctx, scope, openStackMachine, openStackCluster, machine)
-	if err != nil || waitingForServer {
+	machineServer, err := r.reconcileMachineServer(ctx, scope, openStackMachine, openStackCluster, machine)
+	if err != nil {
```
EmilienM, line 365: Line 377, if server isn't ready, InstanceID might be nil, causing panic

EmilienM (changes requested):
The real problem isn't the waiting loop - it's that when OpenStack returns errors about missing images, those errors aren't properly captured, converted to meaningful conditions, or propagated to user-visible status.

Removing the waiting mechanism doesn't improve error visibility and may introduce new issues:
- Controller proceeds with invalid/missing images without proper validation
- Potential runtime panics when accessing non-existent instance details
- No improvement in user-facing error messages

A better approach would focus on the OpenStackServer controller to:
- Add image validation before instance creation - ORC should be done that, maybe CAPO needs to signal this better to the users.
- Implement specific error conditions (ImageNotFound, ImageValidationFailed)
- Improve status reporting with meaningful error messages
- Ensure proper error propagation from server to machine status

The waiting logic serves a valid purpose and should remain until we have proper error detection and reporting mechanisms in place.

lentzi90:
The instance ID that might be `nil` is a good point. We need to handle that.
I don't quite agree with the rest though.

> The real problem isn't the waiting loop - it's that when OpenStack returns errors about missing images, those errors aren't properly captured, converted to meaningful conditions, or propagated to user-visible status.

They do propagate to the OpenStackServer, where the ERROR condition is seen along with the explanation.
Furthermore, we even have propagation in place, but we do not reach it because of that wait.

We are waiting [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/77c09c2a7805586b2321a22d5c05df7263ef9908/controllers/openstackmachine_controller.go#L364-L367) for the OpenStackServer to have `status.ready`. So until it is completely green from ORC point of view, we do not proceed. This means that we never get to the place where we already have the [propagation logic](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/77c09c2a7805586b2321a22d5c05df7263ef9908/controllers/openstackmachine_controller.go#L424-L465). We will never know if the server went into error state or why. The controller will happily sit here waiting for the server to become ready while ORC knows and has already told us that the image is missing or the instance failed to build or there was not enough capacity or the credentials didn't work, or any other reason.

We definitely need to handle the case when the instance ID is missing though. From what I understand, we only need the instance ID for control plane nodes. What if we move the `reconcileMachineState` call above all that? So we would move https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/77c09c2a7805586b2321a22d5c05df7263ef9908/controllers/openstackmachine_controller.go#L412-L415
up to happen immediately after `reconcileMachineServer`. It even has a requeue to wait for the server to become ACTIVE, very similar to the `waitForServer` that is there now.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/77c09c2a7805586b2321a22d5c05df7263ef9908/controllers/openstackmachine_controller.go#L458-L463

To me that would make sense because the OpenStackServer state is fundamentally "lower level" (happens before and is common for all nodes) than the control-plane logic with API loadbalancer.

lentzi90:
Ok I have been testing to see what works and not. There are at least two different cases:

1. Image filter is used and does not match any image.
   This is silent in both ORC and CAPO. There is no visible error on the OSS or OSM, they are just "not ready". Only this log line shows up in CAPO:
   ```
   E0926 11:29:21.102342      97 controller.go:353] "Reconciler error" err="no images were found with the given image filter: name=flatcar_production-v1.33.1, tags=[]" controller="openstackserver" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="OpenStackServer" OpenStackServer="default/development-7058-9kfgq-f9cpz" namespace="default" name="development-7058-9kfgq-f9cpz" reconcileID="4344ee80-7c8e-45a2-afe7-d94564611732"
   ```

2. Image ID is used. This shows up nicely in the OSS but is not propagated to the OSM (because of what we discussed above).

Then I tried moving `reconcileMachineState` up and rerun. This solves the second case quite well (with some room for improvement). The OSM status then looks like this:
```yaml
status:
  conditions:
  - lastTransitionTime: "2025-09-26T11:26:34Z"
    reason: InstanceStateError
    severity: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2025-09-26T11:26:34Z"
    reason: InstanceStateError
    severity: Error
    status: "False"
    type: InstanceReady
  failureMessage: instance state 0xc000f437c0 is unexpected
  failureReason: UpdateError
```
We could improve this to show the state instead of the memory address, and propagate the reason.

The first case is not solved by this. In fact we get panic because `reconcileMachineState` tries to check the `InstanceState` of the OSS, and we do not have an instance yet. I think this should be fixed in both CAPO and ORC. ORC should set `InstanceCreateFailed` (as it does in case 2) and CAPO should check if the instance state is available and set a condition to show if it is not. We should also propagate the InstanceCreateFailed condition from ORC when it is available.
I have checked that the panic goes away if we add the check for the instance state.

Edit: I also tested using imageRef. It behaves the same as case 1.

Edit 2: Note that we are not using ORC for OpenStackServer yet, so this is all CAPO. But we will need to check and fix also in ORC if we find issues. Otherwise we will diverge.

smoshiur1237:
No activities from my side.  I will open a new PR if needed . But we can close this PR for now

PR #2751, opened 2025-10-06
🌱 Dockerfile: add the ARG defaulting within the Dockerfile

fmuyassarov:
Add the default value for `GO_VERSION` within the Dockerfile to fix build checks warnings that come from an empty variable. This doesn’t break existing behavior; version control is still managed by the Makefile
```
 - InvalidDefaultArgInFrom: Default value for ARG golang:${GO_VERSION} results in empty or invalid base image name (line 17)
```

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 # Build the manager binary
 ARG GO_VERSION
-FROM golang:${GO_VERSION} AS builder
+FROM golang:${GO_VERSION:-1.24.7} AS builder
 WORKDIR /workspace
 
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/lgtm

smoshiur1237:
> /approve The only downside is we should remember to update this also when bumping go.

I will update soon to 1.24.8 when it releases, I will remember to do that

fmuyassarov:
> /approve The only downside is we should remember to update this also when bumping go.

Right. But even if it is missed accidentally, it should not cause a problem assuming the version will be passed from the Makefile.

Issue #2742, opened 2025-09-30
update-codegen.sh broken with go1.25 on release-0.11 and release-0.12


lentzi90:
Presubmit tests are broken on release-0.11 and release-0.12.
We are not really supporting release-0.11 anymore so that is less of an issue, but release-0.12 should definitely be fixed.

I am quite confident that they broken when the kubekins-e2e image was bumped to one with go1.25. The image is set in https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-openstack/cluster-api-provider-openstack-presubmits.yaml.

The error happens when running `make generate-codegen`, which calls `hack/update-codegen.sh` and looks like this:
```
 + go install k8s.io/kube-openapi/cmd/openapi-gen
go: downloading k8s.io/klog/v2 v2.120.1
go: downloading golang.org/x/tools v0.21.0
go: downloading github.com/emicklei/go-restful/v3 v3.11.0
go: downloading google.golang.org/protobuf v1.33.0
go: downloading github.com/go-logr/logr v1.4.1
go: downloading golang.org/x/sync v0.7.0
go: downloading golang.org/x/mod v0.17.0
# golang.org/x/tools/internal/tokeninternal
../../golang.org/x/tools@v0.21.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
make[1]: *** [Makefile:64: generate-codegen] Error 1
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-openstack/orc'
make: *** [Makefile:319: generate-orc] Error 2 
```

The issue (as far as I understand) is that the script uses a script (`kube_codegen.sh`) in the upstream package that builds openapi-gen outside of CAPO. This means that we have no control over the dependencies there and thus cannot fix them to work with go1.25. If I do the same `go install k8s.io/kube-openapi/cmd/openapi-gen` in CAPO, it works.

CAPI has a totally different way of doing this. I wonder if we should try to change to that way. See:
https://github.com/kubernetes-sigs/cluster-api/blob/a0bb17c9c2647c5aaa80acc3385bb1adfceeea36/Makefile#L1441-L1445
https://github.com/kubernetes-sigs/cluster-api/blob/a0bb17c9c2647c5aaa80acc3385bb1adfceeea36/Makefile#L456-L465

lentzi90:
[...]

PR #2761, opened 2025-10-07
🌱 Refactor generate-codegen

lentzi90:

This commit changes how to run and build codegen. Instead of using a script inside the openapi-gen package, we now build and call the binary directly. This avoids issues where the upstream package is incompatible with the go version we use (as seen currently on release-0.12).

Related to (but does not fix) https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2742
We need to do a manual backport of this to fix that issue.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90:
/hold cancel
/cc @smoshiur1237 @adilGhaffarDev 
Can you check if this makes sense?
The reason I want this merged is that the current way of doing this has broken tests on release-0.12 (https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2742). Main is unaffected at this time, but it would not make sense to do code generation differently on a release branch than on main. So I want to refactor here on main and then backport (see https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2762) to fix the release branch.

adilGhaffarDev:
LGTM

lentzi90:
Thank you both!
/approve

PR #2789, opened 2025-10-16
✨ Add trunk subports support for OpenStackMachine ports

bnallapeta:
## What this implements
This adds the ability to specify subports when creating trunk ports. Users can now define VLAN subports that will be attached to the trunk when the port is created.

Example usage:
```
  ports:
  - network: 
       filter:
         name: foo_vlan_300
     trunk: true 
     subports:
     - network:
          filter:
            name: bar_vlan_301
        segmentationID: 301
        segmentationType: "vlan"
```

- Added SubportOpts and ResolvedSubportSpec types to API
- Since sub-ports and ports are essentially the same entities CommonPortOpts and CommonResolvedPortSpec types were introduced as a basis for regular ports and trunk subports
- Added SubPortStatus to keep track of ports which are associated with trunk
- Introduced separate method for adding supports to trunk
- Added AddSubports method to NetworkClient interface
- Added validation to ensure subports require trunk enabled
- Updated port normalization to resolve subports

## Testing done
- unit test
- tests on real env


lentzi90:
/ok-to-test
This is so large that I think we should at least have an issue, perhaps even a proposal to discuss and design it first

bnallapeta:
Closing this due to https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2862

PR #2791, opened 2025-10-20
 🌱Release automation workflow

smoshiur1237:
Release automation workflow to improve the release work for future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -448,14 +448,13 @@ staging-manifests:
 ## --------------------------------------
 ##@ Release
 ## --------------------------------------
-
 ifneq (,$(findstring -,$(RELEASE_TAG)))
     PRE_RELEASE=true
 endif
 PREVIOUS_TAG ?= $(shell git tag -l | grep -E "^v[0-9]+\.[0-9]+\.[0-9]+$$" | sort -V | grep -B1 $(RELEASE_TAG) | head -n 1 2>/dev/null)
 ## set by Prow, ref name of the base branch, e.g., main
 RELEASE_DIR := out
-RELEASE_NOTES_DIR := _releasenotes
+RELEASE_NOTES_DIR := releasenotes
 
 .PHONY: $(RELEASE_DIR)
 $(RELEASE_DIR):
@@ -478,7 +477,7 @@ list-image:
 	gcloud container images list-tags $(STAGING_REGISTRY)/$(IMAGE) --filter="tags=('$(RELEASE_TAG)')" --format=json
 
 .PHONY: release
-release: $(RELEASE_NOTES) clean-release $(RELEASE_DIR)  ## Builds and push container images using the latest git tag for the commit.
+release: $(RELEASE_NOTES) $(RELEASE_DIR)  ## Builds and push container images using the latest git tag for the commit.
 	@if [ -z "${RELEASE_TAG}" ]; then echo "RELEASE_TAG is not set"; exit 1; fi
 	@if ! [ -z "$$(git status --porcelain)" ]; then echo "Your local git repository contains uncommitted changes, use git clean before proceeding."; fi
 	git checkout "${RELEASE_TAG}"
@@ -486,6 +485,7 @@ release: $(RELEASE_NOTES) clean-release $(RELEASE_DIR)  ## Builds and push conta
 	$(MAKE) manifest-modification REGISTRY=$(PROD_REGISTRY)
 	$(MAKE) release-manifests
 	$(MAKE) release-templates
+	$(MAKE) generate-release-notes
 
 .PHONY: manifest-modification
 manifest-modification: # Set the manifest images to the staging/production bucket.
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,44 +1,118 @@
-name: release
+# This code is borrowed from https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/release.yaml
+name: Create Release
 
 on:
   push:
-    # Sequence of patterns matched against refs/tags
-    tags:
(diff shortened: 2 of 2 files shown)


hack/tools/release/notes.go
lentzi90, line 1: Do we really need this? I was thinking that we can still use it directly from CAPI.
Currently we have it in the Makefile here:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/b9fa78bbfef0c2a64cdba8af3eec65d0926178ff/Makefile#L552
and the binary is built here:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/b9fa78bbfef0c2a64cdba8af3eec65d0926178ff/hack/tools/Makefile#L163-L165

smoshiur1237: Good catch, I missed it totally. I didn't notice that it is already using the notes. So It won't be needed in CAPO

Makefile
```diff
@@ -448,14 +448,14 @@ staging-manifests:
 ## --------------------------------------
 ##@ Release
 ## --------------------------------------
-
+RELEASE_TAG ?= $(shell git describe --abbrev=0 2>/dev/null)
```
lentzi90: This is already defined here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/b9fa78bbfef0c2a64cdba8af3eec65d0926178ff/Makefile#L97
It is ok to move but we should not keep both.

Makefile
```diff
@@ -633,10 +634,6 @@ clean-temporary: ## Remove all temporary files and folders
 	rm -f kubeconfig
 	rm -rf _artifacts
 
-.PHONY: clean-release
-clean-release: ## Remove the release folder
-	rm -rf $(RELEASE_DIR)
-
```
lentzi90, line 639: I think we could still keep this in case someone want's to manually generate artifacts and clean up.

lentzi90 (commented):
Two small comments, otherwise this looks good 👍

Makefile
smoshiur1237: Ok I kept the original one

lentzi90 (commented):
/approve

lentzi90:
/cc @bnallapeta 
@smoshiur1237 the description still says WIP, can you edit if it is ready now? Feel free to remove hold also.

smoshiur1237:
@lentzi90  thanks for your review. We can take this in. I will update doc in separate PR.

smoshiur1237:
> Looks good overall. A few questions/clarifications:
> 
> 1. Will we actually create the releasenotes directory during next release? I think we should create it right now as part of this PR.
> 2. Need to clearly document this in [release.md](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/RELEASE.md) as this is a complete change in process.
1. I think releasenote directory will be created on the first release after this PR, if it is not present. We just need to push the chages with a new PR with each release. The newly created `md` file,  for example `v0.14.1.md`. As a maintainer, we need to add a commit with the change and push the commit.   The automation will create tags and push them to upstream. 
2. I will add the documentation to the `release.md`  in separate PR, if it is ok?

smoshiur1237:
/hold 
It may need few changes in `makefile` file. Will update the makefile

bnallapeta:
> > Looks good overall. A few questions/clarifications:
> > 
> > 1. Will we actually create the releasenotes directory during next release? I think we should create it right now as part of this PR.
> > 2. Need to clearly document this in [release.md](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/RELEASE.md) as this is a complete change in process.
> 
> 1. Release note will be created and saved in separate folder named `releasenotes` and it will add  the newly created `md` file,  for example `v0.14.1.md` file. As a maintainer, you need to add a commit with the change and push the commit.   The automation will create tags and push them to upstream.
> 2. I will add the documentation to the `release.md`  in separate PR, if it is ok?

1. So, for the first time, someone needs to create the folder and then add the file. Can we not create the folder right now and add a readme.md with basic info and refine it later in a separate PR? (maybe a .gitkeep too)
2. I guess its alright, since we don't have any releases soon. But please prioritize that soon.

smoshiur1237:
> > > Looks good overall. A few questions/clarifications:
> > > 
> > > 1. Will we actually create the releasenotes directory during next release? I think we should create it right now as part of this PR.
> > > 2. Need to clearly document this in [release.md](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/RELEASE.md) as this is a complete change in process.
> > 
> > 
> > 
> > 1. Release note will be created and saved in separate folder named `releasenotes` and it will add  the newly created `md` file,  for example `v0.14.1.md` file. As a maintainer, you need to add a commit with the change and push the commit.   The automation will create tags and push them to upstream.
> > 2. I will add the documentation to the `release.md`  in separate PR, if it is ok?
> 
> 1. So, for the first time, someone needs to create the folder and then add the file. Can we not create the folder right now and add a readme.md with basic info and refine it later in a separate PR? (maybe a .gitkeep too)
> 2. I guess its alright, since we don't have any releases soon. But please prioritize that soon.

1. I am calling in the makefile with the following which is added under` make release` :  `generate-release-notes: $(RELEASE_NOTES_DIR) $(RELEASE_NOTES)`, which will check if the directory is present or not. So it is not needed to create the folder manually. I think that should be ok.
2. yes I will add all necessary instructions to complete this.

lentzi90:
Folder is fine to add with the first release notes, when we make the next release.
Doc update is needed and can be done separately (but please sooner rather than later).

smoshiur1237:
I have checked the makeFile, It looks ok
/hold cancel

PR #2792, opened 2025-10-21
🌱 E2E: Use kind cluster for clusterctl upgrade tests

### LINKED ISSUE #2098: E2E: Use kind cluster for clusterctl upgrade tests
Traditionally, the clusterctl upgrade tests created a workload cluster to be used as secondary management cluster, with the old versions deployed. From that cluster, another workload test cluster was created to check that the old versions worked, before upgrading and finally verifying that the upgraded versions worked.
Creating two workload clusters in this way is quite resource intensive compared to how all the other tests use kind  for management and only create a single test cluster. 

CAPI has now [made it possible](https://github.com/kubernetes-sigs/cluster-api/issues/7613) to use kind also for the secondary management cluster! We should adopt this as it will lower the resource requirement and potentially allow us to dump parallelism to 2 for our optional tests.

lentzi90:

This enables the UseKindForManagementCluster flag for all clusterctl upgrade tests, which makes them use a kind cluster as the secondary management cluster instead of creating an OpenStack workload cluster.

This significantly reduces the resource requirements for these tests by using lightweight kind containers instead of full VMs, and it also speeds up the test execution.

Fixes #2098

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -218,6 +218,7 @@ variables:
   CNI: "../../data/cni/calico.yaml"
   CCM: "../../data/ccm/cloud-controller-manager.yaml"
   EXP_CLUSTER_RESOURCE_SET: "true"
+  IP_FAMILY: "ipv4"
   OPENSTACK_BASTION_IMAGE_NAME: "cirros-0.6.1-x86_64-disk"
   OPENSTACK_BASTION_IMAGE_URL: https://storage.googleapis.com/artifacts.k8s-staging-capi-openstack.appspot.com/test/cirros/2022-12-05/cirros-0.6.1-x86_64-disk.img
   OPENSTACK_BASTION_IMAGE_HASH: 0c839612eb3f2469420f2ccae990827f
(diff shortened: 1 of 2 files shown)


lentzi90:
Looks like it works, and saves a few minutes for the test also!
/hold cancel

lentzi90:
There is a bit of cleanup we could do after this also, but let's take it separately in case we want to revert for some reason.
The things we can cleanup is this:
- Image upload: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/b9fa78bbfef0c2a64cdba8af3eec65d0926178ff/scripts/ci-e2e.sh#L100-L104
- Image download patches: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/data/kustomize/components/upgrade-patches/ci-hack-kct.yaml

These are used to upload the CAPO image that is being tested into the management cluster. All other tests use kind clusters where the image is loaded directly, but the clusterctl upgrade was using a full CAPO cluster so we used this to get the image into it.

lentzi90:
/approve
This is ready to take in

PR #2799, opened 2025-10-25
:bug: Define subnetID on LB member when networks differ

### LINKED ISSUE #2790: Loadbalancer member not created with subnet id
/kind bug

Create a new cluster with a loadbalancer with a VIP subnet in one network while the cluster resides on another network with this configuration:
```
kind: OpenStackCluster
...
spec:
  apiServerLoadBalancer:
    enabled: true
    provider: ovn
    network:
      filter:
        name: network1
    subnets:
    - id: b9be113f-7c72-40a8-9a8f-51413a43a4f6
  ...
  network:
    filter:
      name: network2
  subnets:
  - id: 794735f4-6d6b-4b79-9784-2484d8f33109
```

Reconciling the loadbalancer for the control plane fails with a `LoadBalancerMemberError` with the following error:
```
Provider 'ovn' does not support a requested option: Subnet is required, or Loadbalancer associated with Pool must have a subnet, for Member creation with OVN Provider Driver if it is not the same as LB VIP subnet
```

I reproduced the error with this command:
```
$ openstack loadbalancer member create --address 10.2.0.222 --protocol-port 6443 82b18fb4-3c11-44dc-8b64-9f13af12b36f
Provider 'ovn' does not support a requested option: Subnet is required, or Loadbalancer associated with Pool must have a subnet, for Member creation with OVN Provider Driver if it is not the same as LB VIP subnet (HTTP 501) (Request-ID: req-76ddc667-989b-4c00-b659-98cd98c61ef1)
```

An OVN loadbalancer can have a VIP subnet that is in a different network than the pool members.
It is possible to add a new member by providing the subnet id:
```
$ openstack loadbalancer member create --address 10.2.0.222 --protocol-port 6443 --subnet-id 794735f4-6d6b-4b79-9784-2484d8f33109 82b18fb4-3c11-44dc-8b64-9f13af12b36f
```

The loadbalancer then connects to the Kubernetes control plane machine.

When registering a loadbalancer member the subnet id should be provided if the LB VIP subnet differs from the cluster subnet. 

I traced this to these lines:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/pkg/cloud/services/loadbalancer/loadbalancer.go#L706-L711

The subnet id of the OpenstackMachine is available when this method is called.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.12.5 
- Cluster-API version: v1.10.7
- OpenStack version: 2024.1
- Minikube/KIND version: N/A
- Kubernetes version (use `kubectl version`): N/A

nikParasyr:
Define subnetID on ovn LB member creation when the user is using different networks for the cluster and the loadbalancer

Fixes #2790

In general I have some concerns with the approach, would like input from your side:
1. I'm using `openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.ID`. While working on this i noticed that the field is not always populated ( related [PR](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2798) )
2. I'm checking whether the provider is "ovn", amphora also supports different subnet IDs (although i've never seen it in action). I could drop this check but it would require point 1 being resolved ( in theory we could always set subnetID on member create opts == `openStackCluster.Status.Network.Subnets[0].ID` -- see point 3 )
3. The subnetID is set to `openStackCluster.Status.Network.Subnets[0].ID`. A control-plane machine can be attached to different subnets ( if `OSM.spec.ports`. are defined ). I couldnt find a way to get the subnetID from the spec though as a user can define just a `spec.ports[0].network.id` (or a filter) and `OSM.status` doesn't provide info on subnetID. So **the assumption is that the control-plane machine is using the cluster subnet**.

**TODOs**:

- [X] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold

### DIFF
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -710,6 +710,10 @@ func (s *Service) ReconcileLoadBalancerMember(openStackCluster *infrav1.OpenStac
 			Tags:         openStackCluster.Spec.Tags,
 		}
 
+		if openStackCluster.Status.Network.ID != openStackCluster.Status.APIServerLoadBalancer.LoadBalancerNetwork.ID {
+			lbMemberOpts.SubnetID = openStackCluster.Status.Network.Subnets[0].ID
+		}
+
 		if _, err := s.waitForLoadBalancerActive(lbID); err != nil {
 			return err
 		}
--- a/pkg/cloud/services/loadbalancer/loadbalancer_test.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/providers"
 	. "github.com/onsi/gomega" //nolint:revive
 	"go.uber.org/mock/gomock"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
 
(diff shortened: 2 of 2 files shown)


bnallapeta (commented):
@nikParasyr Thanks for the fix! I am not sure if checking for each provider is really optimal; especially in the long run.

Currently, if I am not wrong, in Octavia, member SubnetID is optional and can be provided. If it’s omitted, Octavia will infer a suitable subnet from the member port’s network. However, in cross-network scenarios (e.g., VIP network != member network), OVN requires an explicit SubnetID for the member to be created correctly. Amphora accepts SubnetID too, so providing it is generally safe.

I can see two options that could be provider-agnostic.

1. Always set SubnetID based on the member’s actual subnet
2. Set SubnetID whenever the VIP network differs from the member’s network

Either path avoids per-provider conditionals and works even when the default provider is OVN without being specified in the spec. Let me know what you think.

cc @lentzi90

nikParasyr:
> With #2798 merged, can we drop the check for `ovn`? I think that would be good to do. The other check for LB vs cluster network ID is good IMO. You are right that we could just unconditionally set the ID but I think there is a point also relying on the defaults as we have been doing so far.

Done.

PR #2800, opened 2025-10-26
:sparkles: Enable setting QoSPolicy on ports

### LINKED ISSUE #2672: Add QoSPolicyID to OpenstackCluster gateway settings
/kind feature

Be able to configure neutron ports with qos profile via id and/or filter.

Something similar had been already filled [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1007) with a proposed solution [here](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1008) but seemed to have been abandoned

nikParasyr:
Enable users to set QoSPolicyID on ports. This can be done by either defining the ID or a filter to query the QoS Policy.

Fixes #2672

- Chose to add both ID and Filter to keep UX consistent with other fields
- ~~I have not added e2e tests, did a quick look and qos extension doesnt seem enabled. Let me know if this is required~~

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold

### DIFF
--- a/hack/ci/cloud-init/controller.yaml.tpl
+++ b/hack/ci/cloud-init/controller.yaml.tpl
@@ -38,7 +38,7 @@
 
     # Neutron
     enable_plugin neutron https://github.com/openstack/neutron stable/${OPENSTACK_RELEASE}
-    ENABLED_SERVICES+=,q-svc,neutron-trunk,ovn-controller,ovs-vswitchd,ovn-northd,ovsdb-server,q-ovn-metadata-agent
+    ENABLED_SERVICES+=,q-svc,neutron-trunk,ovn-controller,ovs-vswitchd,ovn-northd,ovsdb-server,q-ovn-metadata-agent,q-qos
 
     DISABLED_SERVICES=q-agt,q-dhcp,q-l3,q-meta,q-metering
     PUBLIC_BRIDGE_MTU=${MTU}
@@ -197,6 +197,9 @@
     openstack quota set --secgroup-rules 1000 demo
     openstack quota set --secgroups 100 admin
     openstack quota set --secgroup-rules 1000 admin
+
+    # Validate qos extension is set
+    openstack extension list --network | grep qos
 - path: /root/devstack.sh
   permissions: "0755"
   content: |
--- a/pkg/cloud/services/networking/qospolicy.go
+++ b/pkg/cloud/services/networking/qospolicy.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/cloud/services/networking/qospolicy_test.go
+++ b/pkg/cloud/services/networking/qospolicy_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ import (
 
 	"github.com/go-logr/logr/testr"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/qos/policies"
-	. "github.com/onsi/gomega" //nolint:revive
+	. "github.com/onsi/gomega"
 	"go.uber.org/mock/gomock"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
--- a/test/e2e/data/kustomize/multi-network/patch-machine-template-networks.yaml
+++ b/test/e2e/data/kustomize/multi-network/patch-machine-template-networks.yaml
@@ -3,9 +3,17 @@
   path: /spec/template/spec/ports
   value:
   - description: "primary"
+    qosPolicy:
+      id: "${QOS_POLICY_ID}"
   - description: "Extra Network 1"
     network:
       id: "${CLUSTER_EXTRA_NET_1}"
+    qosPolicy:
+      filter:
+        name: "${QOS_POLICY_NAME}"
   - description: "Extra Network 2"
     network:
       id: "${CLUSTER_EXTRA_NET_2}"
+    qosPolicy:
+      filter:
+        shared: true
--- a/test/e2e/shared/openstack.go
+++ b/test/e2e/shared/openstack.go
@@ -42,6 +42,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/loadbalancers"
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/layer3/routers"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/qos/policies"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/groups"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/rules"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/trunks"
@@ -65,6 +66,11 @@ import (
 type ServerExtWithIP struct {
 	servers.Server
 	ip string
+}
+
+type PortWithQoS struct {
+	ports.Port
+	policies.QoSPolicyExt
 }
 
 // ensureSSHKeyPair ensures A SSH key is present under the name.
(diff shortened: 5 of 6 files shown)


docs/book/src/clusteropenstack/configuration.md
```diff
@@ -571,6 +572,52 @@ spec:
         ...
 ```
 
+### Port QoS Policy
+
+Asigning a QoS Policy to a port can be done by specifying the id or a filter. More details about the filter can be found in [QoSPolicyParam](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1beta1/types.go). When using filters to look up a QoS policy, please note that exactly one policy must be returned
```
bnallapeta: Asigning => Assigning

api/v1beta1/types.go
```diff
@@ -290,6 +290,32 @@ type AllocationPool struct {
...
+	// ID is the ID of the QoS policy to use. If ID is provided, filter cannot be provided. Must be in UUID format.
+	// +kubebuilder:validation:Format:=uuid
+	// +optional
+	ID optional.String `json:"id,omitempty"`
+
+	// Filter specifies a filter to select an OpenStack QoS policy. If provided, cannot be empty.
+	Filter *QoSPolicyFilter `json:"filter,omitempty"`
+}
+
+// QoSPolicyFilter specifies a query to select an OpenStack router. At least one property must be set.
```
bnallapeta: select an OpenStack router => select an OpenStack QoS Policy

bnallapeta: Remember to regenerate docs after you change this.

pkg/cloud/services/networking/port.go
```diff
@@ -240,6 +241,15 @@ func (s *Service) EnsurePort(eventObject runtime.Object, portSpec *infrav1.Resol
 		builder = portSecurityOpts
 	}
 
+	if portSpec.QoSPolicyID != nil {
```
bnallapeta, line 266: this only works in the Create scenario. There is no handling for Update cases. 

There is already an existing `ensurePortTagsAndTrunk` function that handles the update scenario. But that takes care of just the port tags and trunk. I think we should modify this to handle the update cases for QoS as well. 
Rename it to something like `ensurePortProperties` and add QoS as well.

pkg/cloud/services/networking/port.go
```diff
@@ -487,6 +497,15 @@ func (s *Service) normalizePorts(ports []infrav1.PortOpts, clusterResourceName,
 			return nil, err
 		}
 
+		// Resolve QoS Policy ID
+		if port.QoSPolicy != nil {
```
bnallapeta, line 543: I think this would cause issues if the QoS is not enabled in OpenStack no? 

How about adding a helper function here? something like `IsQoSSupported` and we could define that in qospolicy.go to check if the extension is enabled or not.

api/v1beta1/types.go
```diff
@@ -422,6 +453,10 @@ type ResolvedPortSpec struct {
 	// +listType=atomic
 	SecurityGroups []string `json:"securityGroups,omitempty"`
 
+	// QoSPolicyID is the ID of the qos policy the port will use.
+	// +optional
+	QoSPolicyID *string `json:"qosPolicyID"`
```
bnallapeta: `omitempty` missing?

pkg/cloud/services/networking/port.go
nikParasyr: Done. Similarly to how trunk was done but merged some code to not have duplicate

pkg/cloud/services/networking/port.go
nikParasyr: Updated to also handle the Update scenario. Also added a unit test for that use case.
Question: Do I also need to update webhooks to allow mutations for qosPolicy? and if so are there some pointers for that?

api/v1beta1/types.go
nikParasyr: indeed. fixed

api/v1beta1/types.go
nikParasyr: done

docs/book/src/clusteropenstack/configuration.md
nikParasyr: done

bnallapeta:
> @bnallapeta thanks for the review. Your suggestions were implemented, have a minor question on the update comment. Moreover, I updated ci and e2e to cover this case ( decided to add it on the existing multi-network scenario since it allows to test various ways of setting qos policy and didnt want to make another scenario just for this )

Awesome! thanks for this. I have replied on that comment. Looks good overall.

nikParasyr:
the e2e tests fails on the multi-network one. looking into it but it might take a while

nikParasyr:
@bnallapeta / @lentzi90 can i get some help on where to look/troubleshoot?

The 1st control-plane node doesnt come up.

What i have checked this far:
1. `qos` on devstack seems to be configured correct:
- i can see it in the cloud-init logs, and various ml2 extensions are added correctly automatically
- qos policy gets created ( grep controller devstack logs for `qos-policy` => shows the creation + couple of gets )
- the checks on extension on the code dont fail

2. `qosPolicyID` gets resolved properly:
- no matter if i pass it by id/filter.name/filter.shared the OpenStackServer.status.resolved of the control-plane machine have the correct qosPolicyID

3. Only 1 port gets created, i can see that both on logs and ports.json:
- i can see that it contains qos_policy in the request body ( and no err returned )
- not sure why only 1 port gets created and not 3

4. The OpenStackServer for the control-plane doesnt get created:
- the status is .status.Ready: false
- it is not dumped on servers.json
- i cannot find any logs on the controller or worker devstack => **Given that I have the port ID and server name I would expect that i can at least find the request but i cant**
- k-orc logs are completely empty

It sort of feels to me that this might be k-orc for some reason not creating the server, but not sure if im correct in this and why it wouldn't. (do we need downstream support for QoS on k-orc maybe? while working on this it didnt feel like that but maybe im wrong)

Any insights/help would be much appreciated

bnallapeta:
@nikParasyr curious.. why did you comment out the qos [here](https://github.com/nikParasyr/cluster-api-provider-openstack/blob/qos_policy/test/e2e/data/kustomize/multi-network/patch-machine-template-networks.yaml)? I think the first time the e2e ran on this PR, these were enabled yes? And still the same failure? 

Also, instead of using shared, how about we give the same policy for all ports and see if that works. I'm thinking that if there are multiple policies and for some reason, the policy resolution becomes an issue, the port won't come up and that causes the whole thing to fail. Perhaps you can test this locally. 

I will try to find some time tomorrow to look into this deeply.

hack/ci/cloud-init/controller.yaml.tpl
```diff
@@ -197,6 +197,9 @@
     openstack quota set --secgroup-rules 1000 demo
     openstack quota set --secgroups 100 admin
     openstack quota set --secgroup-rules 1000 admin
+
+    # Validate qos extension is set and create a policy
```
lentzi90: Policy is created later in the e2e test
```suggestion
    # Validate qos extension is set
```

lentzi90:
The missing logs from ORC (and also very minimal from CAPO) is strange and disturbing. Maybe it is because the test gets interrupted and fails to cleanup properly?

Regarding ORC, we are actually not using it for the OpenStackServers yet :slightly_frowning_face: 
See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2814.
This is also an issue because we are now introducing changes in CAPO that diverges from ORC. You will probably need to check and think about how to implement this in ORC as well, or instead of CAPO.

For progressing here though, I think we need to figure out why we are not getting proper logs from CAPO and fix that. Then it will be clear why the server is not created. Maybe try focusing on only the multi-network test? It will be easier (and faster) to see what is going on.

pkg/cloud/services/networking/qospolicy.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
lentzi90: Probably copied another file? Anyway, let's put the current year since this is a new file.
```suggestion
Copyright 2025 The Kubernetes Authors.
```

pkg/cloud/services/networking/qospolicy_test.go
```diff
@@ -0,0 +1,121 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
lentzi90:
```suggestion
Copyright 2025 The Kubernetes Authors.
```

pkg/cloud/services/networking/qospolicy_test.go
```diff
@@ -0,0 +1,121 @@
...
+*/
+
+package networking
+
+import (
+	"testing"
+
+	"github.com/go-logr/logr/testr"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/qos/policies"
+	. "github.com/onsi/gomega" //nolint:revive
```
lentzi90: Is the nolint really needed? We should be covered by the config here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ee4be02e787d9bd401376037731bd3dd06fe4f79/.golangci.yml#L163-L166
or if we are not covered by that we should update it to cover

nikParasyr:
@bnallapeta 

> why did you comment out the qos [here](https://github.com/nikParasyr/cluster-api-provider-openstack/blob/qos_policy/test/e2e/data/kustomize/multi-network/patch-machine-template-networks.yaml)?

This is on one of my later runs to see if i still get the same problems with 1 port with qosPolicy instead of all 3. I have made ci runs with either defining the same qosPolicy.id on all 3 ports or with the uncommented lines => they all fail the same way. History of runs is here: https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=cluster-api-provider-openstack&pr=2800

> I think the first time the e2e ran on this PR, these were enabled yes? And still the same failure?

Initially e2e tests were not added, i added them after tackling your input ( validating qos extension is enabled etc ). And since then they consistently fail.

> Also, instead of using shared, how about we give the same policy for all ports and see if that works. I'm thinking that if there are multiple policies and for some reason, the policy resolution becomes an issue, the port won't come up and that causes the whole thing to fail. 

I am passing the same policy on all 3 ports, just with different ways ( to test that the filter works as expected ). On the openstackServer.Status.Resolved i can see that all 3 are resolved properly to the same qosPolicyID.


--- 

@lentzi90 

> The missing logs from ORC (and also very minimal from CAPO) is strange and disturbing. Maybe it is because the test gets interrupted and fails to cleanup properly?

I have noticed in general that orc doesn't create many logs ( we had servers fail due to quotas multiple times and orc logs are silent -- probably because as you mention below k-orc is still not used by capo for servers... ). 

> Regarding ORC, we are actually not using it for the OpenStackServers yet 🙁
See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2814.
This is also an issue because we are now introducing changes in CAPO that diverges from ORC. You will probably need to check and think about how to implement this in ORC as well, or instead of CAPO.

Ok this is good to know. If this is to happen, its probably better to wait for it and implement it once in k-orc ( and apis etc here ), instead of doing it now on capo, then on k-orc and patching capo again.  from my side this doesn't have the highest priority atm and can wait for a bit for the migration to occur.  I'll check a bit more in-depth the related ticket and k-orc development (maybe i can start implementing qos there and we meet in the middle :P )

> For progressing here though, I think we need to figure out why we are not getting proper logs from CAPO and fix that. Then it will be clear why the server is not created. Maybe try focusing on only the multi-network test? It will be easier (and faster) to see what is going on.

So this PR on the last commit has multiple runs (because i was amending ) https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=cluster-api-provider-openstack&pr=2800. All of them fail on the multi-network (which is the only e2e that i "touched"). CAPO logs there are consistently empty (other than the first 50 lines of starting controllers etc). Unless im checking on the wrong place

hack/ci/cloud-init/controller.yaml.tpl
nikParasyr: done

pkg/cloud/services/networking/qospolicy.go
nikParasyr: done.

pkg/cloud/services/networking/qospolicy_test.go
nikParasyr: done.

pkg/cloud/services/networking/qospolicy_test.go
nikParasyr: done

lentzi90:
If it not urgent for you, it probably makes more sense to focus on ORC as you said. Maybe sync with @mandre since he picked up https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2814

mandre:
> If it not urgent for you, it probably makes more sense to focus on ORC as you said. Maybe sync with @mandre since he picked up #2814

That would be great. We should first start by adding a QoS controller in ORC as we don't have that yet, then update the Port controller to have ability to reference a QoS, and finally use it in CAPO.
@nikParasyr let me know if you want to start looking at working on a QoS controller in ORC, I'd be happy to help you get started. I've just created https://github.com/k-orc/openstack-resource-controller/issues/573 for that purpose.

nikParasyr:
> @nikParasyr let me know if you want to start looking at working on a QoS controller in ORC, I'd be happy to help you get started. I've just created https://github.com/k-orc/openstack-resource-controller/issues/573 for that purpose.

I’ll reach out once I’m back from a trip ( early December )

nikParasyr:
Closed this one as it requires support on k-orc and a couple more things that i will work on a later date

PR #2808, opened 2025-10-28
🌱 E2E: Fix devstack by pinning pip

lentzi90:

This is an attempt to get E2E back on track. I used copilot for this and have to verified yet if the changes makes sense.

/hold


lentzi90:
This won't work.

PR #2833, opened 2025-11-13
🐛 Fix OpenStackServer reconciliation stuck when cluster is unpaused

### LINKED ISSUE #2824: OpenstackServer doesn't retry when Cluster was paued
/kind bug

I pivoted an openstack cluster from a management cluster to another. Due to the nature of this process, the parent CAPI cluster initializes `paused`, all children get moved and then the cluster gets resumed.

When an OpenstackServer is linked to a paused CAPI cluster, reconciliation ends without requeue. As there are not watchers for the parent cluster in OpenstackServer reconciler, the resource will not be reconciled anymore  (unless restarting CAPO or modifying the object).

CAPO would either detect the parent cluster of an openstack server changed (thus reconciling it again) or it would retry with Requeue in 

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/33d80a2cfabdbc338196ba788dc1d976a5ecb160/controllers/openstackserver_controller.go#L112-L117

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built):v0.12.5
- Cluster-API version: v1.9.11
- OpenStack version:
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): v1.340

bnallapeta:
When a cluster is paused (e.g., during a pivot operation), OpenStackServer resources stop reconciling. However, when the cluster is unpaused, they don't resume because the controller doesn't watch for cluster pause/unpause events.

This PR adds a watch on Cluster resources so OpenStackServers are re-queued when their parent cluster transitions from paused to unpaused state.

Fixes #2824 

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/controllers/openstackserver_controller.go
+++ b/controllers/openstackserver_controller.go
@@ -39,6 +39,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/annotations"
 	v1beta1conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -50,7 +51,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	orcv1alpha1 "github.com/k-orc/openstack-resource-controller/v2/api/v1alpha1"
-	"github.com/k-orc/openstack-resource-controller/v2/pkg/predicates"
+	orcpredicates "github.com/k-orc/openstack-resource-controller/v2/pkg/predicates"
 
 	infrav1alpha1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1alpha1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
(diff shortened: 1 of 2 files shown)


controllers/openstackserver_controller.go
```diff
@@ -752,3 +759,50 @@ func OpenStackServerReconcileComplete(log logr.Logger) predicate.Funcs {
...
+			panic(fmt.Sprintf("Expected a Cluster but got a %T", o))
+		}
+
+		log := log.WithValues("objectMapper", "clusterToOpenStackServer", "namespace", c.Namespace, "cluster", c.Name)
+
+		// Don't handle deleted clusters
+		if !c.DeletionTimestamp.IsZero() {
+			log.V(4).Info("Cluster has a deletion timestamp, skipping mapping.")
+			return nil
+		}
```
lentzi90: Don't we need to clean up if it is deleting? I am thinking that there could be a situation where the user paused the cluster, then deleted (forgetting to unpause first), then unpause.

bnallapeta: hmm, to address this edge case, here's what I did: 
I've moved the pause check after the deletion timestamp check (similar to OpenStackCluster controller). Now the reconciliation flow is:

1. Check if server is being deleted -> proceed with deletion (regardless of pause state)
2. If not deleted, check if cluster is paused -> skip reconciliation  
3. Otherwise, proceed normally

This ensures deletion always proceeds even if the cluster was paused when deletion started, then later unpaused.

bnallapeta: @lentzi90 ptal ^^

lentzi90:
I am a bit worried that this will break clusterctl move. We will probably need to handle all these cases:
1. Cluster paused and has deletion timestamp: do nothing <- This is now broken I think (we delete even when paused)
2. Cluster unpaused and has deletion timestamp: `reconcileDelete` <- This was missing in first version I think
3. Cluster paused and no deletion timestamp: do nothing :heavy_check_mark: 
4. Cluster unpaused and no deletion timestamp: `reconcileNormal` :heavy_check_mark: 

The full test is testing clusterctl move, so let's check

PR #2834, opened 2025-11-14
🐛 Handle SOFT_DELETED and DELETED states in server deletion

### LINKED ISSUE #2618: Can not delete k8s cluster when OpenStack’s soft delete feature is enabled
/kind bug

When OpenStack’s soft delete feature is enabled, the delete method called by CAPO (Cluster API Provider OpenStack) cannot fully remove the virtual machines and their associated resources.
As a result, the Kubernetes cluster deletion process gets stuck at a certain step because the instances remain in a “soft deleted” state instead of being completely removed.

bnallapeta:
When OpenStack's soft-delete is enabled, deleted servers enter SOFT_DELETED state instead of being immediately purged, causing CAPO's deletion poll to timeout and stall cluster cleanup.
This PR updates the deletion poll to treat SOFT_DELETED (and DELETED) as success, allowing reconciliation to proceed while respecting OpenStack's reclaim policy.

** Special Notes **

**Design Approach**
Updated `DeleteInstance` to treat servers in `SOFT_DELETED` or `DELETED` state as successfully deleted. This allows cluster deletion to complete when OpenStack has soft delete enabled, while respecting the cloud admin's recovery policy. CAPO proceeds with cleanup, and OpenStack handles permanent deletion per its configured reclaim interval.

Fixes #2618

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -792,6 +792,11 @@ var (
 	// InstanceStateDeleted is the string representing an instance in a deleted state.
 	InstanceStateDeleted = InstanceState("DELETED")
 
+	// InstanceStateSoftDeleted is the string representing an instance in a soft-deleted state.
+	// This state occurs when OpenStack is configured with a reclaim_instance_interval > 0,
+	// allowing recovery of deleted instances within the reclaim period.
+	InstanceStateSoftDeleted = InstanceState("SOFT_DELETED")
+
 	// InstanceStateUndefined is the string representing an undefined instance state.
 	InstanceStateUndefined = InstanceState("")
 )
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -491,10 +491,15 @@ func (s *Service) DeleteInstance(eventObject runtime.Object, instanceStatus *Ins
 		if err != nil {
 			return false, err
 		}
-		if i != nil {
-			return false, nil
+		// Server not found means it has been permanently deleted
+		if i == nil {
+			return true, nil
 		}
-		return true, nil
+		// Server in SOFT_DELETED or DELETED state means deletion succeeded. This respects OpenStack's soft delete policy.
+		if i.State() == infrav1.InstanceStateSoftDeleted || i.State() == infrav1.InstanceStateDeleted {
+			return true, nil
+		}
+		return false, nil
 	})
 	if err != nil {
 		record.Warnf(eventObject, "FailedDeleteServer", "Failed to delete server %s with id %s: %v", instance.Name, instance.ID, err)
--- a/pkg/cloud/services/compute/instance_test.go
+++ b/pkg/cloud/services/compute/instance_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/go-logr/logr/testr"
 	"github.com/google/go-cmp/cmp"
+	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/keypairs"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers"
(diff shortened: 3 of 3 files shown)


mandre (commented):
While this allows DeleteInstance() to complete when OpenStack is configured with soft-deletion, I'm a bit unclear what happens for all the resources the server depends on (volumes, trunk ports). Are we leaking them? 
Also, what happens when the server is brought back to life?

Were you able to check this change against an OpenStack cloud configured with soft-deletion?

bnallapeta:
@mandre 
> I'm a bit unclear what happens for all the resources the server depends on (volumes, trunk ports). Are we leaking them?

When CAPO creates servers with volumes, it sets DeleteOnTermination=true. So, when the server is deleted, OpenStack automatically deletes the volumes. This is not on CAPO. 
Ports and trunks are explicitly cleaned up by CAPO after DeleteInstance returns, regardless of soft delete. So, behavior isn't changed with this change.


> Also, what happens when the server is brought back to life?

Afaik, once CAPO is done with its cycle, it removes all the labels, finalizers etc and thus, from CAPO's perspective, the server becomes an orphaned resource. Even if it is brought back later, it won't be associated with CAPO. (@lentzi90 could you please chime in here?)

> Were you able to check this change against an OpenStack cloud configured with soft-deletion?

No.

mandre:
> @mandre
> 
> > I'm a bit unclear what happens for all the resources the server depends on (volumes, trunk ports). Are we leaking them?
> 
> When CAPO creates servers with volumes, it sets DeleteOnTermination=true. So, when the server is deleted, OpenStack automatically deletes the volumes. This is not on CAPO. Ports and trunks are explicitly cleaned up by CAPO after DeleteInstance returns, regardless of soft delete. So, behavior isn't changed with this change.

However, because the server is still there, I expect the port cleanup to fail, so perhaps we're creating new problems with this change, leading to resource leaks. It would be good to double check against a real environment. Or are we simply considering the server and associated resources as "not CAPO's problem anymore?". In which case, I think it would deserve being stated explicitly in the docs, because it would then be the user's responsibility to clean the resources after the soft-deletion period ended.

bnallapeta:
@mandre let me test this on a real env and get back.

bnallapeta:
@mandre so I tested this with microstack. Here's what I found:

#### Soft Delete Configuration

```bash
$ sudo grep -r "reclaim_instance_interval" /var/snap/microstack/common/etc/nova/
/var/snap/microstack/common/etc/nova/nova.conf.d/soft-delete.conf:reclaim_instance_interval = 300
/var/snap/microstack/common/etc/nova/nova.conf:reclaim_instance_interval = 300
```
Soft delete is enabled with 300 second reclaim interval

**1. Port deletion succeeds even when server is in `SOFT_DELETED` state.**

#### Test:

1. Create and Delete Test Server

```bash
$ microstack.openstack server create --flavor m1.tiny --image cirros --network test test-vm --wait
$ microstack.openstack server delete test-vm
```

2. Server enters `SOFT_DELETED` state as expected:
```bash
$ microstack.openstack server list --deleted --long
| ID                                   | Status       | Power State | Networks             |
| 7949ff24-2f8a-445b-acca-b7c81b91f1cf | SOFT_DELETED | Shutdown    | test=192.168.222.113 |
```

3. Port remains attached to `SOFT_DELETED` server:
```bash
$ microstack.openstack port list --device-id 7949ff24-2f8a-445b-acca-b7c81b91f1cf
| ID                                   | Fixed IP Addresses      | Status |
| 00c8dda1-2310-4919-9a2c-0b408259c860 | 192.168.222.113         | DOWN   |
```

4. **Port deletion succeeds** (exit code 0):
```bash
$ microstack.openstack port delete 00c8dda1-2310-4919-9a2c-0b408259c860
$ echo $?
0
```

**2. Volumes are automatically handled by OpenStack.**

As noted, CAPO sets `DeleteOnTermination=true` when creating servers with volumes. OpenStack automatically deletes volumes when the server is deleted, regardless of soft delete state. No CAPO cleanup action needed.

**3. Trunk ports use the same deletion API as regular ports.**

Microstack's OVN setup didn't have full trunk support for testing, but trunk ports use the same Neutron port deletion API. Since regular ports can be deleted from `SOFT_DELETED` servers, trunk ports should behave identically.


tl;dr
OpenStack's behavior allows resources to be cleaned up even when the server is in `SOFT_DELETED` state. CAPO's cleanup logic will work correctly with the PR changes.

bnallapeta:
@mandre could you please take a look at this when you find time?

mandre:
Sorry I missed your update. I'll take a look a little later today.

bnallapeta:
@lentzi90 @smoshiur1237 ^^

bnallapeta:
@smoshiur1237 ^^

mandre:
@bnallapeta This isn't merging because of the `/hold` from the PR description. Your PR has all the needed labels, and needs you to drop the hold with a `/hold cancel` comment whenever you feel the PR's ready to go.

lentzi90:
Yup this is good to go! Feel free to remove the hold :slightly_smiling_face: 
It is perhaps surprising that we add hold in the PR template, but it can be quite nice to avoid accidentally merging things.

bnallapeta:
aha, thanks guys. I thought it was because it was missing the lgtm.

Issue #2814, opened 2025-10-30
Migrate from CAPO Internal OpenStackServer Controller to External ORC Controller


lentzi90:
The Cluster API Provider OpenStack (CAPO) project has its own internal OpenStackServer CRD and controller for legacy reasons. The project has started transitioning to use the external [OpenStack Resource Controller (ORC)](https://github.com/openstack-k8s-operators/cluster-api-provider-openstack-orc) instead.

The initial step of this migration was completed in [PR #2346: 🌱 Remove ORC from CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2346). In that PR, the internal ORC integration and related code were removed so that CAPO would rely on an external, out-of-tree ORC deployment. However, the transition was not fully completed, and the largest remaining task is to get rid of the internal OpenStackServer implementation.

We need to be extra careful about the upgrade and transition from the internal OpenStackServer to the ORC implementation. We will need at least one minor release where both are supported and where CAPO actively migrates resources. Some kind of test for this is also needed.

Why do we need to do this? We risk diverging from ORC, we have to maintain our own controller. It would simply be preferable to adopt ORC and have one less thing to worry about.

mandre:
I can take this one. I'll first need to check if there's a feature gap between what CAPO exposes for the server creation and what ORC provides.
/assign

mandre:
OK, we're not there yet 😅. Here's a gap analysis between what CAPO offers and what ORC currently supports.

| CAPO's OpenStackMachineSpec | ORC | Note |
|-|-|-|
| Flavor | ✔️ (FlavorRef) ||
| Image | ✔️ (ImageRef) | They're effectly the same as CAPO leverages ORC for it |
| SSHKeyName | ❌ | KeyPair controller is WIP |
| Ports | ❌ | See details below |
| Port Network | ✔️ |  |
| Port Description | ✔️ |  |
| Port FixedIPs | ✔️ (Port's Addresses) |  |
| Port SecurityGroups | ✔️ (Port's SecurityGroupRefs) | |
| Port Tags | ✔️ (Port's Tags) | |
| Port Trunk | ❌ | Need to extend the Port controller |
| Port AdminStateUp | ❌ | Need to extend the Port controller |
| Port MACAddress | ❌ | Need to extend the Port controller |
| Port AllowedAddressPairs | ✔️ (Port's AllowedAddressPairs) | |
[...]

PR #2835, opened 2025-11-14
✨ Shutdown VMs before Deletion

### LINKED ISSUE #1973: Option to Shutdown VM before deleting it
/kind feature

I think the teardown process of VMs would benefit from an option to shutdown the OpenStack Machine before deleting it. This way it would be possible to benefit from teardown process in the VMs itself. Myself encountered this problem with NVIDIA licensees, which are `borrowed` on VM startup, but never properly released, because the Service on the VM was never stopped properly and therefore had no chance of releasing the license.

I think on way of implementing this, would be to add something like an `waitForShutdown` to the OpenStackMachine CRD. It could be defaulted to `0s` to keep the old behavior and it would be a single configuration to have the possibility to let the VM properly shutdown and have a timeout for the shutdown, so that it does not hang forever in an error case.

If you are interested in such a feature, I would be happy to help implementing it.

shaardie:
Tries to shutdown the OpenStack VM before deleting it. This way even Pods form Daemonsets are shut down more gracefully and services like license daemons on the VMs can be properly shutdown.

Fixes #1973

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1beta1/conditions_consts.go
+++ b/api/v1beta1/conditions_consts.go
@@ -38,6 +38,8 @@ const (
 	InstanceDeletedReason = "InstanceDeleted"
 	// InstanceNotReadyReason used when the instance is in a pending state.
 	InstanceNotReadyReason = "InstanceNotReady"
+	// InstanceStopFailedReason used when stopping the instance failed.
+	InstanceStopFailedReason = "InstanceStopFailed"
 	// InstanceDeleteFailedReason used when deleting the instance failed.
 	InstanceDeleteFailedReason = "InstanceDeleteFailed"
 	// OpenstackErrorReason used when there is an error communicating with OpenStack.
--- a/controllers/openstackserver_controller.go
+++ b/controllers/openstackserver_controller.go
@@ -257,6 +257,14 @@ func (r *OpenStackServerReconciler) reconcileDelete(scope *scope.WithLogger, ope
 			return fmt.Errorf("delete volumes: %w", err)
 		}
 	} else {
+		// Try to stop the VM, if active before deletion
+		if instanceStatus.State() == infrav1.InstanceStateActive {
+			if err := computeService.StopInstance(openStackServer, instanceStatus); err != nil {
+				v1beta1conditions.MarkFalse(openStackServer, infrav1.InstanceReadyCondition, infrav1.InstanceStopFailedReason, clusterv1beta1.ConditionSeverityError, "Stopping instance failed: %v", err)
+				return fmt.Errorf("stop instance: %w", err)
+			}
+		}
+
 		if err := computeService.DeleteInstance(openStackServer, instanceStatus); err != nil {
 			v1beta1conditions.MarkFalse(openStackServer, infrav1.InstanceReadyCondition, infrav1.InstanceDeleteFailedReason, clusterv1beta1.ConditionSeverityError, "Deleting instance failed: %v", err)
 			return fmt.Errorf("delete instance: %w", err)
--- a/pkg/clients/compute.go
+++ b/pkg/clients/compute.go
@@ -61,6 +61,7 @@ type ComputeClient interface {
 	ListFlavors() ([]flavors.Flavor, error)
 	CreateServer(createOpts servers.CreateOptsBuilder, schedulerHints servers.SchedulerHintOptsBuilder) (*servers.Server, error)
 	DeleteServer(serverID string) error
+	StopServer(serverID string) error
 	GetServer(serverID string) (*servers.Server, error)
 	ListServers(listOpts servers.ListOptsBuilder) ([]servers.Server, error)
 
@@ -138,6 +139,12 @@ func (c computeClient) CreateServer(createOpts servers.CreateOptsBuilder, schedu
 func (c computeClient) DeleteServer(serverID string) error {
 	mc := metrics.NewMetricPrometheusContext("server", "delete")
 	err := servers.Delete(context.TODO(), c.client, serverID).ExtractErr()
+	return mc.ObserveRequestIgnoreNotFound(err)
+}
+
+func (c computeClient) StopServer(serverID string) error {
+	mc := metrics.NewMetricPrometheusContext("server", "stop")
+	err := servers.Stop(context.TODO(), c.client, serverID).ExtractErr()
 	return mc.ObserveRequestIgnoreNotFound(err)
 }
 
@@ -227,6 +234,10 @@ func (e computeErrorClient) CreateServer(_ servers.CreateOptsBuilder, _ servers.
 }
 
 func (e computeErrorClient) DeleteServer(_ string) error {
+	return e.error
+}
+
+func (e computeErrorClient) StopServer(_ string) error {
 	return e.error
 }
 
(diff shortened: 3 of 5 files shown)


lentzi90 (commented):
/ok-to-test
Thinking about unnecessary API calls, should we skip trying to shut it down all together if the timeout is 0?

shaardie:
> /ok-to-test Thinking about unnecessary API calls, should we skip trying to shut it down all together if the timeout is 0?

Which timeout do you mean exactly? `timeoutInstanceDelete` seems to be hardcoded to `5min`.

lentzi90:
Oh right, I got the 0 from the issue description. But the question is still relevant. I think users should be able to opt out of this, especially since this adds more API calls.

shaardie:
> Oh right, I got the 0 from the issue description. But the question is still relevant. I think users should be able to opt out of this, especially since this adds more API calls.

So you suggest a new configuration option via CRD?

lentzi90:
Hmm let me gather some second opinions. I want to have more than a gut feeling before we start modifying the CRDs :smile:

mnaser:
I'm not sure how I feel about failing if the system doesn't shut down, I feel like it would be better if the it tries to shut it down for 5 minutes, and if it doesn't shut down, it moves on to termination.

Anyways, OpenStack will flip from a graceful to hard shutdown after 60s by default:

https://docs.openstack.org/nova/rocky/configuration/config.html#DEFAULT.graceful_shutdown_timeout

So the 5 minute timeout seems overkill as well, unless something is seriously wrong (or the cloud has that config changed).

shaardie:
I can also change the PR to continue with deleting the VM instead of failing after the period of time.

For me personally 60s would also be okay for a timeout, but I can think of situations where this can be a little bit short. For example, if there are some custom mounts of nfs, cifs, gpfs, what so ever. This can easily take more than 60s to shutdown.

Maybe you should first decide, if you want to have this value configurable via CRD?

lentzi90:
I have checked with my downstream and they do not have any concerns with the feature (always enabled).

However, it sounds like there are quite many ways to do things and people will want different things. Some do not care about the shutdown and definitely want to force it or just straight delete. Some want to make sure everything is properly shut down, rather error than force. And some will want a different timeout.

So how should we do this? I can see it working with either a flag or CRD field(s).

Then we have one more thing to consider. We want to make use of ORC for managing the servers. See https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2814 for more details. Hopefully we can get this done sooner rather than later, which means that this feature would make more sense to implement in ORC directly. Otherwise we will end up having to migrate it later.

shaardie:
I am not quite sure what you want me to do honestly. I would be happy to change stuff on this PR, if you tell me what you want to have.

If you want to migrate to your new setup first, I would probably use my patched version for now and see, if I re-write the whole thing again, when you migration to ORC is done.

Atomsoldat:
> So how should we do this?

## Human interaction analogy

Let's go through the scenarios where Alice (User) wants Bob (CAPO) to delete a VM in a "regular" talk to your colleague kind of interaction:

- If Alice tells Bob "Please shut down this VM for me, i need one less now" and gives no further details, Bob will have to go with what the best practice is, and will perform an ordinary shutdown, giving the OS some time to properly terminate processes.
  - If the normal shutdown does not proceed as planned, a human operator would typically ask Alice for confirmation whether she would agree to a forced shutdown
  - An automated process can not do this, so it has to do what minimises the deviation from the desired state (VM off, no improperly shut down processes causing trouble) while maximising the velocity of reaching the desired state. We know, that we can not maximise both, but (at least from my perspective) there are many things that can go wrong when we immediately shut everything off without proper cleanup, while the repercussions of a delay of 60 seconds before we  get the big stick seem rather less stark in comparison. So any delay should be better than no delay in the majority of cases
- If Alice knows, that the VM will take a long time to shut down normally, she should tell Bob this information, so that he does not get surprised.
- If Alice wants Bob to immediately pull the plug on the machine instead of performing the usual shutdown, she should tell him this beforehand, because she can not expect this to be his regular modus operandi

In all of the above cases, Alice [should](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) provide Bob with the information he needs to proceed in an ideal way. To me, this hints in the direction of Alice (User) providing this information to Bob (CAPO) beforehand, in a way Bob understands (CRD field). If Bob tries to have one solution that applies to all possible use cases (Configuration Flag) he might get some cases wrong, in which Alice has different requirements. 

You might also have the case that you have one CAPO instance managing VMs that you want to be deleted immediately  as well as VMs that you want to give time for an orderly shutdown. That would also make the CRD field approach more desirable.

## Which value should the feature use by default

In my opinion, the Venn diagram representing the group of people for whom one minute of additional VM runtime would be more than even a minor inconvenience (which could then be fixed easily) and the group of people who would be caught unaware of such a change should have a very small intersection.

Whereas with the way things currently work, the Venn diagram representing the group of people for whom an immediate VM termination would be more than even a minor inconvenience (which may or may not be easily remedied) and the group of people who might be bitten by this in the future probably has a larger intersection (in my opinion).


So i think under those conditions, there is no need to treat the previous default (which is unusual and can definetly cause headaches) with a lot of reverence. I think `60` seconds before forcing termination (which may then be adjusted for individual VMs with special considerations) is a reasonable default. If people never want their VMs to be force terminated, set it to `-1`, and if they want them terminated immediately, set it to`0`.

But this is just my opinion, just trying to give some input to give you a perspective on the choices you mentioned.

lentzi90:
> I am not quite sure what you want me to do honestly. I would be happy to change stuff on this PR, if you tell me what you want to have.
> 
> If you want to migrate to your new setup first, I would probably use my patched version for now and see, if I re-write the whole thing again, when you migration to ORC is done.

I am basically saying that I think we need an option to either turn this feature off or to allow more granular configuration of it. I do not have a strong opinion on how to do that so I am leaving it up to you to propose what to do. The issue description already suggests a `waitForShutdown` field. That sounds reasonable and other people seem to agree also.

If you don't need this urgently, I also suggest looking into ORC first so that we can get an implementation that will work with it. Otherwise we risk breaking this feature later.

smoshiur1237:
As we are expecting to have graceful shut down.  Should we follow the following steps so that ungraceful shutdown doesn't happen:  
1. If VM is running,  
2. Issue a "stop" or "poweroff" command via the OpenStack API.
3. Wait for the VM to reach "SHUTOFF" state.
4. Delete the instance as usual. 
I can see the stop server option, but is it syncing with the deletion of the instance?

shaardie:
> I can see the stop server option, but is it syncing with the deletion of the instance?

I am not quite sure, what you mean. Currently the VM is only deleted without any shutdown. With this PR the controller first triggers a `server stop`, waits for it to finish and then deletes the VM.

shaardie:
> I am basically saying that I think we need an option to either turn this feature off or to allow more granular configuration of it. I do not have a strong opinion on how to do that so I am leaving it up to you to propose what to do. The issue description already suggests a `waitForShutdown` field. That sounds reasonable and other people seem to agree also.
> 
> If you don't need this urgently, I also suggest looking into ORC first so that we can get an implementation that will work with it. Otherwise we risk breaking this feature later.

Okay, so for myself, I have now a running version with shutdown. So I do not need this ASAP. Since you currently update to ORC, I will wait for this transition to happen and update this PR then with a Version which includes a configuration.

PR #2842, opened 2025-11-19
[release-0.9] Fix instance deletion order

mandre:
Delete the OpenStack instance before deleting its network ports. This prevents "Policy doesn't allow compute:detach primary port" errors that occur when attempting to delete a primary NIC while the instance still exists.

Deleting the instance automatically detaches all interfaces, so the explicit detach call is no longer needed.

Related to https://issues.redhat.com/browse/OCPBUGS-65712.

### DIFF
--- a/pkg/cloud/services/compute/instance.go
+++ b/pkg/cloud/services/compute/instance.go
@@ -657,12 +657,13 @@ func (s *Service) DeleteInstance(openStackCluster *infrav1.OpenStackCluster, eve
 		return err
 	}
 
-	// get and delete trunks
-	for _, port := range instanceInterfaces {
-		if err = s.deleteAttachInterface(eventObject, instanceStatus.InstanceIdentifier(), port.PortID); err != nil {
-			return err
-		}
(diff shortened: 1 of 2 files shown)


mandre:
I'm still testing it resolves the linked issue. I'll report back here.

mandre:
Unfortunately none of the clouds I have access to (and capable of handling an openshift deployment) have the policy that prevents deleting the primary NIC. That's probably part of the reason why we're only finding this now.

stephenfin:
> Unfortunately none of the clouds I have access to (and capable of handling an openshift deployment) have the policy that prevents deleting the primary NIC. That's probably part of the reason why we're only finding this now.

I noted this on the downstream bug, but I suspect this is a custom policy on OpenTelekom Cloud as `compute:detach primary port` is not a policy supported by upstream Nova or Red Hat's variant.

This being the case, it fixes someone's issue, is low risk, and doesn't change anything operationally since, as noted above, [nova will do this for us during instance deletion](https://github.com/openstack/nova/blob/53aadaf967b708bfb03616535d45f6378a21cae0/nova/compute/api.py#L2690).

PR #2845, opened 2025-11-20
🌱Add yamlling workflow to check yaml files in pull request

smoshiur1237:
This would be an interesting workflow to check on new pull request YAML files. This will be a basic check and I will add more improvement on linter check in separate PRs such as k8s way of indentation, check on snipped yaml code in md files.


.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+
+    - name: yaml-lint
+      uses: ibiqlik/action-yamllint@2576378a8e339169678f9939646ee3ee325e845c # v3.1.1
+      with:
+        config_file: .yamllint.yaml
```
lentzi90: Should we add also file_or_dir? I am not sure it will check any files otherwise.

```suggestion
        config_file: .yamllint.yaml
        file_or_dir: .
```

.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,19 @@
...
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+
+    - name: yaml-lint
+      uses: ibiqlik/action-yamllint@2576378a8e339169678f9939646ee3ee325e845c # v3.1.1
```
lentzi90: This looks not active. No changes since 2022.
Are you sure it is good to use? I would at least want to know what version of yamllint that it is using.

.github/workflows/yamllint.yaml
smoshiur1237: It should check all the the .yaml, .yml and .yamllint  files in the repo. If we have a specific folder or we want to check fixed folder we can add `file_or_dir`, otherwise current format will check all the files according to the rules added in .yamllint.yaml

.github/workflows/yamllint.yaml
smoshiur1237: We are using it in  project-infra

smoshiur1237: Here I am also testing some rules with same file and it is catching issues : https://github.com/metal3-io/project-infra/pull/1156    https://github.com/metal3-io/project-infra/actions/runs/19536808298/job/55932817800?pr=1156

lentzi90: Good then it is probably ok :+1: 
We should check if it is stuck on an old yamllint version though. Not sure how to do that.

smoshiur1237:
> /approve It is better than nothing so I don't see a reason to block. My only remaining concern is what version of yamllint this is using, considering that the action has not been updated in 3 years.

I have found the following line for yamllint version in repo:   ` pip install 'yamllint>=1.25.0'` does it taking the latest or any version from 1.25.0

lentzi90:
Yeah the question is if the action is using the latest version of yamllint that was available in 2022 when it was built. I am not sure. Anyway I think this is better than nothing as I said.
/CC @bnallapeta

bnallapeta:
I share the same concern as @lentzi90. Looking at the actual [yamllint](https://github.com/adrienverge/yamllint) its using, the project is at `v1.37.1`

I think its better to use yamllint directly as I don't see huge benefits to use this wrapper one. Yes, we would need to put in some effort for configuration instead of passing them as config, but at least, we would be using the latest versions.

lentzi90:
That is a good point. We can just `apt-get install yamllint`. Then run it with `yamllint -c .yamllint.yaml`.

smoshiur1237:
It is using the latest yaml lint. This is just the beginning with yamllint. It will be more useful when we add more rules to it and at least I can confirm that it works perfectly in our other projects.

lentzi90:
How do you know what version it is using?
It is not strange that it is working fine even with the older version. The question is, would it be even better if we get a newer version? And will the old version break at some point? Will the github action that hasn't been updated in three years break?

smoshiur1237:
I mixed with the other project. I have added an installation of yaml lint with the following step : 
```
- name: Install yamllint
  run: sudo apt-get update && sudo apt-get install -y yamllint
```
I haven't add it here yet. I thought it will install the latest yamllint but now I checked the details on the run and it is using the following version:
```
noble/universe amd64 yamllint all 1.33.0-1
Setting up yamllint (1.33.0-1) 
```

smoshiur1237:
I will update the PR according to our other project.

lentzi90:
Thanks!

smoshiur1237:
I have tried to add the yamllint for snipped code and it was asking to change lots of files which may cause bigger issues for auto-generated docs. It was taking long time to change and debug the issue. That is why I have removed that part for checking snipped yaml code but added k8s indentation style. I will fix the indentation error in separate commit.

smoshiur1237:
@bnallapeta now I have added rules for indentation and fixed all the yaml issue in CAPO. Please review

.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+    - name: Install yamllint
+      run: sudo apt-get update && sudo apt-get install -y yamllint
+
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+
+    - name: yaml-lint
+      uses: ibiqlik/action-yamllint@2576378a8e339169678f9939646ee3ee325e845c # v3.1.1
+      with:
+        config_file: .yamllint.yaml
```
lentzi90: I think we can just do this?
```suggestion
    - name: yaml-lint
      run: yamllint -c .yamllint.yaml
```

smoshiur1237: I can add it

PR #2875, opened 2025-12-02
🐛 Update CAPI contract mapping on CRDs

lentzi90:

I think we should have updated this already with v0.13.0 but for some reason it slipped through.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -23,7 +23,7 @@ providers:
     contract: v1beta2
     replacements:
     - old: --metrics-addr=127.0.0.1:8080
-      new: --metrics-addr=:8080
+      new: "--metrics-addr=:8080\n        - --v=4"
     files:
     - sourcePath: "../data/shared/capi/metadata.yaml"
   # For clusterctl upgrade test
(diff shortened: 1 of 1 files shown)


bnallapeta:
looks good in general. Will add the label once tests are fixed.

bnallapeta:
ah, I think more files are updated than the last time I checked. I will re-review it once tests pass.

lentzi90:
Yeah I am testing a few things. There is something making the e2e tests fail (no Machines are created) when I update the labels. Need to debug a bit.

lentzi90:
Slack discussion for reference: https://kubernetes.slack.com/archives/C8TSNPY4T/p1764932889047829
My understanding is that it currently works on main because CAPI reads the label and sees v1beta1. It then uses v1beta1 contract even though we put v1beta2 in the metadata.
We won't be able to update the label until we properly support v1beta2 by actually doing breaking changes to the API and introducing a new API version.

Now the question is, what to do with v0.13. We already did two releases advertising v1beta2 and things seems to be working, But in fact we are not fully implementing that contract. Should we fix the 0.13 release by reverting the contract change? Should we keep the inconsistency and quickly start working on a new API version to introduce in v0.14?

lentzi90:
Closing this since the conclusion is basically that we cannot do this until we have properly implemented v1beta2.

PR #2876, opened 2025-12-03
🌱Add new config to pr-link check

smoshiur1237:
Adding new config for link check on PRs.

1. --insecure = ignore ssl problems 
2. --exclude-all-private = private addresses are unreachable anyways
3. --no-progress = don't print progress


smoshiur1237:
> The other I understand, but why `--insecure`. Do we have links to sites with broken TLS? I am thinking that we should probably check if we link to such sites.

Actually I have seen issues in other projects where the links are correct but it was giving error in link check with broken TLS.. That's why I am thinking to add it here for future check

smoshiur1237:
If you don't want this insecure tag.  I am also ok, not to add it here. Just let me know @lentzi90

PR #2878, opened 2025-12-03
📖Update release documentation

smoshiur1237:
Some additional info and example to ease future release work

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -19,6 +19,7 @@
 A release version string is: `vX.Y.Z`.
 
 A pre-release version string additionally has a suffix:
+
 - `alpha` for an alpha release
 - `beta` for a beta release
 - `rc` for a release candidate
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -75,8 +76,8 @@ There is an [issue template](.github/ISSUE_TEMPLATE/new_release.md) to help trac
      - The commit and PR title should be 🚀 Release v1.x.y:
        -`git commit -S -s -m ":rocket: Release vX.Y.Z"`
        -`git push -u origin release-notes-X.Y.Z`
-     - Important! The commit should only contain the release notes file, nothing
-     else, otherwise automation will not work.
+     - Important! The `origin` must be directed to the fork (for example: nordix) otherwise it will have error. Also, the commit should only contain
```
lentzi90: Let's keep the most important first. Pushing through the fork is normal flow and should not be surprising. Also, maintainers can push directly to upstream (but it is better to go through the fork, since any branch in upstream will get branch protections).
```suggestion
     - Important! The commit should only contain the release notes file, nothing
       else, otherwise automation will not work. Push as normal, through your fork (`origin`).
```

PR #2881, opened 2025-12-04
🌱Bump go to 1.24.11

smoshiur1237:
Uplift go to 1.24.11

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 # Build the manager binary
 ARG GO_VERSION
-FROM golang:${GO_VERSION:-1.24.9} AS builder
+FROM golang:${GO_VERSION:-1.24.11} AS builder
 WORKDIR /workspace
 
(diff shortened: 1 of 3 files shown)


lentzi90:
/approve cancel
Please don't change PRs so much. It is better to open a new when completely changing what a PR does.

Where do you see that govulncheck says we are not affected? In the description I don't see it. In previous cases we saw this:
> Your code is affected by 0 vulnerabilities.
This scan also found 1 vulnerability in packages you import and 2
vulnerabilities in modules you require, but your code doesn't appear to call
these vulnerabilities.

We may still have to ignore these if they require a newer go version, but we should not claim we are unaffected unless that is true.

tuminoid (commented):
You ARE affected by the stdlib issues, if you use certificates, like govulncheck says.

smoshiur1237:
Now it is working
/cc @bnallapeta @lentzi90

PR #2894, opened 2025-12-09
🌱 Bump golanci-lint to v2.7.2

lentzi90:

Golanci-lint is updated to the newest version.

Two new linting errors appeared:
1. The errors package conflicts with standard library package.
2. The shared package name is considered "meaningless".

I have renamed these packages to capoerrors and e2eshared.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1beta1/zz_generated.deepcopy.go
+++ b/api/v1beta1/zz_generated.deepcopy.go
@@ -23,7 +23,7 @@ package v1beta1
 import (
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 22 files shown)


lentzi90:
One thing to discuss here is if we should adjust the paths also. For `errors` I don't think we should. For `e2eshared` perhaps there could be better ways to structure the code. I am open to suggestions.

lentzi90:
Thanks! Taking it in then. We need the bump to get dependabot PRs merged and we can change details later if needed.
/approve
/hold cancel

PR #2915, opened 2025-12-17
🌱 Fix DWARF debugging info

smoshiur1237:
Fixing the DWARF debugging info to have proper version output

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,13 +36,13 @@ COPY ./ ./
 # Build
 ARG package=.
 ARG ARCH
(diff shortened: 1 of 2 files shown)


lentzi90:
Adding some cross references for the future.
The issue we discovered was that the `--version` flag no longer worked. See https://github.com/kubernetes/k8s.io/pull/8884#issuecomment-3664779561.
We simply didn't pass the needed ldflags in to the docker build.

mandre (commented):
/lgtm

PR #2947, opened 2026-01-13
⚠️ Add api/v1beta2 types with CAPI v1beta2 contract

bnallapeta:
**What this PR does / why we need it**:
This PR implements the first step of migrating to the CAPI v1beta2 contract by creating the `api/v1beta2` package with updated types that use standard `[]metav1.Condition` instead of the deprecated `clusterv1beta1.Conditions` format.

### API Types Created
- Created `api/v1beta2/` package with all types copied and modified from v1beta1
- Updated condition format: `clusterv1beta1.Conditions` → `[]metav1.Condition`
- Removed deprecated status fields:
  - `status.Ready` (derive from Ready condition)
  - `status.FailureReason` (use conditions)
  - `status.FailureMessage` (use conditions)
- Updated CAPI type references to v1beta2:
  - `clusterv1.APIEndpoint` (from `sigs.k8s.io/cluster-api/api/core/v1beta2`)
  - `[]clusterv1.FailureDomain` (changed from map to slice in CAPI v1beta2)
- Updated method signatures:
  - `GetConditions()` returns `[]metav1.Condition`
  - `SetConditions()` accepts `[]metav1.Condition`
- Removed `SetFailure()` methods (relied on deprecated fields)
- Removed `+kubebuilder:storageversion` markers from v1beta2
- Removed Ready printcolumn from kubebuilder markers

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2939 part of #2895 

**Special notes for your reviewer**:
**v1beta1 remains the Hub and storage version.** The v1beta2 types exist in the codebase but are not served by the API server. This staged approach allows the types to be reviewed and tested independently before wiring them up.

/hold

Issue #2895, opened 2025-12-09
✨ Implement CAPI v1beta2 contract support with api/v1beta2


bnallapeta:
/kind feature

**Describe the solution you'd like**
Implement full CAPI v1beta2 contract support by creating a new `api/v1beta2` API version with required breaking changes.

**Background**
CAPO v0.13.2 (?) reverted to v1beta1 contract. v0.15 will properly implement v1beta2.

### Tasks

**API Changes:**
- [ ] Create `api/v1beta2/` with breaking changes:
  - Change conditions: `clusterv1.Conditions` → `[]metav1.Condition`
  - Remove deprecated fields: `status.ready`, `status.failureReason`, `status.failureMessage`
  - Update FailureDomains if needed
- [ ] Update all v1beta1 resources: OpenStackCluster, OpenStackMachine, OpenStackClusterTemplate, OpenStackMachineTemplate
- [ ] Run `make generate` to generate deepcopy, CRDs, etc.

**Conversion & Webhooks:**
- [ ] Implement conversion webhooks (v1beta1 ↔ v1beta2)
- [ ] Update validation/defaulting webhooks for v1beta2
- [ ] Update CRD labels: `cluster.x-k8s.io/v1beta2: v1beta2`

**Controllers & Templates:**
- [ ] Update controllers to use conditions instead of deprecated fields
- [ ] Update all cluster templates in `templates/` to v1beta2
- [ ] Update `metadata.yaml` to claim v1beta2 contract

**Documentation & Testing:**
- [ ] Create migration guide: `docs/book/src/topics/crd-changes/v1beta1-to-v1beta2.md`
- [ ] Unit tests for conversion webhooks
- [ ] E2E tests with CAPI v1.11+ (v1beta2 contract)
- [ ] Upgrade tests (v1beta1 → v1beta2)

nikParasyr:
My ideas are mostly around tidying-up/consistency.
I'm not sure whether we should include them in this new api version since we might want to get v1beta2 compatibility asap.
I can allocate some time to work on my proposals but it will require the base of v1beta2 to be in-place.

-----

1. OSM: Update flavor to follow id/filter standard:

Current the OSM allows the following:
```
Kind: OpenStackMachine
...
spec:
  ...
  flavor: <Name of flavor>
  flavorID: <ID of flavor>
  ...
```

This should be updated to the one below so that its consistent with all other fields:
```
Kind: OpenStackMachine
...
spec:
[...]

PR #2958, opened 2026-01-19
⚠️ Implement v1beta1 ↔ v1beta2 conversion webhooks

### LINKED ISSUE #2940: Implement conversion webhooks (v1beta1 ↔ v1beta2)
Add conversion logic between API versions:
- Add `ConvertTo`/`ConvertFrom` methods in v1beta1 types (spoke)
- Add unit tests for all conversion paths

bnallapeta:
- Move Hub markers from v1beta1 to v1beta2, making v1beta2 the conversion hub
- Implement ConvertTo/ConvertFrom for all resource types (OpenStackCluster, OpenStackMachine, and their Templates/Lists)
- Add condition format conversion between CAPI v1beta1 conditions and standard metav1.Condition
- Handle FailureDomains map↔slice conversion and deprecated field derivation

## Test plan
- [x] `make test` passes (348 tests)
- [x] `make lint` passes
- [x] Unit tests for round-trip conversion of all types
- [x] Condition conversion tests verify format mapping

Fixes #2940 part of #2895

To be merged after https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2947

/hold


lentzi90:
Let's see if this fixes the autoscaler test!

bnallapeta:
> Looks like you are not using conversion-gen for this? I am thinking that it would be better to use that since it gives us some automated assurance that we handle all fields. It is also what we used before when we had v1alphaX, and what CAPI uses.

This should be addressed now. I add the required markers and the Makefile target for conversion-gen. PTAL.

bnallapeta:
@lentzi90 things look good on the e2e end

api/v1beta1/conversion.go
```diff
@@ -18,33 +18,264 @@ package v1beta1
...
+		return err
+	}
+	_, err := utilconversion.UnmarshalData(src, dst)
+	return err
+}
+
+// ConvertTo converts this OpenStackClusterTemplate to the Hub version (v1beta2).
+func (src *OpenStackClusterTemplate) ConvertTo(dstRaw ctrlconversion.Hub) error {
+	dst := dstRaw.(*infrav1beta2.OpenStackClusterTemplate)
+	return Convert_v1beta1_OpenStackClusterTemplate_To_v1beta2_OpenStackClusterTemplate(src, dst, nil)
```
lentzi90: Shouldn't we do `utilconversion.MarshallData` here also? I am guessing it is not needed currently, but it would be more consistent and then we don't have to worry about it if we make some changes later.
Same for the rest of the `ConvertTo` and `ConvertFrom` below. All above have it so we might as well add it here.

api/v1beta1/conversion_test.go
```diff
@@ -0,0 +1,390 @@
...
+				CloudName: "openstack",
+			},
+			ManagedSubnets: []SubnetSpec{
+				{
+					CIDR: "192.168.0.0/24",
+				},
+			},
+			ManagedSecurityGroups: &ManagedSecurityGroups{},
+		},
+		Status: OpenStackClusterStatus{
```
lentzi90, line 52: It would be good to include FailureDomains here since that part of the conversion function is done manually.

bnallapeta: Good catch. I added the checks.

api/v1beta1/conversion.go
bnallapeta: Hmm, I did it for the Template types; good future-proofing. But for the List types, I don't think its necessary as they get their own conversion.

bnallapeta:
Ok, I believe the reason for the failing autoscaler tests is cuz the v1beta2 is not registered in the scheme. I planned to do this in a later PR but I realize that conversion webooks are useless without the v1beta2 in the scheme. How will it work if its not even registered? So, I added them now to main. Note that this still does not change the storage version. So we will still be serving v1beta1.

Hopefully, this fixes it 🤞

PR #2985, opened 2026-01-28
🌱 Add OpenStackAuthenticationSucceeded condition

### LINKED ISSUE #2264: No Conditions/Events generated to explain missing clouds.yaml
/kind bug

1. Create a cluster referencing a non-existent secret.
2. The OpenStackCluster is stuck. No events or conditions show up.
3. CAPO logs gives some clues that [auth-url] is missing. (Should not be necessary for users to check logs)

The OpenStackCluster (and OpenStackMachine) should have obvious status messages when the clouds.yaml secret is missing.

**Environment:**

- Cluster-API version: v1.8.5
- OpenStack version:
- Minikube/KIND version:

lentzi90:

- Sets OpenStackAuthenticationSucceededCondition to True/False based on credential validity during reconciliation.
- Adds tests for condition handling on credential errors and missing secrets.

Fixes #2264

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

/hold

### DIFF
--- a/api/v1beta1/conditions_consts.go
+++ b/api/v1beta1/conditions_consts.go
@@ -73,6 +73,14 @@ const (
 )
 
 const (
+	// OpenStackAuthenticationSucceeded reports on the current status of the OpenStack credentials.
+	OpenStackAuthenticationSucceeded clusterv1beta1.ConditionType = "OpenStackAuthenticationSucceeded"
+
+	// OpenStackAuthenticationFailedReason is used when the controller fails to authenticate with OpenStack.
+	OpenStackAuthenticationFailedReason = "OpenStackAuthenticationFailed"
+)
+
+const (
 	// NetworkReadyCondition reports on the current status of the cluster network infrastructure.
 	// Ready indicates that the network, subnets, and related resources have been successfully provisioned.
 	NetworkReadyCondition clusterv1beta1.ConditionType = "NetworkReady"
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -124,8 +124,10 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	clientScope, err := r.ScopeFactory.NewClientScopeFromObject(ctx, r.Client, r.CaCertificates, log, openStackCluster)
 	if err != nil {
+		v1beta1conditions.MarkFalse(openStackCluster, infrav1.OpenStackAuthenticationSucceeded, infrav1.OpenStackAuthenticationFailedReason, clusterv1beta1.ConditionSeverityError, "Failed to create OpenStack client scope: %v", err)
 		return reconcile.Result{}, err
 	}
+	v1beta1conditions.MarkTrue(openStackCluster, infrav1.OpenStackAuthenticationSucceeded)
 	scope := scope.NewWithLogger(clientScope, log)
 
 	// Handle deleted clusters
--- a/controllers/openstackcluster_controller_test.go
+++ b/controllers/openstackcluster_controller_test.go
@@ -224,7 +224,7 @@ var _ = Describe("OpenStackCluster controller", func() {
 		Expect(created.Spec.IdentityRef.Region).To(Equal("RegionOne"))
 	})
 
-	It("should fail when namespace is denied access to ClusterIdentity", func() {
+	It("should fail when namespace is denied access to ClusterIdentity and set OpenStackAuthenticationSucceededCondition to False", func() {
 		testCluster.SetName("identity-access-denied")
 		testCluster.Spec.IdentityRef = infrav1.OpenStackIdentityReference{
 			Type:      "ClusterIdentity",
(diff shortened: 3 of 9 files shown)


lentzi90:
Maybe a better name would be `OpenStackCredentialsExists`?

nikParasyr:
> Maybe a better name would be `OpenStackCredentialsExists`?

I do indeed think that `OpenStackCredentialsReadyCondition` means more that the credentials are there and valid.
So i'd prefer a better name, not sure which one myself:
- `OpenStackCredentialsExists`: a bit better but dont like the exist`s` part that much
- `OpenStackCredentialsResolved`:  sort of the same, still not like the resolve`d`
- `OpenStackCredentialsAvailable`: i think this is a bit clearer

lentzi90:
> OpenStackCredentialsAvailable: i think this is a bit clearer

I like this also! Changing :)

lentzi90 (commented):
/hold cancel

mandre:
Should we try to define a condition that is perhaps a little bit less specific?
From a user point of view, I'd like to know that something is wrong with the credentials without having to look at the logs, as it was captured in [the referenced issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2264). I'd like this condition to points me to the right direction for debugging the issue. That "something is wrong" might not just be "there's no credentials at all" but also "the creds are wrong" or "the clouds.yaml doesn't have the necessary data".

Do you think a more generic condition like `OpenStackAuthError` better captures it?

mandre:
Obviously, if we're using a condition that express "I can't connect to the cloud", we would need to change its status in more places.

lentzi90:
Hmm, yes it would be good to capture more issues, but I am not sure how much more we can do at this early stage. If the secret is missing or malformed, we get a reconciliation error here. Those errors are included in the message with the condition, so the user do get that information.

When there are issues later on, it is usually quite hard to know if they are related to the credentials without parsing the error messages. I am hoping that we do set relevant conditions and messages in most places though.
Checking for the very first time we try the API, I think it is
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9e848d93a00ab80124c623aa28b85cc161394396/controllers/openstackcluster_controller.go#L330-L333

Sigh... we set no condition.
I'll add something here also and test this locally.

lentzi90:
Documenting findings from manual tests. I was wrong about how far we check the credentials when creating the scope. We do notice all kinds of issues here immediately, so that is really great. I'll still create an issue for adding a condition to that `reconcileNetworkComponents` error...

@mandre what do you think about these below? Are they good enough (at least as a first step)?

Here is the condition for a cluster with non-existing credentials secret:
```yaml
status:
  conditions:
  - lastTransitionTime: "2026-01-29T11:05:25Z"
    message: 'Failed to create OpenStack client scope: secrets "dev-test-cloud-config"
      not found'
    reason: OpenStackCredentialsFailed
    severity: Error
    status: "False"
    type: OpenStackCredentialsAvailable
  ready: false
```
This I am quite happy with. It clearly shows which secret it is unhappy about and why. "OpenStack client scope" is perhaps not immediately understandable, but it avoids making assumptions about the kind of error we get. We cannot know if there is some parsing error or something else entirely going on so I think it is quite ok.

Next, I created a secret with matching name but nonsense content. Then I got this:

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-01-29T11:12:59Z"
    message: 'Failed to create OpenStack client scope: OpenStack credentials secret
      dev-test-cloud-config did not contain key clouds.yaml'
    reason: OpenStackCredentialsFailed
    severity: Error
    status: "False"
    type: OpenStackCredentialsAvailable
  ready: false
```

This is also really nice IMO!

Then I did one with proper content but wrong clouds:

```yaml
spec:
  identityRef:
    cloudName: capo-e2e # <-- This does not exist in the clouds.yaml
    name: dev-test-cloud-config
    type: Secret
...
status:
  conditions:
  - lastTransitionTime: "2026-01-29T11:16:19Z"
    message: 'Failed to create OpenStack client scope: auth option failed for cloud
      : Missing input for argument [auth_url]'
    reason: OpenStackCredentialsFailed
    severity: Error
    status: "False"
    type: OpenStackCredentialsAvailable
  ready: false
```

It is looking for capo-e2e but it doesn't exist. Not the best error message, but I am not sure how to improve. Should we add extra validation somewhere? I think this is coming from gophercloud currently.

Next, correct cloud entry except the password was wrong:

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-01-29T11:21:47Z"
    message: 'Failed to create OpenStack client scope: providerClient authentication
      err: Expected HTTP response code [201 202] when accessing [POST http://10.0.3.15/identity/v3/auth/tokens],
      but got 401 instead: {"error":{"code":401,"message":"The request you have made
      requires authentication.","title":"Unauthorized"}}'
    reason: OpenStackCredentialsFailed
    severity: Error
    status: "False"
    type: OpenStackCredentialsAvailable
  ready: false
```

Quite ok I would say.

Finally, wrong URL:

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-01-29T11:23:19Z"
    message: 'Failed to create OpenStack client scope: providerClient authentication
      err: Get "http://10.0.3.150/identity/": EOF'
    reason: OpenStackCredentialsFailed
    severity: Error
    status: "False"
    type: OpenStackCredentialsAvailable
  ready: false
```

Also ok IMO.

All in all, I think this is a clear improvement. Before this PR we did not get any conditions at all. Users had to dig through logs.

mandre:
Thanks for manually checking the different cases. I believe this covers the most important ones indeed.

I'd still rename the condition to `OpenStackAuthenticationSucceded` or something equivalent now that we confirmed it's set to false in more cases than just missing secrets.

lentzi90:
Good point! I guess there is a little risk that authentication fails later, in the middle of operations. Then we probably would not be able to set this condition. But the next reconcile should handle that anyway so should be fine.
I'll update the name.

I also realized now that we should check all other controllers and see if we can do the same there. This one is the most critical, but we could have different credentials per OpenStackMachine and those should also have conditions then.

bnallapeta:
@lentzi90 please update once this is ready for review. Thanks.

lentzi90:
/hold cancel
Now this is up to date and all changes done. PTAL! :slightly_smiling_face:

lentzi90:
I forgot to mention that I didn't include OpenStackMachineTemplate here. It has no conditions at all yet and I didn't want to "mess with the API" just to add these conditions there also. I'll instead do a follow up PR where I add conditions there.

api/v1beta1/conditions_consts.go
```diff
@@ -72,6 +72,15 @@ const (
 	PortCreateFailedReason = "PortCreateFailed"
 )
 
+const (
+	// OpenStackAuthenticationSucceded reports on the current status of the OpenStack credentials.
```
bnallapeta: OpenStackAuthenticationSucceded => OpenStackAuthenticationSucceeded

typo. fix it all places.

api/v1beta1/conditions_consts.go
```diff
@@ -72,6 +72,15 @@ const (
...
+	// Available indicates that the credentials exist, but does not guarantee they are valid or have proper permissions.
```
bnallapeta: Available indicates => Succeeded indicates

controllers/openstackcluster_controller.go
```diff
@@ -124,8 +124,10 @@ func (r *OpenStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	clientScope, err := r.ScopeFactory.NewClientScopeFromObject(ctx, r.Client, r.CaCertificates, log, openStackCluster)
 	if err != nil {
+		v1beta1conditions.MarkFalse(openStackCluster, infrav1.OpenStackAuthenticationSucceded, infrav1.OpenStackAuthenticationFailedReason, clusterv1beta1.ConditionSeverityError, "Failed to create OpenStack client scope: %v", err)
 		return reconcile.Result{}, err
 	}
+	v1beta1conditions.MarkTrue(openStackCluster, infrav1.OpenStackAuthenticationSucceded)
```
bnallapeta: The `MarkTrue` condition is only set for this ctrl - OpenStackCluster where as `MarkFalse` is set for all the ctrls. Perhaps we need to set both true and false for all 4 controllers to stay consistent.

bnallapeta (commented):
Looks good in general. Pending those minor comments.

lentzi90:
Thanks @bnallapeta ! Good findings!

api/v1beta1/conditions_consts.go
lentzi90: Oh actually this one is not true as I found out. We actually check authentication so I'll remove the comment :slightly_smiling_face:

PR #2986, opened 2026-02-02
⚠️ Migrate controllers and types to CAPI v1beta2 conditions

### LINKED ISSUE #2942: Update controllers to use conditions instead of deprecated fields
Migrate controllers away from deprecated status fields:

- Replace `status.FailureReason`/`FailureMessage` usage with condition-based errors
- Update condition utilities to use `metav1.Condition`
- Update controller tests

bnallapeta:
This PR completes the migration to Cluster API v1beta2 conditions contract by updating all controllers, service layer, and tests to use the new conditions API.

_minor additional change: all generated code updated to 2026 copyright year._

Fixes #2942 part of #2895 

- PR # 4 as per #2895 plan
- The migration maintains backward compatibility for v1alpha1 types (e.g., `OpenStackServer.Status.Ready` still set for predicates)
- The v1beta1 types remain as conversion hub (storage version) until full migration complete
- Once the first 3 PRs are merged, I will then do the conversion changes.

/hold


api/v1alpha1/openstackfloatingippool_types.go
```diff
@@ -88,7 +87,7 @@ type OpenStackFloatingIPPoolStatus struct {
 	// +optional
 	FloatingIPNetwork *infrav1.NetworkStatus `json:"floatingIPNetwork,omitempty"`
 
-	Conditions clusterv1beta1.Conditions `json:"conditions,omitempty"`
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
lentzi90, line 90: This would be a change in the existing v1alpha1 CRD. It is "just" conditions, and we don't make much promises about alpha APIs but I am a bit worried how this will work.

controllers/openstackmachine_controller.go
```diff
@@ -456,35 +493,68 @@ func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogg
...
+			Status: metav1.ConditionTrue,
+			Reason: infrav1.ReadyConditionReason,
+		})
 	case infrav1.InstanceStateError:
 		// If the machine has a NodeRef then it must have been working at some point,
 		// so the error could be something temporary.
-		// If not, it is more likely a configuration error so we set failure and never retry.
+		// If not, it is more likely a configuration error so we set terminal error condition.
 		scope.Logger().Info("Machine instance state is ERROR", "id", openStackServer.Status.InstanceID)
-		if !machine.Status.NodeRef.IsDefined() {
```
lentzi90, line 490: Why are we not checking this any more?

kustomize/default/cluster-template.yaml
lentzi90, line 1: nit: We should rename the folder also, or perhaps just move everything one level up. Now we have all these v1beta2 manifests under v1beta1

lentzi90: I am fine leaving this for a follow up PR

lentzi90:
The upgrade error looks real. I am not sure what we are missing, but it seems it doesn't find the v1beta2 API for some reason :thinking:

lentzi90:
Oh now I get it! The issue is that we switch all the yaml to v1beta2 here, but the upgrade test starts from an older version that doesn't know about v1beta2. We will need to keep one kustomization with v1beta1 for the upgrades.

bnallapeta:
You're right. I wonder why it passed twice earlier. 

I will push a commit later today.

lentzi90:
It is a different test. The "normal" e2e suite passes, but we also have "full" suite that runs some additional tests (e.g. upgrade and self-hosted). Those are not automatically triggered on every PR, but I try to make sure they run on PRs that could be affecting upgrades.

api/v1alpha1/openstackfloatingippool_types.go
bnallapeta: hmm, trying to understand the whole picture.

Old type (clusterv1beta1.Condition): has type, status, severity, lastTransitionTime, reason, message

New type (metav1.Condition): has type, status, observedGeneration, lastTransitionTime, reason, message

so, its the severity field that we are dropping. In practice, during the upgrade path, on the first reconcile, controller will silently ignore the unknown severity and it overwrites the conditions with the new format. And since this is status and not spec change for v1alpha1, I think we should be fine.

controllers/openstackmachine_controller.go
bnallapeta: The `NodeRef` check existed to decide whether to call `SetFailure()`, which is gone in v1beta2. 

Also, the old "non-terminal" path returned an empty Result{} with no requeue anyway. It wasn't actually retrying anything. Now we treat all ERROR states as terminal and let `MachineHealthCheck` handle recovery.

kustomize/default/cluster-template.yaml
bnallapeta: I used `git mv` to move everything one folder up. Now, we don't have a subfolder named v1beta1 (or v1beta2). I updated all the references accordingly. 

I checked CAPA, CAPZ for reference and none of them use a subfolder. So, I think its clean this way.

bnallapeta:
looks like the fix worked (v1beta1 templates). This is a different issue. Apparently, an additional failure domain (nova) is being discovered with 0.14 that wasn't originally in 0.13. 

I'm looking into this.

lentzi90:
Thanks! Let me know if you need help!

bnallapeta:
Alright, the resourceVersions didn't stay stable failure has two causes:

v1beta1→v1beta2 conversion iterates a Go map to build the FailureDomains list, producing non-deterministic ordering on each read. Fixed by sorting the converted list by name.

Controller rebuilds status.failureDomains on every reconcile, so the transient nova default AZ appearing in CI causes a status mutation. Fixed by only populating failure domains once (only when the list is empty).

Let's hope it passes this time

lentzi90:
It passed, awesome! I'll review this tomorrow again. I need to check the IPPool stuff

bnallapeta:
finally 🥹 
yes please, let's merge it before it gets tired and breaks again:p

controllers/openstackmachine_controller.go
```diff
@@ -432,31 +467,52 @@ func (r *OpenStackMachineReconciler) reconcileNormal(ctx context.Context, scope
 // reconcileMachineState updates the conditions of the OpenStackMachine instance based on the instance state
 // and sets the ProviderID and Ready fields when the instance is active.
 // It returns a reconcile request if the instance is not yet active or in an error state.
-func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogger, openStackMachine *infrav1.OpenStackMachine, machine *clusterv1.Machine, openStackServer *infrav1alpha1.OpenStackServer) *ctrl.Result {
+func (r *OpenStackMachineReconciler) reconcileMachineState(scope *scope.WithLogger, openStackMachine *infrav1.OpenStackMachine, _ *clusterv1.Machine, openStackServer *infrav1alpha1.OpenStackServer) *ctrl.Result {
```
lentzi90, line 470: nit: We should remove the Machine parameter completely. Can be done in a separate PR.

lentzi90, line 470: nit: We should remove the Machine argument completely. Can be done in a separate PR.

bilbobrovall:
> @huxcrux @bilbobrovall do you have any opinions on this? Do you use the FloatingIPPool? Do you rely on the conditions?

Neither Hugo nor I are working on this anymore. I think @mikaelgron is the one who can answer this.

bnallapeta:
@nikParasyr @smoshiur1237 PTAL.

PR #3007, opened 2026-02-12
:seedling: Bump golangci-lint to 2.9.0

nikParasyr:
Bump golangci-lint to 2.9.0

Had some errors like:
```
pkg/utils/strings/strings_test.go:17:9: var-naming: avoid package names that conflict with Go standard library package names (revive)
package strings
```
opted to use `//nolint:revive` on these lines as i though we probably:
- dont want to rename packages now
- do want to avoid this issue in future packages

(different option would be to use `skip-package-name-collision-with-go-std` in the revive lint [settings](https://golangci-lint.run/docs/linters/configuration/#revive))

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/hack/tools/Makefile
+++ b/hack/tools/Makefile
@@ -15,7 +15,7 @@
 ROOT_DIR_RELATIVE := ../..
 include $(ROOT_DIR_RELATIVE)/common.mk
 
-GOLANGCI_LINT_VERSION ?= v2.8.0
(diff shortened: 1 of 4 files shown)


lentzi90 (commented):
/approve

PR #3037, opened 2026-03-03
:bug: skip reconcile for ipaddress and ipaddressclaim when cluster is paused

### LINKED ISSUE #3029: floating ip address conflict error during migration
/kind bug

I have created floatingippool, and add ipadressclaim for cluster A to assign floating ips for specified machineDeployment.

When I migrate cluster A from management cluster B to management cluster C, I met ip address conflict error:
```
Error: [action failed after 10 attempts: error updating "ipam.cluster.x-k8s.io/v1beta1, Kind=IPAddress" <cluster-A>/<cluster-A>-worker-cloud-alln-1-c-srl9m-fz8lp-floating-ip-address: admission webhook "validation.ipaddress.ipam.cluster.x-k8s.io" denied the request: spec: Forbidden: IPAddress spec is immutable. Diff: v1beta1.IPAddressSpec{
    ClaimRef: {Name: "<cluster-A>-worker-cloud-alln-1-c-srl9m-fz8lp-floating-ip-address"},
    PoolRef:  {APIGroup: &"infrastructure.cluster.x-k8s.io", Kind: "OpenStackFloatingIPPool", Name: "us-txccnci2"},
-   Address:  "10.xxx.xxx.89",
+   Address:  "10.xxx.xxx.230",
    Prefix:   32,
    Gateway:  "",
  }, ...] 
```

The new ipaddress `10.xxx.xxx.230` is created in the new  management cluster C before the old ipaddress `10.xxx.xxx.89` which is attached to the original node migrated to it.

1. When cluster is paused in  new  management cluster C during migration, no new ipdress should be created (reconcile should be paused)
2. After IPAdress migrated, reconcile should ensure the node has exactly the same ipadress.

**Environment:**

- Cluster-API version: v1.10.9
- OpenStack version: v0.12.6
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): v1.33.7
- OS (e.g. from `/etc/os-release`): Ubuntu 24.04.3 LTS"

archerwu9425:

Fixes #3029

This PR is to fix `ipaddressclaim` and `ipaddress` keep reconciling while cluster is paused.
Following the cluster-api ipam provider contract:
https://cluster-api.sigs.k8s.io/developer/providers/contracts/ipam#ipam-provider

Changes include:
1. If the related Cluster is paused, abort reconciliation
2. Add finalizer `openstackfloatingippool.infrastructure.cluster.x-k8s.io` to `ipaddressclaim`
3. For `ipaddress`, if the `ipaddressclaim` defined in `spec.ClaimRef` or the related cluster is paused, abort reconciliation
4. If  `ipaddress` is deleted, update the `status` of the `ipaddressclaim` defined in `spec.ClaimRef`

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


archerwu9425:
@lentzi90  @nikParasyr Could you please help review? Thanks

controllers/openstackfloatingippool_controller.go
```diff
@@ -127,7 +133,32 @@ func (r *OpenStackFloatingIPPoolReconciler) Reconcile(ctx context.Context, req c
...
+				log.Error(err, "Failed to add finalizer to claim", "claim", claim.Name)
+				return ctrl.Result{}, err
+			}
+			continue
+		}
+
+		if annotations.IsPaused(cluster, &claim) {
+			log.V(4).Info("IPAddressClaim or linked Cluster is paused, skipping reconcile", "claim", claim.Name, "namespace", claim.Namespace)
+			continue
+		}
```
lentzi90: Shouldn't we do this before adding the finalizer?

controllers/openstackfloatingippool_controller.go
```diff
@@ -127,7 +133,32 @@ func (r *OpenStackFloatingIPPoolReconciler) Reconcile(ctx context.Context, req c
...
+
 		if !claim.DeletionTimestamp.IsZero() {
+			controllerutil.RemoveFinalizer(&claim, infrav1alpha1.OpenStackFloatingIPPoolFinalizer)
+			if err := r.Client.Update(ctx, &claim); err != nil {
+				return ctrl.Result{}, err
+			}
```
lentzi90, line 182: Generally we should clean up first and then remove the finalizer when we have confirmed that the cleanup worked. Usually we also set a `defer` to handle the updates instead of doing it directly.

controllers/openstackfloatingippool_controller.go
```diff
@@ -288,6 +319,29 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileIPAddresses(ctx context.Con
 			continue
 		}
 
+		// Check if the owning claim or its cluster is paused before processing deletion,
+		// and clear the claim's AddressRef so it can be re-reconciled once unpaused or re-created.
+		if ipAddress.Spec.ClaimRef.Name == "" {
+			continue
```
lentzi90: This is a change in behavior. We are skipping past all the rest of the loop where we handle finalizers and update the pool.

controllers/openstackfloatingippool_controller.go
```diff
@@ -297,6 +351,14 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileIPAddresses(ctx context.Con
 				pool.Status.AvailableIPs = append(pool.Status.AvailableIPs, ipAddress.Spec.Address)
 			}
 		}
+		// Clear AddressRef so the claim will be re-assigned an IP on the next reconcile.
+		if claim != nil && claim.Status.AddressRef.Name != "" {
+			claim.Status.AddressRef.Name = ""
+			if err := r.Client.Status().Update(ctx, claim); err != nil {
+				return fmt.Errorf("failed to clear AddressRef for claim %q: %w", claim.Name, err)
```
lentzi90, line 375: I am not following here. Does this have anything to do with cluster paused or is it trying to fix something else?

lentzi90 (commented):
Thanks for the PR!
I agree that we have a bug in that we reconcile the IPAddresses and claims even when the cluster is paused. This PR seems to do more than just fix this bug though.
It also adds no tests. I think we do need some tests to verify the behavior and I have a few concerns (see below).

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90  I think as long as the IPAdreesClaim has related cluster, we should add finalizer to it first. If the cluster is paused, we can skip reconcile and not create ipadress for it.

controllers/openstackfloatingippool_controller.go
archerwu9425: Good point, I'll update this part. Thanks

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90 My bad, will change this part to just go through the delete flow

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90 Yes, as I mentioned in the PR description:
`4. If ipaddress is deleted, update the status of the ipaddressclaim defined in spec.ClaimRef`

Currently, if we delete ipaddress, it will be deleted, but the ipadressclaim not be notified and status still refer the deleted ipaddress, and will stop the claim to reconcile and create new one. 

Issues we see:
ipaddress is deleted by accident, and it will be never be added back, and we need to manually delete and recreate the ipadressclaim to get a new ipadress.

Let me know if you have any concern for this.

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90 changes done, if `ipAddress.Spec.ClaimRef.Name == ""`, will go on reconcile to delete the `ipAddress`

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90 Changes made:

1. Delete the `ipAddress` the claim refered before remove the finalizer.

2. Not able to defer as the reconcile is for pool and the claim reconcile is in loop (defer in loop will cause lint error and bad practice), so still directly run the update. 

Thanks

controllers/openstackfloatingippool_controller.go
archerwu9425: @lentzi90  Changes done:
Follow your advice, will check cluster paused status before adding the finalizer.

Thanks

archerwu9425:
@lentzi90 changes made based on comment and tests added. Could you please help review again? Thanks

archerwu9425:
@lentzi90 Could you please help review again? Thanks

archerwu9425:
@lentzi90 @nikParasyr  Could you please help review? Thanks

archerwu9425:
@lentzi90 @nikParasyr Could you please help review? Thanks

nikParasyr:
@archerwu9425 i have it on my list, probably end of the week as it is a bit hectic now.
@bnallapeta if you have time in the meantime, would be nice

bnallapeta:
> @archerwu9425 i have it on my list, probably end of the week as it is a bit hectic now. @bnallapeta if you have time in the meantime, would be nice

Will put it on my list. But unfortunately, cannot get it to soon.

controllers/openstackfloatingippool_controller.go
```diff
@@ -127,7 +133,45 @@ func (r *OpenStackFloatingIPPoolReconciler) Reconcile(ctx context.Context, req c
 
 	for _, claim := range claims.Items {
 		log := log.WithValues("claim", claim.Name)
+
+		cluster, err := util.GetClusterFromMetadata(ctx, r.Client, claim.ObjectMeta)
+		if err != nil {
+			log.Error(err, "Failed to get owning cluster, skipping claim", "claim", claim.Name)
+			continue
```
bnallapeta, line 148: Consider a scenario where there's no cluster label no the claim. In that case, we would get ErrNoCluster and since you have a continue here, it would just skip the claim completely. 

This can cause a problem since any IPAddressClaim without a cluster label will never be processed and never be deleted (since finalizer won't be removed). 

I think we should process it normally when label is not found as well as this will avoid the regression. Currently, this is what we do.

controllers/openstackfloatingippool_controller.go
```diff
@@ -288,18 +328,40 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileIPAddresses(ctx context.Con
...
+				claim = nil
 			} else {
-				pool.Status.AvailableIPs = append(pool.Status.AvailableIPs, ipAddress.Spec.Address)
+				cluster, err := util.GetClusterFromMetadata(ctx, r.Client, claim.ObjectMeta)
+				if err != nil && !apierrors.IsNotFound(err) {
+					return fmt.Errorf("failed to get owning cluster for claim %q: %w", claim.Name, err)
+				}
+				if cluster != nil && annotations.IsPaused(cluster, claim) {
+					scope.Logger().V(4).Info("IPAddress owner IPAddressClaim or linked Cluster is paused, skipping deletion", "ipAddress", ipAddress.Name, "claim", claim.Name)
+					continue
```
bnallapeta, line 361: This means the deletion will be skipped. So, the floating IP exists in OpenStack. And the following code at line 367 will mark this as available. This might cause the IP to be allocated to two different claims. 

Fix: add this to ClaimedIPs. That way, this IP won't be utilized by any other resource.

controllers/openstackfloatingippool_controller.go
```diff
@@ -422,14 +508,27 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileFloatingIPNetwork(scope *sc
...
 		panic(fmt.Sprintf("Expected a IPAddressClaim but got a %T", o))
 	}
 	if claim.Spec.PoolRef.Kind != openStackFloatingIPPool {
 		return nil
 	}
+
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, claim.ObjectMeta)
+	if err != nil {
+		log.Error(err, "Failed to get owning cluster, skipping mapping", "claim", claim.Name, "namespace", claim.Namespace)
+		return nil
```
bnallapeta, line 555: Same issue as above. If the claim doesn't have a cluster label, GetClusterFromMetadata returns ErrNoCluster, and this returns nil. Meaning, changes to this claim will never trigger a pool reconciliation. The controller won't even notice the claim was created, updated, or deleted.

fix: If there's no cluster, we should still map the event to the pool so reconciliation proceeds normally.

archerwu9425: @bnallapeta Sorry for the late response due to recent personal issues. I'm wondering in which situation will we have claim that doesn't belongs to a cluster? I thought we shouldn't just let users create a orphan claim and ipaddress just put it there without any change to attach to a node. After all this repo is openstack provider for cluster-api, it should serve a cluster.

Let me know your ideas on this.

Thanks

controllers/openstackfloatingippool_controller.go
```diff
@@ -320,6 +382,30 @@ func (r *OpenStackFloatingIPPoolReconciler) reconcileIPAddresses(ctx context.Con
 	return nil
 }
 
+func (r *OpenStackFloatingIPPoolReconciler) deleteIPAddress(ctx context.Context, scope *scope.WithLogger, pool *infrav1alpha1.OpenStackFloatingIPPool, ipAddress *ipamv1.IPAddress) error {
+	networkingService, err := networking.NewService(scope)
```
lentzi90, line 386: I am a bit concerned about the performance and scalability of this. We would here create a new service client for each IPAddress instead of reusing it for the whole reconciliation.

lentzi90 (commented):
There are some unresolved comments still. Added one more below.
In general, it would be easier to review and get it merged if it focused on just one single fix. It seems to me we are dealing with at least two different things now:
1. skip reconciliation while cluster paused, and
2. clear address ref when IPAddress is deleted.

Please squash the commits

PR #3038, opened 2026-03-03
🌱Update missing go version uplift in netlify.toml file

smoshiur1237:
We missed the uplift of go version 1.25.7 during previous uplift in netlify.toml file.


lentzi90 (commented):
Thanks, good catch!
/approve

PR #3039, opened 2026-03-03
🌱Harden GitHub workflow

smoshiur1237:
Add persist-credentials: false to checkout actions, switch pr-verifier from pull_request_target to pull_request, add explicit permissions to workflows and apply zizmor safe fixes.

### DIFF
--- a/.github/workflows/pr-dependabot.yaml
+++ b/.github/workflows/pr-dependabot.yaml
@@ -21,6 +21,8 @@ jobs:
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v6.0.2
+      with:
+        persist-credentials: false
     - name: Calculate go version
       id: vars
       run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
--- a/.github/workflows/pr-gh-workflow-approve.yaml
+++ b/.github/workflows/pr-gh-workflow-approve.yaml
@@ -5,7 +5,7 @@
 name: Approve GH Workflows
 
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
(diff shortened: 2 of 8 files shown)


bnallapeta:
With `persist-credentials: false`, the GITHUB_TOKEN is not stored in git config, so these git push commands will fail with an authentication error no?

lentzi90:
> With `persist-credentials: false`, the GITHUB_TOKEN is not stored in git config, so these git push commands will fail with an authentication error no?

They already do, unfortunately. We cannot give enough permissions so that it can create PRs from the workflow. I just keep an eye on the workflow and when I see it failing I know it usually means there is a new version

bnallapeta:
> > With `persist-credentials: false`, the GITHUB_TOKEN is not stored in git config, so these git push commands will fail with an authentication error no?
> 
> They already do, unfortunately. We cannot give enough permissions so that it can create PRs from the workflow. I just keep an eye on the workflow and when I see it failing I know it usually means there is a new version

No, I meant this about the release.yaml workflow. Unlike the other workflows that uses third party packages, release.yaml uses direct git commands.

lentzi90:
Ah, good point! I think it is still fine because we put permissions on the job level:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/ecf118b54e754249f4093577dee316c0ef8d2e68/.github/workflows/release.yaml#L16
Unless this change makes the checkout action clean up it all so we have no auth for the rest... :thinking:

bnallapeta:
I think we need to remove that line for the release workflow. So, the reason to set persist creds to false is to not allow any third party modules/packages to misuse our creds.. but in this workflow, its purely Git commands. So, I think the risk is very minimal/nil.

lentzi90:
True, plus this workflow runs *after* merge. So we always have approval from trusted humans before it runs.
@smoshiur1237 could you revert the change for the release workflow?

smoshiur1237:
Here is the revert PR: #3040

PR #3043, opened 2026-03-09
:seedling: Remove update-golangci-lint workflow

nikParasyr:
The workflow requires pull request creation permissions,
which are restricted by an organization-wide policy that
we cannot change. Remove the workflow to avoid failures.

**TODOs**:
- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


nikParasyr:
After looking into it a bit more: 
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/actions/runs/22832899369/job/66223949805#step:6:113

I think the steps gets triggered properly (this is also why the branch exists with a commit etc) but github actions is not authorized to create PRs. So i think the fix is on the settings of the repo, not the workflow definition.

@lentzi90 could you have a look if there is a setting on repo level? if it is on organization level then i suspect they wont be willing to change it but who knows (in the latter case, it might be better to remove the workflow all-together or create an issue instead if actions are allowed to do this)

lentzi90:
Right, I think the issue here is that we would need a PAT for this. There is some limitation for the built-in permissions that prevent a workflow from doing actions that would cause more workflows. If we create a PR, we create more workflows.
I am not completely sure but I think this is the limitation we are hitting.
https://github.com/orgs/community/discussions/25612

mandre:
There's apparently a setting for it: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#preventing-github-actions-from-creating-or-approving-pull-requests

However, the setting's inherited from the organization so it might not be possible for us.

I'd personally wouldn't object to disabling the workflow entirely, rather than updating it to create issues instead. We can live with a slightly older golangci-lint IMO.

mandre:
BTW, the real error can be seen in [this workflow run](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/actions/runs/22787089618/job/66106033022#step:6:108) (as noted by @nikParasyr already :smile:) :
```
  Attempting creation of pull request
  Error: GitHub Actions is not permitted to create or approve pull requests. - https://docs.github.com/rest/pulls/pulls#create-a-pull-request
```

[Subsequent periodic runs](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/actions/runs/22980241298/job/66718194453#step:6:119) would fails because of branch protection rules:
```
  remote: error: GH006: Protected branch update failed for refs/heads/update-golangci-lint-v2.11.3.        
  remote: 
  remote: - 3 of 3 required status checks are expected.        
  remote: 
  remote: - Cannot force-push to this branch        
  To https://github.com/kubernetes-sigs/cluster-api-provider-openstack
   ! [remote rejected]   update-golangci-lint-v2.11.3 -> update-golangci-lint-v2.11.3 (protected branch hook declined)
  error: failed to push some refs to 'https://github.com/kubernetes-sigs/cluster-api-provider-openstack'
  Error: The process '/usr/bin/git' failed with exit code 1
```

nikParasyr:
@mandre / @bnallapeta / @lentzi90 
I think the "silent" agreement here was to remove the workflow altogether due to the organization level policy that we cannot change, so i have updated the PR accordingly.

Let me know if you have a better alternative in mind

lentzi90:
Yes this is ok for me. We will just need to remember to bump it manually every now and then

bnallapeta:
I think losing the automation and relying on manual approach can be tricky. I looked at how CAPI, and other providers are doing this. 

They define the golangci-lint version in [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/pr-golangci-lint.yaml#L33) and have the Makefile extract it from [there](https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L171) at parse time. This way, when Dependabot bumps the golangci-lint-action, the lint version gets bumped automatically. This avoids the whole dependency on PAT or org-level permissions or manual effort. 

We'd need to add a pr-golangci-lint.yaml workflow using golangci/golangci-lint-action, update the Makefile to extract the version from it, and drop the custom ensure-golangci-lint.sh script. Our existing Dependabot github-actions config would then handle auto-bumping going forward.

@nikParasyr @lentzi90  let me know how this sounds.

lentzi90:
Cool! This sounds very good!

bnallapeta:
@nikParasyr do you want to take this up in this PR or another? Or shall I submit a new one?

nikParasyr:
>  @nikParasyr do you want to take this up in this PR or another? Or shall I submit a new one?

Im focusing on the v1beta2 changes these days, so if you can pick it up that would be great

bnallapeta:
@nikParasyr @lentzi90 @mandre PTAL https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/3095

PR #3050, opened 2026-03-10
🌱Add zizmor scanner

smoshiur1237:
Zizmor scans workflow files to keep them hardened. It runs on push to produce a report in security tab, and it runs on PRs and blocks them if they is any errors.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,7 +19,7 @@ jobs:
       release_tag: ${{ steps.release-version.outputs.release_version }}
     if: github.repository == 'kubernetes-sigs/cluster-api-provider-openstack'
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
smoshiur1237, line 1: I was trying to avoid the` persist-credentials: false` in release.yml file. But zizmor scanner is catching it as an error.

lentzi90: You can add `# zizmor: ignore[artipacked]` comments for these.

smoshiur1237: Done

smoshiur1237: It is working with the zizmor ignore comment in release file.

Issue #3011, opened 2026-02-19
Add Fuzz tests for v1beta1 <-> v1beta2 conversions


lentzi90:
As noted in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2958, we should add Fuzz tests to thoroughly exercise the conversion functions and find any corner cases. There should be prior art in the git history and also in [CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/main/util/conversion/conversion.go).

PR #3053, opened 2026-03-11
DONOTMERGE: v1beta2: Standardize flavor in OpenStackMachine

nikParasyr:
Small draft to validate required api changes

@lentzi90 could you check if im missing something:
1. controller/client code, tests, templates, docs are out-of-scope, purely the api changes
2. Had to update:
    - v1beta2 definition, generate manifests
    - conversion webhooks and tests
    - fuzz tests under `pkg/webhooks` (maybe more when https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/3011 is tackled)
 3. question is: Am i missing something? For example are we doing any defaulting/validation outside of kubebuilder? 

/hold


lentzi90 (commented):
From a quick look, this looks solid! I don't think there is anything that you are missing. 👍

PR #3056, opened 2026-03-15
✨ Make allowedAddressPairs on OpenStackMachine ports mutable

jfpucheu:
## Summary

- **OpenStackMachineTemplate webhook**: excludes `allowedAddressPairs` from the immutability check so the field can be updated in-place without triggering a node rolling replacement.
- **OpenStackMachine webhook**: same exclusion to allow the OSMT controller to annotate OSMs without being rejected by the spec-immutability webhook.
- **OpenStackMachine controller**: deep-copies ports before mutation to avoid corrupting the original spec through shared slice backing arrays (which would incorrectly trigger the spec-immutability webhook).
- **Networking service**: adds `UpdateAllowedAddressPairs()` to call `ports.Update` on the Neutron API.
- **OpenStackMachineTemplate controller**: adds `reconcileAllowedAddressPairs` which walks MachineSets → Machines → OpenStackMachines → OpenStackServers and calls `UpdateAllowedAddressPairs()` for each provisioned port. Port IDs are read from `OpenStackServer.Status.Resources.Ports`. Idempotency is tracked via an annotation on each OpenStackMachine (`infrastructure.cluster.x-k8s.io/osmt-allowed-address-pairs`) so only a metadata-only patch is needed, avoiding the spec-immutability webhook entirely.
- **RBAC**: adds `machinesets`, `machines`, and `openstackmachines` get/list/watch/patch permissions to the OSMT controller.
- **Unit tests** for `reconcileAllowedAddressPairs`.
- **E2E test**: creates a cluster with 1 CP and 1 worker (via a dedicated MachineDeployment with an explicit port), then patches the OSMT three times to accumulate `allowedAddressPairs` and verifies via the Neutron API after each patch that the port is updated without server restart.

## Motivation

Neutron supports updating `allowedAddressPairs` on a port without recreating the server. Previously, any change to the OSMT spec (including `allowedAddressPairs`) was rejected by the immutability webhook, forcing operators to recreate nodes to apply network policy changes. This PR enables live updates of `allowedAddressPairs` on existing machines.

This feature is particularly useful in OpenStack environments **without a LoadBalancer** (e.g. `FlavorWithoutLB`), where solutions like **MetalLB** or **kube-vip** are used to expose services via Virtual IPs (VIPs). These tools require adding VIP addresses to the `allowedAddressPairs` of the worker node ports so that OpenStack allows the traffic to pass through. Without this feature, operators had to recreate nodes every time a new VIP was added — defeating the purpose of in-cluster load balancing.

## Test plan

- [ ] Unit tests pass: `go test ./controllers/... ./pkg/webhooks/... ./pkg/cloud/services/networking/...`
- [ ] E2E test passes: `E2E_GINKGO_FOCUS="allowedAddressPairs" make test-e2e`
- [ ] Verify Neutron port `allowedAddressPairs` are updated without Nova server restart

> I am not a developer — this contribution was implemented with the assistance of [Claude Code](https://claude.com/claude-code) because this feature was blocking our production use of MetalLB on OpenStack without LoadBalancer support.

### DIFF
--- a/controllers/openstackmachinetemplate_controller.go
+++ b/controllers/openstackmachinetemplate_controller.go
@@ -73,6 +73,7 @@ type OpenStackMachineTemplateReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachinetemplates,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachinetemplates/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachines,verbs=get;list;watch;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackservers,verbs=get;list;watch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets,verbs=get;list;watch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines,verbs=get;list;watch
 
@@ -218,6 +219,12 @@ func (r *OpenStackMachineTemplateReconciler) reconcileNormal(ctx context.Context
 		openStackMachineTemplate.Status.Capacity[corev1.ResourceStorage] = *resource.NewQuantity(storageBytes, resource.BinarySI)
 	}
 
+	// reconcileAllowedAddressPairs is called independently of the image/flavor logic so that
+	// it is never skipped by an early return (e.g. when imageID is not yet resolvable).
+	if err := r.reconcileAllowedAddressPairs(ctx, scope, clusterName, openStackMachineTemplate); err != nil {
+		return err
+	}
+
 	imageID, err := computeService.GetImageID(ctx, r.Client, openStackMachineTemplate.Namespace, openStackMachineTemplate.Spec.Template.Spec.Image)
 	if err != nil {
 		return err
@@ -239,10 +246,6 @@ func (r *OpenStackMachineTemplateReconciler) reconcileNormal(ctx context.Context
 				openStackMachineTemplate.Status.NodeInfo.OperatingSystem = osType
 			}
 		}
-	}
-
-	if err := r.reconcileAllowedAddressPairs(ctx, scope, clusterName, openStackMachineTemplate); err != nil {
-		return err
 	}
 
 	return nil
--- a/controllers/openstackmachinetemplate_controller_test.go
+++ b/controllers/openstackmachinetemplate_controller_test.go
@@ -42,8 +42,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	infrav1 "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta2"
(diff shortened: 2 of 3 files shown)


jfpucheu:
This is my first PR. I have not been able to test on a multi-AZ environment as I only have a single-AZ DevStack setup available."

lentzi90:
/retitle ✨ Make allowedAddressPairs on OpenStackMachine ports mutable

jfpucheu:
The CI failures are not related to this PR. The flatcar test timed out waiting for the control plane to become ready (infrastructure issue), which cascaded into the other failures. Our changes only affect allowedAddressPairs handling and have no impact on the flatcar bootstrap path.
/retest

controllers/openstackmachinetemplate_controller.go
```diff
@@ -195,6 +212,9 @@ func (r *OpenStackMachineTemplateReconciler) reconcileNormal(ctx context.Context
 	if err != nil {
 		return err
 	}
+	if imageID == nil {
```
bnallapeta, line 232: This early return skips `reconcileAllowedAddressPairs` at the bottom. 

I think `reconcileAllowedAddressPairs` should be called independently from the image/flavor logic. You can either move it before the `imageID` check or split it out.

pkg/webhooks/openstackmachine_webhook.go
bnallapeta, line 1: Let's add additional test cases that include: changing only AAP (should pass), changing AAP + another field (should reject), adding/removing a port (should reject).

controllers/openstackmachinetemplate_controller.go
```diff
@@ -57,6 +71,9 @@ type OpenStackMachineTemplateReconciler struct {
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachinetemplates,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachinetemplates/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackmachines,verbs=get;list;watch;patch
```
bnallapeta, line 75: also add this since we need to read OpenStackServer

```
  // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=openstackservers,verbs=get;list;watch                          
```

jfpucheu: ok added

bnallapeta:
will take a look today.

bnallapeta:
@lentzi90 PTAL when you can.

lentzi90:
On my list! Might have to push it to next week though :slightly_frowning_face:

test/e2e/suites/e2e/e2e_test.go
```diff
@@ -1213,6 +1213,144 @@ var _ = Describe("e2e tests [PR-Blocking]", func() {
...
+			// The OSMT controller requires a direct Cluster owner reference.
+			osmt.Labels = map[string]string{clusterv1.ClusterNameLabel: clusterName}
+			osmt.OwnerReferences = []metav1.OwnerReference{
+				{
+					APIVersion: clusterv1.GroupVersion.String(),
+					Kind:       "Cluster",
+					Name:       cluster.Name,
+					UID:        cluster.UID,
+				},
+			}
```
lentzi90, line 1250: This shouldn't really be "needed", although I can imagine it makes the test faster. CAPI should set these after the first reconciliation.
Anyway, I think we can keep it.

PR #3110, opened 2026-04-09
📖 Replace embedded quick-start with link to upstream CAPI docs

### LINKED ISSUE #2811: Periodically build documentation
Our [getting-started page](https://cluster-api-openstack.sigs.k8s.io/getting-started) is [embedded directly from CAPI](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/getting-started.md).

This is nice as we do not need to duplicate the information here. However, it also means that we do not automatically update our docs when things change in CAPI. As long as we don't merge any PRs with doc changes here, our site is never updated.

We should create periodic job to build the site so that we get the latest changes from CAPI without manual "bump" PRs.

larainema:

The docs site at https://cluster-api-openstack.sigs.k8s.io embeds content from external repositories using `mdbook-embed`. For example, the [getting-started page](https://cluster-api-openstack.sigs.k8s.io/getting-started) is embedded directly from CAPI.

Since the site is deployed via Netlify and only rebuilds on push to `main`, the embedded external content becomes stale if no PRs are merged for some time.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2811

## Solution

Add a weekly scheduled GitHub Actions workflow that triggers a Netlify build via a [build hook](https://docs.netlify.com/configure-builds/build-hooks/). This Add a weekly scheduled GitHub Actions workflow that triggers a Netlify build via a [build hook](https://docs.netlify.com/conmmits.

The workflow also supports `workflow_dispatch` for manual triggering.

## Setup required by maintainers

After merging, a maintainer with Netlify access needs to:
1. Create a **1. Create a **1. Create a **1. Create a **Build & deploy > Build hooks_ (name it e.g. `weekly-docs-rebuild`)
2. Add the hook URL as a GitHub Actions secret named `NETLIFY_BUILD_HOOK`

### DIFF
--- a/docs/book/Makefile
+++ b/docs/book/Makefile
@@ -17,7 +17,6 @@ include $(ROOT_DIR_RELATIVE)/common.mk
 
 # Directories.
 MDBOOK := $(TOOLS_BIN_DIR)/mdbook
-MDBOOK_EMBED := $(TOOLS_BIN_DIR)/mdbook-embed
 MDBOOK_RELEASELINK := $(TOOLS_BIN_DIR)/mdbook-releaselink
 MDBOOK_TABULATE := $(TOOLS_BIN_DIR)/mdbook-tabulate
 BOOK_SRCS :=  $(call rwildcard,.,*.*)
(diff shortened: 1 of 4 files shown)


bnallapeta:
@lentzi90 who has the rights to set this up in netlify? could you share more details on that?

lentzi90:
Hmm I need to check. I am not sure if I even have those rights in netlify.

lentzi90:
I don't think we can do this unfortunately. :disappointed: 
Perhaps it would be better if we just link to https://cluster-api.sigs.k8s.io/user/quick-start instead of embedding it like we do now. Then there is never the issue of keeping it up to date.
What do you think?

larainema:
Thanks for the suggestion @lentzi90! I've pivoted the PR to replace the `embed-github` usage with a direct link to the upstream CAPI quick-start guide instead.

Changes:
- Replaced `{{#embed-github ...}}` in `getting-started.md` with a link to https://cluster-api.sigs.k8s.io/user/quick-start
- Removed the `[preprocessor.embed]` config from `book.toml`
- Removed `mdbook-embed` from both `docs/book/Makefile` and `hack/tools/Makefile`

This way the content always stays up to date (readers go directly to the source) and we avoid the stale embed problem entirely.

lentzi90 (commented):
Thank you! I think this makes sense.
/approve

PR #3111, opened 2026-04-13
📖 Clarify providerID initialization approaches in documentation

### LINKED ISSUE #3042: Clarify preferred providerID initialization in CAPO clusters
/kind feature

Currently [the documentation suggests](https://cluster-api-openstack.sigs.k8s.io/topics/external-cloud-provider) deploying the OpenStack Cloud Controller Manager after the control plane is ready, which then populates `Node.spec.providerID`.

However, it is also possible to [set the provider ID](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/templates/cluster-template.yaml) directly during node bootstrap via kubelet arguments. For example, using instance metadata exposed by OpenStack through cloud-init:

```yaml
apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
kind: KubeadmConfigTemplate
metadata:
  name: ${CLUSTER_NAME}-md-0
spec:
  template:
    spec:
      joinConfiguration:
        nodeRegistration:
          name: '{{ local_hostname }}'
          kubeletExtraArgs:
            cloud-provider: external
            provider-id: openstack:///'{{ instance_id }}'
```

With this approach the node registers with Kubernetes with providerID already set, which allows the cluster to start without requiring OCCM during bootstrap.

In practice this also affects Cluster API reconciliation. When `provider-id` is provided via kubelet arguments, the node registers with `spec.providerID` already populated and the Machine/MachineSet reconciliation completes successfully.

If `provider-id` is not set and OCCM is not deployed, the cluster can still create the underlying OpenStack resources (e.g. `OpenStackMachine` and the VM instance), but reconciliation does not fully complete because the Kubernetes Node never receives a `providerID`. In this state the cluster remains partially reconciled even though the VM and Kubernetes node exist.

The question is which model CAPO considers the preferred or supported approach:

1. OCCM-driven initialization

- kubelet runs with `--cloud-provider=external`
- nodes register without `providerID`
- OCCM populates `Node.spec.providerID`

2. Bootstrap-driven initialization

- kubelet runs with `--cloud-provider=external`
- kubelet receives `--provider-id` from cloud-init metadata
- nodes register with `providerID` already set
- OCCM may still be deployed later if cloud features (e.g. LoadBalancer services) are required

It would be helpful if the documentation clarified whether pre-populating providerID during bootstrap is a supported pattern, or whether OCCM is expected to always be responsible for setting it.

The bootstrap-driven approach can simplify cluster bring-up because it removes the dependency on deploying OCCM just to populate providerID. The cluster can start normally, and OCCM can be installed later if cloud-specific functionality (e.g. Octavia load balancers, node address management, or cloud node lifecycle handling) is needed.

Clarifying the recommended pattern in the CAPO documentation would help operators understand whether both approaches are supported or if OCCM-based initialization is the intended design.

larainema:

The documentation only describes deploying OCCM to populate `Node.spec.providerID`, but all default CAPO cluster templates actually use the bootstrap-driven approach (setting `provider-id` via kubelet args). This has caused confusion about which approach is preferred and whether both are supported.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/3042

## Changes

Adds a new "Setting the providerID on nodes" section to the external cloud provider documentation that clearly documents both supported approaches:

1. **Bootstrap-driven initialization (recommended)**: Set `provider-id` via kubelet arguments using OpenStack instance metadata from cloud-init. This is what all default templates use and what is tested in CI.

2. **OCCM-driven initialization**: Deploy the OpenStack Cloud Controller Manager which populates `providerID` after the control plane is ready.

### DIFF
--- a/docs/book/src/topics/external-cloud-provider.md
+++ b/docs/book/src/topics/external-cloud-provider.md
@@ -3,6 +3,9 @@
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
 - [External Cloud Provider](#external-cloud-provider)
+  - [Setting the providerID on nodes](#setting-the-providerid-on-nodes)
+    - [Option 1: Bootstrap-driven initialization (recommended)](#option-1-bootstrap-driven-initialization-recommended)
+    - [Option 2: OCCM-driven initialization](#option-2-occm-driven-initialization)
   - [Steps of using external cloud provider template](#steps-of-using-external-cloud-provider-template)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Docs looks good!
Could please check the PR description. The last part looks like an AI lost its mind 😅

larainema:
> Docs looks good! Could please check the PR description. The last part looks like an AI lost its mind 😅

Thanks @lentzi90 , updated.

lentzi90 (commented):
/approve

nikParasyr (approved):
/lgtm

PR #3118, opened 2026-04-14
🐛 Fix optional.Bool conversion losing explicit false values

### LINKED ISSUE #2408: Optional Bool Fields Omitted During v1alpha7 → v1beta1 Conversion
/kind bug

When an OpenStackCluster CR is automatically converted from v1alpha7 to v1beta1, the field DisableAPIServerFloatingIP is converted using Convert_bool_To_optional_Bool. This function omits fields that are set to false or nil, leading to their removal from the resulting CR.

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/685b5ee01dfee9ee711fdc65ef5457edaa09b4ac/api/v1alpha7/zz_generated.conversion.go#L983

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/685b5ee01dfee9ee711fdc65ef5457edaa09b4ac/pkg/utils/optional/conversion.go#L105-L112

If `spec.disableAPIServerFloatingIP` field was explicitly set as `false` in v1alpha7, it gets removed during the upgrade to v1beta1.
Since the OpenStackCluster validation webhook does not allow changes to this field after creation, it cannot be set back to false once omitted.
This affects 3rd party controllers/operators managing CAPO CRs, as they fail when attempting to ensure the missing field.

Other optional boolean fields converted in a similar manner are also likely impacted.

Boolean fields that explicitly have a value (even `false`) should not be omitted during conversion.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): 0.11.2
- Cluster-API version: 0.8.4
- OpenStack version: All maintained versions
- Minikube/KIND version:
- Kubernetes version (use `kubectl version`): 1.31
- OS (e.g. from `/etc/os-release`): Ubuntu Jammy

larainema:

`Convert_bool_To_optional_Bool` incorrectly converted `false` to `nil`, losing the distinction between an explicit `false` and an unset value. This caused optional boolean fields like `DisableAPIServerFloatingIP` to be silently dropped during API version conversion when explicitly set to `false`.

Similarly, `RestoreBool` treated an explicit `false` the same as `nil`, overwriting it with the previous value instead of preserving it.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2408

### Bug fixes (`pkg/utils/optional/conversion.go`)

- **`Convert_bool_To_optional_Bool`**: Always preserves the input value. Previously, `false` was converted to `nil`.
- **`RestoreBool`**: Only restores from previous when `dst` is `nil`, not when it is an explicit `false`.

### Tests (`pkg/utils/optional/conversion_test.go`)

Added unit tests for `Convert_bool_To_optional_Bool`, `Convert_optional_Bool_To_bool`, and `RestoreBool` covering:
- `true` is preserved
- `false` is preserved as explicit `false`, not `nil`
- `nil` converts to `false`
- Explicit `false` dst is NOT overwritten by `RestoreBool`

Tests were verified to fail with the old code and pass with the fix.

### DIFF
--- a/pkg/utils/optional/conversion.go
+++ b/pkg/utils/optional/conversion.go
@@ -39,7 +39,7 @@ func RestoreUInt16(previous, dst *UInt16) {
 }
 
 func RestoreBool(previous, dst *Bool) {
-	if *dst == nil || !**dst {
+	if *dst == nil {
 		*dst = *previous
 	}
 }
@@ -103,11 +103,7 @@ func Convert_optional_UInt16_To_uint16(in *UInt16, out *uint16, _ conversion.Sco
 }
 
 func Convert_bool_To_optional_Bool(in *bool, out *Bool, _ conversion.Scope) error {
-	if !*in {
-		*out = nil
-	} else {
-		*out = in
-	}
+	*out = in
 	return nil
 }
 
--- a/pkg/utils/optional/conversion_test.go
+++ b/pkg/utils/optional/conversion_test.go
@@ -1,0 +1,169 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package optional
+
+import (
+	"testing"
+
+	"k8s.io/utils/ptr"
+)
+
+func TestConvert_bool_To_optional_Bool(t *testing.T) {
+	tests := []struct {
+		name     string
+		in       bool
+		wantNil  bool
+		wantBool bool
+	}{
+		{
+			name:     "true is preserved",
+			in:       true,
+			wantNil:  false,
+			wantBool: true,
+		},
+		{
+			name:     "false is preserved as explicit false, not nil",
+			in:       false,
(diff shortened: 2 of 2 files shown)


lentzi90 (commented):
I am a bit confused about this. The issue is for v1alpha7, so to be useful, this fix would need to be targeted at a much older branch (which is no longer supported).
Is there a reason to take this in on main?

I would be more than happy to take it if there is any proof that this is an issue also for v1beta1 <->v1beta2. Then we could also confirm it with a test.

/ok-to-test

larainema:
> I am a bit confused about this. The issue is for v1alpha7, so to be useful, this fix would need to be targeted at a much older branch (which is no longer supported). Is there a reason to take this in on main?
> 
> I would be more than happy to take it if there is any proof that this is an issue also for v1beta1 <->v1beta2. Then we could also confirm it with a test.
> 
> /ok-to-test

Thanks for the review @lentzi90!

You raise a very good point. After investigating further, I found that the functions fixed in this PR (`Convert_bool_To_optional_Bool`, `Convert_optional_Bool_To_bool`, and `RestoreBool`) are actually **dead code on main**. They are not called from anywhere in the current codebase:

- The v1beta1 <-> v1beta2 conversion in `zz_generated.conversion.go` uses direct `unsafe.Pointer` casts for `optional.Bool` fields, not these helper functions
- No other code in the repo calls these functions either

So this fix wouldn't have any practical impact on current conversions.

Would you prefer I:
1. **Close this PR** since the code is dead and the original issue only applies to unsupported v1alpha7 branches
2. **Pivot to a cleanup PR** that removes the unused conversion helpers entirely

Happy to go either way!

lentzi90:
Let's close the PR for now. We may end up using the code still at some point, depending on what we do with v1beta2 :)

larainema:
Sounds good, closing this for now. Thanks for the guidance! 🙏

PR #3119, opened 2026-04-15
🌱 Fuzz conversion for v1beta1 <-> v1beta2

### LINKED ISSUE #3011: Add Fuzz tests for v1beta1 <-> v1beta2 conversions

lentzi90:

- Replace the deprecated gofuzz library with randfill, updating all fuzzer
function signatures and method calls accordingly.
- Add fuzzy conversion tests for v1beta1 API types

Fixes #3011

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90 (commented):
/hold cancel
/approve

nikParasyr (approved):
/lgtm

test/helpers/fuzzerfuncs.go
```diff
@@ -37,9 +39,20 @@ func filterInvalidTags(tags []infrav1.NeutronTag) []infrav1.NeutronTag {
 	return ret
 }
 
-func nonEmptyString(c fuzz.Continue) string {
+func filterInvalidTagsV1Beta2(tags []infrav1.NeutronTag) []infrav1.NeutronTag {
```
bnallapeta: isn't this same as `filterInvalidTags`? I think we could reuse one function without having to create v1beta2 specific?

test/helpers/fuzzerfuncs.go
```diff
@@ -130,58 +174,180 @@ func InfraV1FuzzerFuncs() []interface{} {
...
-					s := c.RandString()
-					if !strings.Contains(s, " ") && s != "" {
-						return s
-					}
-				}
+// InfraV1Beta2FuzzerFuncs returns fuzzer funcs for v1beta2 OpenStack types which:
+// * Constrain the output in ways which are validated by the API server
+// * Constrain fields that are not preserved during v1beta2 <-> v1beta1 round-trip conversion
+// * Add additional test coverage where it is not generated by the default fuzzer.
+func InfraV1Beta2FuzzerFuncs() []any {
```
bnallapeta, line 200: similar to the other one, most of the funcs here are identical copies of the ones in InfraV1FuzzerFuncs. Since slices.Concat merges both lists, the duplicates in InfraV1FuzzerFuncs end up as dead code (last registration wins). 

Could this just contain the funcs that are actually new (the OpenStackCluster/OpenStackMachine/OpenStackMachineTemplate/FailureDomain OpenStackClusterStatus normalizers)?

bnallapeta (changes requested):

lentzi90:
Good points @bnallapeta ! I sort of just assumed there would be differences, but in reality a lot of it is still the same. I'll clean it up!

test/helpers/fuzzerfuncs.go
lentzi90: This made me realize a mistake. The point here was to have different API versions for each function, but now we have the same everywhere. I'll fix this. We will have "duplication" in a way but it will be correct then with v1beta1 for the v1beta1 fuzzing and v1beta2 for the v1beta2 fuzzing

test/helpers/fuzzerfuncs.go
lentzi90: Done

test/helpers/fuzzerfuncs.go
lentzi90: Now I think this is correct. We have separate fuzzers for v1beta1 and v1beta2, so they are no longer identical and won't overwrite each other.

lentzi90:
/hold
Let's take in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/3087 first. I don't want to cause any conflicts

lentzi90:
Ok we are good, no conflict :+1: 
/hold cancel

bnallapeta (approved):
/lgtm

api/v1beta1/conversion.go
```diff
@@ -323,9 +322,10 @@ func Convert_v1beta1_OpenStackMachineSpec_To_v1beta2_OpenStackMachineSpec(
 		}
 
 	default:
-		// This should not be reached due to api validations
-		return fmt.Errorf("cannot convert OpenStackMachineSpec to v1beta2: " +
-			"neither Flavor nor FlavorID is set in v1beta1 object")
+		// This would fail the validation webhook, but we need to handle
+		// conversion anyway. The apiserver may send objects without
+		// a spec e.g. during managedField conversion.
+		// Therefore we return no error.
```
lentzi90, line 328: @nikParasyr I had to change this back to no error. Turns out there are cases when this can happen for "pure conversion", not a "real API object" that would be rejected by the webhook.
CAPI added a test for it, and we caught it here.
Ref. https://github.com/kubernetes-sigs/cluster-api/pull/13409

nikParasyr: this is good to know.
( it will probably create a conflict with https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/3122 but ill rebase as its good to have the fuzz tests there as well to check )

bnallapeta (approved):
/lgtm

PR #3120, opened 2026-04-15
🐛 Use port_trusted_vif extension for trusted VF when available

### LINKED ISSUE #2172: Port trusted VIF moved to a Neutron extension
/kind bug

In recent  Neutron,  setting ``trusted`` key directly in the port's ``binding:profile`` has been deprecated and will be forbidden in future releases.
Dedicated port's attribute ``trusted``, added by the API extension ``port_trusted_vif`` should be used instead.

New ML2 plugin extension ``port_trusted`` is now available. This extension implements the ``port_trusted_vif`` API extension which adds to the port resource a new boolean field called ``trusted``.
This field should be used by admin users to set the port as trusted what was previously possible only through the port's
``binding:profile`` dictionary.

With this extension enabled, it is now possible for admin users to set port as trusted without modifying directly 'binding:profile' field which is supposed to be just for machine to machine communication.

Source: https://review.opendev.org/c/openstack/neutron/+/926068

We need to adapt CAPO with this change bu checking if the extension is enabled.

larainema:
## What this PR does / why we need it

In recent Neutron (2024.2+), setting `trusted` directly in a port's `binding:profile` has been deprecated. A new API extension `port-trusted-vif` adds a dedicated `trusted` boolean field to the port resource, which should be used instead.

This PR adapts CAPO to use the new extension when available, with backward compatibility for older Neutron deployments.

**Which issue(s) this PR fixes:**
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2172

### Port creation (`pkg/cloud/services/networking/port.go`)

- Added `HasPortTrustedVIFExtension()` method that checks if the `port-trusted-vif` Neutron extension is available (follows the same pattern as `GetTrunkSupport()`)
- Modified `EnsurePort` to check for the extension when `TrustedVF` is set. If available, wraps the create opts with `portstrustedvif.PortCreateOptsExt` instead of putting `trusted` in `binding:profile`
- Modified `getPortProfile` to accept a `usePortTrustedVIF` flag so it skips adding `trusted` to the binding:profile when the extension handles it

### Tests (`pkg/cloud/services/networking/port_test.go`)

- Updated existing test to mock `ListExtensions` returning no `port-trusted-vif` extension (backward compatibility path)
- Added new test case "uses port_trusted_vif extension when available instead of binding:profile" that verifies the new code path

### Behavior

| Neutron version | `port-trusted-vif` available? | Behavior |
|---|---|---|
| < 2024.2 | No | `trusted: true` set in `binding:profile` (unchanged) |
| >= 2024.2 | Yes | `trusted: true` set via dedicated port attribute |

## Special notes for your reviewer

1. This PR does NOT change any image versions.
2. The gophercloud v2 library already includes the `portstrustedvif` package, so no dependency updates are needed.
3. The extension availability check uses `ListExtensions()`, the same API used for trunk support checking.

## References

- Neutron change: https://review.opendev.org/c/openstack/neutron/+/926068
- neutron-lib API definition: `ALIAS = 'port-trusted-vif'` in `neutron_lib/api/definitions/port_trusted_vif.py`
- gophercloud support: `gophercloud/v2/openstack/networking/v2/extensions/portstrustedvif`

### DIFF
--- a/pkg/cloud/services/networking/port.go
+++ b/pkg/cloud/services/networking/port.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/portsbinding"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/portsecurity"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/portstrustedvif"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/approve
/cc @mandre

smoshiur1237 (commented):
/lgtm

Issue #3131, opened 2026-04-21
Adopt kube-api-linter and fix findings


lentzi90:
/kind feature

[KAL](https://github.com/kubernetes-sigs/kube-api-linter) is a static analysis linter for Kubernetes API types. It enforces best practices from the [Kubernetes API Conventions](https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md), catching mechanical API review issues so human reviewers can focus on higher-level design.

Upstream CAPI has already adopted KAL (tracked in [kubernetes-sigs/cluster-api#11834](https://github.com/kubernetes-sigs/cluster-api/issues/11834)) and it would make sense for us to follow suit. Adopting KAL during the v1beta2 API work would be ideal.

I think we can split this up into at least three PRs:

1. Add KAL to the repo, but ignore any issues it finds
2. Fix all issues that can be automatically fixed by KAL
3. Manual changes, possibly over multiple PRs depending on findings

nikParasyr:
/remove lifecycle/stale

@lentzi90 I think the consencus here is that we are good with the changes introduced to v1beta2.
The remaining ones can be done/debated on v1beta3 which we have not planned yet.

In this case I guess its good to keep the issue open

PR #3132, opened 2026-04-22
🌱 Add Kube API Linter with everything excluded

lentzi90:

Integrate kube-api-linter (KAL) into the project's linting infrastructure
to validate Kubernetes API conventions. This adds:

- New `.golangci-kal.yml` configuration for KAL with API linters
- Build support for golangci-lint with KAL plugin
- New `lint-api` and `lint-api-fix` make targets
- KAL integrated into main `lint` and `lint-update` targets

This is mostly copied from CAPI. We may want to adjust some settings later on after evaluating.

First part of #3131

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -1,0 +1,96 @@
+version: "2"
+run:
+  go: "1.25"
+  allow-parallel-runners: true
+linters:
+  default: none
+  enable:
+  - kubeapilinter # linter for Kube API conventions
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: KAL is the Kube-API-Linter and lints Kube like APIs based on API conventions and best practices.
+        settings:
+          linters:
+            enable:
+            - "commentstart" # Ensure comments start with the serialized version of the field name.
+            - "conditions" # Ensure conditions have the correct json tags and markers.
+            - "conflictingmarkers" # Ensure there are no conflicting markers.
+            - "duplicatemarkers" # Ensure there are no exact duplicate markers for types and fields.
+            - "forbiddenmarkers" # Ensure that types and fields do not contain any markers that are forbidden.
+            - "integers" # Ensure only int32 and int64 are used for integers.
+            - "jsontags" # Ensure every field has a json tag.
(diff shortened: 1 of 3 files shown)


.golangci-kal.yml
```diff
@@ -0,0 +1,96 @@
...
+            # having the omitempty value in their json tag where appropriate.
+            - "optionalorrequired" # Every field should be marked as optional or required.
+            - "requiredfields" # Required fields should not be pointers, and should not have omitempty.
+            - "ssatags" # Ensure array fields have the appropriate listType markers.
+            - "statusoptional" # Ensure all first children within status should be optional.
+            - "statussubresource" # All root objects that have a status field should have a status subresource.
+            - "uniquemarkers" # Ensure that types and fields do not contain more than a single definition of a marker that should only be present once.
+
+            # Per discussion in July 2024 in CAPI, phase fields are kept for now.
+            # - "nophase" # Phase fields are discouraged by the Kube API conventions, use conditions instead.
```
nikParasyr, line 41: would be nice if we could enable it only on v1beta2, not sure if that is possible.

.golangci-kal.yml
```diff
@@ -0,0 +1,96 @@
...
+
+    ## Exclude all existing API versions for now.
+    ## TODO: Remove exclusions and fix issues incrementally.
+    - path: "api/v1alpha1"
+      linters:
+      - kubeapilinter
+    - path: "api/v1beta1"
+      linters:
+      - kubeapilinter
+    - path: "api/v1beta2"
```
nikParasyr, line 90: Will be interesting when we remove those.

nikParasyr (approved):
/lgtm

.golangci-kal.yml
lentzi90: Yes, we can select what paths should be included. Actually I think it will not be possible to enable it for older APIs since most of the changes would be breaking...

lentzi90 (commented):
/approve
/hold cancel

PR #3154, opened 2026-04-30
🐛 Use lazy evaluation for shell calls in `Makefile`

stephenfin:

Variables assigned using `:=` ([simply expanded variable assignment][1]) are expanded once, immediately when the line is read. This can be problematic when the shell function is used, since [that is expanded when make reads the line][2]. Combined, this means any `$(shell ...)` invocations on a `:=` line run at parse time rather than execution time.

Avoid this by assigning these variables using using `=` ([recursively expanded variable assignment][3]), which causes them to be evaluated lazily. This allows us to tab-complete targets without a 2-3 second delay.

[1]: https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html
[2]: https://www.gnu.org/software/make/manual/make.html#Shell-Function
[3]: https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html

n/a

None.

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ BIN_DIR := bin
 TOOLS_DIR_DEPS := $(TOOLS_DIR)/go.sum $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/Makefile
 TOOLS_BIN_DIR := $(TOOLS_DIR)/$(BIN_DIR)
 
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #3155, opened 2026-04-30
:warning: Rename managedSecurityGroups.allNodesSecurityGroupRules

### LINKED ISSUE #3055: ✨ v1beta2: Rename allNodesSecurityGroupRules to indicate that it only applies for control-plane/worker nodes
/kind feature
/assign

Currently we have [`OSC.spec.managedSecurityGroups.AllNodesSecurityGroupRules`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1beta1/openstackcluster_types.go#L349). This is somewhat problematic because:
- users sometimes expect that bastion security groups should also be part of this (related ticket: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2981)
- introduction of functionality to manage bastion security group rules would make it a bit weird

The intention is to **just rename** `allNodesSecurityGroupRules` to something more clear such as `clusterNodesSecurityGroupRules` (open to suggestions). 
This will:
- clarify to users the scope of this field
- allow **future** addition of `bastionNodeSecurityGroupRules` without weird naming ( im personally interested in this feature but implementation of it has low-prio and will be outside of the scope of this issue, it can added in the future without breaking api changes afaik )

nikParasyr:
Rename `spec.managedSecurityGroups.allNodesSecurityGroupRules` to `spec.managedSecurityGroups.clusterNodesSecurityGroupRules` to clarify that these rules apply only to cluster nodes (control plane and workers), and not to other managed resources such as the bastion host.

Fixes #3055 

**TODOs**:

- [X] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold


pkg/webhooks/openstackcluster_webhook.go
```diff
@@ -273,7 +273,7 @@ func securityGroupRemoteFields(r *infrav1.SecurityGroupRuleSpec) (bool, bool, bo
 func validateManagedSecurityGroupRules(msg *infrav1.ManagedSecurityGroups) field.ErrorList {
 	var allErrs field.ErrorList //nolint:prealloc // We can't know the number of errors upfront
 	allErrs = append(allErrs, validateSecurityGroupRulesRemoteMutualExclusion(
-		msg.AllNodesSecurityGroupRules,
+		msg.ClusterNodesSecurityGroupRules,
 		field.NewPath("spec", "managedSecurityGroups", "allNodesSecurityGroupRules"),
```
bnallapeta: we need to replace `ClusterNodesSecurityGroupRules` here too?

bnallapeta (changes requested):
`allNodesSecurityGroupRules`
  - the same rename change has to occur in e2e and docs too.

pkg/webhooks/openstackcluster_webhook.go
nikParasyr: I believe so yes, updated L277 as well to set it correctly.

nikParasyr:
> the same rename change has to occur in e2e and docs too.

@bnallapeta 
e2e tests were updated already afaik with the exception of the v1beta1 template.
docs were indeed missing updates which i have now added

lentzi90 (commented):
/approve

lentzi90:
@bnallapeta can you make a "proper review comment" so the requested changes clears? I.e. go to the changed files and put the comment/approval through that.

lentzi90:
Ah it was just the hold :sweat_smile: 
I didn't realize prow doesn't care about the requested changes :joy:

PR #3169, opened 2026-05-18
:warning: Group API server fields under spec.apiServer

### LINKED ISSUE #2899: :sparkles: v1beta2 Group apiServer related fields on OpenStackCluster
/kind feature

Group api server related fields

So currently we have a lot of fields for the apiServer that are not grouped.
```
Kind: OpenStackCluster
...
spec:
  ...
  apiServerFixedIP: string
  apiServerFloatingIP: string
  apiServerPort: string
  disableAPIServerFloatingIP: bool
  apiServerLoadBalancer:
    ...lots of things here...
  ...
```

We need to clarify how the above should be best group because each one of them:
- may apply only for the managed LB <- should be part of the apiServerLoadBalancer object
- may apply only when LB is disabled <- need to think if they can be grouped on a seperateObject or not
- may apply on both <- need to think if we should seperate them or not etc

Implementation should be done after core v1beta2 is implemented in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2895

nikParasyr:
Consolidate the various flat API server fields (apiServerFloatingIP, apiServerFixedIP, apiServerPort, disableAPIServerFloatingIP, and apiServerLoadBalancer) into a single structured spec.apiServer object, with apiServerLoadBalancer renamed to managedLoadBalancer within it. Includes conversion from v1beta1 and updated webhook validation.

**Which issue(s) this PR fixes** :
Fixes #2899 

1. ~~Not sure whether [OSC.status](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/79731cc600112fe912c2f49fafb8da4d09407516/api/v1beta2/openstackcluster_types.go#L229) should be updated as well to reflect this change. Open to suggestions~~ => addressed

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold


nikParasyr:
@lentzi90 && @bnallapeta, this is ready from my side

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -352,17 +352,17 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
 			}
 
 			for _, v := range flavors {
-				if v.Enabled && v.Name == *openStackCluster.Spec.APIServerLoadBalancer.Flavor {
+				if v.Enabled && v.Name == *openStackCluster.Spec.APIServer.ManagedLoadBalancer.Flavor {
```
bnallapeta: technically, although this is the same, I think we should change this to lbSpec.Flavour to maintain consistency with the rest of the code.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -352,17 +352,17 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
...
 					lbFlavorID = v.ID
 					break
 				}
 			}
 			if lbFlavorID == "" {
-				record.Warnf(openStackCluster, "OctaviaFlavorNotFound", "Flavor %s specified for Octavia not found, using the default flavor.", *openStackCluster.Spec.APIServerLoadBalancer.Flavor)
+				record.Warnf(openStackCluster, "OctaviaFlavorNotFound", "Flavor %s specified for Octavia not found, using the default flavor.", *openStackCluster.Spec.APIServer.ManagedLoadBalancer.Flavor)
```
bnallapeta: Same as above.

pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -352,17 +352,17 @@ func (s *Service) getOrCreateAPILoadBalancer(openStackCluster *infrav1.OpenStack
...
 			}
 		}
 
-		availabilityZone = openStackCluster.Spec.APIServerLoadBalancer.AvailabilityZone
+		availabilityZone = openStackCluster.Spec.APIServer.ManagedLoadBalancer.AvailabilityZone
```
bnallapeta: Same as above.

bnallapeta (commented):
Looks good overall. Minor comment.

Also, on the status question, I think its best to change the below to maintain consistency. If we don't do this, we might carry the inconsistency in naming in the long run.

`Status.APIServerLoadBalancer → Status.ManagedAPIServerLoadBalancer`

templates/clusterclass-dev-test.yaml
```diff
@@ -223,7 +223,7 @@ spec:
           infrastructureCluster: true
       jsonPatches:
       - op: add
-        path: /spec/template/spec/apiServerLoadBalancer/allowedCIDRs
+        path: /spec/template/spec/apiServer.managedLoadBalancer/allowedCIDRs
```
lentzi90:
```suggestion
        path: /spec/template/spec/apiServer/managedLoadBalancer/allowedCIDRs
```

controllers/openstackmachine_controller.go
```diff
@@ -306,7 +306,7 @@ func removeAPIServerEndpoint(scope *scope.WithLogger, openStackCluster *infrav1.
 	// floating IP to be created and deleted with the cluster. And if the
 	// delete behaviour is correct, we leak it if the instance was
 	// previously deleted.
-	if openStackCluster.Spec.APIServerFloatingIP == nil && instanceStatus != nil {
+	if (openStackCluster.Spec.APIServer == nil || openStackCluster.Spec.APIServer.FloatingIP == nil) && instanceStatus != nil {
```
lentzi90: I think we could use the getter here, right?
```suggestion
	if openStackCluster.Spec.APIServer.GetFloatingIP() == nil && instanceStatus != nil {
```

kustomize/without-lb/patch-without-lb.yaml
```diff
@@ -3,7 +3,7 @@ kind: OpenStackCluster
 metadata:
   name: ${CLUSTER_NAME}
 spec:
-  apiServerLoadBalancer:
+  apiServer:
```
lentzi90, line 6: I guess this works, but shouldn't we go one level down really to get the same as it was before, with the loadbalancer only?

lentzi90 (commented):
Great stuff!
I have basically only one thing that needs fixing. The other two are just minor things.
We probably don't test that specific jsonpatch in the e2e tests so this is why it wasn't detected.

nikParasyr:
@lentzi90 && @bnallapeta:
I think ive addressed all your remarks ( amended on the first commit ).
I also made a second commit the the status change ( `apiServerLoadBalancer => apiServerManagedLoadBalancer`)

lentzi90 (commented):
/approve

PR #3183, opened 2026-05-27
🌱 Enable optionalorrequired linter and add required/optional markers

lentzi90:

This commit enables the `optionalorrequired` linter in `.golangci-kal.yml` and updates all API type definitions in v1beta2 to use the standardized `+required` and `+optional` markers instead of `+kubebuilder:validation:Required`.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/api/v1beta2/types.go
+++ b/api/v1beta2/types.go
@@ -27,11 +27,13 @@ import (
 // OpenStackMachineTemplateResource describes the data needed to create a OpenStackMachine from a template.
 type OpenStackMachineTemplateResource struct {
 	// spec is the specification of the desired behavior of the machine.
+	// +required
 	Spec OpenStackMachineSpec `json:"spec"`
 }
 
 type ResourceReference struct {
 	// name is the name of the referenced resource
+	// +required
 	Name string `json:"name"`
 }
 
@@ -110,8 +112,10 @@ func (f *FlavorFilter) IsZero() bool {
 
 type ExternalRouterIPParam struct {
 	// fixedIP is the FixedIP in the corresponding subnet.
+	// +optional
 	FixedIP string `json:"fixedIP,omitempty"`
 	// subnet is the subnet in which the FixedIP is used for the Gateway of this router.
+	// +required
 	Subnet SubnetParam `json:"subnet"`
 }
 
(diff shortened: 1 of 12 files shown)


api/v1beta2/types.go
```diff
@@ -445,12 +469,15 @@ type ResolvedPortSpecFields struct {
 // ResolvedPortSpec is a PortOpts with all contained references fully resolved.
 type ResolvedPortSpec struct {
 	// name is the name of the port.
+	// +required
 	Name string `json:"name"`
 
 	// description is a human-readable description for the port.
+	// +required
```
lentzi90: This is perhaps the most controversial thing in the PR. We don't have `omitempty`, so I am making it required. If the field was set by humans I would do it the other way, but this is always set by the controller so I think it makes more sense this way.

nikParasyr: The concern I have with required here is that it's a schema-level constraint, but "the controller always sets this" is a runtime guarantee. Those two things operate at different levels. Specifically:
1. Required fields can cause validation errors between object creation and the first reconcile
2. If the controller ever fails mid-reconcile and writes a partial status, required fields could cause issues
3. If at the future for whatever reason we need to change the controller behavior, we will have to go through an api change.

For this specific change, I don't think we are hit by 1 + 2 as this is a sub-struct under status that is not required and I guess all those fields are set by OpenStack once a port is created and therefore the controller can fully populate them, but i do think there might be some edge-cases that we are hit by this. ( Also `macAddress` below is already set to `optional`, I'm wondering why. )

I was looking a bit into k8s/capi best-practises/conventions that go over similar stuff. I found [this](https://github.com/kubernetes-sigs/cluster-api/commit/5b06a59b5fdcda4d9b975b2e859a6e68340fdbd6#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L387) once populated on capi's conventions.

In general I'm not fully confident i can evaluate this and the impact it might have. It feels to me that opting for `required` is a bit "strict" and could cause problems down the road, but I dont have the experience/knowledge to really substantiate it.

nikParasyr (commented):
A bit hesitant to lgtm this one. made a comment with my concerns

api/v1beta2/types.go
lentzi90: Thanks for the input @nikParasyr ! I really appreciate it and I think you pointed out two good reasons to switch this to optional.
1. CAPI guidance that status fields MUST be optional
2. the MAC address below

With this in mind I think it was a mistake that this was missing `omitempty`. We should simply make it optional.

lentzi90: Now changed to optional. The `omitempty` will be required/added when we enable the `optionalfields` linter.

nikParasyr (approved):
/lgtm

lentzi90 (commented):
/approve
/hold cancel

PR #3185, opened 2026-05-29
🌱 Patch govulncheck to support ignoring findings

lentzi90:

This is taking the same patch that CAPI is using. It will allow us to ignore findings that are not relevant or fixable. This way we can keep the signal free from noise.

Ref: https://github.com/kubernetes-sigs/cluster-api/pull/13669

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


lentzi90 (commented):
/approve

nikParasyr (approved):
A bit of an unorthodox approach but i think its safe to follow capi's lead

/lgtm

lentzi90:
Yes I agree. Will try to keep an eye on how it evolves. Maybe we can get the patch upstreamed at some point...
/hold cancel

PR #3189, opened 2026-06-01
:warning: V1beta2 Switch to positive polarity for bool fields

### LINKED ISSUE #2974: ✨ v1beta2: Switch to positive polarity for bool fields in OpenStackCluster
/kind feature

We have multiple boolean fields in `OpenStackCluster` with "negative polarity". Specifically:
```
...
spec:
  disableAPIServerFloatingIP: bool
  disablePortSecurity: bool 
  disableExternalNetwork: bool
```
In general negative polarity is discouraged as it is more difficult to understand double negatives when reading manifests and/or code. For example, `disableX: false` is harder to understand than `enableX: true`. 

This change might be tricky as we have to ensure that conversions can be done properly and user intent is not lost (depends on whether bool/*bool/optional.Bool is used).

Implementation should be done after core v1beta2 is implemented in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2895

nikParasyr:
Switch the polarity of various fields from `disableX` to `enableX`

Fixes #2974 

1. I have done a seperate commit per change, this was to simplify development and (hopefully) reviewing  since i wanted to be extra careful when flipping various "ifs" and not make a mistake there. If preferred i can squash commits before merging
2. Please doublecheck the various spots where i had to change "ifs" etc as a mistake there could alter the behavior of CAPO quite a bit :yum: => edit: had to fix some of them already due to unit tests failing so...

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [X] includes documentation
  - [X] adds unit tests

/hold

### DIFF
--- a/api/v1beta2/openstackcluster_types.go
+++ b/api/v1beta2/openstackcluster_types.go
@@ -31,7 +31,7 @@ const (
 
 // OpenStackClusterSpec defines the desired state of OpenStackCluster.
 // +kubebuilder:validation:XValidation:rule="has(self.enableExternalNetwork) && !self.enableExternalNetwork ? !has(self.bastion) || !has(self.bastion.floatingIP) : true",message="bastion floating IP cannot be set when enableExternalNetwork is false"
-// +kubebuilder:validation:XValidation:rule="has(self.enableExternalNetwork) && !self.enableExternalNetwork ? has(self.apiServer) && has(self.apiServer.disableFloatingIP) && self.apiServer.disableFloatingIP : true",message="apiServer.disableFloatingIP cannot be false when enableExternalNetwork is false"
+// +kubebuilder:validation:XValidation:rule="has(self.enableExternalNetwork) && !self.enableExternalNetwork ? has(self.apiServer) && has(self.apiServer.enableFloatingIP) && !self.apiServer.enableFloatingIP : true",message="apiServer.enableFloatingIP cannot be true when enableExternalNetwork is false"
 type OpenStackClusterSpec struct {
 	// managedSubnets describe OpenStack Subnets to be created. Cluster actuator will create a network,
 	// subnets with the defined CIDR, and a router connected to these subnets. Currently only one IPv4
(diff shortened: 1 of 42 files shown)


nikParasyr:
@lentzi90 && @bnallapeta this is ready from my side

api/v1beta2/openstackcluster_types.go
```diff
@@ -74,25 +74,25 @@ type OpenStackClusterSpec struct {
...
 	//
 	// If ExternalNetwork is not defined or is empty the controller will use any
 	// existing external network as long as there is only one. It is an
 	// error if ExternalNetwork is not defined and there are multiple
-	// external networks unless DisableExternalNetwork is also set.
+	// external networks unless EnableExternalNetwork is also set to false.
 	//
 	// If ExternalNetwork is not defined and there are no external networks
-	// the controller will proceed as though DisableExternalNetwork was set.
+	// the controller will proceed as though EnableExternalNetwork was set to true.
```
lentzi90: I think this should be the other way. If we don't find any external networks, we continue as if external networks were disabled -> enableExternalNetwork: false.
```suggestion
	// the controller will proceed as though EnableExternalNetwork was set to false.
```

lentzi90 (commented):
Thank you for this!
I first checked commit by commit, and found a couple of logic flaws, but they were corrected in the last commit, so never mind this.
I like the changes to external network and floating IP. These feels easier to read now.
However, the port security change does not feel right to me. Even though the default is that security is enabled, it still feels like the naming implies I have to take manual action to configure things securely.
What do you think about leaving `disablePortSecurity` as it was?

nikParasyr:
> However, the port security change does not feel right to me. Even though the default is that security is enabled, it still feels like the naming implies I have to take manual action to configure things securely.
What do you think about leaving disablePortSecurity as it was?

Hmmmm. Fair point. The "implies I have to act to stay secure" feeling makes sense.

My preference is still positive polarity for readability and consistency. Since it's a *bool, nil keeps the secure default either way, so the naming doesn't change behavior. One extra nudge: Neutron's own API is positive too (port_security_enabled, defaults to true [1](https://docs.openstack.org/api-ref/network/v2/?expanded=create-network-detail#networks), [2](https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#ports) ), so disablePortSecurity actually inverts upstream rather than matching it.

Open to reverting to `disablePortSecurity` though if you feel strongly about it.

( i will address the rest of the remarks once we resolve this point )

lentzi90:
> One extra nudge: Neutron's own API is positive too (port_security_enabled, defaults to true [1](https://docs.openstack.org/api-ref/network/v2/?expanded=create-network-detail#networks), [2](https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#ports) )

Ok, this did it for me. I honestly thought it was negative polarity upstream. Then let's keep the positive polarity :slightly_smiling_face:

api/v1beta2/openstackcluster_types.go
nikParasyr: good catch. fixed

lentzi90 (commented):
/approve

lentzi90:
```
failed resolving release url "{go://sigs.k8s.io/cluster-api@v1.12}": error getting file, status code: 504
```
We may just have to wait for things to settle :slightly_frowning_face:

nikParasyr:
@bnallapeta this needs a /lgtm again. update were the remarks from Lennart

PR #3210, opened 2026-06-11
✨Feature/multiple subnets

### LINKED ISSUE #3160: Support multiple IPv4 network on OpenstackCluster
I recently reached the max number of available IPs from a subnet. Then I added a new subnet inside the same network and started to see errors like that in reconciliation process of capo-controller-manager:

err="multiple IPv4 Subnet is not allowed on OpenstackCluster"

I could be able to provision new worker/controller nodes with the new subnet by modifying OpenstackMachineTemplate and setting  the  "ports" section with "fixedIPs" using the ID of the new subnet, but the problem with reconciliation on capo-controller-manager logs persists.

According to the documentation, it's possible to have two subnets max but this must be present on OpenstackCluster at the time of its creation and must be one IPv4 subnet and the other must be IPv6.

Why we can't have more than one IPv4 subnet on OpenstackCluster if we can add multiple IPv4 subnets inside Openstack?

It would be nice to support multiple IPv4 subnets transparently to have better handling on that IP exhaustion scenarios.

memorais:

This creates the possibility of use multiple subnets inside one network. OpenStack provides this functionality and we need this to not break reconciliation process if OpenStack administrators add new subnet inside a network already used by OpenStack cluster-api provider.

Fixes #3160 

**TODOs**:

- [X] squashed commits
- if necessary:
  - [ ] includes documentation
  - [X] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -697,7 +697,14 @@ func openStackMachineSpecToOpenStackServerSpec(openStackMachineSpec *infrav1.Ope
 	}
 
 	var clusterSubnets []infrav1.FixedIP
-	if len(openStackCluster.Spec.Subnets) > 0 {
+	if openStackCluster.Spec.PrimarySubnet != nil {
+		// When a primary subnet is set, restrict node ports to that subnet only.
+		// This allows an administrator to direct new nodes to a specific subnet,
+		// e.g. when migrating away from an exhausted subnet.
+		clusterSubnets = []infrav1.FixedIP{
+			{Subnet: openStackCluster.Spec.PrimarySubnet},
+		}
+	} else if len(openStackCluster.Spec.Subnets) > 0 {
 		clusterSubnets = make([]infrav1.FixedIP, len(openStackCluster.Spec.Subnets))
 		for idx, sn := range openStackCluster.Spec.Subnets {
 			clusterSubnets[idx] = infrav1.FixedIP{
--- a/pkg/cloud/services/loadbalancer/loadbalancer.go
+++ b/pkg/cloud/services/loadbalancer/loadbalancer.go
@@ -55,6 +55,29 @@ const (
 	poolMemberProvisioningStatusActive          = "ACTIVE"
 )
 
+// getPrimarySubnetID returns the ID of the primary cluster subnet for use in
+// load balancer VIP allocation and member registration.
+// If PrimarySubnet is specified in the cluster spec, it is resolved via
+// GetNetworkSubnetByParam scoped to the cluster network, which validates
+// membership and supports both id and filter forms of SubnetParam.
+// Otherwise the first subnet in the cluster status is used.
+func (s *Service) getPrimarySubnetID(openStackCluster *infrav1.OpenStackCluster) (string, error) {
+	subnets := openStackCluster.Status.Network.Subnets
+	if len(subnets) == 0 {
+		return "", fmt.Errorf("no subnets available in cluster status")
+	}
+
+	if openStackCluster.Spec.PrimarySubnet == nil {
+		return subnets[0].ID, nil
+	}
+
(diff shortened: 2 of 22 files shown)


pkg/cloud/services/loadbalancer/loadbalancer.go
```diff
@@ -55,6 +55,27 @@ const (
...
+	subnets := openStackCluster.Status.Network.Subnets
+	if len(subnets) == 0 {
+		return "", fmt.Errorf("no subnets available in cluster status")
+	}
+
+	if openStackCluster.Spec.PrimarySubnet == nil {
+		return subnets[0].ID, nil
+	}
+
+	subnet, err := s.networkingService.GetSubnetByParam(openStackCluster.Spec.PrimarySubnet)
```
bnallapeta: the empty networkID in this function means we resolve primarySubnet against every subnet in the cloud, with no check that its actually one of the cluster's subnets or even in the cluster network. This is risky.

Could we instead match primarySubnet against Status.Network.Subnets? This way, it validates membership and avoids the extra API call on every reconcile.

controllers/openstackmachine_controller.go
```diff
@@ -697,7 +697,14 @@ func openStackMachineSpecToOpenStackServerSpec(openStackMachineSpec *infrav1.Ope
...
 	var clusterSubnets []infrav1.FixedIP
-	if len(openStackCluster.Spec.Subnets) > 0 {
+	if openStackCluster.Spec.PrimarySubnet != nil {
+		// When a primary subnet is set, restrict node ports to that subnet only.
+		// This allows an administrator to direct new nodes to a specific subnet,
+		// e.g. when migrating away from an exhausted subnet.
+		clusterSubnets = []infrav1.FixedIP{
+			{Subnet: openStackCluster.Spec.PrimarySubnet},
+		}
+	} else if len(openStackCluster.Spec.Subnets) > 0 {
```
bnallapeta, line 707: it says for 'migrating away from an exhausted subnet', but that's a runtime action and the field is effectively immutable. The cluster webhook will check every mutable field and then does `reflect.DeepEqual()`. `primarySubnet` is not in that list, so any change to it after creation gets rejected. In order words, this can only be set at create time, which defeats the migration use case. 

We need to add `primarySubnet` to the mutable fields in `ValidateUpdate` function.

bnallapeta (changes requested):

pkg/cloud/services/loadbalancer/loadbalancer.go
memorais: I think I can create a new function (matchPrimarySubnet) instead of calling GetSubnetByParam, which can do a scan on Status.Network.Subnets by ID and return an error if the ID is not in the cluster own subnets. What do you think?

bnallapeta: Yes, I think that would work too. But, since primarySubnet is a SubnetParam, it supports filter as well as id, and we should keep that parity with the rest of the API. A by-ID scan over status can't evaluate the filter (please verify, but I think this only has name/id/tags). 

I think we need to resolve via GetNetworkSubnetByParam(openStackCluster.Status.Network.ID,  primarySubnet), by passing network ID instead of the current "". This gives us id + filter support, and network scoping.

bnallapeta (approved):
/lgtm

bnallapeta:
@lentzi90 PTAL from your end as well.

lentzi90 (commented):
/approve

lentzi90:
Please squash the commits

memorais:
> Please squash the commits

Sorry, I thought that I should squash commits. Should I revert?

lentzi90:
You did right. It removes lgtm when there are any changes. That is expected.
I added the merge-method-squash label because I thought we could automatically squash on merge. Seems like that is disabled, so it complains about it. I'll remove it again now that you squashed :slightly_smiling_face:

PR #2254, opened 2024-11-14
🐛 Requeue when waiting for a server to be deleted

lentzi90:
**What this PR does / why we need it**:

We did not add a requeue when waiting for a server to be deleted. This leads to unnecessary delays when deleting a cluster, since the controller will then wait for the next sync period (default 10m) before it checks again.
This PR adds a requeue similar to what we have in other situations where we expect progress to happen soon.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

PR #3211, opened 2026-06-12
🌱 Replace requeue delay with watch

stephenfin:

Commit 186c11d2e (#2182) added a new OpenStackServerReconcileComplete predicate in order to reduce reconcile churn from many updates to the server. However, the predicate only watched for Ready or error transitions, meaning delete events were no longer delivered. This was addressed in commit 050875940 (#2254), where we added a RequeueAfter to ensure the OpenStackMachine controller polls for an OpenStackServer's disappearance before deleting the OpenStackMachine.

However, this is less efficient (and more complicated) than it can be: the delay means we will only kick the OpenStackMachine deletion on multiples of 10 second, rather than immediately like we would with a watch. Address this by modifying the predicate so that it will also watch for delete events, ensuring we proceed with OpenStackMachine deletion immediately upon OpenStackServer's removal and allowing us to remove the delay.

(Note that I spotted this while trying to debug another, unrelated issue in a downstream product. I believe it's a nice to have rather than a bug)

(n/a)

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/controllers/openstackmachine_controller.go
+++ b/controllers/openstackmachine_controller.go
@@ -69,7 +69,6 @@ const (
 	waitForClusterInfrastructureReadyDuration = 15 * time.Second
 	waitForInstanceBecomeActiveToReconcile    = 60 * time.Second
 	waitForBuildingInstanceToReconcile        = 10 * time.Second
-	deleteServerRequeueDelay                  = 10 * time.Second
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
Nice!
/approve

nikParasyr (approved):
/lgtm

PR #3221, opened 2026-06-16
🌱 Enable requiredfields linter

lentzi90:

Ensure that required fields are not pointers and that the empty values are omitted in json representations.

NOTE: I changed AllowAllInClusterTraffic to optional. The linter suggested to make it a pointer since it had a valid zero value (false). However, in practice this field is optional. If we leave it out, it takes the default value (false). Making it optional doesn't really change anything and is a far smaller/easier fix than making it a pointer.

Fixes #

I thought this would be a breaking PR, but I think actually there were no breaking changes in this. Let me know if you think otherwise. Maybe I missed something?

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -31,7 +31,7 @@ linters:
             # - "optionalfields" # Ensure that all fields marked as optional adhere to being pointers and
             # having the omitempty value in their json tag where appropriate.
             - "optionalorrequired" # Every field should be marked as optional or required.
-            # - "requiredfields" # Required fields should not be pointers, and should not have omitempty.
+            - "requiredfields" # Required fields should have omitempty/omitzero, and should only be pointers when the zero value is valid.
             # - "ssatags" # Ensure array fields have the appropriate listType markers.
             - "statusoptional" # Ensure all first children within status should be optional.
             - "statussubresource" # All root objects that have a status field should have a status subresource.
--- a/api/v1beta2/openstackcluster_types.go
+++ b/api/v1beta2/openstackcluster_types.go
@@ -156,7 +156,7 @@ type OpenStackClusterSpec struct {
 	// to be used when reconciling this cluster. It is also to reconcile
 	// machines unless overridden in the machine spec.
 	// +required
-	IdentityRef OpenStackIdentityReference `json:"identityRef"`
+	IdentityRef OpenStackIdentityReference `json:"identityRef,omitzero"`
 }
 
 type APIServer struct {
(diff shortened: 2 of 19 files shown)


lentzi90:
Part of #3131

api/v1beta2/openstackcluster_types.go
```diff
@@ -340,8 +340,8 @@ type ManagedSecurityGroups struct {
 
 	// allowAllInClusterTraffic allows all ingress and egress traffic between cluster nodes when set to true.
 	// +kubebuilder:default=false
-	// +required
-	AllowAllInClusterTraffic bool `json:"allowAllInClusterTraffic"`
```
lentzi90, line 352: Note this change! As mentioned in the description, I think it makes more sense to make it optional instead of changing it to pointer. In practice it was already optional since kubebuilder applies the default.

nikParasyr: this feels right (would also been picked up by the `defaultOrRequired` linter)

.golangci-kal.yml
```diff
@@ -31,7 +31,7 @@ linters:
             # - "optionalfields" # Ensure that all fields marked as optional adhere to being pointers and
             # having the omitempty value in their json tag where appropriate.
             - "optionalorrequired" # Every field should be marked as optional or required.
-            # - "requiredfields" # Required fields should not be pointers, and should not have omitempty.
+            - "requiredfields" # Required fields should not be pointers, and should not have omitempty.
```
nikParasyr: Nit: the comment is different than what is documented here: https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#requiredfields (actually the opposite for `omitempty/zero`) so its a bit confusing

```suggestion
            - "requiredfields" # Required fields should have omitempty/omitzero, and should only be pointers when the zero value is valid.
```

nikParasyr (changes requested):
I have a small nit, other than that it feels correct (quite an excercise reviewing this and understanding the logic of the linter)...

.golangci-kal.yml
lentzi90: Oh good catch! I copied from CAPI without checking :see_no_evil: 
https://github.com/kubernetes-sigs/cluster-api/blob/44506b37ab04dadc670cbe5463618eeced7d76bc/.golangci-kal.yml#L34

lentzi90: Fixed

nikParasyr (approved):
/lgtm

docs/book/src/api/v1beta2/api.md
```diff
@@ -331,7 +331,7 @@ to first set <code>enabled: false</code> which will remove the bastion and then
 </tr>
 <tr>
 <td>
-<code>identityRef</code><br/>
+<code>identityRef,omitzero</code><br/>
 <em>
```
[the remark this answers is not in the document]
lentzi90: Indeed. This is something we should fix separately. Created an issue for it: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/3222

lentzi90 (commented):
/approve
/hold cancel

PR #3224, opened 2026-06-18
✨ Enable ssatags linter and fix findings

lentzi90:

Add SSA tags on all fields. Enable the linter to ensure we enforce this going forward.

Part of #3131

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


api/v1beta2/types.go
```diff
@@ -346,12 +346,14 @@ type SubnetSpec struct {
 
 	// dnsNameservers holds a list of DNS server addresses that will be provided when creating
 	// the subnet. These addresses need to have the same IP version as CIDR.
+	// +listType=set
```
nikParasyr: I'm not sure this should be a `set`. 
I think neutron treats this as an ordered list and delivers the dns server via DHCP in the same order, these 2 ancient pages/bug go over it:
- https://bugs.launchpad.net/neutron/+bug/1218629
- https://docs.openstack.org/neutron/ocata/devref/dns_order.html

It then depends on the VM OS on how it treats this list ( `rotate` in place or whatever ) but we dont have much control there.

So `semantically` the order here matters (even though i wonder if it will create a real issue on capo side) and thus `+listType=atomic` might be better

(dont ask me why i've spent so much time on how neutron handles the dns_nameservers...)

nikParasyr (commented):
One remark. the rest look good

api/v1beta2/types.go
lentzi90: Thank you for spending that time :smile: 
We don't lose much by going with atomic honestly, so I'll change it. It is not like someone will have lists of dozens of nameservers and try to merge them :joy:

lentzi90: Fixed

lentzi90 (commented):
/approve

PR #3226, opened 2026-06-18
⚠️ Enable optionalfields linter and fix findings

lentzi90:

This enables the [optionalfields](https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#optionalfields) linter. I am not completely sure this is something we need/want yet but I am creating the PR still for discussion and testing.

Note that this changes the OpenStackMachine and OpenStackCluster spec fields to be required. These objects do not make sense without spec. The bastion on the other hand still has an optional spec, since it can be disabled.

Fixes #

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold


.golangci-kal.yml
```diff
@@ -85,6 +85,28 @@ linters:
...
 
+    ## CAPI contract fields: Provisioned bools cannot be changed.
+    - text: "field (ClusterInitialization|MachineInitialization)\\.Provisioned"
+      linters:
+      - kubeapilinter
+
+    ## Bool fields that are established patterns. The zero value (false) is the intended default.
+    - text: "field (OpenStackMachineSpec\\.Trunk|ManagedSecurityGroups\\.AllowAllInClusterTraffic)"
+      linters:
+      - kubeapilinter
```
lentzi90, line 96: These were accidentally included. They are for the nobools linter.

api/v1beta2/types.go
nikParasyr, line 1090: Shouldn't this be converted to `*int32` as well?

api/v1beta1/conversion.go
```diff
@@ -182,19 +183,237 @@ func (dst *OpenStackMachineTemplateList) ConvertFrom(srcRaw ctrlconversion.Hub)
...
+		out.Network = *(*infrav1.NetworkStatusWithSubnets)(unsafe.Pointer(in.Network))
+	}
+	if in.ExternalNetwork != nil {
+		out.ExternalNetwork = *(*infrav1.NetworkStatus)(unsafe.Pointer(in.ExternalNetwork))
+	}
+	if in.Router != nil {
+		out.Router = *(*infrav1.Router)(unsafe.Pointer(in.Router))
+	}
+	if in.APIServerLoadBalancer != nil {
+		out.APIServerManagedLoadBalancer = *(*infrav1.LoadBalancer)(unsafe.Pointer(in.APIServerLoadBalancer))
```
nikParasyr: Not sure if this is fully safe, since we are changing a lot of pointers to structs which change the memory footprint/layout

nikParasyr (commented):
couple of comments inline.

My thoughts on the impact of this:
1. The change from pointers to structs with omitzero => This feels fine and the more modern approach. I dont believe it will impact users (except those that import capo for runtimeSDK or other reasons)
2. The addtion of `MinLength` to multiple structs/strings/... => **This can impact users who have set that as empty before**. Autoconversion wont work afaik as admission will be blocked. This will mean that impacted users will have to rollout a fix on v1beta1 before upgrading (if it is on OSM this is "just" a new node rollout, if its OSC and immutable it gets tricky and shady). We were hit by this actually downstream when capi made similar changes from their v1beta1 to v1beta2.
3. There are some other changes that cant be grouped but they also seemed to not have much impact
4. This change makes the whole v1beta2 change much more heavy. We are not rolling "just" capi v1beta2 support + some readability/consistency changes anymore
---
My general feeling:
The changes the kal-linter proposes make sense in general. They might over-fit some edge cases here and there, but on the whole they're reasonable. In the long run I think the linter will really pay off in keeping the API clean and consistent — especially as the API gets extended, it'll enforce that consistency automatically rather than relying on a reviewer to catch all these details by hand (which is genuinely hard to do reliably). So from that angle I like it, including the specific changes in this PR.
What feels a bit "annoying" is that adopting it now forces a batch of changes whose impact/consequences we can't fully evaluate up front(or at least i cant). On a brand-new controller I'd be 100% in favor of turning this on from day one; adopting it midway is trickier, because we're retrofitting an established, in-use API.

So for me the decision boils down to:
Do we want to increase the consistency, disambiguation, hardening, and convention-enforcement of our API now — and gain a mechanism that keeps it that way going forward — accepting a one-time cost of changes we can't fully derisk? Or do we defer, keep the API as-is for now, and revisit?

I think if we can get multiple pair of eyes on this one, we may lower the risk considerably and be able to push through. I wouldn't base it just on my review, as i dont feel fully confident that i can understand the impact of all these changes fully

lentzi90:
Thanks for the comments @nikParasyr ! I very much agree with you.
On point two, I was also hit by this change in CAPI. It is good that you point it out. We need to be careful and make sure we understand the impact.

@stephenfin @mandre @EmilienM  @bnallapeta @smoshiur1237 
Please, if you have some time, skim through the changes here and tell me what you think. Is it worth it or not to try to take this in before the next release? Should we rather avoid the risk and leave it out? Since these are breaking changes, we would need a new API version if we ever want them.

/retest

api/v1beta1/conversion.go
lentzi90: Fixed

api/v1beta2/types.go
lentzi90: From the linters perspective this is fine as it is because it doesn't have a valid zero value. The minimum is 1, so 0 means unset. The other fields allow zero as value, but I am honestly not sure they should...

bnallapeta:
As noted, the changes look good in general and definitely a good improvement to the overall code base. 

I would recommend that we don't do this in the current release. 
- Don't want it to feel rushed. 
- These changes warrant a new API version.
- Would need careful migration/upgrade docs for existing users and chances of something going wrong is legit.

If required, maybe even split this into two PRs where one would cover the non-breaking changes like (minLength and pointer->value) and the other covering the breaking spec changes (optional->required)

I think we should do this in next release.

lentzi90:
Sounds like we do not want to take this for v0.15. I just want to point out then that if/when we do decide to take it, that will mean another API version (v1beta3 then I guess). I am not sure we want to take on that already for v0.16 but let's see when we get there.

lentzi90:
Closing for now. It will be easier to re-create this from scratch if/when we revisit.

PR #3236, opened 2026-06-24
🌱 Verify cloudbuild image exists and works

lentzi90:

These images are now pruned after some days. We need to verify that we use images that exists and works as part of CI. Otherwise we risk finding out only after pushing a tag...

Related issue https://github.com/kubernetes/k8s.io/issues/9599

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -739,6 +739,10 @@ verify-security: ## Verify code and images for vulnerabilities
 		exit 1; \
 	fi
 
+.PHONY: verify-cloudbuild-image
+verify-cloudbuild-image: ## Verify the gcb-docker-gcloud image in cloudbuild.yaml exists and can build the CAPO container image
+	./scripts/verify-cloudbuild-image.sh
+
 .PHONY: compile-e2e
 compile-e2e: ## Test e2e compilation
 	go test -c -o /dev/null -tags=e2e ./test/e2e/suites/conformance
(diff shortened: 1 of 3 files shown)


lentzi90:
Let's wait for https://github.com/kubernetes/test-infra/pull/37323. Then this should work without the extra commit.

lentzi90 (commented):
/approve
/hold cancel
It works 🎉 
However, I think it is not fool proof. We merge the release notes to the main branch and then the build happens on the release branch. The pre-submit job for the main branch will not catch an issue on the release branch. But as long as we keep the images in sync between main and the release branches we are quite safe.
There is also the timing issue. If we are unlucky we can see this job pass on a PR and then when it actually merges the image no longer exists.
It is still better than nothing and we can set up perodic jobs to catch build issues on all supported branches daily.

lentzi90:
Created https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/3239 for followup

smoshiur1237 (commented):
/lgtm

lentzi90:
/hold
We may not need this. Looks like they have now done some changes so that we can get stable images to use. See https://github.com/kubernetes/release/pull/4481. I'll check what is needed on our side and probably push a separate PR with that

lentzi90:
This is no longer needed. Closing in favor of #3305.

PR #3240, opened 2026-06-25
🚀 Release v0.15.0-alpha.0

lentzi90:
Create an alpha release for v0.15.
Note that this does NOT include #3226. I think we have consensus to proceed without this linter and all the breaking changes that comes with it. Since this is an alpha release we can also still change our mind.

/hold

### DIFF
--- a/releasenotes/v0.15.0-alpha.0.md
+++ b/releasenotes/v0.15.0-alpha.0.md
@@ -1,0 +1,210 @@
+🚨 This is an ALPHA RELEASE. Use it only for testing purposes. If you find any bugs, file an [issue](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/new).
+<details>
+<summary>More details about the release</summary>
+
+:warning: **ALPHA RELEASE NOTES** :warning:
+## Highlights
+
+- This release introduces the v1beta2 API types, which are now available alongside the existing v1beta1 types.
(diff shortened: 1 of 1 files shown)


releasenotes/v0.15.0-alpha.0.md
```diff
@@ -0,0 +1,210 @@
...
+- Suppress revive var-naming warning for strings pkg (#3024)
+- Update cluster-api/test to v1.13.2 (#3220)
+- Update dependabot and security scan for release-0.14 (#2949)
+- Update dependabot to ignore gophercloud and golang.org/x bumps (#3078)
+- Update external approvers (#2973)
+- Update golangci-lint to v2.8.0 (#2954)
+- Update missing go version uplift in netlify.toml file (#3038)
+- Update Trivy version to 0.69.3 (#3064)
+- Uplift go version to 1.25.8 (#3044)
+- :rocket: Release v0.12.7 (#2968)
```
nikParasyr, line 112: its a bit weird that these are included 🤔

lentzi90: Well, it is because they go on the main branch. So we merge the release notes to the main branch and that triggers the release. In that sense, this is a commit that is part of v0.15.0-alpha.0.
We could quite easily hide them of course but I am not sure that would be better.

lentzi90 (commented):
/approve
/hold cancel

Issue #3209, opened 2026-06-11
Allow customization of default egress security group rules in CAPO managedSecurityGroups


alejandro-ripoll:
/kind feature

**Describe the solution you'd like**

Currently, when Cluster API Provider OpenStack (CAPO) creates security group rules, it defines default egress rules that allows all outbound traffic (0.0.0.0/0). This behavior is implemented here:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/a6f9102a5deafe32ec0f237d38dad301cd2df306/pkg/cloud/services/networking/securitygroups_rules.go#L32-L51

In some environments, this default behavior is too permissive and does not comply with security or business constraints that require restricting outbound traffic to specific CIDRs.

I would like to have a way to customize or restrict the default egress rules generated by CAPO without having to disable `managedSecurityGroups`, since this feature provides other valuable functionality that I want to keep (such as automated lifecycle management of security groups).

A possible solution could be one of the following:

- Allow configuring egress CIDR ranges via OpenStackCluster spec
- Provide a flag to disable the default "allow all egress" rule so that I can add only allowed egress ranges via `OpenStackCluster.spec.managedSecurityGroups.allNodesSecurityGroupRules`

The goal is to support environments where outbound traffic must be explicitly restricted without losing the benefits of managed security groups.

**Anything else you would like to add:**

This limitation makes it difficult to adopt CAPO in environments with strict network segmentation requirements (e.g., regulated industries or zero-trust networks), where unrestricted egress is not allowed even temporarily.

A configurable approach would significantly improve the provider’s flexibility and enterprise readiness.

lentzi90:
Sounds good to me. Would you be able/interested to work on this?
This could be a fairly large task so I think some kind of design would be good as a first step. We could do that either here as comments on the issue (if it doesn't end up being too large) or as a proposal in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/tree/main/docs/proposals.

alejandro-ripoll:
Thanks for your response @lentzi90. I would be interested in working on this.

As a first step, I tried to put together a lightweight design here in the issue, since I think the scope is still small enough to discuss as comments. If it grows too much or maintainers prefer a more formal review process, I can move this into a proposal under docs/proposals.

The high-level idea would be to add a generic opt-out mechanism for CAPO predefined security group rules, instead of solving this only for egress CIDRs.

**Design Proposal**

[...]

PR #3248, opened 2026-07-06
📖 proposal for configurable allow-all outbound traffic in managed SGs

alejandro-ripoll:

This PR adds a proposal for allowing users to restrict outbound traffic while keeping CAPO-managed security group lifecycle.

The proposed approach adds a new `managedSecurityGroups.allowAllOutboundTraffic` field, defaulting to `true` to preserve the current behavior.

When set to `false`, CAPO would stop creating only the default full-open egress rules. Other CAPO-managed infrastructure rules, such as API server, kubelet, etcd, NodePort and bastion SSH rules, would still be created.

The proposal also covers applying the same behavior to bastion managed security groups.

Related to #3209

This is a proposal-only PR. The implementation is intentionally scoped to a minimal additive API change and avoids a generic opt-out for all predefined rules.

/hold

### DIFF
--- a/docs/proposals/20260706-managed-security-groups-egress.md
+++ b/docs/proposals/20260706-managed-security-groups-egress.md
@@ -25,7 +25,7 @@ CAPO managed security group rules are not all the same kind of rule:
 * Infrastructure rules required for normal cluster operation, such as etcd, kubelet, API server, NodePort, and bastion SSH.
 * Permissive default outbound rules that allow all IPv4 and IPv6 egress.
 
-This proposal targets only the second category. Required infrastructure rules remain managed by CAPO, including minimum egress required for cluster operation.
+This proposal targets only the second category. Required infrastructure rules remain managed by CAPO, including scoped in-cluster egress and egress to the API server endpoint required for cluster operation.
 
 ### Goals
 
@@ -86,12 +86,29 @@ Behavior:
 * If `managedSecurityGroups` is set to an empty object, behavior is unchanged.
 * If `allowAllOutboundTraffic` is unset or `true`, CAPO creates the current default allow-all IPv4 and IPv6 egress rules.
 * If `allowAllOutboundTraffic` is `false`, CAPO removes the default allow-all egress rules.
-* When `allowAllOutboundTraffic` is `false`, CAPO creates minimum egress rules between the control plane and worker security groups.
-* When `allowAllOutboundTraffic` is `false`, CAPO creates egress rules to the API server endpoint on the configured API server port.
+
+* When `allowAllOutboundTraffic` is `false`, CAPO creates scoped egress rules from the control plane and worker security groups to the managed control plane, worker, and self security groups. These rules allow all protocols and all ports and are created for both IPv4 and IPv6.
+* When `allowAllOutboundTraffic` is `false`, CAPO creates TCP egress rules from the control plane and worker security groups to the API server endpoint on the configured API server port. The rule `etherType` matches the address family of the API endpoint.
+
 * Required ingress rules remain unchanged.
 * User-defined egress rules continue to be reconciled normally.
 * The same allow-all outbound setting is applied consistently to the node security groups and the bastion security group.
 * When `allowAllOutboundTraffic` is `false`, CAPO creates TCP port 22 egress rules from the bastion security group to the control plane and worker security groups.
+
+The scoped in-cluster egress rules have the following properties:
+
+* `direction: egress`
+* `remoteGroupID`: the managed control plane, worker, or source security group itself
+* no protocol restriction
+* no port restriction
+* equivalent rules for both `IPv4` and `IPv6`
+
+The API server endpoint rule has:
+
+* `direction: egress`
+* `protocol: tcp`
+* destination port equal to the configured API server port
+* an `etherType` matching the API endpoint address family
 
 ## API Changes
 
@@ -153,8 +170,8 @@ func allowAllOutboundTraffic(managedSecurityGroups *infrav1.ManagedSecurityGroup
 ```
 
 4. When building managed security group rules, include the current default allow-all egress rules only if `allowAllOutboundTraffic(...)` returns `true`.
-5. When the helper returns `false`, add scoped egress from both node security groups to the control plane and worker security groups.
-6. When the helper returns `false`, add egress from both node security groups to the API server endpoint on the configured API server port.
+5. When the helper returns `false`, add unrestricted in-cluster egress rules from both node security groups to the managed control plane, worker, and self security groups. Create equivalent rules for IPv4 and IPv6, with no protocol or port restrictions.
+6. When the helper returns `false`, add TCP egress from both node security groups to the API server endpoint on the configured API server port. Set the rule `etherType` according to the API endpoint address family.
 7. Continue applying all user-provided `SecurityGroupRuleSpec` entries unchanged.
 8. Apply the same behavior to the bastion security group so disabling allow-all outbound traffic does not leave a separate unrestricted egress path there.
 9. When `allowAllOutboundTraffic` is `false`, add TCP port 22 egress from the bastion security group to the control plane and worker security groups.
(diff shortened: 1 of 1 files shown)


docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,216 @@
...
+}
+```
+
+Behavior:
+
+* If `managedSecurityGroups` is omitted, behavior is unchanged.
+* If `managedSecurityGroups` is set to an empty object, behavior is unchanged.
+* If `allowAllOutboundTraffic` is unset or `true`, CAPO creates the current default allow-all IPv4 and IPv6 egress rules.
+* If `allowAllOutboundTraffic` is `false`, CAPO skips only the default allow-all egress rules.
+* Required infrastructure rules remain unchanged.
```
bnallapeta: All of CAPO's "required" rules are ingress-only. The allow-all egress is what makes etcd, kubelet→API, and CNI overlay traffic work outbound. A user who sets false and adds only external egress rules (like the DNS example above) gets a broken cluster.

When false, CAPO should auto-create scoped in-cluster egress rules: egress to remoteGroupID = controlplane/worker/self on both node groups, plus egress to the API endpoint on 6443. 
This is CNI-agnostic and adds no reachability. Connections still need an ingress rule on the destination. Users then only specify external egress (DNS, registry, proxy).

docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,216 @@
...
+func allowAllOutboundTraffic(managedSecurityGroups *infrav1.ManagedSecurityGroups) bool {
+    return managedSecurityGroups == nil ||
+        managedSecurityGroups.AllowAllOutboundTraffic == nil ||
+        *managedSecurityGroups.AllowAllOutboundTraffic
+}
+```
+
+4. When building managed security group rules, include the current default allow-all egress rules only if `allowAllOutboundTraffic(...)` returns `true`.
+5. Continue applying all user-provided `SecurityGroupRuleSpec` entries unchanged.
+6. Apply the same behavior to the bastion security group so disabling allow-all outbound traffic does not leave a separate unrestricted egress path there.
```
bnallapeta: With the flag off, the bastion SG has one SSH ingress rule and zero egress. It can't reach the nodes. There's no field to add rules to the bastion SG.

Recommendation: auto-add a bastion->node-groups egress rule. Also document bastion.spec.securityGroups as the escape hatch for anything else.

docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,216 @@
...
+  AllowAllOutboundTraffic *bool `json:"allowAllOutboundTraffic,omitempty"`
+}
+```
+
+Behavior:
+
+* If `managedSecurityGroups` is omitted, behavior is unchanged.
+* If `managedSecurityGroups` is set to an empty object, behavior is unchanged.
+* If `allowAllOutboundTraffic` is unset or `true`, CAPO creates the current default allow-all IPv4 and IPv6 egress rules.
+* If `allowAllOutboundTraffic` is `false`, CAPO skips only the default allow-all egress rules.
```
bnallapeta: Neutron auto-adds allow-all egress to every new SG. CAPO must delete those rules, not just skip creating them. The reconciler already does this. Reword the line so that it reflects rules are removed/deleted instead of skipped.

docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,216 @@
...
+* Required infrastructure rules remain unchanged.
+* User-defined egress rules continue to be reconciled normally.
+* The same allow-all outbound behavior is applied consistently to the node security groups and the bastion security group.
+
+## API Changes
+
+The change is additive:
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
bnallapeta: We are at v1beta2 now. This requires a couple things.
- fix this example to reflect v1beta2
- add the field to both served versions
- add a conversion section to this proposal doc

docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,216 @@
...
+The new behavior is only used when users explicitly set:
+
+```yaml
+managedSecurityGroups:
+  allowAllOutboundTraffic: false
+```
+
+## Validation
+
+No special validation is required for `allowAllOutboundTraffic`.
```
bnallapeta, line 197: what happens if a user sets `allowAllInClusterTraffic: true and allowAllOutboundTraffic: false`

I think we need to consider this edge case and add a validation for this. Perhaps a CEL expression.

bnallapeta (changes requested):
Comments added.

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Addressed.

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Addressed.

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Addressed.

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Addressed.

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Taking into account the previous proposed change, `allowAllOutboundTraffic: false` does not remove all egress rules. CAPO still creates the scoped rules required for cluster operation, including egress between the cluster node security groups.

Therefore, this combination is valid:
```
allowAllInClusterTraffic: true
allowAllOutboundTraffic: false
```

It allows all traffic between cluster nodes while keeping external egress restricted.

Since all combinations have a clear behavior, I do not think a CEL validation is needed. Proposal updated to reflect this.

alejandro-ripoll:
Thanks @bnallapeta, proposal updated.

bnallapeta:
specify the exact contents of the scoped egress rules, since the Validation section's "no CEL needed" argument depends on it:
- direction: egress, remoteGroupID = control plane / worker / self, all protocols, all ports
- both IPv4 and IPv6 etherTypes (the removed default rules cover both; the existing in-cluster rule generators are IPv4-only, so this needs to be explicit)
- the API endpoint egress rule should match the endpoint's address family

"Minimum egress rules" is otherwise open to a port-scoped interpretation, which would break the allowAllInClusterTraffic: true + allowAllOutboundTraffic: false reasoning. Exact rule set can be finalized in the implementation PR.

bnallapeta:
apart from the above mentioned comment, which can be addressed during implementation.
/lgtm

lentzi90:
This is on my list. I'll hopefully have time to review it tomorrow.
/ok-to-test
@bnallapeta please make a "review comment" to get rid of the "requested changes" thing that blocks merging.

docs/proposals/20260706-managed-security-groups-egress.md
```diff
@@ -0,0 +1,236 @@
...
+    - name: allow-https-to-proxy
+      direction: egress
+      etherType: IPv4
+      protocol: tcp
+      portRangeMin: 443
+      portRangeMax: 443
+      remoteIPPrefix: 10.0.10.0/24
+```
+
+The `allowAllOutboundTraffic` field will be added to both the `v1beta1` and `v1beta2` API versions with the same name and type.
```
lentzi90: Is there a point in adding it to v1beta1? I am thinking that we will not backport the feature to older release branches where v1beta1 is the only available API version, so in that sense it is not "needed" in v1beta1. Anyone that wants the feature can use v1beta2.
I don't have an issue with adding it to v1beta1, just wondering if it is worth it.

lentzi90 (commented):
Thank you, this looks solid!
One question below, not a blocker.
/approve

bnallapeta (commented):
looks good.

bnallapeta (approved):
/lgtm

docs/proposals/20260706-managed-security-groups-egress.md
alejandro-ripoll: Probably not worth it if `v1beta1` is only there for compatibility purposes and this is a new feature. Will update proposal to reflect that. Thanks

alejandro-ripoll: Proposal updated.

alejandro-ripoll:
> specify the exact contents of the scoped egress rules, since the Validation section's "no CEL needed" argument depends on it:
> 
> * direction: egress, remoteGroupID = control plane / worker / self, all protocols, all ports
> * both IPv4 and IPv6 etherTypes (the removed default rules cover both; the existing in-cluster rule generators are IPv4-only, so this needs to be explicit)
> * the API endpoint egress rule should match the endpoint's address family
> 
> "Minimum egress rules" is otherwise open to a port-scoped interpretation, which would break the allowAllInClusterTraffic: true + allowAllOutboundTraffic: false reasoning. Exact rule set can be finalized in the implementation PR.

Proposal updated to clarify egress rules. Thanks

PR #3255, opened 2026-07-07
:bug: Share OpenStack HTTP transports by TLS trust config to stop connection leaks

### LINKED ISSUE #3227: capo-controller-manager accumulates unbounded ESTABLISHED connections to the OpenStack API and stops reconciling with "connect: cannot assign requested address" after ~1 hour
/kind bug

## What steps did you take and what happened

Under normal steady-state operation on a management cluster, `capo-controller-manager` stops reconciling after roughly one hour. Its logs fill with connection errors to every OpenStack service, for example:

```
dial tcp <openstack-api-vip>:443: connect: cannot assign requested address
```

The errors hit all OpenStack endpoints at once. Restarting the controller pod restores reconciliation immediately, and the problem recurs after another hour or so. In this environment the OpenStack API is served behind a single VIP on `:443`.

### Diagnosis

The error string is the kernel `EADDRNOTAVAIL`. For an outbound `connect()` it means the controller cannot allocate a local ephemeral port for the connection. The destination IP and port are fixed (the single API VIP on 443) and the source IP is fixed (the pod IP), so the only variable is the source port. The pod has exhausted the ephemeral port range toward that one destination.

I inspected the controller pod's network namespace (via an ephemeral debug container, and cross-checked with `nsenter` from the node) and counted sockets toward the API VIP by TCP state.

Sockets to the API VIP on 443, grouped by state:

```
$ ss -tan dst <openstack-api-vip>:443 | awk 'NR>1 {print $1}' | sort | uniq -c | sort -rn
  20657 ESTAB
```

Total toward the VIP on 443:

```
$ ss -tan dst <openstack-api-vip>:443 | tail -n +2 | wc -l
20657
```

Ephemeral port range in the pod's network namespace:

```
$ cat /proc/sys/net/ipv4/ip_local_port_range
32768   60999
```

Overall socket summary for the pod:

```
$ ss -s
Total: 21428
TCP:   21905 (estab 21422, closed 482, orphaned 0, timewait 10)

Transport Total     IP        IPv6
RAW       0         0         0
UDP       0         0         0
TCP       21423     21422     1
INET      21423     21422     1
FRAG      0         0         0
```

### What this shows

- Every socket toward the API VIP is in `ESTABLISHED` (20,657 of 20,657). There is essentially no `TIME-WAIT` (10 across the whole pod) and no `CLOSE-WAIT`.
- The ephemeral range is 32768 to 60999, about 28,231 usable source ports toward a single destination.
- Almost all of the pod's established connections (20,657 of 21,422) go to the OpenStack API VIP.
- The count grows monotonically and is never reclaimed. This implies a net accumulation of roughly 28,000 / 3600 ~= 8 new established connections per second, which exhausts the range and triggers `EADDRNOTAVAIL` at around the one-hour mark. A pod restart drops all sockets and resets the cycle.

### What I ruled out

- Not connection churn. `TIME-WAIT` is near zero, so kernel tuning such as `net.ipv4.tcp_tw_reuse`, or widening `ip_local_port_range`, does not address the cause. Widening the range only delays exhaustion.
- Not the CNI. `EADDRNOTAVAIL` is returned to the application by the local `connect()` in the pod's own network namespace, before any SNAT or datapath. The symptom would be identical under any CNI.
- Likely not unread or unclosed response bodies. That pattern tends to leave connections in `CLOSE-WAIT`. Here they are cleanly `ESTABLISHED`, which is consistent with connections that are opened, used once, then left open and idle without being reused or closed.

## What did you expect to happen

Connections to the OpenStack API are reused across reconciles, or closed when finished, so the number of established connections from the controller to the API stays bounded over time and the controller does not exhaust the ephemeral port range.

### Suggested further diagnostics

- Run the controller at `--v=6` and observe a single reconcile. If the gophercloud client authenticates and/or opens a new connection on essentially every reconcile, the scope cache is not effective in this setup.
- Watch the `go_goroutines` metric and pod memory. Each idle keep-alive connection retains a read and a write goroutine, so a leak of N connections corresponds to roughly 2N goroutines. A goroutine count climbing in lockstep with the connection count would confirm abandoned HTTP transports.

## Environment

- Cluster API Provider OpenStack version: **v0.14.4**
- gophercloud: **v2** (as shipped with v0.14.4)
- OpenStack Resource Controller (ORC) version: 2.5.0
- Cluster API version: 1.12.8
- OpenStack version / distribution: `2025.1`
- Kubernetes version of the management cluster (`kubectl version`): `1.35.3`
- OS (for example from `/etc/os-release`): `ubuntu 24.04`
- Approximate scale: `195` OpenStackClusters, `743` OpenStackMachines

bartekle:
pkg/scope cached one *http.Transport per credential (via the scope LRU cache), but evicting or refreshing a scope never closed it. net/http has no finalizer on Transport, so an evicted scope's idle keep-alive connections were never released - only accumulated - for the lifetime of the manager process. With one credential per cluster and a small scope cache, this showed up as tens of thousands of stale established connections across a fleet of clusters.

Key transports by TLS trust config (CA bundle + verify mode) instead of by credential: that dimension is low-cardinality and stable across a fleet against the same cloud(s), so it can be cached in a small, explicitly-evicting LRU (closing idle connections on overflow) without reintroducing the leak. The credential-level scope cache can now evict freely since it no longer owns an exclusive connection pool.

Also raises MaxIdleConnsPerHost to 25 since this transport is now shared by every credential in the fleet against a given host, and adds --transport-cache-max-size to make the new cache's bound configurable.

Fixes #3227 

**TODOs**:

- [ ] squashed commits
- if necessary:
  - [ ] includes documentation
  - [ ] adds unit tests

/hold

### DIFF
--- a/pkg/scope/provider.go
+++ b/pkg/scope/provider.go
@@ -18,10 +18,7 @@ package scope
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"fmt"
-	"net/http"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -256,22 +253,11 @@ func NewProviderClient(cloud clientconfig.Cloud, regionName string, caCert []byt
 	ua.Prepend(fmt.Sprintf("cluster-api-provider-openstack/%s", version.Get().String()))
 	provider.UserAgent = ua
 
-	config := &tls.Config{
-		MinVersion: tls.VersionTLS12,
+	transport, err := getOrCreateTransport(caCert, cloud.Verify)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("get transport for cloud %v: %w", cloud.Cloud, err)
 	}
-	if cloud.Verify != nil {
-		config.InsecureSkipVerify = !*cloud.Verify
-	}
-	if caCert != nil {
-		config.RootCAs = x509.NewCertPool()
-		ok := config.RootCAs.AppendCertsFromPEM(caCert)
-		if !ok {
-			// If no certificates were successfully parsed, set RootCAs to nil to use the host's root CA
-			config.RootCAs = nil
-		}
-	}
-
-	provider.HTTPClient.Transport = &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}
+	provider.HTTPClient.Transport = transport
 	if klog.V(5).Enabled() {
 		provider.HTTPClient.Transport = &osclient.RoundTripper{
 			Rt:     provider.HTTPClient.Transport,
--- a/pkg/scope/transport_test.go
+++ b/pkg/scope/transport_test.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scope
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+)
+
+// generateTestCACertPEM returns a self-signed CA certificate in PEM form,
+// suitable for exercising the x509.CertPool.AppendCertsFromPEM success path.
+func generateTestCACertPEM(t *testing.T) []byte {
+	t.Helper()
+
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		t.Fatalf("generate key: %v", err)
+	}
+	template := &x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject:      pkix.Name{CommonName: "test-ca"},
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().Add(time.Hour),
+		IsCA:         true,
+	}
+	der, err := x509.CreateCertificate(rand.Reader, template, template, &key.PublicKey, key)
+	if err != nil {
+		t.Fatalf("create certificate: %v", err)
+	}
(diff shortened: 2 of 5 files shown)


bartekle:
i tried this version on my staging cluster with ~80 clusters and it seems like it helps with number of connections
```sh
ss -tan | awk 'NR>1 {print $1}' | sort | uniq -c | sort -rn
     11 TIME-WAIT
      6 ESTAB
      3 LISTEN
```

nikParasyr:
@bartekle could you please sign the cla as documented here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/CONTRIBUTING.md#sign-the-cla and also update PR title with the correct emoji

bartekle:
i will try to deal with CLA asap

pkg/scope/provider.go
```diff
@@ -256,22 +253,11 @@ func NewProviderClient(cloud clientconfig.Cloud, regionName string, caCert []byt
...
-		if !ok {
-			// If no certificates were successfully parsed, set RootCAs to nil to use the host's root CA
-			config.RootCAs = nil
-		}
+	transport, err := getOrCreateTransport(caCert, cloud.Verify)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("get transport for cloud %v: %w", cloud.Cloud, err)
 	}
-
-	provider.HTTPClient.Transport = &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}
```
lentzi90, line 274: What if we simply set `IdleConnTimeout` here?

```
provider.HTTPClient.Transport = &http.Transport{
  Proxy: http.ProxyFromEnvironment, TLSClientConfig: config
  IdleConnTimeout:                                   90 * time.Second
}
```

Maybe also `MaxIdleConnsPerHost`?

pkg/scope/transport_test.go
```diff
@@ -0,0 +1,267 @@
...
+		t.Fatalf("create certificate: %v", err)
+	}
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: der})
+}
+
+func boolPtr(b bool) *bool { return &b }
+
+// None of these tests use t.Parallel(): they share the package-level,
+// size-bounded transportCache, and running them concurrently would let one
+// test's insertions evict entries another test is mid-assertion on.
```
lentzi90, line 60: I do not like the sound of this. I think we need to consider alternatives. Could the cache be stored in the `providerScopeFactory` together with the `clientCache` instead?

lentzi90 (commented):
Thanks for the PR!
I am a bit concerned with the complexity of this. Package level state (cache), mutex, changes to exported package signatures (`NewFactory`) and a new flag.
Do we need all this to fix the connection leak? Mainly I wonder if we could simply set `IdleConnTimeout` on the existing transport and skip all the caching changes/additions? The caching seems like a separate concern to me and should not be strictly needed for fixing the bug.

PR #3256, opened 2026-07-10
DNM: Prototype delegation of resource management to ORC

mandre:
This is not meant to merge as is, but show what CAPO leveraging ORC for resource management could look like.

For https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2814.

/hold


lentzi90:
Cool stuff! I skimmed through it and there is a lot that I like!

mandre:
> Cool stuff! I skimmed through it and there is a lot that I like!

Yeah, I was curious what it would look like if we were to leverage more of ORC, and I'm also quite happy how it turned out. I was able to use this code to deploy an OpenShift (granted it only makes a limited use of CAPO, but still, it was successful :smile: ). I like how this removes a lot of complexity from CAPO and replaces it with a thin API translation layer. Obviously since the tests are not passing there's still more work to do, but it's a good start. I'll keep working on it on my spare time.

PR #3265, opened 2026-07-16
:bug: allow unconditional providerID updates in OpenStackMachine

simkam:

Remove conditional immutability check that only allowed providerID and instanceID to be set when the old value was nil. This created a race condition where controller reconciliation retries would be rejected by the webhook, causing IPI installation failures.

The fix aligns CAPO with upstream CAPI patterns used by AWS, GCP, and VSphere providers, which all allow providerID to be updated unconditionally. This prevents webhook rejections during normal controller operation while maintaining the same practical behavior since the controller only sets these fields once.

| Provider | File | Restriction |
|----------|------|-------------|
| **AWS (CAPA)** | [awsmachine_webhook.go](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/af0e5ca481bd15349b3d6e3336238152467086a4/webhooks/awsmachine_webhook.go#L128) | Unconditional | 
| **GCP (CAPG)** | [gcpmachine_webhook.go](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/930755247ae023047b1d4b62b656c3112ecdaecc/webhooks/gcpmachine_webhook.go#L88) | Unconditional | 
| **VSphere (CAPV)** | [vspheremachine.go](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/dadbe25b1117ace60df90d3dd9e4c943d39ac4a0/internal/webhooks/vspheremachine.go#L109) | Allowlist | 
| **Azure (CAPZ)** | [azuremachine_webhook.go](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/cd72698e45a63a84a8dbf2f385837e42a5735278/internal/webhooks/azuremachine_webhook.go#L75) | No restriction |

/hold

### DIFF
--- a/pkg/webhooks/openstackmachine_webhook.go
+++ b/pkg/webhooks/openstackmachine_webhook.go
@@ -95,17 +95,13 @@ func (*openStackMachineWebhook) ValidateUpdate(_ context.Context, oldObj, newObj
 		})
 	}
 
-	// allow changes to providerID once
-	if oldOpenStackMachineSpec["providerID"] == nil {
-		delete(oldOpenStackMachineSpec, "providerID")
-		delete(newOpenStackMachineSpec, "providerID")
-	}
+	// allow changes to providerID (matches AWS, GCP, VSphere providers)
+	delete(oldOpenStackMachineSpec, "providerID")
+	delete(newOpenStackMachineSpec, "providerID")
 
-	// allow changes to instanceID once
-	if oldOpenStackMachineSpec["instanceID"] == nil {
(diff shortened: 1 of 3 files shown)


nikParasyr:
/lgtm

and just for my sanity:

simkam:

The failure looks like a flake.

lentzi90 (commented):
Thanks! I was not aware that other providers did this differently.
/approve

PR #3276, opened 2026-07-23
✨ infer IP version from subnet CIDR on creation

### LINKED ISSUE #1587: enable automatic creation of IPv6-subnets
/kind feature

### **Problem Description**
I want to be able to create an IPv6-only cluster with the cluster-api provider openstack. My expectation is that the capo-controller is able to create the network's subnet, security-groups and a loadbalancer for the kube-apiserver's floating IP from an IPv6 ip-range.

This currently fails when the **openstackcluster.spec.nodeCidr** is given an IPv6-subnet string (this one for example: `fde6:9b71:ffd3:192::/64`) with the following error:
```
Failed to create subnet k8s-clusterapi-cluster-cl-test-ipv6-only-cl-test-ipv6-only: Bad request with: [POST https://network.fra.optimist.innovo.cloud/v2.0/subnets], error message: {"NeutronError": {"type": "InvalidInput", "message": "Invalid input for operation: cidr 'fde6:9b71:ffd3:192::/64' does not match the ip_version '4'.", "detail": ""}}
```

### **Howto reproduce**
1. create a cluster objectt like this:
```

Jamstah:
Subnet creation was unconditionally passing IPVersion 4 to OpenStack, which would cause IPv6 subnets to be created with the wrong IP version. The fix parses the CIDR and selects IPv6 when the network address has no IPv4 representation. Tests are extended to cover both the create and reconcile-existing paths for an IPv6 CIDR.

Fixes #1587

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

### DIFF
--- a/pkg/cloud/services/networking/network.go
+++ b/pkg/cloud/services/networking/network.go
@@ -19,6 +19,7 @@ package networking
 import (
 	"errors"
 	"fmt"
+	"net"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/attributestags"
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
Thanks, this looks good!
I would call this a feature though.
/retitle ✨ infer IP version from subnet CIDR on creation
/approve

nikParasyr (approved):
/lgtm

PR #3357, opened 2026-09-12
🐛 Allow the configured API server port in the managed control plane security group

### LINKED ISSUE #3349: Managed security group API server rule ignores configured apiServerPort
/kind bug

Create an `OpenStackCluster` with:
- `apiServerPort` set to a non-default value (for example `8443`)
- `apiServerLoadBalancer.enabled: true`

After the cluster reconciles, the Octavia load balancer listener and pool member are both created on the configured port (`8443` in this example). The managed security group's "Kubernetes API" ingress rule is still created on port `6443`, regardless of `apiServerPort`.

Confirmed against a live cluster with `openstack security group rule list`:

    IP Protocol: tcp
    Port Range: 6443:6443
    Direction: ingress
    IP Range: 0.0.0.0/0

And `openstack loadbalancer listener list` / `openstack loadbalancer member list` for the same cluster:

    name: k8s-clusterapi-cluster-<ns>-<cluster>-kubeapi-8443
    protocol_port: 8443

So the load balancer is reachable on `8443`, but the security group only opens `6443` — the port nothing is actually listening on. Traffic to the real API server port is not allowed by the managed security group at all.

Reproduced on `v0.14.6` and `v0.15.0-rc.0` separately using `spec.apiServer.port`. Same result on both: the security group rule is unaffected by the configured port.

The managed security group's API server ingress rule should open the configured `apiServerPort` (or `spec.apiServer.port` on `v1beta2`), not a hardcoded `6443`.

The rule is built in `getSGControlPlaneHTTPS()` in `pkg/cloud/services/networking/securitygroups_rules.go`, which hardcodes `PortRangeMin`/`PortRangeMax` to `6443`.

Separately, and lower priority: the `apiServerPort`/`apiServer.port` field description ("the port on which the listener on the APIServer will be created") reads as listener-only. In practice this same value also sets the Octavia pool member's port and `ControlPlaneEndpoint.Port`, which might make documentation misleading.

**Environment:**

- Cluster API Provider OpenStack version (Or `git rev-parse HEAD` if manually built): v0.14.6, also reproduced on commit dd4999d3 (v0.15.0-rc.0)
- Cluster-API version: v1.13.4
- OpenStack version: Yoga
- Minikube/KIND version: N/A (management cluster is RKE2)
- Kubernetes version (use `kubectl version`): v1.35.7+rke2r1

shashankvarma499:

The managed control plane security group always allowed ingress to port 6443, regardless of the configured API server port. The Octavia load balancer listener and its pool members are created on the configured port, so a cluster using a non-default `apiServer.port` ended up with a security group that only opened the port nothing is listening on, while the port the API server is actually reachable on stayed closed.

This threads the API server port through to the security group rule generation so the `Kubernetes API` ingress rule matches the port the API server is exposed on.

Fixes #3349

The port is derived with the same precedence already used for the control plane endpoint (control plane endpoint port, then `apiServer.port`, defaulting to 6443), so clusters with an explicitly configured control plane endpoint keep working unchanged.

Tests:
- `TestGetSGControlPlaneHTTPS` asserts the generated rule uses the given port for both the default (6443) and a custom port.
- `TestService_ReconcileSecurityGroupsAPIServerPort` runs `ReconcileSecurityGroups` with a custom port and asserts the created `Kubernetes API` rule opens that port.

Both tests fail if the rule is hardcoded back to 6443.

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes. (No image changes.)

**TODOs**:

- [x] squashed commits
- if necessary:
  - [ ] includes documentation
  - [x] adds unit tests

### DIFF
--- a/controllers/openstackcluster_controller.go
+++ b/controllers/openstackcluster_controller.go
@@ -783,7 +783,11 @@ func reconcileNetworkComponents(scope *scope.WithLogger, cluster *clusterv1.Clus
 		return fmt.Errorf("failed to reconcile loadbalancer network: %w", err)
 	}
 
-	err = networkingService.ReconcileSecurityGroups(openStackCluster, clusterResourceName)
+	// The API server port is configurable, so the control plane security group
+	// must allow traffic to the port the API server actually listens on rather
+	// than assuming the default.
+	apiServerPort := getAPIServerPort(openStackCluster)
+	err = networkingService.ReconcileSecurityGroups(openStackCluster, clusterResourceName, apiServerPort)
 	if err != nil {
 		conditions.Set(openStackCluster, metav1.Condition{
 			Type:    infrav1.SecurityGroupsReadyCondition,
--- a/pkg/cloud/services/networking/securitygroups.go
+++ b/pkg/cloud/services/networking/securitygroups.go
@@ -39,8 +39,10 @@ const (
 	remoteGroupIDSelf  string = "self"
 )
 
-// ReconcileSecurityGroups reconcile the security groups.
-func (s *Service) ReconcileSecurityGroups(openStackCluster *infrav1.OpenStackCluster, clusterResourceName string) error {
+// ReconcileSecurityGroups reconcile the security groups. apiServerPort is the
+// port the core API server is exposed on, which the control plane security
+// group must allow traffic to.
+func (s *Service) ReconcileSecurityGroups(openStackCluster *infrav1.OpenStackCluster, clusterResourceName string, apiServerPort int32) error {
 	s.scope.Logger().Info("Reconciling security groups")
 	if openStackCluster.Spec.ManagedSecurityGroups == nil {
 		s.scope.Logger().V(4).Info("No need to reconcile security groups")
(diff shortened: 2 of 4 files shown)


lentzi90:
Formatting looks strange. There will probably be some gofmt errors I think
/ok-to-test

lentzi90:
> Formatting looks strange. There will probably be some gofmt errors I think

Never mind this. It was my browser/github doing something strange. I had extra newlines everywhere.

lentzi90 (commented):
Thank you for the PR!
I think the logic is sound and this does fix the issue.
However, I have one thing I would want to change. Instead of passing the `apiServerPort` as an argument in each function, I think we should add a method on the OpenStackCluster itself that can be used where needed.
We already have some methods like this here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/5d15f6707ca8b5fb7c8b009bfa209b17cb54479a/api/v1beta2/openstackcluster_types.go#L359-L411

So please add something like this there:
```go
// GetAPIServerPort returns the port the API server is expected to listen on,
// derived from the control plane endpoint if already set, then from
// spec.apiServer.port, defaulting to 6443.
func (c *OpenStackCluster) GetAPIServerPort() int32 {
	switch {
	case c.Spec.ControlPlaneEndpoint != nil && c.Spec.ControlPlaneEndpoint.IsValid():
		return c.Spec.ControlPlaneEndpoint.Port
	case c.Spec.APIServer.GetPort() != nil:
		return int32(*c.Spec.APIServer.GetPort())
	}
	return 6443
}
```

Then remove the old `getAPIServerPort` helper and the `apiServerPort` parameter from the function arguments. We can call `GetAPIServerPort` directly in `generateDesiredSecGroups`.

